PR #1, opened 2022-07-20
Add scripts to rule them all

pezholio:
This adds some simple scripts to aid running in development. We use these in the frontend application, so it’s useful to mirror this approach in the API too, to aid with context switch. Currently the bootsrap and server scripts are stubs, as there’s nothing we (yet) need to do to bootstrap / setup the app, but they’re worth having around as a reminder to update them when we do start introducing external dependencies, like a database etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,42 @@
 # Approved Premises API
 
-This is the backend for the Approved Premises service. Its API is consumed by the corresponding "UI" codebase ([approved-premises-ui](https://github.com/ministryofjustice/approved-premises-ui)). 
+This is the backend for the Approved Premises service. Its API is consumed by the corresponding "UI" codebase ([approved-premises-ui](https://github.com/ministryofjustice/approved-premises-ui)).
+
+## Prerequisites
+
+TBC
+
+## Setup
+
+When running the application for the first time, run the following command:
+
+```bash
+script/setup # TODO - this script is currently a stub
+```
+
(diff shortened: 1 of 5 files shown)


pezholio:
[Add scripts to rule them all for the API](https://trello.com/c/EHsjukQc/470-add-scripts-to-rule-them-all-for-the-api)

README.md
```diff
@@ -1,6 +1,42 @@
...
+
+```bash
+script/setup
+```
+
+If you're coming back to the application after a certain amount of time, you can run:
+
+```bash
+script/bootstrap
+```
```
edavey, line 21: As these don't do anything yet would it be helpful to say as much, e.g. with a "TODO"

script/bootstrap
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+# script/bootstrap: Resolve all dependencies that the application requires to
+#                   run.
+
+set -e
+
+cd "$(dirname "$0")/.."
```
edavey, line 8: should this print out something to acknowledge that this is a placeholder script e.g. "Bootstrap script not yet implemented"

edavey (approved):
Good stuff: thanks for adding these. 

I've made a couple of comments/suggestions asking whether it's worth noting explicitly that bootstrapping isn't yet implemented...

richpjames (approved):
Looks good, agree with both of Ed's points

PR #3, opened 2022-07-26
Document approach to handling validation errors

pezholio:
This documents the approach we'll take to handling validation errors, using the [RFC-7807 problem spec](https://datatracker.ietf.org/doc/html/rfc7807). This is already handled well in Spring Boot and gives us a reliable and predictable way of reporting errors back to the client

### DIFF
--- a/doc/architecture/decisions/0001-use-problem-responses-to-handle-validation-errors.md
+++ b/doc/architecture/decisions/0001-use-problem-responses-to-handle-validation-errors.md
@@ -1,0 +1,59 @@
+# 1. Use `application/problem+json` responses to handle validation errors
+
+Date: 2022-07-25
+
+## Status
+
+Accepted
+
+## Context
+
+We have already decided to
+[handle validation errors at the API level][1]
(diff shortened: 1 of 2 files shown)


richpjames (approved):
Seems sensible, I'm glad it's properly specced out too

doc/architecture/decisions/0001-use-problem-responses-to-handle-validation-errors.md
```diff
@@ -0,0 +1,59 @@
...
+We have already decided to
+[handle validation errors at the API level][1]
+, but we need a to agree a common and predictable way to return validation or other
+errors to the client, so they can be parsed an returned to the user.
+
+## Decision
+
+We will return `application/problem+json` responses, as outlined in [RFC-7807][2]. As well as returning the standard Problem
+response, we will also return details of what fields caused the error in an
+`invalid-params` array. For example:
```
edavey, line 20: Is it worth referencing the "extension members" terminology of the RFC (https://datatracker.ietf.org/doc/html/rfc7807#section-3.2) ?

PR #4, opened 2022-07-27
Update the API response to show errors

pezholio:
I had to upgrade to OpenAPI 3.0 to handle content-type headers on a per-response basis. It makes sense for us to use the latest version anyway.

Error responses now look like this:

image


pezholio:
[Update the Swagger spec in the API repo](https://trello.com/c/uEjDdQVn/504-update-the-swagger-spec-in-the-api-repo)

edavey (approved):
this looks great, Stu 👍

PR #6, opened 2022-07-28
Add Person to Arrival and Booking schemas

richpjames:
We need to return a person's name with a Booking and Arrival
I thought it would make sense to return a Person object which consists of a name and CRN

I'm not sure whether it would make sense to add a Person to the other events whilst I'm here (EG Nonarrival, Departure) - I feel they will need a Person too.


richpjames:
Renamed the branch to include ticket number

PR #7, opened 2022-07-28
(497) Add Person to Arrival and Booking schemas

richpjames:
We need to return a person's name with a Booking and Arrival
I thought it would make sense to return a Person object which consists of a name and CRN

I'm not sure whether it would make sense to add a Person to the other events whilst I'm here (EG Nonarrival, Departure) - I feel they will need a Person too
[Trello ticket](https://trello.com/c/73Y1jvej/497-displaying-person-on-probations-name-against-a-booking-arrival)

### DIFF
--- a/src/main/resources/static/mini-manage-api-stubs.yml
+++ b/src/main/resources/static/mini-manage-api-stubs.yml
@@ -394,8 +394,8 @@ components:
         id:
           type: string
           format: uuid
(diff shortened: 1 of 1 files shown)


pezholio (approved):
LGTM - I think in the future we'll probably only need to pass the CRN in when creating the Booking, but that can come later

PR #8, opened 2022-07-28
Suggested api changes

edavey:
Some suggested tweaks / changes to the interface between the UI and API components.

I also have some suggestions for the `Premises` entity which can follow on....

### DIFF
--- a/src/main/resources/static/mini-manage-api-stubs.yml
+++ b/src/main/resources/static/mini-manage-api-stubs.yml
@@ -213,7 +213,7 @@ paths:
           description: unexpected error
           content: {}
       x-codegen-request-body-name: body
-  /premises/{premisesID}/bookings/{bookingID}/departure:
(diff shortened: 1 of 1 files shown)


richpjames (approved):
These all seem like sensible changes to me!

src/main/resources/static/mini-manage-api-stubs.yml
```diff
@@ -134,7 +134,7 @@ paths:
         content:
           'application/json':
             schema:
-              $ref: '#/components/schemas/BookingBody'
+              $ref: '#/components/schemas/NewBooking'
```
pezholio, line 137: I think `BookingDto` might be a better name here, as this is what we're calling it in the UI

pezholio: Whoops - didn't realise this had been merged!

PR #11, opened 2022-08-02
Add "reference data" endpoints

edavey:
We add endpoints for the following "reference" data:

#### for departures
- `/departure-reasons`
- `/move-on-categories`
- `/destination-providers`

#### for arrivals
- `/supervising-providers`
- `/supervising-teams`
- `/supervising-officers`
- `/key-workers`

#### for lost-beds
- `/lost-bed-reasons`

This PR also adds "tags" to group the endpoints (in the docs) into
endpoints concerned with:

- premises
- operations on premises
- operations on bookings
- reference data

api_aug_2_22

### DIFF
--- a/src/main/resources/static/mini-manage-api-stubs.yml
+++ b/src/main/resources/static/mini-manage-api-stubs.yml
@@ -451,6 +451,29 @@ paths:
         500:
           description: unexpected error
           content: { }
+  /reference-data/supervising-teams:
+    get:
+      tags:
+        - Reference Data
+      summary: Lists all supervising teams for arrivals
+      responses:
+        200:
+          description: successful operation
+          content:
+            'application/json':
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/SupervisingTeam'
+        401:
+          description: not authenticated
+          content: { }
+        403:
+          description: unauthorised
+          content: { }
+        500:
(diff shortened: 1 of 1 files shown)


src/main/resources/static/mini-manage-api-stubs.yml
```diff
@@ -359,6 +377,167 @@ paths:
...
+          description: successful operation
+          content:
+            'application/json':
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/DepartureReason'
+        401:
+          description: not authenticated
+          content: { }
```
brownbearsoftware, line 396: Minor point but just ommitting the content key here achieves the same result as providing an empty map

brownbearsoftware: although I notice this is how it's done elsewhere in the file, so maybe one for a seperate PR

edavey: 👍 thanks @brownbearsoftware

brownbearsoftware (approved):

src/main/resources/static/mini-manage-api-stubs.yml
brownbearsoftware: sorry forgot to actually approve it!

PR #12, opened 2022-08-02
Remove empty _content_ properties from Swagger spec

edavey:
These aren't adding anything.

We will soon be describing our error handling as
content-type `application/problem+json` see ADR:

[Use application/problem+json responses to handle validation errors](https://github.com/ministryofjustice/approved-premises-api/blob/main/doc/architecture/decisions/0001-use-problem-responses-to-handle-validation-errors.md)


brownbearsoftware (approved):

brownbearsoftware:
I updated all of these to the RFC-7808 Problem responses in https://github.com/ministryofjustice/approved-premises-api/pull/14 so don't think we need this anymore :)

PR #13, opened 2022-08-02
Add postgres & flyway

brownbearsoftware:
Added Postgres to the local development flow + a local profile to connect to it from.

Then changed the application itself to connect to it and run migrations via Flyway.

Also added some instructions to the README on running through IntelliJ for debugging.

Changed test job to start up a Postgres database and changed base integration test to clear & run migrations after each integration test.


docker-compose-development-database.yml
```diff
@@ -0,0 +1,15 @@
+version: "3"
```
pezholio, line 1: Can we name these `docker-compose.yml` and `docker-compose.test.yml`, just to be consistent with what we have in the UI?

build.gradle.kts
```diff
@@ -12,6 +12,7 @@ val springDocVersion = "1.6.9"
 dependencies {
   implementation("org.springframework.boot:spring-boot-starter-webflux")
   implementation("org.springframework.boot:spring-boot-starter-data-jpa")
+  implementation("org.flywaydb:flyway-core")
```
pezholio, line 15: Just a comment about migrations - is there any way we can timestamp these `i.e. yyyymmddhhmmdd`, rather than prefixing with `V1` etc, I find this keeps the files order nicely, rather than `V1, V10, V11` etc

src/main/resources/application.yml
```diff
@@ -18,6 +18,10 @@ spring:
       test:
         - "stdout"
 
+  jpa:
```
pezholio, line 21: Does this mean we have to create the flyway migrations manually as SQL? Does JPA othewise try and create schemas from models? Not a problem, just for my own understanding.

pezholio (approved):
Looks good from my POV - just a couple of comments/questions for my own understanding - thanks!

docker-compose-development-database.yml
brownbearsoftware: Sure thing, have rebased in

src/main/resources/application.yml
brownbearsoftware: That's correct yeah - you can just let JPA do it itself but generally seen as too risky on a production database

build.gradle.kts
brownbearsoftware: Sure, have removed the V prefix - so can be named like 20220803__migration_title.sql now

src/main/resources/application.yml
pezholio: 👍

pezholio (approved):

brownbearsoftware:

PR #14, opened 2022-08-03
Openapi generator support

brownbearsoftware:
Added Open API Generator Gradle plugin & configured.  Made compileKotlin Gradle task depend on it so it's impossible to end up in "woops, forgot to run openApiGenerate" situations.

Added a mock implementation for Premises GET to demonstrate it working.

There's some detail in the ADR about the trade-offs of this approach, I was hoping by now they would have added the missing support for the OpenAPI 3 `oneOf` option on response bodies but sadly not - so potentially this could be too much of an issue to justify going this way.

The development flow is like so:

 - Edit mini-manage-api-stubs.yml
 - Run the `openapitools`-> `openApiGenerate` gradle task
 - If you've added an endpoint with a new root path, it will create a new interface that needs to be implemented - these can be seen in build/generated/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi:
image

- In src/main/controller, create an equivalently named controller, e.g. PremisesController which implements the interface
- The interface has a default implementation for all of the methods on it which just returns a 501 "Not Implemented".  To provide the implementation override the method - in IntelliJ just press Alt + Insert and select `Override Methods`
image
- Then select the endpoint you want to implement
image


doc/architecture/decisions/0003-use-open-api-generator-to-scaffold-controllers-from-open-api-file.md
```diff
@@ -0,0 +1,52 @@
...
+the possibility of generating other types of code in the future to reduce repetitive work for clients of the API:
+ - [TypeScript Client](https://openapi-generator.tech/docs/generators/typescript-fetch)
+ - [Kotlin Client](https://openapi-generator.tech/docs/generators/kotlin)
+
+
+Unfortunately, OpenAPI Generator does not support the OpenAPI 3 `anyOf`, `oneOf` options on request/response bodies 
+which are useful if you want to return completely different responses from an endpoint.  Whilst this is on their 
+[roadmap](https://openapi-generator.tech/docs/roadmap/#short-term) it has been there for some time.  The implication of 
+this is that if you want to return either a success or error response from the same endpoint then you need to do so in 
+an enveloped response, e.g.:
```
pezholio, line 32: Does this mean we need to make changes to the OpenAPI schema? Could we do this in this PR, so we don't forget?

pezholio (approved):
LGTM - The outline of the process is really useful too - would you be able to add the documentation on the development flow as a Markdown file in `/doc`, so it doesn't get lost? Thanks! 👍

doc/architecture/decisions/0003-use-open-api-generator-to-scaffold-controllers-from-open-api-file.md
brownbearsoftware: Quite possibly, do we have an idea of the what the shape of the error responses are going to be or is this TBD?

brownbearsoftware: I've just spotted the ADR for problem responses, so all good - will have a read.

pezholio: There's also an example here https://github.com/ministryofjustice/approved-premises-api/blob/main/src/main/resources/static/mini-manage-api-stubs.yml#L156

brownbearsoftware: So this works out quite actually and seems we won't need to do the envelope response after all.

Whilst Open API Generate doesn't support returning different models from the delegate methods (it just picks the first one from the possible responses, which should always be the 2xx response) this won't be a problem since the https://github.com/zalando/problem-spring-web library uses exceptions to convey a failure - the exception handlers aren't bound by the same restrictions in what can be returned.

I have added the problem responses to the other 4/5xx responses in the spec too.

brownbearsoftware: Have also updated the ADR to reflect this finding

brownbearsoftware:

PR #15, opened 2022-08-03
525 - Add get all premises endpoint

brownbearsoftware:
Add migration that creates tables for Premises, Bed, Ap Area, Local Authority Area, Probation Region.
Add JPA entities for those tables.
Implement the GET all premises endpoint.

Also added kotlin-jpa/allOpen plugins which allow JPA to instantiate data classes without us providing a no-args constructor and making all properties nullable.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApAreaEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApAreaEntity.kt
@@ -1,0 +1,17 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity
+
+import java.util.UUID
+import javax.persistence.Entity
+import javax.persistence.Id
+import javax.persistence.OneToMany
+import javax.persistence.Table
+
+@Entity
+@Table(name = "ap_area")
+data class ApAreaEntity(
+  @Id
+  var id: UUID,
+  var name: String,
+  @OneToMany(mappedBy = "apArea")
+  var premises: MutableList<PremisesEntity>
+)
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PremisesEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PremisesEntity.kt
@@ -1,0 +1,28 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity
+
+import java.util.UUID
+import javax.persistence.Entity
+import javax.persistence.Id
+import javax.persistence.JoinColumn
+import javax.persistence.ManyToOne
+import javax.persistence.Table
+
+@Entity
+@Table(name = "premises")
+data class PremisesEntity(
+  @Id
+  var id: UUID,
+  var name: String,
+  var apCode: String,
+  var postcode: String,
+  var totalBeds: Int,
+  @ManyToOne
+  @JoinColumn(name = "probation_region_id")
+  var probationRegion: ProbationRegionEntity,
+  @ManyToOne
+  @JoinColumn(name = "ap_area_id")
+  var apArea: ApAreaEntity,
+  @ManyToOne
+  @JoinColumn(name = "local_authority_area_id")
+  var localAuthorityArea: LocalAuthorityAreaEntity
+)
--- a/src/main/resources/logback-spring.xml
+++ b/src/main/resources/logback-spring.xml
@@ -7,7 +7,7 @@
     <springProperty scope="context" name="app" source="spring.application.name"/>
 
 
-    <springProfile name="dev || stdout">
+    <springProfile name="local || stdout">
         <appender name="consoleAppender" class="ch.qos.logback.core.helpers.NOPAppender"/>
         <appender name="logAppender" class="ch.qos.logback.core.ConsoleAppender">
             <layout class="ch.qos.logback.classic.PatternLayout">
@@ -16,7 +16,7 @@
         </appender>
     </springProfile>
 
-    <springProfile name="!(dev || stdout)">
+    <springProfile name="!(local || stdout)">
         <appender name="consoleAppender" class="ch.qos.logback.core.ConsoleAppender">
             <layout class="ch.qos.logback.classic.PatternLayout">
                 <Pattern>${LOG_PATTERN}</Pattern>
--- a/src/main/resources/static/mini-manage-api-stubs.yml
+++ b/src/main/resources/static/mini-manage-api-stubs.yml
@@ -584,6 +584,10 @@ components:
       properties:
         id:
           type: string
+          format: uuid
+          example: 6abb5fa3-e93f-4445-887b-30d081688f44
+        identifier:
+          type: string
           example: LEEDS
         name:
           type: string
@@ -593,7 +597,8 @@ components:
       properties:
         id:
           type: string
-          example: 1500001234
+          format: uuid
+          example: cd1c2d43-0b0b-4438-b0e3-d4424e61fb6a
         name:
           type: string
           example: Yorkshire & The Humber
@@ -602,7 +607,11 @@ components:
       properties:
         id:
           type: string
-          example: 1500123456
+          format: uuid
+          example: 952790c0-21d7-4fd6-a7e1-9018f08d8bb0
+        identifier:
+          type: string
+          example: NPSNE
         name:
           type: string
           example: NPS North East Central Referrals
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/PremisesTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/PremisesTest.kt
@@ -1,0 +1,115 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.integration
+
+import com.fasterxml.jackson.databind.ObjectMapper
+import org.junit.jupiter.api.Test
+import org.springframework.beans.factory.annotation.Autowired
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.health.api.model.ApArea
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.health.api.model.LocalAuthorityArea
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.health.api.model.Premises
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.health.api.model.ProbationRegion
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApAreaEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.LocalAuthorityAreaEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.PremisesEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ProbationRegionEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.repository.ApAreaTestRepository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.repository.LocalAuthorityAreaTestRepository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.repository.PremisesTestRepository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.repository.ProbationRegionTestRepository
+import java.util.UUID
+
+class PremisesTest : IntegrationTestBase() {
+  @Autowired
+  lateinit var objectMapper: ObjectMapper
+
+  @Autowired
+  lateinit var probationRegionRepository: ProbationRegionTestRepository
+
+  @Autowired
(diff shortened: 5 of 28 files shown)


brownbearsoftware:

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApAreaEntity.kt
```diff
@@ -0,0 +1,16 @@
...
+
+import javax.persistence.Entity
+import javax.persistence.Id
+import javax.persistence.OneToMany
+import javax.persistence.Table
+
+@Entity
+@Table(name = "ap_area")
+data class ApAreaEntity(
+  @Id
```
pezholio, line 11: Can we make the IDs UUIDs please?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PremisesEntity.kt
```diff
@@ -0,0 +1,30 @@
...
+@Entity
+@Table(name = "premises")
+data class PremisesEntity(
+  @Id
+  var id: UUID,
+  var name: String,
+  var apCode: String,
+  var postcode: String,
+  @OneToMany(mappedBy = "premises")
+  var beds: MutableList<BedEntity>,
```
pezholio, line 20: At this point, we won't be modelling the concept of beds (just a total bed count), so this might be a premature optimisation.

src/test/resources/sql/get-all-premises.sql
```diff
@@ -0,0 +1,20 @@
...
+    ('PBREG2', 'PBREG2');
+
+INSERT INTO premises (id, name, ap_code, postcode, probation_region_id, ap_area_id, local_authority_area_id) VALUES
+    ('322e3614-2dc0-4166-8dec-63c016e20a0e', 'Premises One', 'PREMONE', 'ST8ST8', 'PBREG1', 'APAREA1', 'ATAREA1'),
+    ('f04743a7-d9f1-4818-8847-5c15f4874ec0', 'Premises Two', 'PREMTWO', 'ST8ST7', 'PBREG2', 'APAREA2', 'ATAREA2');
+
+INSERT INTO bed (id, premises_id) VALUES
+    ('62a24e1c-7e4a-4dc3-b52e-a1495a3e8937', '322e3614-2dc0-4166-8dec-63c016e20a0e'),
+    ('7984554a-83f9-4651-9d4b-b4898803273b', 'f04743a7-d9f1-4818-8847-5c15f4874ec0'),
+    ('04bc06f3-0efb-448e-a129-5ef858f15cbd', 'f04743a7-d9f1-4818-8847-5c15f4874ec0');
```
pezholio, line 20: Is there an easier way to generate objects here, rather than directly using SQL? Thinking about something like https://github.com/thoughtbot/fishery (in Typescript-land), or https://github.com/thoughtbot/factory_bot (in Ruby land). Just there's a lot of boilerplate that needs to be done to setup/teardown

pezholio (commented):
Just a couple of drive-by comments, not a thorough review!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApAreaEntity.kt
brownbearsoftware: Sure, just for ApArea or also for LocalAuthority and ProbationRegion?

pezholio: All of them please - LocalAuthority and ProbationRegion will also need seperate `identifier` fields for the internal IDs

brownbearsoftware: can do - `identifier` would be a string?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PremisesEntity.kt
brownbearsoftware: Have switched it to just be a total count column on the Premises table

src/test/resources/sql/get-all-premises.sql
brownbearsoftware: I think the closest to those we'll get is using JPA from the tests to save objects - will switch it over so you can take a look.

brownbearsoftware: Have switched it over now - more LOC overall so not sure it really adresses your concern around boilerplate?  I can't think of a way to do this any more succintly though unfortunately.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/PremisesTest.kt
```diff
@@ -0,0 +1,115 @@
...
+        Premises(
+          id = UUID.fromString("322e3614-2dc0-4166-8dec-63c016e20a0e"),
+          name = "Premises One",
+          apCode = "PREMONE",
+          postcode = "ST8ST8",
+          bedCount = 1,
+          probationRegion = ProbationRegion(id = UUID.fromString("d520e141-fb24-4f39-839b-666e69b29f17"), name = "PBREG1"),
+          apArea = ApArea(id = UUID.fromString("73f5aa38-a44a-41ce-b50f-6bb7df51cab7"), name = "APAREA1"),
+          localAuthorityArea = LocalAuthorityArea(id = UUID.fromString("311dc6ec-3925-4f42-aaca-13b6f3de96e0"), name = "ATAREA1")
+        ),
```
pezholio, line 93: It would be good if we could reuse the Premises we define above here, to be more DRY and ensure the tests aren't brittle

src/main/resources/logback-spring.xml
```diff
@@ -7,7 +7,7 @@
     <springProperty scope="context" name="app" source="spring.application.name"/>
 
 
-    <springProfile name="dev || stdout">
+    <springProfile name="local || stdout">
```
edavey, line 10: Just wondering if we are right to be using `local` rather than `dev`? 

Probably I'm not understanding properly but it seems that `dev` is the conventional term at HMPPS?

e.g.

- https://tech-docs.hmpps.service.justice.gov.uk/application-platform.html#dev-namespace
- https://github.com/ministryofjustice/make-recall-decision-api/blob/ce163502e4843b768754c076c8157a6b239e5d80/src/main/resources/application-dev.yml

src/main/resources/static/mini-manage-api-stubs.yml
```diff
@@ -583,6 +583,10 @@ components:
       type: object
       properties:
         id:
+          type: string
+          format: uuid
+          example: 6abb5fa3-e93f-4445-887b-30d081688f44
+        identifer:
```
edavey, line 589: tiny typo here I think: should be `identifier`

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/PremisesTest.kt
edavey: I wonder if it would be clearer if our `expectedJson` were the literal JSON we expect in the response, e.g.

```kt
val expectedJson = """
      [
        {
          "id": "322e3614-2dc0-4166-8dec-63c016e20a0e",
          "name": "Premises One",
          "apCode": "PREMONE",
          "postcode": "ST8ST8",
          "bedCount": 1,
          "probationRegion": {
            "id": "d520e141-fb24-4f39-839b-666e69b29f17",
            "identifier": null,
            "name": "PBREG1"
          },
          "apArea": {
            "id": "73f5aa38-a44a-41ce-b50f-6bb7df51cab7",
            "name": "APAREA1"
          },
          "localAuthorityArea": {
            "id": "311dc6ec-3925-4f42-aaca-13b6f3de96e0",
            "identifer": null,
            "name": "ATAREA1"
          }
        },
        {
          "id": "f04743a7-d9f1-4818-8847-5c15f4874ec0",
          "name": "Premises Two",
          "apCode": "PREMTWO",
          "postcode": "ST8ST7",
          "bedCount": 2,
          "probationRegion": {
            "id": "c1585959-91d7-4ee6-b0c1-30df25cd4983",
            "identifier": null,
            "name": "PBREG2"
          },
          "apArea": {
            "id": "1387e449-c8f7-486a-9f39-3d6cf0210ecd",
            "name": "APAREA2"
          },
          "localAuthorityArea": {
            "id": "6758b2a9-36f6-4b15-aff9-0902e121f5ac",
            "identifer": null,
            "name": "ATAREA2"
          }
        }
      ]

    """.trimIndent()
```

If we use our models (`Premises` etc) and rely on `writeValueAsString` to construct the expectation then we add a level of indirection to the spec I think and we make it harder to be confident in what we're actually asserting?

edavey (commented):
Really great to see all these neat techniques! Thanks a lot for knocking this out so rapidly and carefully.

I have left a couple of comments/suggestions.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/PremisesTest.kt
pezholio: Good call @edavey - I still think there's value in defining the premises as a variable and then reusing the attributes in the JSON, i.e:

```
val premises1 = PremisesEntity(
        id = UUID.fromString("f04743a7-d9f1-4818-8847-5c15f4874ec0"),
        name = "Premises One",
        apCode = "PREMONE",
        postcode = "ST8ST7",
        totalBeds = 2,
        probationRegion = pbReg2,
        apArea = apArea2,
        localAuthorityArea = localAuthorityArea2
      )

....

val expectedJson = """
      [
        {
          "id": premises1.id,
          "name": premises1.name,
          "apCode": premises1.apCode,
          "postcode": premises1.postcode,
...
```

etc, etc

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/PremisesTest.kt
```diff
@@ -0,0 +1,115 @@
...
+        id = UUID.fromString("f04743a7-d9f1-4818-8847-5c15f4874ec0"),
+        name = "Premises Two",
+        apCode = "PREMTWO",
+        postcode = "ST8ST7",
+        totalBeds = 2,
+        probationRegion = pbReg2,
+        apArea = apArea2,
+        localAuthorityArea = localAuthorityArea2
+      )
+    )
```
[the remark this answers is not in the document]
brownbearsoftware: Makes sense, I've had a go with kfactory - the test itself looks a lot cleaner now

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/PremisesTest.kt
brownbearsoftware: So with the change to use kfactory - I'm now taking the values it generates and using them instead - so won't be able to explicitly write the JSON anymore as we don't know the values ahead of time

src/main/resources/logback-spring.xml
brownbearsoftware: Yeah I think you're correct re dev vs local - any objection to me moving it from local->dev in a subsequent PR just to keep this one from balooning?

src/main/resources/db/migration/20220803__premises_tables.sql
```diff
@@ -0,0 +1,36 @@
+CREATE TABLE ap_area (
```
pezholio, line 1: Can we add a time to the migration name here too, just thinking if we generate more than one migration a day, it could get confusing. Might be good to have a generate-migration script to make this easier. Something like:

```
timestamp=`date "+%Y%m%d%H%M%S"`
filename=`echo "$1" | awk '{print tolower($0)}' | tr ' ' '_'`

touch "src/main/resources/db/migration/$timestamp_$filename"
```

Which can then be run with `script/geneate-migration migration name goes here`

src/main/resources/logback-spring.xml
edavey: 👍 sounds sensible to have a PR following on which just does the local -> dev renaming.

src/main/resources/db/migration/20220803__premises_tables.sql
pezholio: Amazing 👍

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/PremisesTest.kt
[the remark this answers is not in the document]
pezholio: Cool, that makes sense - thanks for persisting with me! 👍

pezholio (approved):

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/PremisesTest.kt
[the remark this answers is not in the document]
brownbearsoftware: No worries, best we settle on patterns we're all happy with at the start!

PR #16, opened 2022-08-03
Attempt to start docker daemon when running launcher script

edavey:
Just a convenience: we're doing this in the UI app and it seems to help
reduce the friction a tiny bit.

Not sure if this plays nicely with Windows?

### DIFF
--- a/script/utils/launch-docker.sh
+++ b/script/utils/launch-docker.sh
@@ -1,0 +1,15 @@
+#!/bin/sh
+
+if (! docker stats --no-stream > /dev/null 2>&1  ); then
+  echo "==> Launching Docker..."
+
+  # On Mac OS this would be the terminal command to launch Docker
+  open /Applications/Docker.app
+  printf "Waiting for Docker to launch..."
+ #Wait until Docker daemon is running and has completed initialisation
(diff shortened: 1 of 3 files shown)


script/utils/launch-docker.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
+
+if (! docker stats --no-stream > /dev/null 2>&1  ); then
```
brownbearsoftware, line 3: Re this potentially not playing nice with Windows, I think it also won't play nice with Linux - maybe:

```
if [ "$(uname)" == "Darwin" ]; then
  if (! docker stats --no-stream > /dev/null 2>&1  ); then
    echo "==> Launching Docker..."

    # On Mac OS this would be the terminal command to launch Docker
    open /Applications/Docker.app
    printf "Waiting for Docker to launch..."
   #Wait until Docker daemon is running and has completed initialisation
  while (! docker stats --no-stream > /dev/null 2>&1 ); do
    # Docker takes a few seconds to initialize
    printf '.'
    sleep 1
  done
  fi
else
  echo "Not on Mac so skipping start Docker check"
fi
```

PR #17, opened 2022-08-04
Rename local profile to dev

brownbearsoftware:
To stay in-line with other MoJ projects, rename the local profile to dev + associated gradle task


pezholio (approved):
LGTM 👍

PR #18, opened 2022-08-04
Add required properties to OpenAPI spec

brownbearsoftware:
So that OpenAPI generate creates models with correct nullability, set the required properties for the schemas in mini-manage-api-stubs.yml.


pezholio (approved):
LGTM 👍

PR #21, opened 2022-08-04
Add .trivyignore to .gitignore

brownbearsoftware:
This is auto-generated by one of the gradle plugins


pezholio:
Ah yes, I keep seeing this pop up when I run builds and end up just deleting it 😆

pezholio (approved):

PR #23, opened 2022-08-05
528  - Secure endpoints with HMPPS Auth tokens

brownbearsoftware:
Setup Spring Security to protect endpoints, must now present a valid HMPPS Auth JWT.

Added way of generating arbitrary JWTs with the private key that HMPPS Auth uses in the containerized version to be used in the tests - setup tests to use containerized HMPPS Auth.

Added convenience script to get a client_credentials JWT from containerized HMPPS Auth.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/OAuth2ResourceServerSecurityConfiguration.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/OAuth2ResourceServerSecurityConfiguration.kt
@@ -1,0 +1,114 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.config
+
+import com.fasterxml.jackson.databind.ObjectMapper
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.context.annotation.Bean
+import org.springframework.core.convert.converter.Converter
+import org.springframework.http.HttpMethod
+import org.springframework.security.authentication.AbstractAuthenticationToken
+import org.springframework.security.config.annotation.web.builders.HttpSecurity
+import org.springframework.security.config.annotation.web.configuration.EnableWebSecurity
+import org.springframework.security.config.http.SessionCreationPolicy
+import org.springframework.security.config.web.servlet.invoke
+import org.springframework.security.core.GrantedAuthority
+import org.springframework.security.core.authority.SimpleGrantedAuthority
+import org.springframework.security.oauth2.jwt.Jwt
+import org.springframework.security.oauth2.server.resource.authentication.JwtAuthenticationToken
+import org.springframework.security.oauth2.server.resource.authentication.JwtGrantedAuthoritiesConverter
+import org.springframework.security.web.AuthenticationEntryPoint
+import org.springframework.security.web.SecurityFilterChain
+
+@EnableWebSecurity
+class OAuth2ResourceServerSecurityConfiguration {
+  @Bean
+  @Throws(Exception::class)
+  fun securityFilterChain(http: HttpSecurity, @Autowired objectMapper: ObjectMapper): SecurityFilterChain {
+    http {
+      authorizeHttpRequests {
+        authorize(HttpMethod.GET, "/health/**", permitAll)
+        authorize(HttpMethod.GET, "/swagger-ui/**", permitAll)
+        authorize(HttpMethod.GET, "/favicon.ico", permitAll)
+        authorize(HttpMethod.GET, "/info", permitAll)
+        authorize(anyRequest, authenticated)
+      }
+
+      anonymous { disable() }
+
+      oauth2ResourceServer {
+        jwt { jwtAuthenticationConverter = AuthAwareTokenConverter() }
+
+        authenticationEntryPoint = AuthenticationEntryPoint { _, response, _ ->
+          response.apply {
+            status = 401
+            contentType = "application/problem+json"
+            characterEncoding = "UTF-8"
+
+            writer.write(
+              objectMapper.writeValueAsString(
+                object {
+                  val title = "Unauthenticated"
+                  val status = 401
+                  val detail =
+                    "A valid HMPPS Auth JWT must be supplied via bearer authentication to access this endpoint"
+                }
+              )
+            )
+          }
+        }
+      }
+
+      sessionManagement { sessionCreationPolicy = SessionCreationPolicy.STATELESS }
+    }
+
+    return http.build()
+  }
+}
+
+class AuthAwareTokenConverter : Converter<Jwt, AbstractAuthenticationToken> {
+  private val jwtGrantedAuthoritiesConverter: Converter<Jwt, Collection<GrantedAuthority>> =
+    JwtGrantedAuthoritiesConverter()
+
+  override fun convert(jwt: Jwt): AbstractAuthenticationToken {
+    val claims = jwt.claims
+    val principal = findPrincipal(claims)
+    val authorities = extractAuthorities(jwt)
+    return AuthAwareAuthenticationToken(jwt, principal, authorities)
+  }
+
+  private fun findPrincipal(claims: Map<String, Any?>): String {
+    return if (claims.containsKey(CLAIM_USERNAME)) {
+      claims[CLAIM_USERNAME] as String
+    } else if (claims.containsKey(CLAIM_USER_ID)) {
+      claims[CLAIM_USER_ID] as String
+    } else {
+      claims[CLAIM_CLIENT_ID] as String
+    }
+  }
+
+  private fun extractAuthorities(jwt: Jwt): Collection<GrantedAuthority> {
+    val authorities = mutableListOf<GrantedAuthority>().apply { addAll(jwtGrantedAuthoritiesConverter.convert(jwt)!!) }
+    if (jwt.claims.containsKey(CLAIM_AUTHORITY)) {
+      @Suppress("UNCHECKED_CAST")
+      val claimAuthorities = (jwt.claims[CLAIM_AUTHORITY] as Collection<String>).toList()
+      authorities.addAll(claimAuthorities.map(::SimpleGrantedAuthority))
+    }
+    return authorities.toSet()
+  }
+
+  companion object {
+    const val CLAIM_USERNAME = "user_name"
+    const val CLAIM_USER_ID = "user_id"
+    const val CLAIM_CLIENT_ID = "client_id"
+    const val CLAIM_AUTHORITY = "authorities"
+  }
+}
+
+class AuthAwareAuthenticationToken(
+  jwt: Jwt,
+  private val aPrincipal: String,
+  authorities: Collection<GrantedAuthority>
+) : JwtAuthenticationToken(jwt, authorities) {
+  override fun getPrincipal(): String {
+    return aPrincipal
+  }
+}
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/ExceptionHandling.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/ExceptionHandling.kt
@@ -1,0 +1,26 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.problem
+
+import org.springframework.security.access.AccessDeniedException
+import org.springframework.security.authentication.AuthenticationCredentialsNotFoundException
+import org.springframework.web.bind.annotation.ControllerAdvice
+import org.zalando.problem.StatusType
+import org.zalando.problem.ThrowableProblem
+import org.zalando.problem.spring.common.AdviceTrait
+import org.zalando.problem.spring.web.advice.ProblemHandling
+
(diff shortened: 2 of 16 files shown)


brownbearsoftware:
[Add token authentication to the API](https://trello.com/c/1SUaq2p0/528-add-token-authentication-to-the-api)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/OAuth2ResourceServerSecurityConfiguration.kt
```diff
@@ -0,0 +1,108 @@
...
+
+@EnableWebSecurity
+class OAuth2ResourceServerSecurityConfiguration {
+  @Bean
+  @Throws(Exception::class)
+  fun securityFilterChain(http: HttpSecurity, @Autowired objectMapper: ObjectMapper): SecurityFilterChain {
+    http
+      .authorizeHttpRequests { authorize ->
+        authorize
+          .mvcMatchers(HttpMethod.GET, "/premises").authenticated()
```
mikehalmamoj, line 27: We tend to come at this from the opposite end - list the anonymous endpoints such as /health and open api docs, then require everything else to be authenticated. [Random example](https://github.com/ministryofjustice/create-and-vary-a-licence-api/blob/main/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/config/ResourceServerConfiguration.kt#L15).

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/OAuth2ResourceServerSecurityConfiguration.kt
```diff
@@ -0,0 +1,108 @@
...
+import org.springframework.security.config.http.SessionCreationPolicy
+import org.springframework.security.core.GrantedAuthority
+import org.springframework.security.core.authority.SimpleGrantedAuthority
+import org.springframework.security.oauth2.jwt.Jwt
+import org.springframework.security.oauth2.server.resource.authentication.JwtAuthenticationToken
+import org.springframework.security.oauth2.server.resource.authentication.JwtGrantedAuthoritiesConverter
+import org.springframework.security.web.SecurityFilterChain
+
+@EnableWebSecurity
+class OAuth2ResourceServerSecurityConfiguration {
```
andymarke, line 20: We have tended to reverse this, only allow authenticated requests with but allow only certain requests through unauthenticated (e.g. health)

https://github.com/ministryofjustice/hmpps-nomis-prisoner-api/blob/main/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/config/ResourceServerConfiguration.kt

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/OAuth2ResourceServerSecurityConfiguration.kt
```diff
@@ -0,0 +1,108 @@
...
+      it.jwt().jwtAuthenticationConverter(AuthAwareTokenConverter())
+
+      it.authenticationEntryPoint { _, response, _ ->
+        response.apply {
+          status = 401
+          contentType = "application/problem+json"
+          characterEncoding = "UTF-8"
+
+          writer.write(
+            objectMapper.writeValueAsString(
```
andymarke, line 43: not sure if this is needed since won't Spring return a 401 anyway ?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/ExceptionHandling.kt
```diff
@@ -0,0 +1,26 @@
...
+  override fun toProblem(throwable: Throwable, status: StatusType): ThrowableProblem? {
+    if (throwable is AuthenticationCredentialsNotFoundException) {
+      return UnauthenticatedProblem()
+    }
+
+    if (throwable is AccessDeniedException) {
+      return ForbiddenProblem()
+    }
+
+    return AdviceTraitDefault().toProblem(throwable, status)
```
mikehalmamoj, line 22: Does this return `json+problem` responses? Interesting, I'll keep an eye on this.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/AuthTest.kt
```diff
@@ -0,0 +1,69 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.integration
+
+import org.junit.jupiter.api.Test
+import org.springframework.beans.factory.annotation.Autowired
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.JwtProvider
+
+class AuthTest : IntegrationTestBase() {
```
andymarke, line 7: Not sure if you have a future PR but we would typically also protect endpoints with a role; would should check the token contains the correct role (aka authority) and return a 403 if it doesn't.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/util/JwtProvider.kt
```diff
@@ -0,0 +1,41 @@
...
+import io.jsonwebtoken.SignatureAlgorithm
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.core.io.ByteArrayResource
+import org.springframework.stereotype.Component
+import java.security.KeyStore
+import java.security.PrivateKey
+import java.util.Base64
+
+@Component
+class JwtProvider(
```
andymarke, line 13: Maybe take a look at https://github.com/ministryofjustice/hmpps-nomis-prisoner-api/blob/main/src/test/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/helper/JwtAuthHelper.kt as an alternative approach which means you wouldn't need `hmpps-auth-dev-key` stuff when running tests

andymarke (dismissed):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/OAuth2ResourceServerSecurityConfiguration.kt
```diff
@@ -0,0 +1,108 @@
...
+import org.springframework.security.oauth2.server.resource.authentication.JwtAuthenticationToken
+import org.springframework.security.oauth2.server.resource.authentication.JwtGrantedAuthoritiesConverter
+import org.springframework.security.web.SecurityFilterChain
+
+@EnableWebSecurity
+class OAuth2ResourceServerSecurityConfiguration {
+  @Bean
+  @Throws(Exception::class)
+  fun securityFilterChain(http: HttpSecurity, @Autowired objectMapper: ObjectMapper): SecurityFilterChain {
+    http
```
matt-crowson-moj, line 24: Spring does provide a Kotlin DSL for this kind of thing.  Not essential but is kinda nice IMO:

http {
  authorizeRequests {
    ....
  }
  ...
}

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/OAuth2ResourceServerSecurityConfiguration.kt
brownbearsoftware: So one advantage of doing it this way around is that it becomes impossible for endpoints that require a higher level of authorisation to slip through PR without an entry actually being added here.  e.g. if just having a JWT is sufficient for most endpoints on the service but for a new POST operaiton we also require an authority it's fairly easily done for someone to forget to add the requisite entry here.  Ideally this would be caught by testing + PR review but it is an easy one to miss.  That said, I don't know if that's actually going to be the case for this project yet so happy to switch for consistency sake if you'd like?

Also seems I forgot to add the entries for the healthcheck, swagger etc. so will pop those in.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/ExceptionHandling.kt
brownbearsoftware: It does indeed, we're using https://github.com/zalando/problem-spring-web for this.  It works quite well out of the box but for some of the authentication failure conditions (e.g. expired JWT) I had to override a few bits - this class being part of that

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/AuthTest.kt
matt-crowson-moj: For reference spring boot - [Method Secuity](https://docs.spring.io/spring-security/reference/servlet/authorization/method-security.html)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/OAuth2ResourceServerSecurityConfiguration.kt
brownbearsoftware: @mikehalmamoj had a similar comment, interested in your thoughts on my response also: https://github.com/ministryofjustice/approved-premises-api/pull/23#discussion_r940097738

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/AuthTest.kt
brownbearsoftware: This is still a TBD for us yet (which authority to use, potentially we'll need a new one) so will add once we've come to a decision

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/util/JwtProvider.kt
brownbearsoftware: Interesting, I'm guessing this works because this is overriding the part of Spring Security that would normally use the JWKS from HMPPS Auth with one generated at runtime?

```
@Bean
  fun jwtDecoder(): JwtDecoder = NimbusJwtDecoder.withPublicKey(keyPair.public as RSAPublicKey).build()
```

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/OAuth2ResourceServerSecurityConfiguration.kt
mikehalmamoj: Ah ok, so does that mean you would expect to add the authorities here? We tend to use the `@PreAuthorize` annotation for that which is nice because then the required authority is next to the endpoint it secures. And we'd generally write an integration test for the required authority before adding it to the endpoint making it harder to miss. (You'll almost certainly end up securing your endpoints with a role btw. that's a fairly standard pattern we use).

All that said I don't think there's anything wrong with your approach. And consistency is a good thing generally, but herd mentality isn't 😁

brownbearsoftware:
```
Ah ok, so does that mean you would expect to add the authorities here?
```

We've gone for the approach of using [Open Api Generator](https://github.com/OpenAPITools/openapi-generator/) to create the request/response models and controllers from the OpenAPI spec file so there's no chance of the spec being inconsistent with the actual implementation ([Pull Request here](https://github.com/ministryofjustice/approved-premises-api/pull/14) if interested.)  One downside of that is that we can't easily add @PreAuthorize to the controllers since they're generated so I would expect to add the authorities here.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/OAuth2ResourceServerSecurityConfiguration.kt
brownbearsoftware: Didn't realise this was an option, much nicer!  Have updated.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/OAuth2ResourceServerSecurityConfiguration.kt
mikehalmamoj: Ah interesting, contract first. I tried that generator a few years and it worked ok with some small compromises. Cool makes sense to me.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/util/JwtProvider.kt
brownbearsoftware: Have switched to this method 👍

mikehalmamoj (approved):

PR #27, opened 2022-08-09
526 - Get bookings for Premises

brownbearsoftware:
Implemented the GET /premises/{id}/bookings endpoint & associated tests.

Also moved the kfactory setup into IntegrationTestBase to avoid duplication across the individual test classes.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
@@ -3,16 +3,19 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.controller
 import org.springframework.http.ResponseEntity
 import org.springframework.stereotype.Service
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.PremisesApiDelegate
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.health.api.model.Booking
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.health.api.model.Premises
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.NotFoundProblem
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.PremisesService
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.BookingTransformer
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.PremisesTransformer
 import java.util.UUID
 
 @Service
 class PremisesController(
   private val premisesService: PremisesService,
-  private val premisesTransformer: PremisesTransformer
+  private val premisesTransformer: PremisesTransformer,
+  private val bookingTransformer: BookingTransformer
 ) : PremisesApiDelegate {
   override fun premisesGet(): ResponseEntity<List<Premises>> {
     return ResponseEntity.ok(
@@ -25,5 +28,12 @@ class PremisesController(
       ?: throw NotFoundProblem(premisesId, "Premises")
 
     return ResponseEntity.ok(premisesTransformer.transformJpaToApi(premises))
+  }
+
+  override fun premisesPremisesIdBookingsGet(premisesId: UUID): ResponseEntity<List<Booking>> {
+    val premises = premisesService.getPremises(premisesId)
+      ?: throw NotFoundProblem(premisesId, "Premises")
+
+    return ResponseEntity.ok(premises.bookings.map(bookingTransformer::transformJpaToApi))
   }
 }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ArrivalEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ArrivalEntity.kt
@@ -1,0 +1,39 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity
+
+import java.time.LocalDate
+import java.util.Objects
+import java.util.UUID
+import javax.persistence.Entity
+import javax.persistence.Id
+import javax.persistence.JoinColumn
+import javax.persistence.OneToOne
+import javax.persistence.Table
+
+@Entity
+@Table(name = "arrival")
+data class ArrivalEntity(
+  @Id
+  val id: UUID,
+  val arrivalDate: LocalDate,
+  val expectedDepartureDate: LocalDate,
+  val notes: String?,
+  @OneToOne
+  @JoinColumn(name = "booking_id")
+  var booking: BookingEntity
+) {
+  override fun equals(other: Any?): Boolean {
+    if (this === other) return true
+    if (other !is ArrivalEntity) return false
+
+    if (id != other.id) return false
+    if (arrivalDate != other.arrivalDate) return false
+    if (expectedDepartureDate != other.expectedDepartureDate) return false
+    if (notes != other.notes) return false
+
+    return true
+  }
+
+  override fun hashCode() = Objects.hash(arrivalDate, expectedDepartureDate, notes)
+
+  override fun toString() = "ArrivalEntity:$id"
+}
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BookingEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BookingEntity.kt
@@ -1,0 +1,55 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity
+
+import java.time.LocalDate
+import java.util.Objects
+import java.util.UUID
+import javax.persistence.Entity
+import javax.persistence.Id
+import javax.persistence.JoinColumn
+import javax.persistence.ManyToOne
+import javax.persistence.OneToOne
+import javax.persistence.Table
+
+@Entity
+@Table(name = "booking")
+data class BookingEntity(
+  @Id
+  val id: UUID,
+  @OneToOne(mappedBy = "booking")
+  var person: PersonEntity?,
+  val arrivalDate: LocalDate,
+  val departureDate: LocalDate,
+  val keyWorker: String,
+  @OneToOne(mappedBy = "booking")
+  var arrival: ArrivalEntity?,
+  @OneToOne(mappedBy = "booking")
+  var departure: DepartureEntity?,
+  @OneToOne(mappedBy = "booking")
+  var nonArrival: NonArrivalEntity?,
+  @OneToOne(mappedBy = "booking")
+  var cancellation: CancellationEntity?,
+  @ManyToOne
+  @JoinColumn(name = "premises_id")
+  var premises: PremisesEntity
+) {
+  override fun equals(other: Any?): Boolean {
+    if (this === other) return true
+    if (other !is BookingEntity) return false
+
+    if (id != other.id) return false
+    if (person != other.person) return false
+    if (arrivalDate != other.arrivalDate) return false
+    if (departureDate != other.departureDate) return false
+    if (keyWorker != other.keyWorker) return false
+    if (arrival != other.arrival) return false
+    if (departure != other.departure) return false
+    if (nonArrival != other.nonArrival) return false
+    if (cancellation != other.cancellation) return false
+
+    return true
+  }
+
+  override fun hashCode() = Objects.hash(person, arrivalDate, departureDate, keyWorker, arrival, departure, nonArrival, cancellation)
+
+  override fun toString() = "BookingEntity:$id"
+}
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/DepartureEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/DepartureEntity.kt
@@ -1,0 +1,50 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity
+
+import java.time.OffsetDateTime
+import java.util.Objects
+import java.util.UUID
+import javax.persistence.Entity
+import javax.persistence.Id
+import javax.persistence.JoinColumn
+import javax.persistence.ManyToOne
+import javax.persistence.OneToOne
+import javax.persistence.Table
+
+@Entity
+@Table(name = "departure")
+data class DepartureEntity(
+  @Id
+  val id: UUID,
+  val dateTime: OffsetDateTime,
+  @ManyToOne
+  @JoinColumn(name = "departure_reason_id")
+  val reason: DepartureReasonEntity,
+  @ManyToOne
+  @JoinColumn(name = "move_on_category_id")
+  val moveOnCategory: MoveOnCategoryEntity,
+  @ManyToOne
+  @JoinColumn(name = "destination_provider_id")
+  val destinationProvider: DestinationProviderEntity,
+  val notes: String?,
+  @OneToOne
+  @JoinColumn(name = "booking_id")
+  var booking: BookingEntity
+) {
+  override fun equals(other: Any?): Boolean {
+    if (this === other) return true
+    if (other !is DepartureEntity) return false
+
+    if (id != other.id) return false
+    if (dateTime != other.dateTime) return false
+    if (reason != other.reason) return false
+    if (moveOnCategory != other.moveOnCategory) return false
+    if (destinationProvider != other.destinationProvider) return false
+    if (notes != other.notes) return false
+
+    return true
+  }
+
+  override fun hashCode() = Objects.hash(dateTime, reason, moveOnCategory, destinationProvider, notes)
+
+  override fun toString() = "DepartureEntity:$id"
+}
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/repository/BookingRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/repository/BookingRepository.kt
@@ -1,0 +1,9 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.repository
+
+import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.stereotype.Repository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.BookingEntity
+import java.util.UUID
+
+@Repository
+interface BookingRepository : JpaRepository<BookingEntity, UUID>
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/ArrivalTransformer.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/ArrivalTransformer.kt
@@ -1,0 +1,17 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer
+
+import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.health.api.model.Arrival
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ArrivalEntity
+
+@Component
+class ArrivalTransformer() {
+  fun transformJpaToApi(jpa: ArrivalEntity?) = jpa?.let {
+    Arrival(
+      bookingId = jpa.booking.id,
+      arrivalDate = jpa.arrivalDate,
+      expectedDepartureDate = jpa.expectedDepartureDate,
+      notes = jpa.notes
+    )
+  }
+}
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/BookingTransformer.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/BookingTransformer.kt
@@ -1,0 +1,37 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer
+
+import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.health.api.model.Booking
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.BookingEntity
+import java.lang.RuntimeException
+
+@Component
+class BookingTransformer(
+  private val personTransformer: PersonTransformer,
+  private val arrivalTransformer: ArrivalTransformer,
+  private val departureTransformer: DepartureTransformer,
+  private val nonArrivalTransformer: NonArrivalTransformer,
+  private val cancellationTransformer: CancellationTransformer
+) {
+  fun transformJpaToApi(jpa: BookingEntity) = Booking(
+    id = jpa.id,
+    person = personTransformer.transformJpaToApi(jpa.person!!),
+    arrivalDate = jpa.arrivalDate,
+    departureDate = jpa.departureDate,
+    keyWorker = jpa.keyWorker,
+    status = determineStatus(jpa),
+    arrival = arrivalTransformer.transformJpaToApi(jpa.arrival),
+    departure = departureTransformer.transformJpaToApi(jpa.departure),
+    nonArrival = nonArrivalTransformer.transformJpaToApi(jpa.nonArrival),
+    cancellation = cancellationTransformer.transformJpaToApi(jpa.cancellation)
+  )
+
+  private fun determineStatus(jpa: BookingEntity) = when {
+    jpa.nonArrival != null -> Booking.Status.notMinusArrived
+    jpa.arrival != null && jpa.departure == null -> Booking.Status.arrived
+    jpa.departure != null -> Booking.Status.departed
+    jpa.cancellation != null -> Booking.Status.cancelled
+    jpa.arrival == null && jpa.nonArrival == null -> Booking.Status.awaitingMinusArrival
+    else -> throw RuntimeException("Could not determine status for Booking ${jpa.id}")
+  }
+}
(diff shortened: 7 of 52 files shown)


pezholio (commented):
Would you mind breaking down https://github.com/ministryofjustice/approved-premises-api/pull/27/commits/fb78d6abfdc5b580f7357abf6711f47d15861fd4 into smaller commits, to make it easier to review?

brownbearsoftware:
[Add API endpoints for bookings](https://trello.com/c/z8TqoOuh/526-add-api-endpoints-for-bookings)

brownbearsoftware:
> Would you mind breaking down [fb78d6a](https://github.com/ministryofjustice/approved-premises-api/commit/fb78d6abfdc5b580f7357abf6711f47d15861fd4) into smaller commits, to make it easier to review?

Have done 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/DepartureEntity.kt
```diff
@@ -0,0 +1,43 @@
...
+  Absconded,
+  PlannedMove,
+  Recalled
+}
+
+enum class MoveOnCategory {
+  PrivateHouse,
+  Custody,
+  Deported
+}
```
pezholio, line 43: Could we add all the Reasons / Categories as per https://trello.com/c/0o9HkzJi/434-recording-departures#comment-62d98669beb5dc0130173739 pls?

pezholio: I'm also open to the idea of them being database entries - but I think enums is fine for now

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/BookingTransformer.kt
```diff
@@ -0,0 +1,46 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer
+
+import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.health.api.model.Booking
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.BookingEntity
+import java.lang.RuntimeException
+
+@Component
+class BookingTransformer(
```
pezholio, line 9: Just wondering as there's a bit of logic here, and if it doesn't cause too much pain, is there any value in adding some unit tests for these transformers? Happy to be guided by you, but my usual way of doing things would be to add some tests if anything out of the ordinary was being done, also just for the sake of self-documentation, if nothing else.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/DepartureTransformer.kt
```diff
@@ -0,0 +1,51 @@
...
+        Departure.Reason.absconded
+      }
+      DepartureReason.PlannedMove -> {
+        Departure.Reason.plannedMove
+      }
+      DepartureReason.Recalled -> {
+        Departure.Reason.recalled
+      }
+    }
+  }
```
pezholio, line 36: Can you explain how this works? `DepartureReason` is an enum, and it looks like `Departure.Reason` is a subclass of `Departure`, but I don't see where this is.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/BookingTest.kt
```diff
@@ -0,0 +1,168 @@
...
+        Departure.Reason.absconded
+      }
+      DepartureReason.PlannedMove -> {
+        Departure.Reason.plannedMove
+      }
+      DepartureReason.Recalled -> {
+        Departure.Reason.recalled
+      }
+    }
+  }
```
pezholio, line 153: Is there any way we can reuse this logic from the transformers to save repeating ourselves? Or, as I think we talked about before (I think me and Ed had a difference of opinion here!) being more prescriptive about the shape of JSON we expect. We could make this easier by only returning one booking, and then testing the edge cases for transformation in unit tests for the transformer. What do you think?

pezholio (commented):
Looks good, just some questions / suggestions. Not going to put an explicit approval on yet, because I'd like some Kotlin folks to have a peek too 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/DepartureEntity.kt
brownbearsoftware: Given the use of commas, slashes, brackets etc. in a lot of the names I think we might be better off going for database entries - will make it a lot simpler if we need to reword some of them post-launch too since they can be keyed off a surrogate key rather than their literal names.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/DepartureTransformer.kt
```diff
@@ -0,0 +1,51 @@
...
+
+@Component
+class DepartureTransformer(private val destinationProviderTransformer: DestinationProviderTransformer) {
+  fun transformJpaToApi(jpa: DepartureEntity?): Departure? {
+    if (jpa == null) return null
+
+    return Departure(
+      bookingId = jpa.booking.id,
+      dateTime = jpa.dateTime,
+      reason = transformDepartureReason(jpa.reason),
```
pezholio, line 17: Would we return these as a JSON object, now we're using database entried?

brownbearsoftware: Yeah that's right, I've just pushed the change to return objects

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/DepartureEntity.kt
brownbearsoftware: Have pushed this change now

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/DepartureTransformer.kt
brownbearsoftware: This is no longer in the code - but to answer the question - DepartureReason is an enum in the JPA package that I had written by hand wherreas Departure.Reason was in the generated API response models package that OpenAPI generate produces.  Departure.Reason also being an enum but as a nested class rather than at the package level.  e.g.

I'd written as:
```
package xyz.jpa

class DepartureEntity { ... }
enum class DepartureReason { enum1, enum2 .... }
```

OpenAPI had generated as:
```
package xyz.models
class Departure {
   ....
   enum class Reason { enum1, enum2 }
}
```

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/BookingTest.kt
brownbearsoftware: I had a similar internal debate on this one when writing - some thoughts:

```
Is there any way we can reuse this logic from the transformers to save repeating ourselves?
```

We could do this by just injecting the transformers into the test.  I'm averse to this though because we're then esentially just trusting that the transformer is correct since it will be testing its own work.  On the other hand, doing it like this we are basically just writing the same code in two different places which in practice probably isn't much better.

```
Or, as I think we talked about before (I think me and Ed had a difference of opinion here!) being more prescriptive about the shape of JSON we expect
```

I don't think writing the JSON manually provides any benefit in our case:

```
data class MyThing(
   val id: String,
   val name: String
)

val factoryProducedObject = factory<MyThing>.produceAndPersist()
val expectedJson = """
{
   "id": "${factoryProducedObject.id}",
   "name": "${factoryProducedObject.name}"
}
"""

webTestClient.get()
      .uri("/my-thing/${factoryProducedObject.id}")
      .exchange()
      .expectStatus()
      .isOk
      .expectBody()
      .json(expectedJson)
```

The `.json(expectedJson)` call doesn't just do an `expectedJson == responseBodyJson`, because we've written our JSON in a human readable way and the endpoint by default will return it without superfulous whitespace/newlines.  We would potentially also hit problems with the ordering of properties etc.  So it still needs to parse our hand-written JSON and compare it to a parsed version of the response body JSON.

With this in mind, I don't think there is an advantage in doing this over trusting Jackson's ObjectMapper (which is widely used and tested) to convert our API model (which is generated from the OpenAPI spec so we know it's the same model the endpoint is going to return.)

We could instead just get the response body as a string and compare it to our hand-written JSON.  We'd have to ensure we are ordering our hand-written JSON in the same way and not adding superfulous whitespace but I feel this would make the tests overly proscriptive  - e.g. should a Jackson (the serializer used by Spring Boot) update that alters the order in which properties are written break all of our integration tests?

We could provide all of the values as literals to the factory which would give more confidence that nothing strange is happening with us transforming betwen the generated JPA model and the API spec model:

```
data class MyThing(
   val id: String,
   val name: String
)

val factoryProducedObject = factory<MyThing>
   .withId("thing-id")
   .withName("thing-name")
   .produceAndPersist()

val expectedResponseBody = MyThing(
   id = "thing-id",
   name = "thing-name"
)

val expectedResponseBodyJson = objectMapper.writeValueAsString(expectedResponseBody)

webTestClient.get()
      .uri("/my-thing/${factoryProducedObject.id}")
      .exchange()
      .expectStatus()
      .isOk
      .expectBody()
      .json(expectedJson)
```

But this seems to invalidate the point in using KFactory to begin with - we'd be writing way more code than if we were just writing the SQL by hand but for no added benefit.


```We could make this easier by only returning one booking, and then testing the edge cases for transformation in unit tests for the transformer.```

Agreed on this, I will add some transformer unit tests so we don't need to do all the edge cases here.  I think the above discussion is still valid even if we only do one Booking in the integration test though.

Comment (pezholio) [reply src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/BookingTest.kt]: What do you think about injecting the transformers into the test, as long as we unit testing the transformers when there's any logic beyond simple marshalling?

File: src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/DepartureTransformer.kt (hunk 3 above)
Comment (pezholio) [reply src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/DepartureTransformer.kt]: Ah ha! That makes sense. Thanks for the explanation 👍

File: src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/BookingTest.kt (hunk 4 above)
Comment (brownbearsoftware) [reply src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/BookingTest.kt]: Fair point, seems reasonable enough if they've been unit tested :)

File: src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt [hunk 6]
```diff
@@ -3,16 +3,19 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.controller
...
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.PremisesApiDelegate
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.health.api.model.Booking
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.health.api.model.Premises
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.NotFoundProblem
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.PremisesService
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.BookingTransformer
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.PremisesTransformer
 import java.util.UUID
 
 @Service
```
matt-crowson-moj, line 14: Looks like you are using the wrong stereotype.  This is a controller isn't it?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ArrivalEntity.kt
```diff
@@ -0,0 +1,42 @@
...
+import java.time.LocalDate
+import java.util.UUID
+import javax.persistence.Entity
+import javax.persistence.Id
+import javax.persistence.JoinColumn
+import javax.persistence.OneToOne
+import javax.persistence.Table
+
+@Entity
+@Table(name = "arrival")
```
matt-crowson-moj, line 12: I guess this is more of a convention/style thing but tend to pluralise table names.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ArrivalEntity.kt
```diff
@@ -0,0 +1,42 @@
...
+data class ArrivalEntity(
+  @Id
+  var id: UUID,
```
matt-crowson-moj, line 15: Heavy use of `var` instead of `val` on entities.  Ideally would go for immutable objects where possible or mutable only where needed.  Obviously I don't know how these are used.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
mikehalmamoj: I think if you use `@RestController` then you don't have to be explicit about the `ResponseEntity` and Spring will wire that up for you. However I suspect your approach here is dictated by the contract first approach? 

I think this is why we don't use contract first - so we can make use of the various Spring annotations to decorate the endpoints. Here's a [recent example of a controller](https://github.com/ministryofjustice/hmpps-nomis-prisoner-api/blob/main/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisprisonerapi/resource/VisitResource.kt) of ours and you can see we use lots of annotations!

(As a side note I thought contract first was the way to go until I realised all of the limitations. I found the generators to be quite limited, but was hoping they would have matured by now)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ArrivalEntity.kt
```diff
@@ -0,0 +1,42 @@
...
+import java.util.UUID
+import javax.persistence.Entity
+import javax.persistence.Id
+import javax.persistence.JoinColumn
+import javax.persistence.OneToOne
+import javax.persistence.Table
+
+@Entity
+@Table(name = "arrival")
+data class ArrivalEntity(
```
matt-crowson-moj, line 13: This is more of potential gotcha and not saying the affects you.

With data classes you get toString for free. You may (or may not) want to override this in case you have any PII on your entities that could be leaked to the logs by mistake.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/repository/BookingRepository.kt
```diff
@@ -0,0 +1,9 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.repository
+
+import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.stereotype.Repository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.BookingEntity
+import java.util.UUID
+
+@Repository
```
mikehalmamoj, line 8: I've started defining the repository in the same file as the entity - they are very much coupled anyway and the first thing I always do with a Repository is click on the Entity!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/ArrivalTransformer.kt
```diff
@@ -0,0 +1,19 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer
+
+import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.health.api.model.Arrival
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ArrivalEntity
+
+@Component
+class ArrivalTransformer() {
+  fun transformJpaToApi(jpa: ArrivalEntity?): Arrival? {
```
matt-crowson-moj, line 9: This looks like very java way of doing things `if (jpa == null) return null`

Should just be able to do something like this ` fun transformJpaToApi(jpa: ArrivalEntity?): Arrival? = jpa?.let { Arrival( ... ) }`

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/repository/BookingRepository.kt
matt-crowson-moj: +1 to this idea.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/ArrivalTransformer.kt
```diff
@@ -0,0 +1,19 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer
+
+import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.health.api.model.Arrival
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ArrivalEntity
+
+@Component
+class ArrivalTransformer() {
```
mikehalmamoj, line 8: We've found that a nice pattern is to add a constructor to the API class that accepts an Entity as the single parameter. Saves on yet another component to inject for a single use with little benefit (transformations like this are trivial).

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ArrivalEntity.kt
brownbearsoftware: Have ammended 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/ArrivalTransformer.kt
matt-crowson-moj: What Mike says above actually what I would also do.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
brownbearsoftware: It's sort-of-a-controller, the actual Controller is in `build/generated/src/main/kotlin/uk/gov/..../api/PremisesApiController.kt` which just delegates to PremisesApiDelegate which this class implements.  So I think as far as the stereotypes go this is either a @Service or perhaps a @Component 

image

```
However I suspect your approach here is dictated by the contract first approach?
```
That's correct, hands are a little tied on this one.  Do you two think that the contract-first approach deviates too much from the norm at MoJ?

```
(As a side note I thought contract first was the way to go until I realised all of the limitations. I found the generators to be quite limited, but was hoping they would have matured by now)
```

We used this a lot on a project at a previous work place - one big issue I found with it was lack of support for oneOf on response bodies when wanting to return entirely distinct responses for failure/error.  We've sidestepped that one here by using the problem-spring-web library.  Were there any other gotchas you remember?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ArrivalEntity.kt
brownbearsoftware: Ah hadn't realised the convention, since we've got some existing tables will leave it as-is for this PR then switch them all over with a migration in a follow-up PR.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/ArrivalTransformer.kt
brownbearsoftware: This is another case where generating the API models from the spec means we can't unfortunately.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/util/RandomData.kt
```diff
@@ -21,3 +24,9 @@ fun randomPostCode() = randomStringUpperCase(2) + randomNumberChars(1) + " " +
   randomStringUpperCase(2) + randomNumberChars(1)
 
 fun randomInt(min: Int, max: Int) = Random.nextInt(min, max)
+
+fun randomDateAfter(point: LocalDate): LocalDate = point.plusDays(randomInt(1, 14).toLong())
+fun randomDateBefore(point: LocalDate): LocalDate = point.minusDays(randomInt(1, 14).toLong())
+
+fun randomDateTimeAfter(point: OffsetDateTime): OffsetDateTime = point.plusMinutes(randomInt(1, 60 * 24 * 14).toLong()).truncatedTo(ChronoUnit.SECONDS)
+fun randomDateTimeBefore(point: OffsetDateTime): OffsetDateTime = point.minusMinutes(randomInt(1, 60 * 24 * 14).toLong()).truncatedTo(ChronoUnit.SECONDS)
```
mikehalmamoj, line 32: These feel like they would make nice extension functions.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ArrivalEntity.kt
matt-crowson-moj: Unless this has gone live do you need a migration to rename them?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BookingEntity.kt
```diff
@@ -0,0 +1,63 @@
...
+    var result = id.hashCode()
+    result = 31 * result + (person?.hashCode() ?: 0)
+    result = 31 * result + arrivalDate.hashCode()
+    result = 31 * result + departureDate.hashCode()
+    result = 31 * result + keyWorker.hashCode()
+    result = 31 * result + (arrival?.hashCode() ?: 0)
+    result = 31 * result + (departure?.hashCode() ?: 0)
+    result = 31 * result + (nonArrival?.hashCode() ?: 0)
+    result = 31 * result + (cancellation?.hashCode() ?: 0)
+    return result
```
andrewrlee, line 61: meh is it possible to use [the built in java method](https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#hash-java.lang.Object...-)

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/util/RandomData.kt
brownbearsoftware: Good shout, have switched them to extension functions

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/ArrivalTransformer.kt
brownbearsoftware: Have switched these to expression bodied for now, hands are a bit tied on adding a constructor to the API models since they're generated but will go for that approach if we switch away from Open API Generator

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ArrivalEntity.kt
brownbearsoftware: I mean that since we already have other tables from previous PRs that are named in the singular rather than the plural (and have been deployed to dev) it would make sense to move the entire project from singular->plural in one PR

matt-crowson-moj: Ah right sorry.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ArrivalEntity.kt
brownbearsoftware: This is a really good point, have changed the toString on them all

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BookingEntity.kt
brownbearsoftware: Indeed, I was lazy and let IntelliJ do these - have switched

matt-crowson-moj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ArrivalEntity.kt
brownbearsoftware: This was addressed in https://github.com/ministryofjustice/approved-premises-api/pull/30

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/repository/BookingRepository.kt
brownbearsoftware: This was addressed in https://github.com/ministryofjustice/approved-premises-api/pull/30

PR #34, opened 2022-08-12
Create booking

brownbearsoftware:
Implement create booking endpoint + tests

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/ExceptionHandling.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/ExceptionHandling.kt
@@ -1,15 +1,22 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.problem
 
-import org.springframework.security.access.AccessDeniedException
+import com.fasterxml.jackson.databind.exc.InvalidFormatException
+import org.springframework.http.HttpStatus
+import org.springframework.http.ResponseEntity
+import org.springframework.http.converter.HttpMessageNotReadableException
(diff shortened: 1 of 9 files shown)


brownbearsoftware:
[Add API endpoints for bookings](https://trello.com/c/z8TqoOuh/526-add-api-endpoints-for-bookings)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/ExceptionHandling.kt
```diff
@@ -21,6 +28,32 @@ class ExceptionHandling : ProblemHandling {
...
+  override fun handleMessageNotReadableException(
+    exception: HttpMessageNotReadableException,
+    request: NativeWebRequest
+  ): ResponseEntity<Problem> {
+    val responseBuilder = Problem.builder()
+      .withStatus(Status.BAD_REQUEST)
+      .withTitle("Bad Request")
+
+    if (exception.cause is InvalidFormatException) {
+      val invalidFormatException = exception.cause as InvalidFormatException
```
mikehalmamoj, line 41: I'm surprised you need this cast, I thought it would be smartcasted in the `if`.

mikehalmamoj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/ExceptionHandling.kt
brownbearsoftware: I think it's because smart casts only work on local variables or immutable members, otherwise you could get a race condition with another thread changing the value of cause

PR #39, opened 2022-08-16
546 - Create lost bed

brownbearsoftware:
Implement POST /premises/{premises id}/lost-bed endpoint + integration tests.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/LostBedsEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/LostBedsEntity.kt
@@ -1,0 +1,60 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity
+
+import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.stereotype.Repository
+import java.time.LocalDate
+import java.util.Objects
+import java.util.UUID
+import javax.persistence.Entity
+import javax.persistence.EnumType
+import javax.persistence.Enumerated
+import javax.persistence.Id
+import javax.persistence.JoinColumn
+import javax.persistence.OneToOne
+import javax.persistence.Table
+
+@Repository
+interface LostBedsRepository : JpaRepository<LostBedsEntity, UUID>
+
+@Entity
+@Table(name = "lost_beds")
+data class LostBedsEntity(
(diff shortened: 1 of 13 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/LostBedsEntity.kt
```diff
@@ -0,0 +1,60 @@
...
+    if (this === other) return true
+    if (other !is LostBedsEntity) return false
+
+    if (id != other.id) return false
+    if (startDate != other.startDate) return false
+    if (endDate != other.endDate) return false
+    if (numberOfBeds != other.numberOfBeds) return false
+    if (reason != other.reason) return false
+    if (referenceNumber != other.referenceNumber) return false
+    if (notes != other.notes) return false
```
pezholio, line 45: This seems a bit long-winded. Can we make hashes of the objects to compare and compare them that way?

pezholio (approved):
LGTM 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/LostBedsEntity.kt
brownbearsoftware: I don't think equals can rely on hashcode incase of a hash collision there still needs to be a way of deciding for definite if two objects are equal or hashmaps would break

brownbearsoftware: for ref, I didn't write this by hand - alt + insert -> `equals and hashcode()` on IntelliJ does them like this

pezholio: Ah, fair enough. Was just an idle thought 😃

pezholio: Oooh! Shiny!

brownbearsoftware:
[Write API endpoint for creating a lost bed](https://trello.com/c/0ItEq9Pf/546-write-api-endpoint-for-creating-a-lost-bed)

PR #40, opened 2022-08-16
Fix for ADR-0001

brownbearsoftware:
Change from map to array as specified in the ADR


pezholio (approved):
LGTM - just had to check it was the code that was wrong and not the ADR, but yes, this is  the type of response we expect in the frontend 👍

PR #41, opened 2022-08-16
(607) Add a cancellation reasons api endpoint

pezholio:
This adds an API endpoint for Cancellation Reasons, following the pattern we already have for reference data. My first Kotlin PR, so be kind! 😄


pezholio:
[Add a cancellation reasons API endpoint](https://trello.com/c/rLLiN6Ii/607-add-a-cancellation-reasons-api-endpoint)

brownbearsoftware (approved):
Spot on, nice

PR #43, opened 2022-08-16
606 - Get AP Capacity endpoint

brownbearsoftware:
Add endpoint to get AP capacity per day, from today until the furthest-in-the-future Booking or Lost Beds

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
@@ -5,6 +5,7 @@ import org.springframework.stereotype.Service
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.PremisesApiDelegate
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.health.api.model.Arrival
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.health.api.model.Booking
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.health.api.model.DateCapacity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.health.api.model.LostBed
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.health.api.model.NewArrival
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.health.api.model.NewBooking
@@ -23,6 +24,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.ArrivalTrans
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.BookingTransformer
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.LostBedsTransformer
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.PremisesTransformer
+import java.time.LocalDate
 import java.util.UUID
 
 @Service
@@ -149,5 +151,32 @@ class PremisesController(
     val lostBed = premisesService.createLostBeds(lostBedsTransformer.transformApiToJpa(body, premises))
 
     return ResponseEntity.ok(lostBedsTransformer.transformJpaToApi(lostBed))
+  }
+
+  override fun premisesPremisesIdCapacityGet(premisesId: UUID): ResponseEntity<List<DateCapacity>> {
+    val premises = premisesService.getPremises(premisesId)
+      ?: throw NotFoundProblem(premisesId, "Premises")
+
+    val lastBookingDate = premisesService.getLastBookingDate(premises)
+    val lastLostBedsDate = premisesService.getLastLostBedsDate(premises)
+
+    val capacityForPeriod = premisesService.getAvailabilityForRange(
+      premises,
+      LocalDate.now(),
+      maxOf(
+        LocalDate.now(),
+        lastBookingDate ?: LocalDate.now(),
+        lastLostBedsDate ?: LocalDate.now()
+      )
+    )
+
+    return ResponseEntity.ok(
+      capacityForPeriod.map {
+        DateCapacity(
+          date = it.key,
+          availableBeds = it.value.getFreeCapacity(premises.totalBeds)
+        )
+      }
+    )
   }
 }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BookingEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BookingEntity.kt
@@ -16,7 +16,10 @@ import javax.persistence.Table
 @Repository
 interface BookingRepository : JpaRepository<BookingEntity, UUID> {
   @Query("SELECT b FROM BookingEntity b WHERE b.premises.id = :premisesId AND b.arrivalDate <= :endDate AND b.departureDate >= :startDate")
-  fun findAllByPremisesIdAndOverlappingDate(premisesId: UUID, startDate: LocalDate, endDate: LocalDate): MutableList<BookingEntity>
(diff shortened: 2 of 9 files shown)


brownbearsoftware:
[Write API endpoint to get the capacity of an AP](https://trello.com/c/vHBidefy/606-write-api-endpoint-to-get-the-capacity-of-an-ap)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/Availability.kt
```diff
@@ -0,0 +1,12 @@
...
+import java.time.LocalDate
+
+data class Availability(
+  val date: LocalDate,
+  val pendingBookings: Int,
+  val arrivedBookings: Int,
+  val nonArrivedBookings: Int,
+  val cancelledBookings: Int,
+  val lostBeds: Int
+)
```
pezholio: I think we just said to return the `availableBeds` here? I think this info might be interesting though, so maybe return the `availableBeds` in addition to the count?

pezholio: Oh wait, I'm getting ahed of myself here - so to be clear, this is a data class that has all the different types of bookings, which we then sum when calculating returning in the endpoint?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
```diff
@@ -150,4 +152,31 @@ class PremisesController(
...
+    val lastLostBedsDate = premisesService.getLastLostBedsDate(premises)
+
+    val capacityForPeriod = premisesService.getAvailabilityForRange(
+      premises,
+      LocalDate.now(),
+      maxOf(
+        LocalDate.now(),
+        lastBookingDate ?: LocalDate.now(),
+        lastLostBedsDate ?: LocalDate.now()
+      )
```
pezholio, line 170: Love it!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/Availability.kt
```diff
@@ -9,4 +9,6 @@ data class Availability(
   val nonArrivedBookings: Int,
   val cancelledBookings: Int,
   val lostBeds: Int
-)
+) {
+  fun getFreeCapacity(totalBeds: Int) = totalBeds - pendingBookings - arrivedBookings - lostBeds
```
pezholio, line 13: Is it worth bracketing this to be clearer?

```suggestion
  fun getFreeCapacity(totalBeds: Int) = ((totalBeds - pendingBookings) - arrivedBookings) - lostBeds
```

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/CapacityTest.kt
```diff
@@ -0,0 +1,224 @@
...
+      .header("Authorization", "Bearer $jwt")
+      .exchange()
+      .expectStatus()
+      .isOk
+      .expectBodyList<DateCapacity>()
+      .hasSize(0)
+  }
+
+  @Test
+  fun `Get Capacity with booking in future returns OK with list entry for each day until the booking ends`() {
```
pezholio, line 105: I dunno if it'd be easier and quicker to break all these edge cases into unit tests, which would be cheaper / faster than having integration tests for all the possible cases. What do you think?

pezholio (commented):
Looks good from my POV. Would like some Kotlin folks to review it though. Might also be good to unit test `getAvailabilityForRange` to prevent having a load of integration tests.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/Availability.kt
brownbearsoftware:
```
Oh wait, I'm getting ahed of myself here - so to be clear, this is a data class that has all the different types of bookings, which we then sum when calculating returning in the endpoint?
```

that's right yeah, I suspect we'll need these counts elsewhere so figured the service layer should return them all then just transform it for the endpoint's response

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
brownbearsoftware: elvis operator up there with the best kotlin features :D

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/Availability.kt
brownbearsoftware: Have done

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/CapacityTest.kt
brownbearsoftware: Yeah agreed, I've removed some of the integration tests and added some unit ones

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
```diff
@@ -150,4 +152,31 @@ class PremisesController(
 
     return ResponseEntity.ok(lostBedsTransformer.transformJpaToApi(lostBed))
   }
+
+  override fun premisesPremisesIdCapacityGet(premisesId: UUID): ResponseEntity<List<DateCapacity>> {
+    val premises = premisesService.getPremises(premisesId)
```
andymarke, line 157: very very minor - I wonder if there is any point in letting the exception handler process these by throwing, since you are having to deal with ResponseEntity here anyway you could just do something like `return ResponseEntity.notFound(some error )`

Obviously you would have to restructure this sine you want to return early etc etc

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BookingEntity.kt
```diff
@@ -13,7 +14,13 @@ import javax.persistence.OneToOne
 import javax.persistence.Table
 
 @Repository
-interface BookingRepository : JpaRepository<BookingEntity, UUID>
+interface BookingRepository : JpaRepository<BookingEntity, UUID> {
+  @Query("SELECT b FROM BookingEntity b WHERE b.premises.id = :premisesId AND b.arrivalDate <= :endDate AND b.departureDate >= :startDate")
+  fun findAllByPremisesIdAndOverlappingDate(premisesId: UUID, startDate: LocalDate, endDate: LocalDate): MutableList<BookingEntity>
```
andymarke, line 19: just curious, why does this have to return a `MutableList` ?

andymarke (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
brownbearsoftware: So we could do but because we're returning an https://datatracker.ietf.org/doc/html/rfc7807 problem+json response we'd end up with quite a lot of duplication everywhere constructing the response bodies

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BookingEntity.kt
brownbearsoftware: Would seem it doesn't, have switched to List

andymarke (approved):
lgtm

PR #44, opened 2022-08-16
Use kfactories in unit tests

brownbearsoftware:
Changes to allow us to use the existing factories in unit tests by not providing a repository - i.e. just produce the object rather than also saving it to JPA.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/factory/ApAreaEntityFactory.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/factory/ApAreaEntityFactory.kt
@@ -8,8 +8,10 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.randomStringUpperCa
 import java.util.UUID
 
 class ApAreaEntityFactory(
-  apAreaTestRepository: ApAreaTestRepository
+  apAreaTestRepository: ApAreaTestRepository?
 ) : PersistedFactory<ApAreaEntity, UUID>(apAreaTestRepository) {
+  constructor() : this(null)
+
   private var id: Yielded<UUID> = { UUID.randomUUID() }
   private var name: Yielded<String> = { randomStringMultiCaseWithNumbers(8) }
   private var identifier: Yielded<String> = { randomStringUpperCase(3) }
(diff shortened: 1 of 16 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/factory/PersistedFactory.kt
```diff
@@ -3,7 +3,9 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.factory
 import io.github.bluegroundltd.kfactory.Factory
 import org.springframework.data.jpa.repository.JpaRepository
 
-abstract class PersistedFactory<EntityType : Any, PrimaryKeyType : Any>(private val repository: JpaRepository<EntityType, PrimaryKeyType>) : Factory<EntityType> {
-  fun produceAndPersist(): EntityType = repository.saveAndFlush(this.produce())
+abstract class PersistedFactory<EntityType : Any, PrimaryKeyType : Any>(private val repository: JpaRepository<EntityType, PrimaryKeyType>?) : Factory<EntityType> {
```
pezholio, line 6: If this is called `PersistedFactory`, is there an argument for having a `Factory` or `NonPersistedFactory` class instead?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/factory/ApAreaEntityFactory.kt
```diff
@@ -8,8 +8,10 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.randomStringUpperCa
 import java.util.UUID
 
 class ApAreaEntityFactory(
-  apAreaTestRepository: ApAreaTestRepository
+  apAreaTestRepository: ApAreaTestRepository?
 ) : PersistedFactory<ApAreaEntity, UUID>(apAreaTestRepository) {
```
pezholio, line 12: I dunno if this is legit Kotlin, but if we had two factory types could set the return type like this:

```suggestion
 ) : apAreaTestRepository ? PersistedFactory<ApAreaEntity, UUID>(apAreaTestRepository) : Factory<ApAreaEntity, UUID>(apAreaTestRepository) {
```

PR #47, opened 2022-08-17
549 - Add available beds for today field to Premises responses

brownbearsoftware:
Change get all premises and get premises by ID endpoints to include availableBedsForToday field in response


brownbearsoftware:
[Return the total number of available beds for a premises in the API](https://trello.com/c/pYxxzEHt/549-return-the-total-number-of-available-beds-for-a-premises-in-the-api)

pezholio (approved):
LGTM 👍

PR #49, opened 2022-08-17
599 - Extend Bookings

brownbearsoftware:
Add Extension tables to DB - extend the GET /premises/{id}/bookings endpoint to return the current extensions in the body.

Add POST /premises/{id}/bookings/{booking id}/extensions endpoint to create an extension:
 - Extension must be to a date later than the current departure date

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
@@ -8,10 +8,12 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.health.api.model.Arrival
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.health.api.model.Booking
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.health.api.model.Cancellation
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.health.api.model.DateCapacity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.health.api.model.Extension
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.health.api.model.LostBed
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.health.api.model.NewArrival
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.health.api.model.NewBooking
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.health.api.model.NewCancellation
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.health.api.model.NewExtension
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.health.api.model.NewLostBed
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.health.api.model.Premises
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ArrivalEntity
@@ -109,6 +111,7 @@ class PremisesController(
         departure = null,
         nonArrival = null,
         cancellation = null,
+        extensions = mutableListOf(),
         premises = premises
       )
     )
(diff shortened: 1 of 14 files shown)


pezholio:
fedb60362ab615557fc4e3f030486520d6a3d461 seems to be an empty commit BTW 😃

src/main/resources/static/mini-manage-api-stubs.yml
```diff
@@ -256,6 +256,59 @@ paths:
...
+          $ref: '#/components/responses/403Response'
+        404:
+          description: invalid premises ID or booking ID
+          content:
+            'application/json':
+              schema:
+                $ref: '#/components/schemas/Problem'
+        500:
+          $ref: '#/components/responses/500Response'
+      x-codegen-request-body-name: body
```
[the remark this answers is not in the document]
brownbearsoftware: Looks like it changes the generated parameters name: https://openapi-generator.tech/docs/swagger-codegen-migration/#body-parameter-name

Just added it for consistency as it was already on the others though

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BookingEntity.kt
```diff
@@ -30,7 +30,7 @@ data class BookingEntity(
   val id: UUID,
   var crn: String,
   val arrivalDate: LocalDate,
-  val departureDate: LocalDate,
+  var departureDate: LocalDate,
```
pezholio, line 33: What is the significance of `val` vs `var`? Is it similar to `let` vs `const` in JS?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
```diff
@@ -195,7 +198,35 @@ class PremisesController(
...
+        previousDepartureDate = booking.departureDate,
+        newDepartureDate = body.newDepartureDate,
+        notes = body.notes,
+        booking = booking
+      )
+    )
+
+    booking.departureDate = body.newDepartureDate
+    booking.extensions.add(extension)
+    bookingService.updateBooking(booking)
```
pezholio, line 227: Should we do this in a transaction, so if one query fails we don't end up with a weird state where a booking has an extension, but that departure date hasn't been changed or vice versa?

pezholio (approved):
LGTM - just a coupla questions and a point about transaction safety

src/main/resources/static/mini-manage-api-stubs.yml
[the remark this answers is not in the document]
pezholio: Cool 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BookingEntity.kt
brownbearsoftware: Yeah same concept as `let` vs `const` - `var` can be reassigned later whereas `val` cannot

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
brownbearsoftware: Good point, have ammended with a transaction

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
```diff
@@ -8,16 +8,31 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.BookingEntity
...
+  private val cancellationRepository: CancellationRepository,
+  private val extensionRepository: ExtensionRepository
 ) {
   fun createBooking(bookingEntity: BookingEntity): BookingEntity = bookingRepository.save(bookingEntity)
+  fun updateBooking(bookingEntity: BookingEntity): BookingEntity = bookingRepository.save(bookingEntity)
   fun getBooking(id: UUID) = bookingRepository.findByIdOrNull(id)
   fun createArrival(arrivalEntity: ArrivalEntity): ArrivalEntity = arrivalRepository.save(arrivalEntity)
   fun createCancellation(cancellationEntity: CancellationEntity): CancellationEntity = cancellationRepository.save(cancellationEntity)
+
+  @Transactional
```
pezholio, line 29: Oh wow! That's all you have to do? So it'll roll back if one or the other transaction fails?

brownbearsoftware: Yeah - Spring creates a proxy subclass of BookingService when it detects this annotation then that's what actually gets injected into PremisesController - so when you call this method it actually calls the subclass first which adds the transaction handling/recovery which in turn calls this method

pezholio: Nice!

pezholio (approved):

PR #53, opened 2022-08-22
597 - Search for Person by CRN

brownbearsoftware:
As this is the first time we've needed to call an upstream service (Community API), first laid some groundwork to make writing clients easier (namely BaseHMPPSClient) - this deals with fetching a client_credentials token from HMPPS Auth and also makes deserializing to a generic type (such as `List<T>`) easier.

The actual implementation for the search endpoint is relatively simple - perhaps of most note is the LAO (Limited Access Offenders) checks - if the response from Community API indicates that the CRN belongs to an LAO we then call the userAccess endpoint on Community API to see if this user is allowed to view the LAO.  If the JWT making the request was not generated for a user authenticated via delius then we return a 403 immediately since this is the only type of user that the LAO settings work with.

This PR also changes the security config to require ROLE_PROBATION for all protected endpoints.

**For this to work we will need a `client-id` and `client-secret` for each of the environments to be injected via environment variables.  We will also need to set the URLs for Community API - Stuart has requested the clients so this is underway.  The clients will need the `SCOPE_IGNORE_DELIUS_INCLUSIONS_ALWAYS` and `SCOPE_IGNORE_DELIUS_EXCLUSIONS_ALWAYS` authorities so that we can enforce the LAO restrictions ourselves**


src/main/resources/static/mini-manage-api-stubs.yml
```diff
@@ -468,6 +468,35 @@ paths:
         500:
           $ref: '#/components/responses/500Response'
       x-codegen-request-body-name: body
+  /crn/search:
```
pezholio, line 471: I think I'd prefer `/person/search` here, as the `person` is the entity we're searching for

src/main/resources/static/mini-manage-api-stubs.yml
```diff
@@ -1296,3 +1325,23 @@ components:
...
         - availableBeds
+    CrnSearch:
+      type: object
+      properties:
+        crn:
+          type: string
+          example: X320741
+      required:
+        - crn
+    CrnSearchResult:
```
pezholio, line 1336: This might be better expressed as a `person`. It also gives us flexibility to return more attributes later

src/main/resources/static/mini-manage-api-stubs.yml
```diff
@@ -468,6 +468,35 @@ paths:
         500:
           $ref: '#/components/responses/500Response'
       x-codegen-request-body-name: body
+  /crn/search:
+    post:
```
pezholio, line 472: I think as this is a search, I'd expect this to be a `GET` with a query string.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/OffenderService.kt
```diff
@@ -0,0 +1,32 @@
...
+
+import org.springframework.http.HttpStatus
+import org.springframework.stereotype.Service
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.ClientResult
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.CommunityApiClient
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.community.OffenderDetailSummary
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.ForbiddenProblem
+
+@Service
+class OffenderService(private val communityApiClient: CommunityApiClient) {
```
pezholio, line 11: I think I'd like to see this unit tested, so we can test all the edge cases at that point rather than having to maintain lots of expensive integration tests.

pezholio (commented):
Just a couple of comments. Looking good though!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/OffenderService.kt
brownbearsoftware: Have moved some of the edge cases to unit tests instead

docker-compose.yml
```diff
@@ -23,5 +23,26 @@ services:
...
+  community-api:
+    image: quay.io/hmpps/community-api:latest
+    container_name: community-api
+    ports:
+      - "9590:8080"
+      - "9592:9592"
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://localhost:8080/health"]
+    entrypoint: [
+        "java",
```
andymarke, line 35: as a matter of interest why do you need this entry pointy section? Is something missing from the community-api image?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/client/BaseHMPPSClient.kt
```diff
@@ -0,0 +1,116 @@
...
+import org.springframework.http.HttpHeaders
+import org.springframework.http.HttpMethod
+import org.springframework.http.HttpStatus
+import org.springframework.http.RequestEntity
+import org.springframework.web.client.RestTemplate
+import org.springframework.web.context.request.RequestContextHolder
+import org.springframework.web.context.request.ServletRequestAttributes
+import java.net.URI
+
+abstract class BaseHMPPSClient(
```
andymarke, line 14: Might be worth thinking why this is better than using different webClient configurations and injecting those into the service.

Most services would use the clientRegistrationId and spring config to control this for instance https://github.com/ministryofjustice/hmpps-prisoner-from-nomis-migration/blob/5eb6af449b1a9a9551f31e57db0c5e5490f03989/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/config/WebClientConfiguration.kt

docker-compose.yml
brownbearsoftware: This was just to allow me to attach a remote debugger to Community API as I was unsure of a few bits around LAO.  Just forgot to remove it when committing.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/client/HMPPSAuthClient.kt
```diff
@@ -0,0 +1,59 @@
...
+import org.springframework.stereotype.Component
+import org.springframework.web.client.RestTemplate
+import org.springframework.web.client.exchange
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.config.AuthAwareAuthenticationToken
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.hmppsauth.GetTokenResponse
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.InternalServerErrorProblem
+import java.net.URI
+
+@Component
+class HMPPSAuthClient(
```
andymarke, line 18: I'm trying to picture how this will work if you have multiple client credentials? e.g one for each downstream service

andymarke (approved):
LGTM

andymarke:
Are you sure you need to make that userAccess call, it looks to me the `/offenders/crn/{can}` endpoint is one of the few endpoints that does the userAccess stuff on your behalf

`    public OffenderDetailSummary getOffenderSummaryByCrn(final @PathVariable("crn") String crn, Authentication authentication) {
        userAccessService.checkExclusionsAndRestrictions(crn, authentication.getAuthorities());

        return offenderService.getOffenderSummaryByCrn(crn)
            .orElseThrow(() -> new NotFoundException(String.format("Offender with crn %s not found", crn)));
    }`

So it looks like the 2nd call is redundant. I guess you will have to make sure the end user's username is passed through on the token

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/client/HMPPSAuthClient.kt
brownbearsoftware: Is this a done thing in the HMPPS Auth world?  Guessing a least privilege thing?

andymarke: Yes, since different APIs require different roles and scopes. And roles are not (technically) unique across services. So it is best practice - HOWEVER IT IS AN UNWRITTEN BEST PRACTICE

docker-compose.yml
brownbearsoftware: Have removed this now

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/client/BaseHMPPSClient.kt
brownbearsoftware: Have switched over for consistency sake

brownbearsoftware:
[Fetch person by CRN in API](https://trello.com/c/NQrVjPxt/597-fetch-person-by-crn-in-api)

src/main/resources/static/mini-manage-api-stubs.yml
```diff
@@ -507,6 +507,35 @@ paths:
         500:
           $ref: '#/components/responses/500Response'
       x-codegen-request-body-name: body
+  /person/search:
```
pezholio, line 510: Can this be `/people/search` in the spirit of RESTfulness?

pezholio (approved):
LGTM 👍

src/main/resources/static/mini-manage-api-stubs.yml
brownbearsoftware: Will rectify this in a follow on PR once this and https://github.com/ministryofjustice/approved-premises-api/pull/63 are merged

src/main/resources/static/mini-manage-api-stubs.yml
```diff
@@ -507,6 +507,35 @@ paths:
...
+    get:
+      summary: Searches for a Person by their CRN
+      parameters:
+        - name: crn
+          in: query
+          description: CRN to search for
+          required: true
+          schema:
+            type: string
+      responses:
```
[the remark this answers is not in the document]
brownbearsoftware: Good point, have added

pezholio (approved):

PR #54, opened 2022-08-23
610 - Create departure

brownbearsoftware:
Adds a POST endpoint: /premises/{premises id}/booking/{booking id}/departures

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
@@ -8,11 +8,13 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.health.api.model.Arrival
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.health.api.model.Booking
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.health.api.model.Cancellation
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.health.api.model.DateCapacity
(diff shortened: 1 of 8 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
```diff
@@ -193,6 +205,65 @@ class PremisesController(
...
+    }
+
+    val destinationProvider = destinationProviderRepository.findByIdOrNull(body.destinationProviderId)
+    if (destinationProvider == null) {
+      validationIssues["destinationProviderId"] = "Destination Provider does not exist"
+    }
+
+    if (validationIssues.any()) {
+      throw BadRequestProblem(validationIssues)
+    }
```
pezholio, line 250: Just wondering if this can be broken out into the `createDeparture` service method, or something like `validateDeparture`. This can then be unit tested. I'm generally a proponent of the "Skinny Controller" approach, and I tend to get a bit nervous when I see too much logic in controllers

pezholio (approved):
LGTM - just some philosophical musings on code design. Happy to have a chat if you don't think it's the right way to do things, or if it diverges more from your usual way of doing things

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
brownbearsoftware: Will tackle this alongside the existing endpoints in a follow up PR

pezholio (approved):

PR #55, opened 2022-08-23
611 - Create Non Arrival for Booking

brownbearsoftware:
Implement create Non Arrival endpoint for Bookings + tests

Added own table for the reasons, similar to how Cancellations work

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
@@ -15,12 +15,16 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.health.api.model.NewBook
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.health.api.model.NewCancellation
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.health.api.model.NewExtension
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.health.api.model.NewLostBed
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.health.api.model.NewNonarrival
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.health.api.model.Nonarrival
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.health.api.model.Premises
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ArrivalEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.BookingEntity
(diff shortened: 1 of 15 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/factory/NonArrivalReasonEntityFactory.kt
```diff
@@ -0,0 +1,31 @@
...
+
+  fun withIsActive(isActive: Boolean) = apply {
+    this.isActive = { isActive }
+  }
+
+  override fun produce(): NonArrivalReasonEntity = NonArrivalReasonEntity(
+    id = this.id(),
+    name = this.name(),
+    isActive = this.isActive()
+  )
```
pezholio, line 30: Just thinking about these factories - is there a way we can provide defaults for some of the values? Would save a lot of the repetition in tests where we don't actually care about the values.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
```diff
@@ -175,6 +182,45 @@ class PremisesController(
...
+  ): ResponseEntity<Nonarrival> {
+    val premises = premisesService.getPremises(premisesId)
+      ?: throw NotFoundProblem(premisesId, "Premises")
+
+    val booking = bookingService.getBooking(bookingId)
+      ?: throw NotFoundProblem(bookingId, "Booking")
+
+    if (booking.premises.id != premises.id) {
+      throw NotFoundProblem(bookingId, "Booking")
+    }
```
pezholio, line 198: Again, as with #54, it'd be good to reuse this stuff

pezholio (approved):
LGTM - some similar questions to #54 though

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/factory/NonArrivalReasonEntityFactory.kt
brownbearsoftware: Yeah so the default values are set when we declare the variables at the top of the class
```
  private var id: Yielded<UUID> = { UUID.randomUUID() }
  private var name: Yielded<String> = { randomStringMultiCaseWithNumbers(6) }
  private var isActive: Yielded<Boolean> = { true }
```

So you don't need to specify any of them in this case if you don't want to, e.g. just do:

```
NonArrivalEntityFactory().produce()
```

which will give you a NonArrival with random id/name and isActive of true

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
brownbearsoftware: Will tackle this alongside the existing endpoints in a follow up PR

PR #57, opened 2022-08-24
Remove health in generated models

brownbearsoftware:


pezholio (dismissed):
Yay! This was bugging me too 😃

PR #59, opened 2022-08-24
Move validation from Controllers into Service Layer

brownbearsoftware:
Just wanting to get some early feedback on the pattern before proceeding.

Still need to do some unit tests for Departures/remove some integration tests + move the other types into service layer too.


pezholio (dismissed):
Yeah, this is a really nice approach. Happy with this 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/ValidatableActionResult.kt
```diff
@@ -0,0 +1,7 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.model
+
+sealed interface ValidatableActionResult<ResultType> {
+  data class Success<ResultType>(val result: ResultType) : ValidatableActionResult<ResultType>
```
pezholio, line 4: I think `entity` and `EntityType` would be better names here, i.e:

```suggestion
  data class Success<EntityType>(val entity: EntityType) : ValidatableActionResult<EntityType>
```

pezholio (dismissed):
Looks great, thanks so much for this 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
```diff
@@ -291,7 +291,7 @@ class PremisesController(
   }
 
   private fun <ResultType> extractResultEntityOrThrow(result: ValidatableActionResult<ResultType>) = when (result) {
-    is ValidatableActionResult.Success -> result.result
+    is ValidatableActionResult.Success -> result.entity
```
pezholio, line 294: Could you squash these commits into the others? I'm a big fan of https://github.com/tummychow/git-absorb for this. No worries if it causes a big mess.

pezholio (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
brownbearsoftware: sure thing, have sorted

pezholio (approved):

PR #66, opened 2022-09-12
704 - extend person details

brownbearsoftware:
Additional details added: date of birth, sex, status (in custody/in community), noms number, nationality, religion/belief, gender identity, prison name


pezholio (dismissed):
This is great work, thanks! Do we need to add the Prisons API secrets to the Helm setup before merging?

PR #71, opened 2022-09-13
717 - Get all applications for logged in user

brownbearsoftware:
**Implement GET /applications + tests**

Returns all of the applications for the current user.  Json schema validation will follow in another PR.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ApplicationsController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ApplicationsController.kt
@@ -1,0 +1,24 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.controller
+
+import org.springframework.http.ResponseEntity
+import org.springframework.stereotype.Service
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.ApplicationsApiDelegate
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Application
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.ApplicationService
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.HttpAuthService
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.ApplicationsTransformer
+
+@Service
+class ApplicationsController(
+  private val httpAuthService: HttpAuthService,
+  private val applicationService: ApplicationService,
+  private val applicationsTransformer: ApplicationsTransformer
+) : ApplicationsApiDelegate {
+  override fun applicationsGet(): ResponseEntity<List<Application>> {
+    val deliusPrincipal = httpAuthService.getDeliusPrincipalOrThrow()
+    val username = deliusPrincipal.name
+    val applications = applicationService.getAllApplicationsForUsername(username)
+
+    return ResponseEntity.ok(applications.map(applicationsTransformer::transformJpaToApi))
+  }
+}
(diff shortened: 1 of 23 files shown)


pezholio (dismissed):
LGTM - just a couple of minor things

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ProbationOfficerEntity.kt
```diff
@@ -0,0 +1,24 @@
...
+
+@Entity
+@Table(name = "probation_officers")
+data class ProbationOfficerEntity(
+  @Id
+  val id: UUID,
+  val name: String,
+  val distinguishedName: String,
+  var isActive: Boolean,
+  @OneToMany(mappedBy = "createdByProbationOfficer")
```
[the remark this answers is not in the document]
brownbearsoftware: Yeah can be used in other JPA models too

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ApplicationsController.kt
```diff
@@ -0,0 +1,23 @@
...
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.ApplicationsTransformer
+
+@Service
+class ApplicationsController(
+  private val httpAuthService: HttpAuthService,
+  private val applicationService: ApplicationService,
+  private val applicationsTransformer: ApplicationsTransformer
+) : ApplicationsApiDelegate {
+  override fun applicationsGet(): ResponseEntity<List<Application>> {
+    val userDistinguishedName = httpAuthService.getDeliusPrincipalOrThrow().name
```
pezholio, line 18: Can we have this set as a seperate `val` too, so:

```
val applications = applicationService.getAllApplications(deliusPrincipal.name)
```

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
```diff
@@ -84,9 +84,11 @@ class PremisesController(
     val premises = premisesService.getPremises(premisesId)
       ?: throw NotFoundProblem(premisesId, "Premises")
 
+    val deliusPrincipal = httpAuthService.getDeliusPrincipalOrThrow().name
```
pezholio, line 87: Again, I feel like this might be better fetching the principal and then getting the name. Even though I know it'll throw if it doesn't work, chaining methods like this makes me a bit nervous. Also means that if we get some strange behaviour, it's harder to track down.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ApplicationService.kt
```diff
@@ -0,0 +1,23 @@
...
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApplicationEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApplicationRepository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ProbationOfficerRepository
+
+@Service
+class ApplicationService(
+  private val probationOfficerRepository: ProbationOfficerRepository,
+  private val applicationRepository: ApplicationRepository
+) {
+  fun getAllApplications(userDistinguishedName: String): List<ApplicationEntity> {
```
pezholio, line 13: Could this be called `getAllApplicationsForUserName`? It's just clearer what we're doing then.

pezholio (approved):
LGTM - just another couple of things. Sorry!

PR #74, opened 2022-09-14
Add lost bed reasons

pezholio:
This adds the missing `reference-data/lost-bed-reasons` endpoint, as well as seed data to import the reasons.


brownbearsoftware (approved):
Spot on, LGTM

PR #80, opened 2022-09-15
436 Draft domain events

edavey:
In order to support the flow of data back to Delius the new AP service will
publish [Domain Events](https://dsdmoj.atlassian.net/wiki/spaces/NDSS/pages/3732570166/ADR+-+Domain+Event+Schema)

A key design point is that the Domain Event `messages` published via
SNS/SQS should be "skinny" and provide a link (`"detailUrl"`) to full
Event information exposed via our service's API.

In this PR we make a first attempt at designing AP Domain Events which 
can support Delius' needs. Our understanding is that the highest priority
events are:

1. `application.assessed` -> the decision on eligibility for an AP place must be recorded in Delius
2. `person.arrived` -> the address must be updated in Delius 
3. `person.not-arrived` -> the COM needs to be notified within Delius 
4. `person.departed` -> the address must be updated in Delius

We add the specification for the Domain Events API as a separate
OpenAPI spec as for now we don't want to include it in the `openApiGenerate`
code generation gradle task.

Trello: https://trello.com/c/yYzrHuuR/436-move-delius-features-to-new-ap-service

ap_domain_events

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/OAuth2ResourceServerSecurityConfiguration.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/OAuth2ResourceServerSecurityConfiguration.kt
@@ -40,6 +40,7 @@ class OAuth2ResourceServerSecurityConfiguration {
         authorize(HttpMethod.GET, "/swagger-ui/**", permitAll)
         authorize(HttpMethod.GET, "/v3/api-docs/swagger-config", permitAll)
         authorize(HttpMethod.GET, "/api.yml", permitAll)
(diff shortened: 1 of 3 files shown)


edavey:
@brownbearsoftware Could you take a quick look at this? I'd like to get it merged so that we can all view the swagger specs on the dev environment during our workshop on Weds. 

I've tweaked the swagger config to keep this draft spec separate from the active docs used for code generation:

```
springdoc:
  swagger-ui:
    urls:
      - name: API
        url: "api.yml"
      - name: Domain events
        url: "domain-events-api.yml"
```

brownbearsoftware (approved):
LGTM - I've not been through all the spec since that's what the meeting is for but doesn't look like it'll cause any issues being merged

PR #82, opened 2022-09-20
Return detailed problem responses when request body does not conform to OpenAPI spec

brownbearsoftware:
**Return more detailed problem responses when request bodies that don't conform to the OpenAPI spec are provided**

Jackson (the deserializer used by Spring) fails fast when deserializing into POJO, so it's only possible out of the box to provide the first error encountered during deserialization back to the caller.  This catches those exceptions then uses reflection to check each property and provide more detailed responses.  For this to work, we need to read the input stream more than once, so this also uses ContentCachingRequestWrapper to read the input stream once and cache it for reuse.

**Update existing validation responses to return a JSON Path parameter name**

Currently, all request bodies are simple non-nested objects, but its likely this won't always be the case.  The type validation code returns JSON Path names since it works to arbitrary levels so this needs to be consistent across the rest of the API.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -30,6 +30,8 @@ dependencies {
   implementation("org.springframework.boot:spring-boot-starter-oauth2-resource-server")
   implementation("org.springframework.boot:spring-boot-starter-oauth2-client")
 
+  implementation(kotlin("reflect"))
+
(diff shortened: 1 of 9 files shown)


pezholio (dismissed):
LGTM - took a bit of getting my head around it, but I think this will really help. Just a couple of clarity-related suggestions 👍

pezholio (approved):

PR #84, opened 2022-09-20
Handle 403 from Community API userAccess point correctly

brownbearsoftware:
Will now return a 403 rather than a 500 when LAO restricted offender requested


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/OffenderService.kt
```diff
@@ -41,6 +41,13 @@ class OffenderService(
       val access =
         when (val accessResponse = communityApiClient.getUserAccessForOffenderCrn(userDistinguishedName, crn)) {
           is ClientResult.Success -> accessResponse.body
+          is ClientResult.StatusCodeFailure -> {
+            if (accessResponse.status == HttpStatus.FORBIDDEN) {
+              return AuthorisableActionResult.Unauthorised()
```
pezholio, line 46: How do we differential from when a request is forbidden because there's something wrong with the the way the request has been formed and when a request is forbidden because there's we're requesting a limited access offender? Because I think when it's the former, a 500 is legit, because the problem is at the server end, but if it's the latter, then we recording a 403 is more appropriate.

brownbearsoftware: Good point, have just pushed a change to handle this case too

pezholio (approved):
Yeah, this is much better - thanks! 👍

PR #86, opened 2022-09-21
Rename NewBooking arrival and departure dates

pezholio:
This ensures they are consistent with the values in the actual Booking model


brownbearsoftware (approved):
LGTM

PR #87, opened 2022-09-21
717 - Validate json schema

brownbearsoftware:
**Validate against JSON schema when fetching applications**

When retrieving all JSON schemas for a user, test any that were created with an earlier version of the applications schema against the newest version of the application schema.  If they pass the validation, bump their version to the newest schema.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ApplicationService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ApplicationService.kt
@@ -8,16 +8,14 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ProbationOffi
 @Service
 class ApplicationService(
   private val probationOfficerRepository: ProbationOfficerRepository,
-  private val applicationRepository: ApplicationRepository
+  private val applicationRepository: ApplicationRepository,
+  private val jsonSchemaService: JsonSchemaService
 ) {
   fun getAllApplicationsForUsername(userDistinguishedName: String): List<ApplicationEntity> {
     val probationOfficerEntity = probationOfficerRepository.findByDistinguishedName(userDistinguishedName)
       ?: return emptyList()
 
     return applicationRepository.findAllByCreatedByProbationOfficer_Id(probationOfficerEntity.id)
-      .map {
-        // TODO: For any entries where schema version is lower than newest - validate against newest and promote if possible
-        it.apply { schemaUpToDate = true }
-      }
+      .map(jsonSchemaService::attemptSchemaUpgrade)
   }
 }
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ApplicationService.kt
```diff
@@ -3,21 +3,40 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.service
...
+        return application.apply { schemaUpToDate = false }
       }
+
+      application.schemaVersion = newestSchema
+      applicationRepository.save(application)
+    }
+
+    application.schemaUpToDate = true
+
+    return application
```
pezholio, line 40: I wonder if this belongs in the json Schema Service?

pezholio (approved):
LGTM 👍

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/JsonSchemaServiceTest.kt
```diff
@@ -0,0 +1,108 @@
...
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.factory.ApplicationEntityFactory
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.factory.ApplicationSchemaEntityFactory
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.factory.ProbationOfficerEntityFactory
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApplicationEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApplicationRepository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApplicationSchemaRepository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.JsonSchemaService
+import java.util.UUID
+
+class JsonSchemaServiceTest {
```
pezholio, line 17: Was there a test for `validate` before? Would be good to have both

brownbearsoftware: So this isn't actually called by anything other than `attemptSchemaUpgrade` - had meant to make it private hence the lack of tests.  Have made private now 👍

pezholio: Ah, gotcha. I imagine it'll be used elsewhere at some point though right?

pezholio (approved):

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/JsonSchemaServiceTest.kt
brownbearsoftware: Yeah will need it during creation, will pop the unit tests in for it then

PR #89, opened 2022-09-22
718 - Get single application

brownbearsoftware:
**Implement GET /applications/{id} endpoint**

User must be the creator of the application to retrieve it


pezholio:
[Get an application](https://trello.com/c/wBicNNOQ/718-get-an-application)

pezholio (approved):
Lovely stuff, looks great 👍

PR #91, opened 2022-09-22
752 - Create new application

brownbearsoftware:
**Implement POST /applications endpoint**

This will automatically create an entry for the Probation Officer if one does not already exist.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ProbationOfficerService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ProbationOfficerService.kt
@@ -1,0 +1,28 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.service
+
+import org.springframework.stereotype.Service
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ProbationOfficerEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ProbationOfficerRepository
+import java.util.UUID
+
+@Service
+class ProbationOfficerService(
+  private val httpAuthService: HttpAuthService,
+  private val probationOfficerRepository: ProbationOfficerRepository
+) {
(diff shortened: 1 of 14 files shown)


pezholio:
[Create an application](https://trello.com/c/f5oMzilC/752-create-an-application)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ProbationOfficerService.kt
```diff
@@ -0,0 +1,28 @@
...
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ProbationOfficerEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ProbationOfficerRepository
+import java.util.UUID
+
+@Service
+class ProbationOfficerService(
+  private val httpAuthService: HttpAuthService,
+  private val probationOfficerRepository: ProbationOfficerRepository
+) {
+  fun getProbationOfficerForRequestUser(): ProbationOfficerEntity {
```
pezholio, line 13: Is this something we can use elsewhere, would prevent a lot of calls to the API and ensure we have a source of truth in our database that we use everywhere

pezholio (approved):
LGTM 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ProbationOfficerService.kt
brownbearsoftware: What else do you think it would be useful for?  I had thought this might potentially be superseded by the outcome of:  https://trello.com/c/0PoVG3S2/710-clarify-source-of-ap-team-data-supervising-teams-officers-and-keyworkers-needed-when-recording-arrivals

pezholio: Yeah, you might be right. I guess going forward there's no point in having two repositories for the same data, especially where things like name etc may change

PR #92, opened 2022-09-22
Use new cloud platform namespace

pezholio:
Since we updated the repo name, the deploys have been failing, rather than put in a bunch of fixes that will hang around forever, I thought it best to create a whole new namespace that fits the naming conventions and allows us to host both services in the same namespace. I’ve copied the secrets over from our old namespace to the new one too.


pezholio:
Keeping this as draft until all the pieces are in place

brownbearsoftware (approved):
LGTM

PR #94, opened 2022-09-26
Validation errors refactor

brownbearsoftware:
**Added value type for ValidationErrors**

To make it clear that the MutableMap<String, String> we use in quite a lot of places are for validation errors

**Added a DSL for validation**

Reduces boilerplate around validation somewhat and provides a more semantic syntax.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/ValidationErrors.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/ValidationErrors.kt
@@ -5,4 +5,19 @@ value class ValidationErrors(private val errorMap: MutableMap<String, String>) :
   constructor() : this(mutableMapOf())
 }
 
-fun singleValidationErrorOf(propertyNameToMessage: Pair<String, String>) = ValidationErrors().apply { this[propertyNameToMessage.first] = propertyNameToMessage.second }
+private fun singleValidationErrorOf(propertyNameToMessage: Pair<String, String>) = ValidationErrors().apply { this[propertyNameToMessage.first] = propertyNameToMessage.second }
+
+class ValidatedScope<EntityType> {
+  val validationErrors = ValidationErrors()
+
+  val fieldValidationError: ValidatableActionResult.FieldValidationError<EntityType> = ValidatableActionResult.FieldValidationError(validationErrors)
+
+  infix fun success(entity: EntityType) = ValidatableActionResult.Success(entity)
+  infix fun generalError(message: String) = ValidatableActionResult.GeneralValidationError<EntityType>(message)
+  infix fun String.hasValidationError(message: String) = validationErrors.put(this, message)
+  infix fun String.hasSingleValidationError(message: String) = ValidatableActionResult.FieldValidationError<EntityType>(singleValidationErrorOf(this to message))
+}
+
+inline fun <EntityType> validated(scope: ValidatedScope<EntityType>.() -> ValidatableActionResult<EntityType>): ValidatableActionResult<EntityType> {
+  return scope(ValidatedScope())
+}
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/ValidationErrors.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/ValidationErrors.kt
@@ -1,3 +1,8 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.model
 
-typealias ValidationErrors = MutableMap<String, String>
+@JvmInline
+value class ValidationErrors(private val errorMap: MutableMap<String, String>) : MutableMap<String, String> by errorMap {
+  constructor() : this(mutableMapOf())
+}
+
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ApplicationService.kt
```diff
@@ -8,7 +8,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ProbationOffi
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.AuthorisableActionResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.ValidatableActionResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.ValidationErrors
-import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.singleValidationErrorOf
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.validated
```
pezholio, line 11: Am I being dim? Where is this model defined?

pezholio: Yes I am 😆

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/ValidationErrors.kt
```diff
@@ -6,3 +6,21 @@ value class ValidationErrors(private val errorMap: MutableMap<String, String>) :
...
+
+  val validationErrorsPresent: Boolean
+    get() = validationErrors.any()
+
+  val accumulatedFieldValidationError: ValidatableActionResult.FieldValidationError<EntityType> = ValidatableActionResult.FieldValidationError(validationErrors)
+
+  infix fun success(entity: EntityType) = ValidatableActionResult.Success(entity)
+  infix fun generalError(message: String) = ValidatableActionResult.GeneralValidationError<EntityType>(message)
+  infix fun String.hasValidationError(message: String) = validationErrors.put(this, message)
+  infix fun String.hasSingleValidationError(message: String) = ValidatableActionResult.FieldValidationError<EntityType>(singleValidationErrorOf(this to message))
```
pezholio, line 21: Do we even need `this singleValidationErrorOf` now? As we're initializing an empty `validationErrors` in the scope?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/ValidationErrors.kt
```diff
@@ -6,3 +6,21 @@ value class ValidationErrors(private val errorMap: MutableMap<String, String>) :
...
 
 fun singleValidationErrorOf(propertyNameToMessage: Pair<String, String>) = ValidationErrors().apply { this[propertyNameToMessage.first] = propertyNameToMessage.second }
+
+class ValidatedScope<EntityType> {
+  private val validationErrors = ValidationErrors()
+
+  val validationErrorsPresent: Boolean
+    get() = validationErrors.any()
+
+  val accumulatedFieldValidationError: ValidatableActionResult.FieldValidationError<EntityType> = ValidatableActionResult.FieldValidationError(validationErrors)
```
pezholio, line 16: I feel this naming could be simplified to just `fieldValidationError`

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ApplicationService.kt
brownbearsoftware: This is one of the reasons I'm on the fence about whether this is better than after the first commit.  Perhaps a bit too confusing at first glance?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/ValidationErrors.kt
```diff
@@ -6,3 +6,21 @@ value class ValidationErrors(private val errorMap: MutableMap<String, String>) :
 }
 
 fun singleValidationErrorOf(propertyNameToMessage: Pair<String, String>) = ValidationErrors().apply { this[propertyNameToMessage.first] = propertyNameToMessage.second }
+
+class ValidatedScope<EntityType> {
+  private val validationErrors = ValidationErrors()
+
+  val validationErrorsPresent: Boolean
+    get() = validationErrors.any()
```
pezholio, line 14: Can we get the validation errors directly from the scope instead of calling `validationErrorsPresent`. Calling `validationErrors.any()` directly feels more expressive to me and cuts down on boilerplate

pezholio (approved):
The more I look at this, the more I like it! A couple of suggestions to aid clarity, but on the whole LGTM 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/ValidationErrors.kt
brownbearsoftware: It's still used in this file for `hasSingleValidationError`, it can't be succintly inlined though because of the apply call vs the receiver function in the infix both being a `this` so think it's clearest to keep it like this.  Will make it `private` though.

PR #95, opened 2022-09-27
Feature/deploy to temporary accommodation namespace

tahb:
WIP

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -60,6 +60,28 @@ workflows:
 #            - approved-premises-api-prod
 #          requires:
 #            - request-prod-approval
+  temporary-accommodation-build-test-and-deploy:
+    jobs:
+      - hmpps/helm_lint:
+          name: helm_lint
+      - hmpps/build_multiplatform_docker:
+          name: build_docker
+          filters:
+            branches:
+              only:
+                - main
+      - hmpps/deploy_env:
+          name: deploy_dev
+          env: "dev"
+          context: hmpps-common-vars
+          chart_name: hmpps-temporary-accommodation-api
+          filters:
+            branches:
+              only:
+                - main
+          requires:
+            - build_docker
+            - helm_lint
 
   security:
     triggers:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -60,6 +60,28 @@ workflows:
...
+          name: build_docker
+          filters:
+            branches:
+              only:
+                - main
+      - hmpps/deploy_env:
+          name: deploy_dev
+          env: "dev"
+          context: hmpps-common-vars
+          chart_name: hmpps-temporary-accommodation-api
```
tahb, line 77: We hope that passing the `chart_name` here is enough to influence [the following orb script](https://github.com/ministryofjustice/hmpps-circleci-orb/blob/145a06ecbeeb3fb80de0dc6a79dd2fc07b307e24/src/scripts/deploy_env.sh#L22) which invokes Helm?

helm_deploy/hmpps-temporary-accommodation-api/values.yaml
```diff
@@ -0,0 +1,60 @@
...
+    SPRING_JPA_DATABASE: postgresql
+    SPRING_DATASOURCE_URL: "jdbc:postgresql://${DB_HOST}/${DB_NAME}"
+
+  # Pre-existing kubernetes secrets to load as environment variables in the deployment.
+  # namespace_secrets:
+  #   [name of kubernetes secret]:
+  #     [name of environment variable as seen by app]: [key of kubernetes secret to load]
+
+  namespace_secrets:
+    hmpps-temporary-accommodation-api:
```
tahb, line 35: We copied this file and renamed every instance of `hmpps-approved-premises-api` with `hmpps-temporary-accommodation-api`. We aren't super confident about this so would appreciate another pair of eyes. Should any of these values reflect the repository name instead?

tahb:
Haven't forgotton about this. We may still need to do this but for now we're working on a release branch until we get confirmation on our approach.

PR #96, opened 2022-09-27
Config flag to log authorities & scopes

brownbearsoftware:
Add configuration option `log-client-credentials-jwt-info` which logs the authorities, scope and expiry of client_credentials tokens retrieved from HMPPS Auth.

This is useful for troubleshooting as we have no way currently of knowing for certain which authorities/scopes our service accounts have been given.


edavey (approved):
this will be super-useful!

this

PR #98, opened 2022-09-27
Use keys for error messages

pezholio:
Rather than using sentences for error messages, this updates the error response `errorType`s  to return keys. This makes translating messages in the frontend easier, and gives us more flexibility.


brownbearsoftware (approved):
LGTM

PR #99, opened 2022-09-28
Return accurate error for special string types

brownbearsoftware:
**Return accurate error when failing to deserialize special string types**

LocalDate, LocalDateTime, OffsetDateTime and UUID are all represented as strings in JSON so their deserialization process is different to the other types - handle this case by attempting to parse the values and returning "invalid" if a non-parsable value has been provided.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/DeserializationValidationService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/DeserializationValidationService.kt
@@ -10,6 +10,7 @@ import java.math.BigDecimal
 import java.time.LocalDate
 import java.time.LocalDateTime
 import java.time.OffsetDateTime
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/DeserializationValidationService.kt
```diff
@@ -113,3 +135,46 @@ class DeserializationValidationService {
...
+    return doesNotThrow { UUID.fromString(jsonNode.textValue()) }
+  }
+}
+
+private fun doesNotThrow(block: () -> Unit) = try {
+  block()
+  true
+} catch (e: Exception) {
+  false
+}
```
pezholio, line 180: Ah, this is a lovely neat thing!

pezholio (approved):
LGTM - thanks for doing this 👍

PR #102, opened 2022-09-29
(762) Add person detail to Application responses

pezholio:
This adds the detail of a specific person to all of the Application responses. This involves us having to call the Community and Prisons APIs, so we have guards in place for if either API has problems returning the person detail(s).

This is done using a private method in the controller, the implementation of which is nicked from the BookingsController. I do wonder if this would be better placed in a service, but I think this is fine for now.

I’ve also added the `mockInmateDetailPrisonsApiCall` function, which is used in some other integration tests to the IntegrationTestBase file, so we can use it everywhere.


pezholio:
[Get person details in an application](https://trello.com/c/fLmjmQe0/762-get-person-details-in-an-application)

brownbearsoftware (approved):
LGTM, good stuff.

One thing, not specific to this PR but this brought it to the front of my mind again, we should think about is a nicer place of putting the common mock http call functions.  Perhaps a file per Client or something like that?

pezholio:
Yeah, that's a good call. Would be a neater place to put it

PR #103, opened 2022-09-29
CAS3-396: create new schemas for persisting & returning a new cas3 pr…

ian-wells-nimble:
Defining the schema for a new cas3 premises

### DIFF
--- a/src/main/resources/static/api.yml
+++ b/src/main/resources/static/api.yml
@@ -5,6 +5,37 @@ info:
 servers:
 - url: /
 paths:
+  /cas3_premises:
+    post:
+      tags:
+        - Operations on cas3 premises
+      summary: Adds a new premises
+      requestBody:
+        content:
+          'application/json':
+            schema:
+              $ref: '#/components/schemas/NewCas3Premises'
+        required: true
+      responses:
+        201:
+          description: new premises added
+          content:
+            'application/json':
+              schema:
+                $ref: '#/components/schemas/Cas3Premises'
+        400:
+          description: invalid request
+          content:
+            'application/problem+json':
+              schema:
+                $ref: '#/components/schemas/ValidationError'
+        401:
+          $ref: '#/components/responses/401Response'
+        403:
+          $ref: '#/components/responses/403Response'
+        500:
+          $ref: '#/components/responses/500Response'
+      x-codegen-request-body-name: body
   /premises:
     get:
       tags:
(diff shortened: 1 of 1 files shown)


ian-wells-nimble:
> I think it would be good if this was more generic, and we had some way of differentiating between the type of premises (so a `type` field or similar). I feel a bit uneasy about adding new endpoints that are specifically for a given service.

Fair point yeah. I'll leave this here for discussion as James wants to chat about it tomorrow also

tahb:
I expect the shape of a premise resource to stay broadly the same. There may be some data changes and/or divergent behaviour in the future. One way we could control those is as Stu says with a new piece of state like `type`. 

As a general principle of our collaboration on the API I think we'd do well to take a default stance of sharing over separation. When we find that to be too hard then we still have the option to separate. This situation is a good example, CAS3 need the POST Premise endpoint now but we know that CAS1 have that same feature on their roadmap.

It would be good to understand any concerns that drive this separation as we might be able to factor those in to the work we need to do to support this.

Floppy (changes requested):
I agree that in general we should keep the same endpoints unless we *have to* split.

I agree that a `type` parameter  (or maybe `category`, computers get sniffy about the word "type") would be a good way of separating the different premises. We could add that to the CAS1 database diagram at https://dbdiagram.io/d/6200fa0285022f4ee54d7569 probably?

The other feedback I have is on the address field. The County/Town/Address split only comes from one bed tracker, it's not a general thing. The areas are all doing it different ways, so we can think about the _right_ way instead of using _their_ way. The current CAS1 diagram uses a single string field for address, and I think this is fine for now - we can always split it up later on based on identified needs.

brownbearsoftware:
```
We could add that to the CAS1 database diagram at https://dbdiagram.io/d/6200fa0285022f4ee54d7569 probably?
```

This link is from Discovery I think but it's not actually reflective of the current state of the schema (e.g. one notable discrepency is that schema has individual beds modelled whereas due to us deciding to go for mini-manage first we just have a bed count on the premises table.)

For reference, here's an entity-relation diagram of the current state of the schema generated by DBeaver:

er

```
I agree that in general we should keep the same endpoints unless we have to split.
```

Is there an overview of the MVP requirements for CAS3 somewhere?  If we're going to share the repo it's going to be pertinent that both teams are aware of the requirements for the other team when adding/changing anything so I'm just wondering what this would look like from a process perspective?

edavey:
what kind of information do you plan to represent in `CAS3Premises.type`? 

And might that be better suited to some other (join) table e.g. `cas3_premises_types` or more generally something like `premises_properties`? -- but then we really are jumping with 2 feet into a general purpose design...

edavey:
I suggest we try to make this work on the existing `premises` table and see what impact this has on our code...

tahb:
From what I understand, Ian [forked off](https://github.com/ian-wells-nimble/hmpps-approved-premises-api/pull/1), we got some more input from Joe and Ed and we got this work running locally. Our next steps are to get this work back into this repository in our `temporary-accommodation-main` branch. 

I suspect this can probably be closed?

PR #105, opened 2022-09-30
710 - Get staff endpoint

brownbearsoftware:
Add GET /premises/{id}/staff endpoint


brownbearsoftware:
[Changes to how Key Workers are handled](https://trello.com/c/0PoVG3S2/710-changes-to-how-key-workers-are-handled)

edavey (approved):
very clear and looks correct to me, this

PR #106, opened 2022-10-03
757 - Get case notes

brownbearsoftware:
Implement GET /people/{crn}/prison-case-notes endpoint

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/prisonsapi/CaseNotes.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/prisonsapi/CaseNotes.kt
@@ -1,0 +1,34 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.model.prisonsapi
+
+import java.time.LocalDateTime
+
+data class CaseNotesPage(
+  val totalElements: Int,
+  val totalPages: Int,
+  val number: Int,
+  val content: List<CaseNote>
+)
+
+data class CaseNote(
+  val caseNoteId: Long,
+  val bookingId: Long,
+  val type: String,
+  val typeDescription: String?,
+  val subType: String,
+  val subTypeDescription: String?,
+  val source: String?,
+  val creationDateTime: LocalDateTime,
+  val occurrenceDateTime: LocalDateTime,
+  val staffId: Long,
+  val authorName: String,
+  val text: String,
+  val originalNoteText: String,
+  val agencyId: String?,
+  val amendments: List<CaseNoteAmendment>
+)
+
+data class CaseNoteAmendment(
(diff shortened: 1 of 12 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/prisonsapi/CaseNotes.kt
```diff
@@ -0,0 +1,34 @@
...
+  val text: String,
+  val originalNoteText: String,
+  val agencyId: String?,
+  val amendments: List<CaseNoteAmendment>
+)
+
+data class CaseNoteAmendment(
+  val creationDateTime: LocalDateTime,
+  val authorName: String,
+  val additionalNoteText: String
```
edavey, line 33: nice, I haven't seen an example of this so far in my travels but thanks for digging out the necessary schema so we're ready...

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/OffenderServiceTest.kt
```diff
@@ -407,6 +419,91 @@ class OffenderServiceTest {
...
+        nomsNumber = nomsNumber,
+        from = LocalDate.now().minusDays(prisonCaseNotesConfigBindingModel.lookbackDays!!.toLong()),
+        page = 1
+      )
+    } returns ClientResult.Success(
+      HttpStatus.OK,
+      CaseNotesPage(
+        totalElements = 4,
+        totalPages = 2,
+        number = 1,
```
edavey, line 495: is `number` always going to be `1`, until our API implements implements its own pagination?

As far as I understand, the consumer supplies only a noms number, and our API traverses all the PrisonAPI pages necessary to collect all of the case notes for the configured time period, and our API returns all the case notes in a single response. If so should we make it clearer in our tests what `number` is and how it behaves?

I wonder if -- in order to retrieve the contents of multiple pages from the Prison API in a timely fashion -- we'll need to experiment with different page sizes? i.e. we ought to add a third config element to `prison-case-notes` in `application.yml`?

edavey (approved):
👍 just one question about the page `number` property which wasn't very clear to me...

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/OffenderServiceTest.kt
brownbearsoftware: This should have been 2 rather than 1 (have updated.)

But yes you are correct, our endpoint just takes noms number - then we follow the pages on Prison API until we've got all of the data.

Good shout on making the page size configurable - I've added that too

PR #107, opened 2022-10-04
765 - Redis caching for Staff and Offender data

brownbearsoftware:
**Add a distributed cache backed by Redis that sits over the Client layer**

The ClientResult values are serialized and stored in Redis.  The cache name is prefixed with the version of the application so that we don't have any issues when deploying a new version where the models we're deserializing into may have changed etc.

 - Add a cache over Staff info for 6 hours
 - Add a cache over Offender Details (Community API), User Access (Community API), Inmate Details (Prisons API) for 20 minutes

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/RedisConfiguration.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/RedisConfiguration.kt
@@ -2,6 +2,7 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.config
 
 import com.fasterxml.jackson.core.type.TypeReference
 import com.fasterxml.jackson.databind.ObjectMapper
+import org.springframework.beans.factory.annotation.Value
 import org.springframework.boot.autoconfigure.cache.RedisCacheManagerBuilderCustomizer
 import org.springframework.boot.info.BuildProperties
 import org.springframework.cache.annotation.EnableCaching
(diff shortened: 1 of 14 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/RedisConfiguration.kt
```diff
@@ -1,16 +1,138 @@
...
+        )
+      }
+      is ClientResult.Success -> {
+        SerializableClientResult(
+          discriminator = ClientResultDiscriminator.SUCCESS,
+          status = clientResult.status,
+          body = objectMapper.writeValueAsString(clientResult.body),
+          exceptionMessage = null,
+          type = clientResult.body::class.java.typeName,
+          method = null,
```
edavey, line 88: are these `discriminator` values the key to whether the call to the cache is a hit or a miss?

Ought this logic to be described in a unit test?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/RedisConfiguration.kt
```diff
@@ -31,6 +34,9 @@ class RedisConfiguration {
     return RedisCacheManagerBuilderCustomizer { builder: RedisCacheManagerBuilder ->
       builder.clientCacheFor<List<StaffMember>>("staffMembersCache", Duration.ofHours(6), version, objectMapper)
         .clientCacheFor<StaffMember>("staffMemberCache", Duration.ofHours(6), version, objectMapper)
+        .clientCacheFor<OffenderDetailSummary>("offenderDetailsCache", Duration.ofMinutes(20), version, objectMapper)
+        .clientCacheFor<UserOffenderAccess>("userAccessCache", Duration.ofMinutes(20), version, objectMapper)
+        .clientCacheFor<InmateDetail>("inmateDetailsCache", Duration.ofMinutes(20), version, objectMapper)
```
edavey, line 39: should these expiry values be defined in a config file perhaps?

edavey (approved):
Nice work! There's a lot of magic going on with the `RedisCacheManagerBuilderCustomizer` but I'm just about following along. Two questions:

- do we need unit testing of the `ClientResultRedisSerializer` behaviour?
- should the various cache expiry times (6 hrs and 20 mins) be held separately in config? (assuming that we don't want unit test assertions that our cache manage builder instance has been called with a duration of 20 mins for the cache named 'userAccess' etc? )

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/RedisConfiguration.kt
brownbearsoftware: These identify the type of ClientResult that's being cached:
 - Success (HTTP response was 2xx response)
 - StatusCodeFailure (HTTP response was 4xx, 5xx)
 - OtherFailure (Broken pipe, Unknown host etc.)

This allows us to construct the correct implementation of ClientResult fom what's stored in Redis when the cache is hit.  Doing this lets us continue to write code as though the cache isn't even there - it appears to the Client call-sites that the request has been made.

The actual logic of hit/miss/TTL etc. is handled by Spring Cache so nothing for us to do there.

brownbearsoftware: I have added some unit tests around this though

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/RedisConfiguration.kt
brownbearsoftware: Have extracted into properties

brownbearsoftware:
[Redis caching for Staff Members and Offenders](https://trello.com/c/ZQUBkeqO/765-redis-caching-for-staff-members-and-offenders)

PR #127, opened 2022-10-06
Add Tilt for running UI and API together

brownbearsoftware:
Login with: JimSnowLdap & secret

Usable CRN: X320741

----

To start run `tilt up` - this starts the docker-compose.yml up with some services not started.

Need to run the api project locally.

Go to http://localhost:10350/ for the Tilt Web UI where you can enable or disable services.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,5 +1,52 @@
 version: "3"
 services:
+  frontend:
+    image: quay.io/hmpps/hmpps-approved-premises-ui:latest
+    container_name: approved-premises-ui-dev
+    environment:
+      - REDIS_HOST=approved-premises-redis-dev
+      - REDIS_PORT=6379
+      - REDIS_TLS_ENABLED=false
+      - REDIS_AUTH_TOKEN=
+      - HMPPS_AUTH_EXTERNAL_URL=http://localhost:9091/auth
+      - HMPPS_AUTH_URL=http://hmpps-auth:8080/auth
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -1,5 +1,52 @@
...
+        - SYSTEM_CLIENT_SECRET=clientsecret
+        - SESSION_SECRET=app-insecure-default-session
+        - NODE_ENV=development
+        - TOKEN_VERIFICATION_API_URL=http://hmpps-auth:8080/verification
+        - APPROVED_PREMISES_API_URL=http://host.docker.internal:8080
+        - INGRESS_URL=http://localhost:3000
+      ports:
+        - 3000:3000
+        - 9229:9229
+      entrypoint: "node --inspect-brk=0.0.0.0 dist/server.js | bunyan"
```
edavey, line 48: I think this either needs to be commented out or shifted right if we want to run this container also?

brownbearsoftware: right you are!  Not sure what happened there, have sorted

brownbearsoftware:
[Add Tilt to UI and API repos so whole system can be run together locally](https://trello.com/c/CtLyRDld/769-add-tilt-to-ui-and-api-repos-so-whole-system-can-be-run-together-locally)

pezholio:
Gonna close this, as it's now in https://github.com/ministryofjustice/hmpps-approved-premises-tools

PR #142, opened 2022-10-11
Fix/docker container errors in dev

tahb:
These changes are the ones I found were needed to get both apps running using docker-compose. [There is a similar pull request on the frontend](https://github.com/ministryofjustice/hmpps-approved-premises-ui/pull/237).

We've got:

- removing container names when not needed to remove name clashes with the frontend
- exposing redis on a non-default port to avoid clashing with the frontend (that still uses the default)
- an addition to the script/server to shut any old containers down when starting up. The script can leave things hanging when we term the process. We found maing changes to the docker-compose and restarting the server was clashing with existing containers running without this

With these changes and a similar set on the frontend both apps work together when running `script/server` which is nice. I'm aware you're experimenting with Tilt right now however I think these changes are still useful.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,7 +2,6 @@ version: "3"
 services:
   database:
     image: "postgres"
-    container_name: approved-premises-postgres-dev
     environment:
       - POSTGRES_USER=localdev
       - POSTGRES_PASSWORD=localdev_password
@@ -14,7 +13,6 @@ services:
 
   hmpps-auth:
     image: quay.io/hmpps/hmpps-auth:latest
-    container_name: hmpps-auth
     ports:
       - "9091:8080"
     healthcheck:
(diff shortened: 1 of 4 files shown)


tahb:
Though I think I've changed the redis port I'm not sure what features/test coverage there is that this Redis is still functioning. Could use a double check from someone in the know please 🙏

docker-compose.yml
```diff
@@ -53,7 +49,6 @@ services:
```
tahb, line 49: The docker hosts `wiremock` and `community-api` should still work here given the name of the service still matches. Removing the `container_name` should not break it and we've got the apps running locally to test.

tahb:
Whilst the changes in this proposal fix several errors we continue to notice problems authenticating from the frontend. This is probably because of how hmpps-auth is set up requiring us to pass a jwt token around(?). 

[We've now looked at the work that's been done in the last week to use Tilt](https://github.com/ministryofjustice/hmpps-approved-premises-tools). This seems to work well and removes a category of authorisation error we had been seeing.

This work may be of some minor value or may be superceded by the Tilt work in which case the docker-compose work in the individual repos may be worth removing instead?

PR #145, opened 2022-10-13
OpenAPI spec for Assessments

brownbearsoftware:
Derived from: https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/4189880327/Tech+Day+in+Birmingham+Oct+12+2022#API-Design-for-%22Assess%22-transaction

One notable difference is using probationOfficerId (UUID) rather than Staff Id as we currently use this for the Applications.  This table is populated when an application is created.  We would likely need to switch this so that it's populated by an endpoint that is called each time a user with ROLE_PROBATION logs in to the application - perhaps the node backend could make this call to the API after the authorization code flow is completed?

### DIFF
--- a/src/main/resources/static/api.yml
+++ b/src/main/resources/static/api.yml
@@ -837,6 +837,41 @@ paths:
           $ref: '#/components/responses/403Response'
         500:
           $ref: '#/components/responses/500Response'
+  /applications/{applicationId}/allocations:
+    post:
+      tags:
+        - Operations on applications
+      summary: Reallocates an application
+      parameters:
+        - name: applicationId
+          in: path
+          description: ID of the application
+          required: true
+          schema:
+            type: string
+            format: uuid
+      requestBody:
+        content:
+          'application/json':
+            schema:
+              $ref: '#/components/schemas/Reallocation'
+        required: true
+      responses:
+        200:
+          description: successful operation
+        400:
+          description: invalid params
+          content:
+            'application/problem+json':
+              schema:
+                $ref: '#/components/schemas/ValidationError'
+        401:
+          $ref: '#/components/responses/401Response'
+        403:
+          $ref: '#/components/responses/403Response'
+        500:
+          $ref: '#/components/responses/500Response'
+      x-codegen-request-body-name: body
   /reference-data/departure-reasons:
     get:
       tags:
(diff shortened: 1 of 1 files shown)


brownbearsoftware:
@edavey further regarding the probation_officer table - we could use this to lookup staff identifier/staff code from Delius too on the /secure/staff/username/JimSnowLdap endpoint and either denormalise into our table or lookup on demand when we need to send those in domain events?

src/main/resources/static/api.yml
```diff
@@ -997,6 +1032,154 @@ paths:
...
+  /assessments:
+    get:
+      tags:
+        - Assessment data
+      summary: Gets assessments the user is authorised to view
+      parameters:
+        - in: query
+          name: filter
+          required: false
+          description: all (default) returns all assessments the user is authorised to view.  Allocated returns only those allocated to the user.
```
pezholio, line 1044: I wonder whether an alternative could be:

```
GET /assessments - return all assessments the user is allowed to view
GET /assessments?probationOfficerId=some-uuid
```

When a "normal" user is logged in, it's likely that the only assessments they're allowed to see are the ones they're allocated to. Adding the `probationOfficerId` query string is then a way for "admin" users to filter by a particular probation officer.

src/main/resources/static/api.yml
```diff
@@ -997,6 +1032,154 @@ paths:
...
+              schema:
+                $ref: '#/components/schemas/Assessment'
+        401:
+          $ref: '#/components/responses/401Response'
+        403:
+          $ref: '#/components/responses/403Response'
+        500:
+          $ref: '#/components/responses/500Response'
+  /assessments/{assessmentId}/notes:
+    post:
```
pezholio, line 1124: Given the user may be coming back and updating a note, then we might want to have a `PUT` action too?

pezholio: Actually, scratch that. For audit purposes, it's probably better to make these uneditable isn't it?

src/main/resources/static/api.yml
```diff
@@ -1823,3 +2006,93 @@ components:
...
+        - assignedAt
+        - data
+        - clarificationNotes
+    UpdateAssessment:
+      type: object
+      properties:
+        data:
+          type: object
+          additionalProperties:
+            $ref: '#/components/schemas/UpdateAssessment'
```
pezholio, line 2056: Is this going to cause an infinite loop?

pezholio (approved):
LGTM - just a few comments / clarifications / random musings.

src/main/resources/static/api.yml
edavey: yes, we didn't talk about it, but this design accommodates having multiple assessments without difficulty I think?

src/main/resources/static/api.yml
edavey: I think having a url param available to `WORKGROUP_MANAGER` sounds good but I think the resource we're trying specify is an "assessor" so this ought to be `assessorId` perhaps? i.e. I think we're also going to want a list of assessors to implement the "Workflow manager re-allocates an application for assessment" story?

`GET /assessors`

where the response includes a representation of their "qualifications" (authorisation to assess PIPE / women etc)? Not to mention their location...

src/main/resources/static/api.yml
```diff
@@ -1823,3 +2006,93 @@ components:
...
+          format: uuid
+        schemaVersion:
+          type: string
+          format: uuid
+        outdatedSchema:
+          type: boolean
+        createdAt:
+          type: string
+          format: date-time
+        assignedAt:
```
edavey, line 2028: `allocatedAt` would conform better perhaps to our terminology? Do we need "assign" at all?

src/main/resources/static/api.yml
```diff
@@ -1823,3 +2006,93 @@ components:
...
         - note
+    Assessment:
+      type: object
+      properties:
+        id:
+          type: string
+          format: uuid
+        application:
+            $ref: '#/components/schemas/Application'
+        assignedToProbationOfficer:
```
edavey, line 2017: I wonder if an `assessor` property / schema would be clearer here than referring to `assignedToProbationOfficer`? 

And probably it would be better to avoid introducing a synonym for "allocate" here?

I'm thinking that the "assessment" has an "assessor" -- that's the person to whom the task of "assessment" has been "allocated"?

src/main/resources/static/api.yml
```diff
@@ -997,6 +1032,154 @@ paths:
...
+                type: array
+                items:
+                  type: string
+                  enum:
+                    - AP_ASSESSOR
+                    - AP_MATCHER
+                    - AP_MANAGER
+                    - AP_WORKFLOW_MANAGER
+                    - AP_APPLICANT
+                    - AP_ROLE_ADMIN
```
edavey, line 1176: assuming that our service manages these roles (they aren't in the JWT) do you think it's necessary to have the `AP_*` prefix? Would it have value in system wide logging or auditing do you think? Or would it just be cruft?

edavey (approved):
Nicely done!

I've done a bit of nit-picking on language:

- I'm not sure we need "assign" when we have the term "allocate", and 
- I think we can wrap probation officer up as an assessor for a more natural API perhaps

edavey:
> @edavey further regarding the probation_officer table - we could use this to lookup staff identifier/staff code from Delius too on the /secure/staff/username/JimSnowLdap endpoint and either denormalise into our table or lookup on demand when we need to send those in domain events?

yes this sounds good -- I'm not totally clear on what you have in mind, but some kind of long term cache of Delius users is going to be needed... Maybe we need a Trello card for spiking this?

src/main/resources/static/api.yml
brownbearsoftware: I think this depends on how the workflow manager will go about doing their reassignments.  If it's a case of just viewing all of the assignments for a staff member who is on holiday then probationOfficer as query param makes most sense.  If it's a case of viewing all of them in one go to see which have been pending for too long, probably the filter=all vs allocated would be the way.

src/main/resources/static/api.yml
brownbearsoftware: Immutable would be easiest.  I suppose if we wanted to make them editable but auditable we could store a domain event each time one is edited?

src/main/resources/static/api.yml
brownbearsoftware: Ah right you are, was meant to be AnyValue

src/main/resources/static/api.yml
brownbearsoftware: Or should it perhaps be a filter over all assessments on a variable number of days pending?

src/main/resources/static/api.yml
brownbearsoftware: Agreed on the synonym point!

On the probation officer part, assessors are a subet of probation officers though right?  So maybe `allocatedToProbationOfficer`?

src/main/resources/static/api.yml
brownbearsoftware: Yeah they do seem redundant, will remove the prefix

src/main/resources/static/api.yml
pezholio: I think let's do immutable. This will make the model easy to reuse when we move to sending notes back and forth between AP staff and PPs

src/main/resources/static/api.yml
brownbearsoftware: No query param - will let frontend deal with it

PR #146, opened 2022-10-13
790 - Allow submitting document alongside application

brownbearsoftware:
Add a `document` opaque JSON field to applications.

Return this field on the GET endpoints and accept on the PUT.

If `submitted_at` is set, then `document` must also be set.


pezholio (approved):
LGTM 👍

PR #148, opened 2022-10-14
Update `alertSeverity` field to Prometheus Slack route for dev

brownbearsoftware:


pezholio (approved):

helm_deploy/values-dev.yaml
```diff
@@ -23,4 +23,4 @@ generic-service:
 # CloudPlatform AlertManager receiver to route prometheus alerts to slack
 # See https://user-guide.cloud-platform.service.justice.gov.uk/documentation/monitoring-an-app/how-to-create-alarms.html#creating-your-own-custom-alerts
 generic-prometheus-alerts:
-  alertSeverity: digital-prison-service-dev
+  alertSeverity: hmpps-approved-premises
```
pezholio, line 26: Should this be `hmpps-approved-premises-dev`?

brownbearsoftware: That perhaps would be clearer but this is the name the Cloud Platform team gave it so I aren't able to change.  I suspect the pre-prod and prod ones will be qualified when we get to that point.

PR #149, opened 2022-10-14
790 - Allocations prerequisites

brownbearsoftware:
**Add migrations and entities for assessments**

Add a table to store assessments + the respective JPA entities

**Add explicit fields to applications**

Add isWomens and isPipe fields to applications so that the Node app can send these alongside the opaque JSON blob.  This is so that the API can use these values without having to deserialize the entire blob or use JSON paths.


src/main/resources/db/migration/all/20221014113823__add_assessments_table.sql
```diff
@@ -0,0 +1,31 @@
+CREATE TABLE assessment_schemas (
+    id UUID NOT NULL,
+    added_at TIMESTAMP WITH TIME ZONE NOT NULL,
+    schema TEXT NOT NULL,
+    PRIMARY KEY (id)
+);
+
+INSERT INTO assessment_schemas (id, added_at, "schema") VALUES
```
pezholio, line 8: I'm wondering if we should just have one `schemas` table as we've already got a `data` schema for the Applications, as we'll have a `document` schema too, and a `document` schema for assessments too

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ApplicationService.kt
```diff
@@ -101,6 +101,14 @@ class ApplicationService(
       validationErrors["$.document"] = "empty"
     }
 
+    if (submittedAt != null && isWomensApplication == null) {
+      validationErrors["$.isWomensApplication"] = "empty"
+    }
+
+    if (submittedAt != null && isPipeApplication == null) {
+      validationErrors["$.isPipeApplication"] = "empty"
+    }
```
pezholio, line 110: This will cause the update to fail when we save a partially completed application at the point when we don't know this information. We should probably make this nullable and only do the check at the point of submission

pezholio (changes requested):

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ApplicationService.kt
brownbearsoftware: So this is the point of submission as I understand it (setting submittedAt to a value) - seems the most RESTful way of doing it?

I did have a similar thought though around losing any updated data but as I understand it the frontend will call the endpoint after each section (where submittedAt would be null) then once again after the check your answers section (where submittedAt is given a value) so should be OK?

pezholio (approved):
LGTM - sorry, I have lost the ability to read code 😊

src/main/resources/db/migration/all/20221014113823__add_assessments_table.sql
brownbearsoftware: Good shout, have updated

brownbearsoftware:

PR #154, opened 2022-10-18
Feature/implement create new premises post endpoint

ian-wells-nimble:
This PR is to do a comprehensive first pass of implementing a new POST endpoint for adding a new premises. 

The request body is specific to CAS3, however the response body is using the existing Premises schema which has had to be augmented to add some new fields:

- address
- notes
- service

Address is only a single string, notes are optional and service in this case is 'CAS3'. This is so that when a call is made to get only CAS3 properties, they can be queried on this value. 

In PremisesService you will see:

`// TODO: dummy data for now - remove asap
    val apAreaEntity = ApAreaEntity(
      id = UUID.randomUUID(),
      name = "arbitrary_ap_area",
      identifier = "arbitrary_identifier",
      probationRegions = mutableListOf()
    )

    val probationRegion = ProbationRegionEntity(
      id = UUID.fromString("afee0696-8df3-4d9f-9d0c-268f17772e2c"), // Wales in db
      name = "arbitrary_probation_region",
      apArea = apAreaEntity,
      premises = mutableListOf()
    )
    // end of dummy data`

This is dummy data (for now) for just ap area and probabtion region as they are not being passed through but are required as part of creating a new premises using the shared schema. 

Also just below that you will see:

`// TODO: on the next iteration the identifier and name will be retrieved by the id
    val localAuthorityArea = LocalAuthorityAreaEntity(
      id = UUID.fromString(requestBody.localAuthority),
      identifier = "arbitrary_identifier",
      name = "arbitrary_local_authority_area",
      premises = mutableListOf()
    )`

This is the initial first pass for this structure as the UUID is being passed down from the client and a subsequent story will pass both identifier and name. 

An outside in approach has been used to test the creation of a new premises and also the validation scenarios.


brownbearsoftware:
Around the ktLint stuff:

- The gradle task `addKtlintCheckGitPreCommitHook` will add a check to stop you comitting code that's going to upset it
- The gradle task `ktLintFormat` will sort out most issues for you
- There's an IntelliJ plugin `Ktlint (unofficial)` by `Nick Bedal` that adds error highlighting and a context item to format - seems to be a bit flaky (doesn't always trigger) but I've found it to be fairly useful nonetheless

brownbearsoftware:
On the commits, generally if you're correcting something from a previous commit (e.g. formatting or changing a comment etc.) the way seems to be to rebase the original commit: https://playbook.dxw.com/tech/how-we-do-development/#7-tidy-up-your-commits

e.g. the `CAS3-443: reformat code to address linting` commits should probably be part of the commits that changed those lines to begin with rather than additional ones.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/PremisesService.kt
```diff
@@ -91,4 +95,74 @@ class PremisesService(
 
       return success(lostBedsEntity)
     }
+
+  fun createNewPremises(requestBody: NewPremises) = validated<PremisesEntity> {
+
+    // TODO: dummy data for now - remove asap
```
brownbearsoftware, line 101: Are `ApArea` and `ProbationRegion` populated like this because CAS3 Premises don't follow the same organisational heirarchy as CAS1 Premises?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/PremisesService.kt
```diff
@@ -91,4 +95,74 @@ class PremisesService(
 
       return success(lostBedsEntity)
     }
+
+  fun createNewPremises(requestBody: NewPremises) = validated<PremisesEntity> {
```
brownbearsoftware, line 99: Elsewhere we've avoided passing the generated-from-OpenAPI request models into the service layer so that if we need another external interface later (e.g. an MQ) the service layer is agnostic to that.  So I'd perhaps just pass down the individual properties you need as top level instead.

Thinking about it as I wrote this, there is at least one area we've fallen foul of this already (the JSON Path keys when building a validation response being the example that comes to mind.)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/PremisesService.kt
ian-wells-nimble: Yeah it's because, for now, only local authority is passed through from the FE when creating a new premises. If there's a better default way to address this I'd be interested to know about that

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/PremisesService.kt
ian-wells-nimble: Oh okay, so should I still just pass down the properties needed and not the request model?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/PremisesService.kt
```diff
@@ -91,4 +95,74 @@ class PremisesService(
...
+    if (requestBody.postcode.isEmpty()) {
+      return generalError("Please provide a postcode")
+    }
+
+    if (requestBody.service.isEmpty()) {
+      return generalError("Please provide a service type")
+    }
+
+    try {
+      UUID.fromString(requestBody.localAuthorityId)
```
brownbearsoftware, line 130: If you specify the type of `localAuthortiyId` in api.yml as:

```
type: string
format: uuid
```

OpenAPI generate will type the property as a UUID in the model it generates.  Then you should get this behaviour for free - e.g. here I just added a `aUUID` property to the create booking request

image

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/PremisesService.kt
```diff
@@ -91,4 +95,74 @@ class PremisesService(
...
+    val probationRegion = ProbationRegionEntity(
+      id = UUID.fromString("afee0696-8df3-4d9f-9d0c-268f17772e2c"), // Wales in db
+      name = "arbitrary_probation_region",
+      apArea = apAreaEntity,
+      premises = mutableListOf()
+    )
+    // end of dummy data
+
+    if (requestBody.address.isEmpty()) {
+      return generalError("Please provide an address")
```
brownbearsoftware, line 118: I think these should be field validation errors rather than general ones, e.g.

```
if (requestBody.address.isEmpty()) {
      "$.address" hasValidationError "Please provide an address"
}
```

then further down do a 
```
if (validationErrors.any()) {
      return fieldValidationError
    }
```

This way you'll get a more comprehensive output broken down by property rather than just the first issue.

Also on this - for the field validation stuff there's a mapping of error keys -> translations in our frontend somewhere.  I'm not super clear on how this actually works in the Node App, Stu or Rich should be able to provide more info if needed.  I would think your frontend will need to work the same way since you'll be using endpoints we've already built that return these keys (and we'll perhaps need to use this endpoint in the future too.)

tahb:
As discussed, I was testing this out with the frontend and the /premises endpoint now seems to 500 with a problem seemingly related to `address`?

```
java.lang.NullPointerException: Parameter specified as non-null is null: method uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Premises.<init>, parameter address
	at uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Premises.<init>(Premises.kt)
```

src/main/resources/db/migration/all/20220929141644__alter_premises_table_to_add_extra_columns.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE premises ADD COLUMN service TEXT;
```
brownbearsoftware, line 1: I'm assuming this is to discriminate between CAS1 and CAS3 premises?  If so presume this is going to be required on every Premises moving forwards.  So maybe something like:

```
ALTER TABLE premises ADD COLUMN service TEXT NOT NULL DEFAULT 'CAS1';
```

src/main/resources/static/api.yml
```diff
@@ -1039,6 +1071,12 @@ components:
         availableBedsForToday:
           type: integer
           example: 20
+        service:
+          type: string
```
brownbearsoftware, line 1075: perhaps an enum since we know all of the possible values?  So:

```
type: string
enum:
- CAS1
- CAS3
```

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PremisesEntity.kt
```diff
@@ -20,9 +20,12 @@ data class PremisesEntity(
   val id: UUID,
   val name: String,
   val apCode: String,
+  val address: String,
```
tahb, line 23: In our [wireframes](https://miro.com/app/board/uXjVPQkTpYI=/) we have "address line 1" and "address line 2", I wonder if we should call this field `address_line_1` just in case we want to add that second field in the future. I think this is what we decided on our trello card.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PremisesEntity.kt
```diff
@@ -20,9 +20,12 @@ data class PremisesEntity(
...
   var postcode: String,
   var totalBeds: Int,
   val deliusTeamCode: String,
+  val service: String,
+  val notes: String,
```
tahb, line 28: Is there a data type such as Text that better supports longer form data than String? 

Edit: I see later the migration adds a field with `TEXT` so this might be ok.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/PremisesService.kt
```diff
@@ -91,4 +95,74 @@ class PremisesService(
...
+    }
+
+    val premisesEntity = premisesRepository.save(
+      PremisesEntity(
+        id = UUID.randomUUID(),
+        name = premisesName,
+        apCode = "CAS3_default_apcode",
+        address = requestBody.address,
+        postcode = requestBody.postcode,
+        deliusTeamCode = "arbitrary_delius_team_code", // is there a default?
```
brownbearsoftware, line 155: This is used by the /premises/{premises id}/staff endpoint - it hits the Community API `/secure/teams/$deliusTeamCode/staff` endpoint which returns a Delius Team.  As I understand it these aren't specific to APs so presumably there also exists a team for each CAS3 premises.  Probably one to investigate, if there is one for each CAS3 premises this should come from the request - if there isn't then this probably needs to be feature flagged in some way?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/PremisesService.kt
```diff
@@ -91,4 +95,74 @@ class PremisesService(
...
+    val premisesName = when (requestBody.name.isNullOrEmpty()) {
+      true -> "Unknown"
+      false -> requestBody.name
+    }
+
+    val premisesEntity = premisesRepository.save(
+      PremisesEntity(
+        id = UUID.randomUUID(),
+        name = premisesName,
+        apCode = "CAS3_default_apcode",
```
brownbearsoftware, line 152: I would think we'd want to change this to be nullable in the DB and kotlin models rather than providing it a sentinel value?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/PremisesTest.kt
```diff
@@ -19,6 +20,167 @@ class PremisesTest : IntegrationTestBase() {
...
+        )
+      )
+      .exchange()
+      .expectStatus()
+      .is4xxClientError
+      .expectBody()
+      .jsonPath("title").isEqualTo("Bad Request")
+      .jsonPath("detail").isEqualTo("Please provide a local authority ID in valid UUID format")
+  }
+
```
tahb, line 183: Nice coverage ⭐

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/PremisesService.kt
```diff
@@ -91,4 +95,74 @@ class PremisesService(
...
+        address = requestBody.address,
+        postcode = requestBody.postcode,
+        deliusTeamCode = "arbitrary_delius_team_code", // is there a default?
+        probationRegion = probationRegion,
+        localAuthorityArea = localAuthorityArea,
+        bookings = mutableListOf(),
+        lostBeds = mutableListOf(),
+        service = requestBody.service,
+        notes = requestBody.notes,
+        totalBeds = 0 // default
```
brownbearsoftware, line 162: Is this not coming from the request body because CAS3 needs to model on a bed level from the get-go?

tahb (commented):
Some minor reflections from me. 

As Joe suggests the git history looks to need some tidying. 

As mentioned in my previous comment, it'd be good to understand the new error we're seeing and get to a point where we can see this working locally.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/PremisesService.kt
brownbearsoftware: I this because 'ApArea' and 'ProbationRegion' aren't relevant to a CAS3 premises though?  i.e. they will never be passed from the Frontend because they're not properties a CAS3 premises has?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/PremisesService.kt
brownbearsoftware: Yeah that's right - bit of pedantry I suppose but may as well be consistent :)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/PremisesService.kt
ian-wells-nimble: Yeah indeed - a property is created and then beds can be added as separate entities

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/PremisesService.kt
ian-wells-nimble: Now done

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/PremisesService.kt
ian-wells-nimble: At the moment yeah but I can't say for definite if they won't be needed at some point

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/transformer/BookingTransformerTest.kt
```diff
@@ -89,7 +89,10 @@ class BookingTransformerTest {
       premises = mutableListOf()
     ),
     bookings = mutableListOf(),
-    lostBeds = mutableListOf()
+    lostBeds = mutableListOf(),
+    address = "artbitrary, address",
```
edavey, line 93: typo on "arbitrary"

edavey (commented):
Looks good, thks. I've left a couple of small comments.

With regards to the general structure of the PR it would be easier to review if each commit were a discrete step in the overall implementation and could then be reviewed individually, rather than needing to look at the entire PR.

But quite possibly this was a long-lived branch and it wasn't practical to rewrite the history into a neat narrative?

tahb:
With the agreement from Sharon [last week about working on the same API](https://mojdt.slack.com/archives/C0463S9537E/p1665667175227169) I think we can now look to merge this into `main` rather than our release branch. 

That said, given the work has been reviewed already it probably makes sense to merge this in as originally intended and then look to propose a merge from `temporary-accommodation-main` into `main`. We can then shut down `temporary-accommodation-main`.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/PremisesService.kt
brownbearsoftware: Does your product owner or a domain expert from CAS3 perhaps know?  Keen that we make anything that isn't actually relevant for both types of premises is nullable rather than just filled with placeholder data

src/main/resources/static/api.yml
```diff
@@ -1049,12 +1087,35 @@ components:
...
+    NewPremises:
+      type: object
+      properties:
+        name:
+          type: string
+        address:
+          type: string
+        postcode:
+          type: string
+        localAuthorityId:
```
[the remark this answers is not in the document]
ian-wells-nimble: Now changed to localAuthorityAreaId :-)

ian-wells-nimble:
@edavey @tahb @brownbearsoftware I have forced pushed up a revision.

Here's some of my observations:

This is has been my first PR getting used to the ways of working so a lot has been learnt from it so can be applied to future work and PR's. 

I have tidied up the history - it is still long but that's the nature of being a longlived branch as you mentioned @edavey. I can avoid this in future. 

There are perhaps a couple of things on here for future discussion rather than trying to resolve now such as the enum for a service type. As we just get a string from the client atm that's enough for a first pass with a further story for making it better. 

The hardcoded strings are okay for now? As a first pass to get this infrastructure in and a story to be raised to address these as this PR really needs to be in

brownbearsoftware:
Not one for this PR specifically but one other thing that's just sprung to mind; I'm guessing this will need an acompanying change to the GET /premises endpoint to return only CAS1 or CAS3 Premises so the frontends can supply the relevant value.  Or alternatively, just a change on each frontend to filter out the other type when displaying lists of Premises.

ian-wells-nimble:
> 

I've raised it with Kath to have a story follow closely behind to address these hard-coded values

brownbearsoftware:
Leaving ApCode, Delius Team Code and Ap Area as placeholder/dummy data for the time being makes me nervous as I would argue the first & foremost priority should be making sure we have a coherent model that makes sense for both use cases.  Appreciate there are other factors at play around timing though so will leave to @edavey / @pezholio from our end.

tahb:
@edavey In regards to [Joe's comment](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/154#issuecomment-1285474883) there, I'm wondering if we might in future want to split any AP specific fields like ApArea and DeliusTeam code out in a similar way we plan to do for Characteristics?

pezholio:
@tahb That's a good idea - I don't _think_ we search by any of those fields (at least, not yet), so that might be a good option. Thoughts? @edavey @brownbearsoftware?

edavey:
> @edavey In regards to [Joe's comment](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/154#issuecomment-1285474883) there, I'm wondering if we might in future want to split any AP specific fields like ApArea and DeliusTeam code out in a similar way we plan to do for Characteristics?

I'm not seeing why we would need to split these attributes out of the `premises` table. As far as  I understand they are required attributes of a CAS3 premises in the same way that they are required attributes of a CAS1 premises.

Is the problem here actually that we don't have values yet for the CAS3 seed data? 

Is there a reason why you can't make up realistic values for these fields? I think that's what we've done in CAS1...

pezholio:
> Is there a reason why you can't make up realistic values for these fields? I think that's what we've done in CAS1...

Yeah, we have just very basic seed data in at the mo

tahb:
I'm not sure that ApArea, ApCode or Delius Team Code have come up for CAS3 yet. Right now, based on our bedspace trackers, this is what we imagine we'll be collecting about premises:

Manage Beta design exploration 👀 - Frame 1

I'm not familiar with what those fields mean for CAS1 but I suspect this will be something that may come up later for us but @edavey  you mentioned you had some understanding that it's relevant?

edavey:
> I'm not sure that ApArea, ApCode or Delius Team Code have come up for CAS3 yet. Right now, based on our bedspace 
...
> I'm not familiar with what those fields mean for CAS1 but I suspect this will be something that may come up later for us but @edavey you mentioned you had some understanding that it's relevant?

hmm, it might be quicker to have a chat about this in person to understand the differences properly and agree a way forward? 

And in the meantime, you could just seed your data with valid values?

tahb:
We had a quick chat about this and concluded:

`apCode` - this field is a new internal reference for a premises, though CAS3 don't have a direct need we can create our own reference by creating a random sha. This should be quick to do now @ian-wells-nimble so I think we should do so. In the future this column should be renamed to something more generic.

`deliusTeamCode` - we are not sure if CAS3 will ever have probation officers on site, we suspect not but should confirm this. If we're right, this field is AP specific and we should think of a way to redesign this model to allow it to be optional for CAS3. This work should be done as soon as a follow-up and I think reasonably should come before the next bits of work. For now we can keep `"UNKNOWN"` or similar to allow this pull request to be merged.

`apAreaEntity` - this may or may not be needed by CAS3 in future. CAS3 don't have a good handle on it yet though it may be related to asking users which PDU a premises resides within(?). CAS1 are currently only using this field with the intention of reporting. We should probably figure out a way we can add dummy data here the short term. 

@edavey Could we allow the `apAreaEntity` relationship to be optional? It feels misleading to assign our premises to a real but random ap area. The data would look real but would be inaccurate for a premises. Adding a fake ApArea might work but may interfere if CAS1 users are able to see it in a list they can pick from.

edavey:
Further to our chat, I've had more of a think about how to handle these divergencies and have some proposals which would mean a bit of work on the CAS1 side but would avoid us having to take on technical debt and feel soiled at such an early stage in the process of developing a shared API.

### premises.ap_code

These are unique, human-readable codes which have been assigned to each premises during the manual auditing exercise which Dan Stretton is leading. They are strings like:

- `NEALB` for "North East: Albion", and
- `MIWOR` for "Midlands: Wordsworth"

This system has been extended to identify individual beds within APs e.g

- `MIWOR03`
- `NEALB18`

The value of these codes is purely in their human-readability and we expect to use them only in views and in reporting. They are not used as keys in database joins. Premises and beds use autogenerated UUIDs as their primary keys.

I suggest that `apCode`:

- should be renamed to `premisesCode`
- should be implemented by CAS3 premises for now. In the future we (CAS1) could extract this to some other table like `friendly_ids`


### premises.delius_team_code_id

It seems clear that not all (any?) CAS3 premises will be staffed by Delius teams.

I suggest we (CAS1) extract these Delius staff codes to a join table called something like `premises_staffing`. Once CAS3 know more about their staffing needs they can implement a compatible staffing entity.


### probation_regions.ap_area_id

APs are associated with probation_regions (`premises.probation_region_id`), approx 13 in total. 

Each probation region belongs to one of about 8 "AP areas" (`probation_regions.ap_area_id`). 

Each probation region has many Probation Delivery Units (aka LDUs) which are currently missing from our schema.

Each premises is also associated with a Local Authority Area (`premises.local_authority_area_id`)

As it seems clear that not all of these categorisations will make sense for CAS3, perhaps we (CAS1) should extract all this "organisational" information into something like `OrganisationalUnit` entities, so that each type of premises can have the types of `organisational_unit_memberships` which are meaningful for that service e.g. from available:

**organisational_units**
- ap areas
- probation regions
- probation delivery units
- local authority areas

CAS1 can redefine these properties through a join table:

**organisational_unit_memberships**
joining organisational_units to premises

Once CAS3 know more about their needs they can use these organisational units or add their own.

I think we ought to stick to doing the job right and not try to fudge our differences as I suspect that as we proceed we're going to discover small but significant differences round most corners and that we need to let those differences drive maintainable design.

So my view is that we should i) merge this PR and ii) and proceed right away with some refactoring to remove the hacks and fudges.

edavey (approved):
Happy to get this merged now in the interest of moving forward, but I think we need to get cracking right away on addressing some of the differences  / disconnects which are fudged in this PR.

PR #155, opened 2022-10-18
Fix for create booking 500 when CRN not found

brownbearsoftware:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
```diff
@@ -164,7 +164,7 @@ class PremisesController(
 
     val offenderResult = offenderService.getOffenderByCrn(body.crn, httpAuthService.getDeliusPrincipalOrThrow().name)
     if (offenderResult is AuthorisableActionResult.Unauthorised) throw ForbiddenProblem()
-    if (offenderResult is AuthorisableActionResult.NotFound) validationErrors["crn"] = "Invalid crn"
+    if (offenderResult is AuthorisableActionResult.NotFound) throw BadRequestProblem(mapOf("crn" to "Invalid crn"))
```
pezholio, line 167: Should there be an updated test here?

brownbearsoftware: We moved the validation edge cases out of integration tests and into unit tests on the service layer - so this particular case isn't covered since it's in the controller

PR #156, opened 2022-10-18
Add ADR for generating Domain Events

edavey:


doc/architecture/decisions/0004-generate-domain-events-when-ap-data-is-changed.md
```diff
@@ -0,0 +1,128 @@
...
+  }
+}
+```
+
+### 3. Delius integration updates NDelius
+
+This information is used to make changes to Delius, e.g. updating the person's address,
+creating a "contact" records within NDelius etc
+
+## Consequences
```
pezholio, line 107: Should we talk about the need to spin up new infrastructure for this?

doc/architecture/decisions/0004-generate-domain-events-when-ap-data-is-changed.md
```diff
@@ -0,0 +1,128 @@
+# 4. Generate Domain Events when AP data is changed
+
+Date: 2022-10-18
+
+## Status
+
+Approved
+
+## Context
```
pezholio, line 9: Is it worth talking about alternative approaches - for example, I know we were talking about some kind of CSV format during alpha

doc/architecture/decisions/0004-generate-domain-events-when-ap-data-is-changed.md
pezholio: It's also worth pointing out this will have some impact on how quickly we deliver other features. I know we've factored this in, but worth documenting.

edavey: I don't think we need to spin up any infrastructure for this. My understanding is that the event messages will be published to an existing AWS SNS queue called `hmpps-domain-events` and the full Domain Event representations will be exposed via our existing API, through an `/events/` namespace.

I totally agree that doing this will require effort have an impact on our overall capacity but I think that impact is reduced as this is a largely completely separate work-stream and doesn't complicate delivering apply, assess, match etc

doc/architecture/decisions/0004-generate-domain-events-when-ap-data-is-changed.md
edavey: Do you mean alternative approaches to:

1.  **communicating changes to Delius?** We've had a very strong steer than using Domain Events is the right way to do it.  Or were you referring to 

2. **communicating metrics to the performance hub?** Maybe I should be explicit that Events won't be consumed to produce KPIs -- instead we'll generate a CSV for the performance hub by querying our stored events.

doc/architecture/decisions/0004-generate-domain-events-when-ap-data-is-changed.md
pezholio: 👍

pezholio (approved):

PR #157, opened 2022-10-18
Rename probation worker to user

brownbearsoftware:
Based on recent discussions around this we've decided to go for `User` over `Probation Worker`, so rename existing usages for consistency.


brownbearsoftware:

pezholio (approved):
LGTM. I ummed and ah-ed about whether to call the `createdByUser` just `createdBy`, but I think this is better because we can reason about the type we're going to get back much more easily

PR #158, opened 2022-10-18
Get user info from Community API

brownbearsoftware:
**Whenever we need a User's details or ID, get their info from Community API if we don't already have it.**

Their name and staff identifier specifically is info that won't change and we will need access to outside the context of them making a request (for allocation.)

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
@@ -1,6 +1,8 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.service
 
 import org.springframework.stereotype.Service
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.ClientResult
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.CommunityApiClient
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserEntity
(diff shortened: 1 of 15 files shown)


edavey (approved):
Thanks, this is really easy to follow and all looks correct.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
```diff
@@ -1,26 +1,36 @@
...
           id = UUID.randomUUID(),
-          name = username,
-          distinguishedName = username,
-          isActive = true,
+          name = "${staffUserDetails.staff.forenames} ${staffUserDetails.staff.surname}",
+          deliusUsername = username,
+          deliusStaffIdentifier = staffUserDetails.staffIdentifier,
           applications = mutableListOf()
         )
       )
```
edavey, line 36: I was expecting that `communityApiClient.getStaffUserDetails(username)` would only be called if `userRepository.findByDeliusUsername(username)` came back blank. Probably I'm not reading this properly?

I see the unit tests in `UserServiceTest` and I wonder if they'd be clearer if they described the way the service interacts with the repository rather than asserting what comes back from the repo, which I think is stubbed in any case?

e.g.

- the test `getUserForRequest returns existing User when exists` asserts that the repo is NOT sent a `save()` message but only a `findByDeliusUsername()` message, and 

- the test `getUserForRequest returns new User when one does not already exist` asserts that the repo IS sent a `save()` as well as the `findByDeliusUsername()` message

brownbearsoftware: You're quite right, end of day brain haze 🥴 Have amended!

PR #159, opened 2022-10-19
Fix for missing move on category

brownbearsoftware:
Fix for bug where missing move on category not detected properly + split out validation unit tests as combined test is what allowed this to sneak through


edavey (approved):
thanks for adding detailed tests here.

PR #160, opened 2022-10-19
Store roles & qualifications for Users

brownbearsoftware:
**Store roles & qualifications for Users**
Add two new tables for storing roles & qualifications

**Implement GET /profile endpoint**
This returns info for the logged in user for the frontend to use

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/UserEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/UserEntity.kt
@@ -4,7 +4,11 @@ import org.springframework.data.jpa.repository.JpaRepository
 import org.springframework.stereotype.Repository
 import java.util.UUID
 import javax.persistence.Entity
+import javax.persistence.EnumType
+import javax.persistence.Enumerated
 import javax.persistence.Id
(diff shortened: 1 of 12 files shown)


brownbearsoftware:

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/UserEntity.kt
```diff
@@ -22,5 +26,53 @@ data class UserEntity(
...
+  val role: UserRole
+)
+
+enum class UserRole {
+  ASSESSOR,
+  MATCHER,
+  MANAGER,
+  WORKFLOW_MANAGER,
+  APPLICANT,
+  ADMIN
```
pezholio, line 57: Should these enums be lowercase too?

pezholio (approved):
LGTM - really clear 👍 Dunno if I'm on the right track with those enums - feel free to ignore me!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/UserEntity.kt
pezholio: Ah, that's fine then 👍

PR #163, opened 2022-10-20
Feature/447 implement get list of all local authorities

ian-wells-nimble:
**Context:**

When a user is creating a new temporary accommodation property, they need to be able to select a local authority area from a list. This list of course needs to be the actual names of the areas. When all of the data is passed to the server to create a new property, it will include the ID of the area which will be persisted.

This change exposes a new endpoint that the client can call to retrieve a full list of area objects with all of their attributes.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ReferenceDataController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ReferenceDataController.kt
@@ -6,16 +6,19 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.ReferenceDataApiDele
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.CancellationReason
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.DepartureReason
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.DestinationProvider
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.LocalAuthorityArea
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.LostBedReason
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.MoveOnCategory
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.CancellationReasonRepository
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.DepartureReasonRepository
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.DestinationProviderRepository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.LocalAuthorityAreaRepository
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.LostBedReasonRepository
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.MoveOnCategoryRepository
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.CancellationReasonTransformer
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.DepartureReasonTransformer
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.DestinationProviderTransformer
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.LocalAuthorityAreaTransformer
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.LostBedReasonTransformer
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.MoveOnCategoryTransformer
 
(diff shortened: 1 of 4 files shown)


src/main/resources/static/api.yml
```diff
@@ -1032,6 +1032,26 @@ paths:
           $ref: '#/components/responses/403Response'
         500:
           $ref: '#/components/responses/500Response'
+  /reference-data/local-authorities:
```
edavey, line 1035: I think maybe we ought to be consistent with the term "local authority **areas**"?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ReferenceDataController.kt
```diff
@@ -3,21 +3,9 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.controller
...
-import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.LostBedReasonRepository
-import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.MoveOnCategoryRepository
-import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.CancellationReasonTransformer
-import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.DepartureReasonTransformer
-import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.DestinationProviderTransformer
-import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.LostBedReasonTransformer
-import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.MoveOnCategoryTransformer
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.*
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.*
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.*
```
edavey: I think I've seen elsewhere that using wildcard imports is not the done thing at HMPPS, but Joe will know much better than me?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ReferenceDataController.kt
```diff
@@ -3,9 +3,24 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.controller
...
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.DestinationProviderRepository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.LocalAuthorityAreaRepository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.LostBedReasonRepository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.MoveOnCategoryRepository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.CancellationReasonTransformer
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.DepartureReasonTransformer
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.DestinationProviderTransformer
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.LocalAuthorityAreaTransformer
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.LostBedReasonTransformer
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.MoveOnCategoryTransformer
```
edavey, line 23: ah, ok, I see you've gone back to individual imports here. Would you mind correcting in the previous commit? That way the history is clearer and presents an easy narrative -- as we are not squashing commits when we merge to main...

edavey (commented):
👍 nice!

 just a couple of points on:

- naming of "local authority area"
- rewriting of commits to keep the commit narrative clear and allow commit-by-commit reviewing

tahb:
If possible [I think we're ready](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/154#issuecomment-1282679489) to put this straight into `main`

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ReferenceDataController.kt
brownbearsoftware: Indeed, in IntelliJ if you go `File`->`Settings`->`Editor`->`Code Style`->`Kotlin` and set it up like so (making sure to remove anything in the bit in red) you should be on ktLint's good side :)

image

ian-wells-nimble:
@edavey @brownbearsoftware I have pushed up a revision

ian-wells-nimble:
> If possible [I think we're ready](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/154#issuecomment-1282679489) to put this straight into `main`

Oh okay shall I edit this to merge straight in to main once approved?

tahb:
@ian-wells-nimble I'd suggest that we get https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/154 into `main` first as you mentioned there would should be some conflicts. Then we can change the target branch of this pull request to `main`.

brownbearsoftware (approved):

PR #167, opened 2022-10-20
Fix for 500 when requesting Limited Access Offender

brownbearsoftware:
Failing to deserialize because of additional property and default Jackon object mapper - the easiest route is just to include the missing property


edavey (approved):
👍

PR #168, opened 2022-10-20
Add 'X-Service-Name' header parameter to allow filtering of premises by service

danlivings-dxw:
> See [ticket #445 on the CAS3 Trello board](https://trello.com/c/7gUPFXcB/445-cas1-properties-are-not-viewable-in-view-list-of-properties).

Currently the temporary accommodation landing page displays all properties in the database, regardless of whether they are intended for use by the CAS1 or CAS3 service.

~~This PR provides a way for clients of the API to filter premises by the desired service through the use of a query parameter on the `GET /premises/` endpoint:~~
- ~~`/premises/?service=cas1` returns only premises for CAS1~~
- ~~`/premises/?service=cas3` returns only premises for CAS3~~

This PR provides a way for clients of the API to filter premises by the desired service through the use of a header parameter on the `GET /premises/` endpoint:
- `X-Service-Name: approved-premises` returns only premises for CAS1
- `X-Service-Name: temporary-accommodation` returns only premises for CAS3

This ~~query~~ header parameter is optional, and if not supplied, the existing behaviour is kept for backwards compatibility. However, both the CAS1 and the CAS3 frontends should be updated to use this parameter after this PR is merged to avoid premises from the wrong service being displayed.

~~This PR depends on #154 as that adds service information to the premises in the database.~~ #154 has been merged into `main` and this PR has been rebased accordingly.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
@@ -67,8 +67,14 @@ class PremisesController(
   private val staffMemberService: StaffMemberService
 ) : PremisesApiDelegate {
   override fun premisesGet(xServiceName: ServiceName?): ResponseEntity<List<Premises>> {
+    val premises = if (xServiceName == null) {
+      premisesService.getAllPremises()
+    } else {
+      premisesService.getAllPremisesForService(xServiceName)
+    }
+
     return ResponseEntity.ok(
-      premisesService.getAllPremises().map {
+      premises.map {
         val availableBedsForToday = premisesService.getAvailabilityForRange(it, LocalDate.now(), LocalDate.now().plusDays(1))
           .values.first().getFreeCapacity(it.totalBeds)
 
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
@@ -21,6 +21,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.NewNonarrival
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.NewPremises
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Nonarrival
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Premises
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ServiceName
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.StaffMember
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.BookingEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.ValidationErrors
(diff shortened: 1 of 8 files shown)


tahb:
[Related Slack conversation about the value names we want to send](https://mojdt.slack.com/archives/C0463S9537E/p1666622511328229) which may prompt us to rename here.

tahb:
I think some good people for review here would be @ian-wells-nimble and @brownbearsoftware.

@pezholio @richpjames you may want to be aware that we are putting in this filter and that you might like to use it on your list of premises to **explicitly** filter out premises that belong to temporary accommodation.

edavey:
The API is going to have to implement service-specific behaviour in numerous points. For example, the API needs to understand:

- to which service a premises belongs
- which characteristics which can be applied to premises or rooms of a specific service
- whether or not a Delius team code should be associated with a premises
- whether to calculate occupancy based on `premises.bedCount` or via `premises.beds.count`

Asking the CAS3 and CAS1 UIs to pass a `service=CAS1|2` url param is perhaps going to be a noisy and error-prone way of triggering these expected behaviours? A couple of other, perhaps more reliable, options we might consider:

- for each UI app to set a `header` in its base REST client class e.g. `X-CAS-SERVICE` with a value of `1` or `3` (or `approved-premises` or `temporary-accommodation`)
- for the API to read the service from a role set in the `client_credentials` JWT from the UI authentication with the API e.g. (`ROLE_CAS_1` | `ROLE_CAS_3`)

What do you think?

brownbearsoftware:
```
Asking the CAS3 and CAS1 UIs to pass a service=CAS1|2 url param is perhaps going to be a noisy and error-prone way of triggering these expected behaviours?
```

I think this endpoint should be the only place we need to let the client specify which type of Premises it's concerned with, elsewhere we can do the service-specific stuff by looking at the `type` property of the Premises entity that we're dealing with instead so it's impossible for the client to trigger the wrong approach.  e.g.

```
when(premises.type) {
   "CAS1" -> cas1Behaviour()
   "CAS3" -> cas3Behaviour()
   else -> throw RuntimeException("Unsupported Premises Type")
}
```

brownbearsoftware:
That being said, I think the header could still be a good thing to send with every request so we can pop it into the Sentry context - this should cut down on any "who's bug is this anyway" between the two teams post launch: https://docs.sentry.io/platforms/java/enriching-events/context/

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
```diff
@@ -65,9 +66,15 @@ class PremisesController(
   private val staffMemberTransformer: StaffMemberTransformer,
   private val staffMemberService: StaffMemberService
 ) : PremisesApiDelegate {
-  override fun premisesGet(): ResponseEntity<List<Premises>> {
+  override fun premisesGet(service: ServiceCategory?): ResponseEntity<List<Premises>> {
     return ResponseEntity.ok(
-      premisesService.getAllPremises().map {
+      premisesService.getAllPremises().filter {
```
brownbearsoftware, line 71: Maybe premature optimisation but this would be better handled by the SQL server I think - in the PremisesRepository interface (lives in `src/main/kotlin/..../jpa/entity/PremisesEntity.kt` if you add a method like this:

```
@Repository
interface PremisesRepository : JpaRepository<PremisesEntity, UUID> {
  fun findAllByType(type: String): List<PremisesEntity>
}
```

spring-data-jpa will do the query for you (see https://docs.spring.io/spring-data/jpa/docs/current/reference/html/#repositories.query-methods.details)

danlivings-dxw: This has been implemented.

tahb:
Nice! @brownbearsoftware this should be good for another review of the implementation if you have time please? We have a few things blocked by this so if not let us know and we can try and reach out to kotlin-dev.

brownbearsoftware (approved):
LGTM

PR #170, opened 2022-10-24
OpenAPI spec for needs endpoint

brownbearsoftware:


src/main/resources/static/api.yml
```diff
@@ -1964,6 +1993,55 @@ components:
...
+          type: integer
+      required:
+        - section
+        - name
+        - overThreshold
+        - riskOfHarm
+        - flaggedAsNeed
+        - severity
+        - identifiedAsNeed
+        - needScore
```
edavey, line 2044: I think we're missing `riskOfReoffending`?

brownbearsoftware: Right you are, have added

edavey (approved):
Looks good, thank you

PR #166, opened 2022-10-20
ADR for CAS teams working together on the same API

tahb:
This proposal seeks to document the decisions that have recently been made in relation to CAS1/2/3 setting out to reuse the same API.

PR #171, opened 2022-10-24
Chore/adr for room and bed modelling

tahb:
Based atop https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/166 so please review that first.

### DIFF
--- a/doc/architecture/decisions/0006-rooms-and-beds-for-both-cas-services.md
+++ b/doc/architecture/decisions/0006-rooms-and-beds-for-both-cas-services.md
@@ -1,0 +1,108 @@
+# 6. rooms-and-beds-for-both-cas-services
+
+Date: 2022-10-24
+
+## Status
+
+Accepted
+
+## Context
+
+The CAS1 team have a longer term vision of the data layer needed for their
+property management service[1]. At the time of writing this part of the service
+is partially complete and called "mini-manage". It includes Premises but not
+Rooms or Beds.
+
+CAS1 have introduced a temporary bed concept by allowing users to store a
+`bed_count` on Premises directly. It was imagined that this would eventually
+change over.
+
+CAS3 are looking to extend mini-manage now in way that works for CAS3 but is
+also compatible for CAS1. Both teams built a shared understanding going into
+this work:
+
+- Both CAS1 and CAS3 need to represent rooms in some fashion and associate them
+  with bookings in future
+- CAS1 will have rooms with one or multiple beds
+- CAS3 will have one bed per room, it is highly unlikely but still possible
+  that in future this might change
+- Physical spaces such as Premises and Rooms look to have different
+  characteristics for CAS1 and CAS3. For example, CAS3 needs to specify whether
+  a room has a shared kitchen but CAS1 may not, however CAS1 might need to know
+  if it has an en suite.
+
+[1] <https://dbdiagram.io/d/6200fa0285022f4ee54d7569>
+
+## Decision
+
+- We will attempt to build and share the core room and bed resources
+- We will separate out the storage of unique characteristics from
+  rooms and beds in a way that supports future flexibility for both services
+- We imagine[2] we can create a new Characteristics table where records can be
+  created to represent boolean attributes. These could be associated with
+  premises or rooms via two further join tables:
+
+         ,---------------------------------------.
+         |characteristics                        |
+         |---------------------------------------|
+         |*id: uuid                              |
+         |--                                     |
+         |*name: string                          |
+         |*service_scope: string <<default: "*">>|
+         |*model_scope: string <<default: "*">>  |
+         `---------------------------------------'
+                 |                |
+                 |                |----------`
+                 |                          .|.
+                 |           ,---------------------------.
+                 |           |premises_characteristics   |
+                 |           |---------------------------|
+                 |           |*premise_id: uuid          |
+                 |           |*characteristic_id: uuid   |
+                 |           `---------------------------'
+                 |                           .|.
+                 |                            |
+                 |                            |
+                .|.                  ,---------------------------.
+      ,---------------------------.  |premises                   |
+      |room_characteristics       |  |---------------------------|
+      |---------------------------|  |*id: uuid                  |
+      |*room_id: uuid             |  |--                         |
+      |*characteristic_id: uuid   |  |address: string            |
+      `---------------------------'  |postcode: string           |
+             .|.                     `---------------------------'
+              |-------`                      |
+                     |             `----------
+                     |             .|.
+               ,---------------------------.
+               |rooms                      |
+               |---------------------------|
+               |*id: uuid                  |
+               |--                         |
+               `---------------------------'
+                              |
+                             .|.
+               ,---------------------------.
+               |Bed                        |
+               |---------------------------|
+               |*id: uuid                  |
+               |--                         |
+               `---------------------------'
+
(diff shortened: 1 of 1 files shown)


doc/architecture/decisions/0005-rooms-and-beds-for-both-cas-services.md
```diff
@@ -0,0 +1,106 @@
...
+  The implementation for this will need to be considered to allow CAS1 to do
+  this across two different transactions
+- Splitting out characteristics into a new table, potentially with join tables,
+  will require more set up work but this cost should be reasonable
+- Admin may in future need to modify the list of characteristics, having these
+  stored in a table rather than hardcoded should enable that
+- CAS1 will have to maintain their temporary `Premises.bed_count` mechanism for
+  beds alongside this extension work until such time as it can be refactored
+- We may learn of a Characteristic in future that isn't a boolean that may
+  require us to revist this plan to accommodate multiple values
```
edavey, line 106: Would it be helpful to document the fact that we neither service currently anticipates the need to model the characteristics of an individual bed?

doc/architecture/decisions/0005-rooms-and-beds-for-both-cas-services.md
```diff
@@ -0,0 +1,106 @@
...
+                 |                            |
+                 |                            |
+                .|.                  ,---------------------------.
+      ,---------------------------.  |Premise                    |
+      |RoomCharacteristic         |  |---------------------------|
+      |---------------------------|  |*id: integer <<autonumber>>|
+      |*room_id: integer          |  |--                         |
+      |*characteristic_id: integer|  |address: string            |
+      `---------------------------'  |postcode: string           |
+             .|.                     `---------------------------'
```
edavey, line 73: just to nit-pick on the language. Our usage of "premises": 
> a house or building, together with its land and outbuildings ...

is a plural noun -- i.e. "one premises", "two premises" etc

There's another v different usage of the word which does have the singular noun form "premise":

> a previous statement or proposition from which another is inferred or follows as a conclusion

`</ pedantry>`

doc/architecture/decisions/0005-rooms-and-beds-for-both-cas-services.md
```diff
@@ -0,0 +1,106 @@
...
+  premises or rooms via two further join tables:
+
+         ,---------------------------------------.
+         |Characteristic                         |
+         |---------------------------------------|
+         |*id: integer <<autonumber>>            |
+         |--                                     |
+         |*name: string                          |
+         |*service_scope: string <<default: "*">>|
+         |*model_scope: string <<default: "*">>  |
```
edavey, line 52: I'm curious about how you would use these scopes? 

Is it to display a list of characteristics which could (but have not yet) been applied to a CAS3 room? The list of characteristics which have been used can readily be surfaced with a join, so perhaps this is anticipating the need to apply previously unused characteristics through the frontend UI?

edavey (approved):
Thanks Tom, this looks good. I've made a couple of minor queries / nitpicks.

I'm keen for @brownbearsoftware to have a look also...

doc/architecture/decisions/0004-reuse-api-for-cas1-and-cas3.md
```diff
@@ -0,0 +1,86 @@
...
+Both CAS teams will aim to share and build upon the same API.
+
+Relatedly there is a frontend repository[3] that CAS3 will fork and develop
+independently.
+
+[3] <https://github.com/ministryofjustice/hmpps-approved-premises-ui>
+
+## Consequences
+
+- a single technical architect that's shared across both projects to maintain
```
brownbearsoftware, line 63: This is @edavey I'm guessing?

doc/architecture/decisions/0004-reuse-api-for-cas1-and-cas3.md
```diff
@@ -0,0 +1,86 @@
...
+  oversight and improve the chances of delivering a joined up service
+- both teams may have to collaborate and design parts of the service together,
+  hopefully the tech leads and technical architect can do this for the most part
+  to protect the rest of the teams time
+- we may find that working on the same API proves too logistically difficult in
+  which case CAS3 could fork away at that point
+- we may find there isn't as much shared behaviour as we anticipate, in which
+  case CAS3 could work away at that point
```
brownbearsoftware, line 71: should this be `fork away` rather than `work away`?

doc/architecture/decisions/0005-rooms-and-beds-for-both-cas-services.md
```diff
@@ -0,0 +1,106 @@
...
+- We will separate out the storage of unique characteristics from
+  rooms and beds in a way that supports future flexibility for both services
+- We imagine[2] we can create a new Characteristics table where records can be
+  created to represent boolean attributes. These could be associated with
+  premises or rooms via two further join tables:
+
+         ,---------------------------------------.
+         |Characteristic                         |
+         |---------------------------------------|
+         |*id: integer <<autonumber>>            |
```
brownbearsoftware, line 48: We've used UUID surrogate keys everywhere else so far would be good to do so here too for consistency sake

doc/architecture/decisions/0005-rooms-and-beds-for-both-cas-services.md
brownbearsoftware: On this - all of our tables names since `20220811100325__pluralize_table_names.sql` are pluralized & snake case so would be good to have these the same (e.g. `room_characteristics`)

doc/architecture/decisions/0004-reuse-api-for-cas1-and-cas3.md
tahb: Yes I believe it is 👍

doc/architecture/decisions/0004-reuse-api-for-cas1-and-cas3.md
tahb: Yup, thanks for the spot!

doc/architecture/decisions/0005-rooms-and-beds-for-both-cas-services.md
tahb: Yes absolutely. I'll change this to uuid.

doc/architecture/decisions/0005-rooms-and-beds-for-both-cas-services.md
```diff
@@ -0,0 +1,106 @@
...
+## Consequences
+
+- For CAS3, the new 'create a Room' endpoint will also need to create a bed.
+  The implementation for this will need to be considered to allow CAS1 to do
+  this across two different transactions
+- Splitting out characteristics into a new table, potentially with join tables,
+  will require more set up work but this cost should be reasonable
+- Admin may in future need to modify the list of characteristics, having these
+  stored in a table rather than hardcoded should enable that
+- CAS1 will have to maintain their temporary `Premises.bed_count` mechanism for
```
brownbearsoftware, line 103: To allow CAS1 to continue with a per-Premises bed count but have CAS3 do bed-level modelling, Bookings and Lost Beds will need to have optional references to Beds with some switching around Premises type in various places:

- GET /premises & GET /premises/{id} (`availableBedsForToday` calculation for this will be slightly different I think as would need a Lost Bed entry for each individual Bed?)
- GET /premises/{premises id}/capacity (similar to above)
- POST /premises/{premises id}/booking (validate that a bed reference has/has-not been provided based on the type of premises.  Validate that you're not attempting to double book a bed?)
- POST /premises/{premises id}/booking/{booking id}/extensions (there's some potential complexity here if you need to extend someone's stay if the bed has already been booked for someone else after the current length of the booking - could probably just return an error in these cases and force the Probation Practitioner to make a new booking)

Might be good to make a note of some of these cases here and how the tables affected/logic will change?

doc/architecture/decisions/0005-rooms-and-beds-for-both-cas-services.md
tahb: Absolutely, I'll make those updates.

doc/architecture/decisions/0005-rooms-and-beds-for-both-cas-services.md
tahb: Yes that was something we noticed wasn't it 👍

doc/architecture/decisions/0005-rooms-and-beds-for-both-cas-services.md
tahb: Thanks for this context, we will definitely need to plan for these details and beleive that you and Ed have just run through them.

PR #175, opened 2022-10-24
Feature/add characteristics to a property

ian-wells-nimble:
Refers to [this story](https://trello.com/c/CqNOUrof/444-a-user-can-add-additional-characteristics-about-a-property-when-adding-a-property)

A property can have characteristics associated with it. A list of these can be retrieved as reference data so that the client can display them to the end user to select accordingly. 

This PR is a first iteration to set up the necessary infrastructure to be able to persist characteristics (that can be shared across services) and to be able to get a list of them. 

A further iteration on a story being done next is to add a service type param so characteristics can be returned that only a particular service cares about.

This first pass sets up the Characteristic table which will eventually have link entity tables to use it for example with premises. So there would be a link table between this and Premises

### DIFF
--- a/src/main/resources/db/migration/all/20220803121541__characteristic_table.sql
+++ b/src/main/resources/db/migration/all/20220803121541__characteristic_table.sql
@@ -1,0 +1,7 @@
+CREATE TABLE characteristic (
+    id UUID NOT NULL,
+    name TEXT NOT NULL,
+    service_scope TEXT NOT NULL DEFAULT '*',
+    model_scope TEXT NOT NULL DEFAULT '*',
+    PRIMARY KEY (id)
+);
--- a/src/main/resources/db/migration/local+dev/20220913131809__mock_characteristics.sql
+++ b/src/main/resources/db/migration/local+dev/20220913131809__mock_characteristics.sql
@@ -1,0 +1,10 @@
+INSERT INTO characteristics (id, name, service_scope) VALUES
+('952790c0-21d7-4fd6-a7e1-9018f08d8bb0','Park nearby','CAS3'),
+('199334d3-fabb-432f-84c3-0e92eaf13f24','Pub nearby','CAS3'),
+('1dac11c5-a1b4-4313-b0f7-41b2e57f2404','School nearby','CAS3'),
+('94021062-f692-4877-b6e8-f36c7ff87a18','Not suitable for registered sex offenders (RSO)','CAS3'),
+('16b52729-be75-43b4-b711-77fe5cbcb477','Not suitable for arson offenders','CAS3'),
+('7846fbf2-b423-4ccc-b23b-d8b866f86bde','Men only','CAS3'),
+('7cd8dd2c-a14d-4a78-8d72-48449cc5aaa3','Women only','CAS3'),
+('d9f52ae8-4c01-4751-8975-992efacd5260','Floor level access','CAS3'),
+('a862be08-a96a-4337-9f46-26286db8015f','Wheelchair accessible','CAS3');
--- a/src/main/resources/static/api.yml
+++ b/src/main/resources/static/api.yml
@@ -1118,6 +1118,26 @@ paths:
           $ref: '#/components/responses/403Response'
         500:
           $ref: '#/components/responses/500Response'
+  /reference-data/characteristics:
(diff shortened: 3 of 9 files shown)


src/main/resources/db/migration/all/20220803121541__characteristic_table.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE characteristic (
```
brownbearsoftware, line 1: Just for consistency sake with the other tables, could we name this `characteristics` instead?

src/main/resources/db/migration/all/20220803121541__characteristic_table.sql
```diff
@@ -0,0 +1,6 @@
...
+    id UUID NOT NULL,
+    name TEXT NOT NULL,
+    service_scope TEXT DEFAULT '*',
+    model_scope TEXT DEFAULT '*'
+);
```
brownbearsoftware, line 6: Missing a `PRIMARY KEY (id)`

src/main/resources/db/migration/local+dev/20220913131809__mock_characteristics.sql
```diff
@@ -0,0 +1,18 @@
+INSERT INTO characteristic (id, name, service_scope)
+VALUES ('952790c0-21d7-4fd6-a7e1-9018f08d8bb0','Park nearby','CAS3');
```
brownbearsoftware, line 2: Again, not really an issue but I would also consider generating the UUIDs rather than just changing the last few characters so that they're easier to tell apart when debugging.  There's a good plugin for IntelliJ named "UUID Generator" by Leo Million that adds an option to the `Alt+Insert` menu to generate a UUID.  This works with multi-line selections so if you go

```
INSERT INTO characteristic (id, name, service_scope) VALUES
   ('', 'a', 'CAS3'),
   ('', 'b', 'CAS3'),
   ('', c', 'CAS3')
```

Then hold down alt and drag down the empty IDs you can generate them all in one go

src/main/resources/static/api.yml
```diff
@@ -1334,6 +1354,8 @@ components:
           $ref: '#/components/schemas/ApArea'
         localAuthorityArea:
           $ref: '#/components/schemas/LocalAuthorityArea'
+        characteristics:
+          $ref: '#/components/schemas/Characteristic'
```
brownbearsoftware, line 1358: Should this be an array?

src/main/resources/db/migration/all/20220803121541__characteristic_table.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE characteristic (
+    id UUID NOT NULL,
+    name TEXT NOT NULL,
+    service_scope TEXT DEFAULT '*',
```
brownbearsoftware, line 4: these should perhaps be `NOT NULL` too since I'm assuming `*` means apply to all services / models?

src/main/resources/static/api.yml
```diff
@@ -1415,6 +1437,21 @@ components:
...
+          format: uuid
+          example: 952790c0-21d7-4fd6-a7e1-9018f08d8bb0
+        name:
+          type: string
+        serviceScope:
+          type: string
+        modelScope:
+          type: string
+      required:
+        - name
```
brownbearsoftware, line 1454: `id` should be required since it's always present.

Potentially also `serviceScope` and `modelScope` depending on the response to this comment: https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/175/files#r1004275092

src/main/resources/db/migration/all/20220803121541__characteristic_table.sql
tahb: I think I could have over communicated [the naming choices more Ian](https://mojdt.slack.com/archives/D042D4ME75F/p1666619714158769). We learnt about the pluralised snake cased naming and [documented all the models here](https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/main/doc/architecture/decisions/0006-rooms-and-beds-for-both-cas-services.md#decision).

brownbearsoftware (approved):
LGTM

PR #176, opened 2022-10-25
831 - Service specific Premises types

brownbearsoftware:
**Split out PremisesEntity into abstract base class + concrete derived classes**

Common properties remain on PremisesEntity but two derived classes `ApprovedPremisesEntity` and `TemporaryAccommodationPremisesEntity` are created.

There are several ways of achieving this as outlined here: https://thorben-janssen.com/complete-guide-inheritance-strategies-jpa-hibernate

Ultimately went with the `JOINED` option as:
 - It allows us to keep referential integrity at the database level for other tables with foreign keys to the premises table - the `TABLE_PER_CLASS` option would have forced us to abandon the FK constraints
 - It's neater to navigate and reason about than the `SINGLE_TABLE` option as you can easily see which properties belong to which type
 - Performance impact of the join should be acceptable given the number of users/traffic is not particularly high in the scheme of things.

We store the additional properties in two new tables, when JPA retrieves the data it joins to the relevant table as determined by the `service` field.

We can then distinguish between the types of property via type assertion, e.g.

```
when(somePremises) {
   is ApprovedPremisesEntity -> cas1SpecificStuff()
   is TemporaryAccommodationEntity -> cas3SpecificStuff()
}
```

From the perspective of the actual code, the same `PremisesRepository` that we know and love can be used.  It will return the correct subclass based on the value of the `service` column.  The extra properties can then be accessed through type assertion as above when required or treated as the base class when behaviour is the same for both types of Premises.

In the test code, we need to deal with concrete classes when inserting so we now have two distinct factories for producing & persisting Premises entities.  e.g. 

```
approvedPremisesEntityFactory.produceAndPersist { ... }
temporaryAccommodationEntityFactory.produceAndPersist { ... }
```

_The next step would be to move deliusTeamCode from PremisesEntity down into ApprovedPremisesEntity, however we're not 100% sure that CAS3 won't need this for some publicly run Premises yet in which case a staff-details strategy would be more appropriate so will leave that for now_

_I've also made the POST Premises endpoint only work for CAS3 Premises for now, will sort this in a subsequent PR_

### DIFF
--- a/doc/architecture/decisions/0007-storing-service-specific-properties.md
+++ b/doc/architecture/decisions/0007-storing-service-specific-properties.md
@@ -1,0 +1,54 @@
+# 7. storing-service-specific-properties
+
+Date: 2022-10-26
+
+## Status
+
+Accepted
+
+## Context
+
+Whilst many of the properties of a Premises are shared between CAS1 (Approved Premises) 
+and CAS3 (Temporary Accommodation) there are some known divergences (e.g. Ap Code being
+specific to CAS1).  There are also likely to be unknown divergences.
+
+This requires us to have a way of sharing most properties but allowing for service specific 
+properties too.
+
+## Decision
+
+- We will use JPA's inheritance support[1] to create a common base class (PremisesEntity) from 
+  which a class for each service (`ApprovedPremisesEntity` and `TemporaryAccommodationPremisesEntity`) 
+  will derive.
+- We will use the `JOINED` strategy to model this in the database (i.e. one `premises` table storing 
+  the common fields with an additional table for each of the derived classes `approved_premises` and 
+  `temporary_accommodation_premises`) to store their specialised fields.
+  - There is a performance impact from using this approach as JPA must generate a JOIN instead of 
+    a single table select.  It's unlikely that this will be significant enough to cause an issue at 
+    our anticipated level of load.
+  - The `SINGLE_TABLE` approach was also considered but decided against as:
+    - The properties of derived classes must be nullable in the table, but potentially not-nullable in 
+      the Kotlin entity models.  Consistency between the two is preferable.
+    - A single table with mixed fields becomes harder to reason about over time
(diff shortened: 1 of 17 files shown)


edavey (approved):
Very interesting, this looks workable to me.

If you have any links to relevant Spring documentation or illustrative blog posts, it might be a good idea to include in a commit message or the PR narrative.

If we try this out and it looks like a winning approach then we'll want to write this up as an ADR.

brownbearsoftware:
> I wonder if this PR ought to add a fourth commit, in which we move premises.apCode and premises.deliusTeamCode to the ApprovedPremisesEntity.
> 
> In that way we can remove the UNKNOWN placeholder text from PremisesService: https://github.com/ministryofjustice/approved-premises-api/blob/main/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/PremisesService.kt#L173-L176

Happy to press on with this, will do in a follow up PR though :)

> If you have any links to relevant Spring documentation or illustrative blog posts, it might be a good idea to include in a commit message or the PR narrative.
> 
> If we try this out and it looks like a winning approach then we'll want to write this up as an ADR.

Sure, have updated the PR description a bit and added an ADR whilst it's fresh in my mind.  Will see if any issues emerge with the ApCode/DeliusTeamCode work

edavey:
Thanks Joe, makes sense to continue in a further PR.

Thanks for the additional narrative and the link to the blog post, and for incorporating this and more into an ADR.

Generally I do think it's really important to capture background thinking and decision making in commit messages, even where separate ADRs aren't needed. PR conversations are ultimately ephemeral whereas the Git history should be an enduring documentary resource!

brownbearsoftware:
[Spike: Clean approach for CAS1/CAS3 specific fields in API](https://trello.com/c/Sid7SfX9/831-spike-clean-approach-for-cas1-cas3-specific-fields-in-api)

PR #182, opened 2022-10-27
rename characteristic table to characteristics

ian-wells-nimble:
The creation of the table 'characteristic' should actually be 'characteristics'


danlivings-dxw (approved):
LGTM 👍

PR #183, opened 2022-10-27
Split Premises response body into AP/TA types

brownbearsoftware:
In a nutshell, split the response type from the get all premises and get premises by ID endpoints into 3 entities:
- One containing common properties
- One for AP specific properties that extends the common one
- Another for TA specific properties that extends the common one

Commit messages explain in more detail.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -49,11 +49,27 @@ dependencies {
   testRuntimeOnly("io.jsonwebtoken:jjwt-jackson:0.11.5")
 
   implementation("com.fasterxml.jackson.core:jackson-databind") {
-    version { strictly("2.14.0-rc1") }
+    version { strictly("2.14.0-rc2") }
   }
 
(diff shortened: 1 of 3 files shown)


brownbearsoftware:
[Spike: Clean approach for CAS1/CAS3 specific fields in API](https://trello.com/c/Sid7SfX9/831-spike-clean-approach-for-cas1-cas3-specific-fields-in-api)

edavey (approved):
This is a thing of beauty!

Is it too good to be true? 

Do you anticipate complexity with more involved differences, e.g. the need to calculate occupancy based on `premises.bedCount` or via `premises.beds.count`?

brownbearsoftware:
> Do you anticipate complexity with more involved differences, e.g. the need to calculate occupancy based on premises.bedCount or via premises.beds.count?

I think the `availableBedsForToday` property should be OK.  CAS3 might need to do a similar thing for the `/premises/{premisesId}/capacity` endpoint though to return a per-bed response for each day instead

PR #185, opened 2022-10-27
WIP: Generate domain events from the Application service

edavey:
In this initial work we tie Domain Event "generators" to services. e.g.:

- `ApplicationService` will use an `ApplicationEventGenerator`, passing in
  an instance of `Application` each time
- `BookingService` will use a `BookingEventGenerator`, passing in an instance
  of `Booking` with each message sent to` generate()`
- `AssessmentService` will send an `Assessment` to its `AssessmentEventGenerator`

etc

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/domain/events/ApplicationEventGenerator.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/domain/events/ApplicationEventGenerator.kt
@@ -1,0 +1,12 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.domain.events
+
+import org.springframework.stereotype.Service
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApplicationEntity
+
+@Service
+class ApplicationEventGenerator {
+
+  fun generate(eventName: String, application: ApplicationEntity): String {
+    return "Generating an $eventName domain event for application ${application.id}"
+  }
+}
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/domain/events/BookingEventGenerator.kt
```diff
@@ -0,0 +1,11 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.domain.events
+
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.BookingEntity
+
+class BookingEventGenerator {
+  fun generate(eventName: String, booking: BookingEntity): String {
+    val message = "Generating a $eventName domain event for booking ${booking.id}"
+    println(message)
```
brownbearsoftware, line 8: Generally try to avoid using `println` as it goes straight to stdout - so miss out on logging levels, log shipping etc.:

```
class BookingEventGenerator {

  fun generate(eventName: String, booking: BookingEntity): String {
    val message = "Generating a $eventName domain event for booking ${booking.id}"
    log.info("Generating a $eventName domain event for booking ${booking.id}")
    return message
  }
}
```

edavey: 👍

pezholio:
@edavey This has been around for a while - I assume this work has been done in a different PR now?

edavey:
yes, this is very much superseded!

PR #187, opened 2022-10-27
Feature/get characteristics by service type

ian-wells-nimble:
This is a further iteration on getting property characteristics, only now you can get them by service type or without which will return everything.


brownbearsoftware (approved):
LGTM

PR #192, opened 2022-10-31
Store service name against a property

ian-wells-nimble:
So that properties can be differentiated, a service name is required to be stored against them on creation of a new one. 

This was a little more involved than just passing down the service name, that is now in the header rather than the body, as there has been some work done previously to split out the concrete implementations of a TA property or an AP property.

Where the POST method deals with the service name in the header, is calls down to the service layer which calls a save method on the repo, creating specifically an instance of a TA premises. At this point there's no need to tell it what service to persist as it knows it's a TA instance and so it saves the relevant service name by default.

### DIFF
--- a/src/main/resources/db/migration/all/20220803121542__revise_service_column_default_value.sql
+++ b/src/main/resources/db/migration/all/20220803121542__revise_service_column_default_value.sql
@@ -1,0 +1,7 @@
+UPDATE premises
+SET service = 'approved-premises'
+WHERE service = 'CAS1';
+
+UPDATE premises
+SET service='temporary-accommodation'
+WHERE service = 'CAS3';
--- a/src/main/resources/static/api.yml
+++ b/src/main/resources/static/api.yml
@@ -36,6 +36,13 @@ paths:
       tags:
         - Premises
       summary: Add a new premises
+      parameters:
+        - name: X-Service-Name
+          in: header
+          required: false
+          description: If given, persist the service name against this property
+          schema:
+            $ref: '#/components/schemas/ServiceName'
       requestBody:
         content:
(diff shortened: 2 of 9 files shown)


src/main/resources/static/api.yml
```diff
@@ -36,6 +36,13 @@ paths:
       tags:
         - Premises
       summary: Add a new premises
+      parameters:
+        - name: X-Service-Name
+          in: header
+          required: false
```
brownbearsoftware, line 42: Since the endpoint can never return a 2xx if this header isn't provided, feels like it should be `required: true`

brownbearsoftware: ignore me, just saw that it defaults to approved-premises if header is missing

src/main/resources/db/migration/all/20220803121542__revise_service_column_default_value.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE premises ALTER COLUMN service SET DEFAULT 'approved-premises';
```
brownbearsoftware, line 1: This won't work as intended because the rows already exist and already have a value of 'CAS1' - so needs to be something like:
```
UPDATE premises SET service = 'approved-premises' WHERE service='CAS1';
UPDATE premises SET service='temporary-accommodation' WHERE service='CAS3';
```

ian-wells-nimble: Doh of course

brownbearsoftware (approved):
LGTM

danlivings-dxw (approved):
LGTM 👍

PR #193, opened 2022-10-31
819 - assessment decision

brownbearsoftware:
Whilst thinking about which further endpoints we'd need for assessments realised that the current GET endpoint is missing a decision which would need to be a first class property so adding that here.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/AssessmentEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/AssessmentEntity.kt
@@ -6,6 +6,8 @@ import org.springframework.stereotype.Repository
 import java.time.OffsetDateTime
 import java.util.UUID
 import javax.persistence.Entity
+import javax.persistence.EnumType
+import javax.persistence.Enumerated
 import javax.persistence.Id
 import javax.persistence.JoinColumn
 import javax.persistence.ManyToOne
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/AssessmentEntity.kt
```diff
@@ -42,8 +44,16 @@ data class AssessmentEntity(
...
+  @Enumerated(value = EnumType.STRING)
+  var decision: AssessmentDecision?,
 
   @Transient
   var schemaUpToDate: Boolean
 )
+
+enum class AssessmentDecision {
+  SUITABLE,
+  UNSUITABLE
```
edavey, line 58: would `ACCEPTED | REJECTED` be more a natural pair of values for the outcome of assessing an application? Or are we thinking that this is a an assessment of a person's suitability for placement?

One of the reasons for a negative outcome would be that the applicant has failed to provide answers to follow up requests for further information. That makes me think that ACCEPTED/REJECTED is a better choice as what we ultimately need to record is a decision on the application rather than the suitability of the person being referred...

src/main/resources/static/api.yml
```diff
@@ -2383,6 +2385,11 @@ components:
         - allocatedAt
         - data
         - clarificationNotes
+    AssessmentDecision:
+      type: string
+      enum:
+        - suitable
+        - unsuitable
```
edavey, line 2392: see other comment

edavey (approved):
well remembered!

I've suggested that accepted/rejected might be a slightly better set of enums....

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/AssessmentEntity.kt
brownbearsoftware: makes sense 👍

PR #194, opened 2022-10-31
Add support for room characteristics when getting a list of rooms/creating a new room

danlivings-dxw:
> See [ticket #495 on the CAS3 Trello board](https://trello.com/c/YgJCtrWY/495-room-characteristics-can-be-saved).

Currently, when adding a new room to a premises, only a name and some optional notes can be given. To provide additional information about a room, the API will need to provide a way to attach characteristics to a room.

This PR provides support for room characteristics through the following API changes:
- The `NewRoom` schema used to define the request body of the `POST /premises/{premisesId}/rooms` endpoint has been extended to allow the client to specify a list of UUIDS representing characteristics.
- The `Room` schema used in the response body of the `POST /premises/{premisesId/rooms` and the `GET /premises/{premisesId}/rooms` endpoints have been extended to specify the active characteristics of the room.

A characteristic has two ways in which to limit what it can be applied to, known as 'scopes'. The model scope determines which type of entity the characteristic is valid for (currently `"room"` or `"premises"`). The service scope determines which service the characteristic is valid for (currently `"approved-premises"` or `"temporary-accommodation"`). It is possible for a single characteristic to be shared across model or service scopes, although this is beyond the scope of this PR.

To add a characteristic to a room, the `RoomService` enforces that the characteristic has the correct scopes. A characteristic must have a model scope of `"room"`, and the service scope must match the service the premises is for. If these are incorrect, they are reported as validation errors.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/RoomService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/RoomService.kt
@@ -42,6 +42,7 @@ class RoomService(
         notes = notes,
         premises = premises,
         beds = mutableListOf(),
+        characteristics = mutableListOf(),
       )
     )
 
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/RoomService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/RoomService.kt
@@ -1,8 +1,11 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.service
 
+import org.springframework.data.repository.findByIdOrNull
 import org.springframework.stereotype.Service
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApprovedPremisesEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.BedEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.BedRepository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.CharacteristicRepository
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.PremisesEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.RoomEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.RoomRepository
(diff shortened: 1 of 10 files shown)


ian-wells-nimble:
@brownbearsoftware I've just had a chat with Dan, my thinking was that when saving characteristics they could be optional as the user may not select any, however we'd still need to add defensive code to deal with that if they're optional. In this PR Dan has made them required and if there isn't any passed through then an empty list is generated. There's for's and against's for either approach but I'm happy with the way Dan has done it.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/RoomService.kt
```diff
@@ -1,31 +1,57 @@
...
+      } else {
+        if (entity.modelScope != "room") {
+          "$.characteristics[$index]" hasValidationError "incorrectCharacteristicModelScope"
+        }
+        if (premises is TemporaryAccommodationPremisesEntity && entity.serviceScope != "temporary-accommodation") {
+          "$.characteristics[$index]" hasValidationError "incorrectCharacteristicServiceScope"
+        }
+        if (premises is ApprovedPremisesEntity && entity.serviceScope != "approved-premises") {
+          "$.characteristics[$index]" hasValidationError "incorrectCharacteristicServiceScope"
+        }
```
ian-wells-nimble, line 49: I was looking to see if these ifs can be consolidated into something a when but I don't think they can due to checking the state of both modelScope & serviceScope

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/RoomService.kt
```diff
@@ -1,31 +1,57 @@
...
     premises: PremisesEntity,
     roomName: String,
     notes: String?,
+    characteristics: List<UUID>,
   ): ValidatableActionResult<RoomEntity> = validated {
     if (roomName.isEmpty()) {
       "$.name" hasValidationError "empty"
     }
 
+    val characteristicEntities = characteristics.mapIndexed { index, uuid ->
```
brownbearsoftware, line 35: I'd try and steer clear of java.util.Optional since there's so much cross-over with kotlin's built in nullability:

```
val characteristicEntities = characteristics.mapIndexed { index, uuid ->
      val entity = characteristicRepository.findByIdOrNull(uuid)

      if (entity == null) {
        "$.characteristics[$index]" hasValidationError "doesNotExist"
      } else {
        if (entity.modelScope != "room") {
          "$.characteristics[$index]" hasValidationError "incorrectCharacteristicModelScope"
        }
        if (premises is TemporaryAccommodationPremisesEntity && entity.serviceScope != "temporary-accommodation") {
          "$.characteristics[$index]" hasValidationError "incorrectCharacteristicServiceScope"
        }
        if (premises is ApprovedPremisesEntity && entity.serviceScope != "approved-premises") {
          "$.characteristics[$index]" hasValidationError "incorrectCharacteristicServiceScope"
        }
      }

      entity
    }

    if (validationErrors.any()) {
      return fieldValidationError
    }

    val automaticallyCreateBed = when (premises) {
      is TemporaryAccommodationPremisesEntity -> true
      else -> false
    }

    val room = roomRepository.save(
      RoomEntity(
        id = UUID.randomUUID(),
        name = roomName,
        notes = notes,
        premises = premises,
        beds = mutableListOf(),
        characteristics = characteristicEntities.map { it!! }.toMutableList(),
      )
    )
```

Outcome is pretty much the same except the double bang operator throws a NullPointerException instead of NoSuchElementException like Optional

brownbearsoftware (approved):
LGTM

ian-wells-nimble (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/RoomService.kt
danlivings-dxw: I did try to avoid `Optional` but I couldn't quite figure out how, it seems that the `!!` operator was the missing piece of the puzzle. I'll update this.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/PremisesTest.kt
```diff
@@ -597,4 +598,196 @@ class PremisesTest : IntegrationTestBase() {
...
+          name = "test-room",
+          characteristics = mutableListOf(characteristic),
+        )
+      )
+      .exchange()
+      .expectStatus()
+      .is4xxClientError
+      .expectBody()
+      .jsonPath("title").isEqualTo("Bad Request")
+      .also { it.returnResult().responseBodyContent?.let { it1 -> println(String(it1)) } }
```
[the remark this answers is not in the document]
danlivings-dxw: Yes, I'll remove it.

PR #195, opened 2022-11-01
(797) ADR for first class citizen fields in Applications

pezholio:
This is a first draft of an ADR for hadling first-class citizen fields in Applications. I've spelt out some other possible options, as well as the way forward I've chosen. This is by no means set in stone, and I'd like to get as much feedback as possible before deciding on a way forward, so please, if you think this is rubbish, please tell me!

### DIFF
--- a/doc/architecture/decisions/0008-add-first-class-citizen-fields-to-applications.md
+++ b/doc/architecture/decisions/0008-add-first-class-citizen-fields-to-applications.md
@@ -1,0 +1,139 @@
+# 8. Add First-Class Citizen Fields to Applications
+
+Date: 2022-10-31
+
+## Status
+
+Accepted
+
+## Context
+
+Approved Premises applications are currently stored in the database as a JSON
+object. This allows the format of the questionnaire to be flexible, simple, and
+take into account any changes in the questionnaire format, whilst also
+preserving applications that have been submitted in an older format.
+
+However, when processing an application, there are some situations where we need
+easy access to particular fields within the JSON application object. These include
+(but are not limited to):
+
+- If an application is for a women-only AP
+- If an application is for a PIPE placement
+- The target location
+- The desired start date
+
+These are important fields that need to be queried on, and made into first class
+citizens within the database table, without sacrificing flexibility.
+
+## Decision
+
+At the point of submission of a completed application, we will use [JsonLogic](https://jsonlogic.com)
+to fetch the important responses from the JSON object and populate the first
+class citizen fields with those responses.
+
+The JsonLogic rules for each individual field will be stored in the database
+alongside the corresponding schema. This means that we don't have to alter code
+when a schema changes.
+
+### Examples
+
+Assuming there is a `sentenceType` field in the JSON document
+that looks like this:
+
+```json
+{
+   "basic-information": {
+      "sentence-type": {
+         "sentenceType": "extendedDeterminate"
+      }
+   }
+}
+```
+
+Then there will be a `sentenceTypeQuery` field in the `application_schemas`
+table with the value:
+
+```json
+{ "var": "basic-information.sentence-type.sentenceType" }
+```
+
+When the application is sumbitted, we will set the `sentenceType` field in the
+`applications` table like so (pseudocode):
+
+```text
+query = application.schema.sentenceTypeQuery # returns '{ "var": "basic-information.sentence-type.sentenceType" }'
+sentenceType = JSONLogic.apply(application.body, query) # returns `extendedDeterminate`
+application.sentenceType = sentenceType
+```
+
+If the schema changes, for example:
+
+```json
+{
+   "basic-info": {
+      "question1": {
+         "sentenceType": "extendedDeterminate"
+      }
+   }
+}
+```
+
+Then there will be a `sentenceTypeQuery` field in the new schema row with
+the value:
+
+```json
+{ "var": "basic-info.question1.sentenceType" }
+```
+
+There will be some situations where some first-class citizen fields (such
+as start date), may have to rely on multiple fields. In the case of start
+date, we will hav a `releaseDateQuery` field like this:
+
+```json
+{
+   "if":[
+      {
+         "==":[
+            {
+               "var":"basic-information.placement-date.startDateSameAsReleaseDate"
+            },
+            "yes"
+         ]
+      },
+      {
+         "var":"basic-information.release-date.releaseDate"
+      },
+      {
+         "var":"basic-information.placement-date.startDate"
+      }
+   ]
+}
+```
+
+This returns the value of `basic-information.release-date.releaseDate` if
+`basic-information.placement-date.startDateSameAsReleaseDate` is `yes`,
+otherwise it will be the value of `basic-information.placement-date.startDate`.
+
+Again, we can set this first-class citizen field at the point of submission
+like so (pseudocode):
+
+```text
+query = application.schema.releaseDateQuery
+releaseDate = JSONLogic.apply(application.body, query)
+application.sentenceType = releaseDate
+```
+
+There are a couple of JsonLogic
+parsers we can use in a Kotlin application, such as [JSON Logic Java](https://github.com/jamsesso/json-logic-java),
+and [JSON Logic Kotlin](https://github.com/advantagefse/json-logic-kotlin).
+
+## Consequences
+
+This will ensure that changing schema will be a deliberate process, and we will
+be able to update the signposting to the first-class citizen fields in the same
+place as the schema, reducing the risk of anything getting missed, and making
+schema changes easier.
+
+There may be some situations when the change to a schema may mean some first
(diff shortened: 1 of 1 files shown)


pezholio:
[Discuss and agree a way forward for first-class citizen fields](https://trello.com/c/H3rSKvxP/797-discuss-and-agree-a-way-forward-for-first-class-citizen-fields)

doc/architecture/decisions/0008-add-first-class-citizen-fields-to-applications.md
```diff
@@ -0,0 +1,122 @@
...
+
+All of these approaches come with advantages and drawbacks:
+
+- saving the additional fields will require work at the API side, we would have to query the JSON object whenever it's updated to draw out the relevant fields. We would also have to consider what happens when the schema is updated
+- updating the UI code would further complicate the code at this level. We would have to update our page objects to pull out this data and marshal it into the correct format to send to the API, which would require extra conditionals and complexities. We also have no guarantee that the data would remain in sync.
+- using the JsonLogic statement in the transformer would require work to ensure this was updated every time the schema changed
+
+## Decision
+
+With all these approaches in mind, we have decided to use the transformers to
```
brownbearsoftware, line 34: Since the frontend already understands the `data` field of an application (and will do for assessments too) I don't think we'd need to return these fields again seperately in the response.  So we'd only need to use the JSON Logic stuff when the API actually needs to make use of a value (e.g. understanding if the application been submitted on a PUT) rather than in the transformer?

pezholio: Hmmm - that's an interesting point actually - maybe we should be using the JSONlogic in the frontend?

pezholio: No, it's just any object at the mo

doc/architecture/decisions/0008-add-first-class-citizen-fields-to-applications.md
```diff
@@ -0,0 +1,122 @@
...
+         "var":"basic-information.release-date.releaseDate"
+      },
+      {
+         "var":"basic-information.placement-date.startDate"
+      }
+   ]
+}
+```
+
+This returns the value of `basic-information.release-date.releaseDate` if
```
richpjames, line 109: I find this examples incredibly helpful, thanks Stu 🙏
Using the first example: 
```json
{
   "basic-information": {
      "sentence-type": {
         "sentenceType": "extendedDeterminate"
      }
   }
}
```
What if instead of `sentence-type` changing, `sentenceType` changed? Would the schema update in the same way?

doc/architecture/decisions/0008-add-first-class-citizen-fields-to-applications.md
richpjames: I'm thinking we should probably start typing `data` for our own convinience... if we did would that help with what we're discussing in this ADR? I'm struggling to see how but probably missing something

brownbearsoftware: Ah I see, the frontend still "understands" the data though right (i.e. that understanding is dynamically typed but it's still present in the source code) so it would feel a little strange to use normal dot/bracket notation access half of the time then JSON Logic sometimes?

pezholio: I think the motivation here is that we want one source of truth for where these fields will come from, and something that is compatible with schema changes. Keeping the two tightly coupled will ensure we're only changing them in one place when we upgrade the schema.

pezholio:
@brownbearsoftware Thinking a bit more on this - I think the use case I was thinking of here was more for returning fields for display, rather than fields for logic and querying. Are the fields like `isWomenOnly` and `isPipePlacement` likely to be things we're going to want to query on? Or are they just going to be things that will be used within the backend code? If it's the latter, then I wonder if there's space for some kind of model that sits between the database model and the transformer that marshals those fields into first-class entities. Apologies if I'm not making sense!

doc/architecture/decisions/0008-add-first-class-citizen-fields-to-applications.md
```diff
@@ -0,0 +1,122 @@
...
+
+```json
+{
+   "basic-information": {
+      "sentence-type": {
+         "sentenceType": "extendedDeterminate"
+      }
+   }
+}
+```
```
edavey, line 56: Probably I'm not understanding properly as this feels like it might be coming across as more complex than it is?

Is it worth stepping through an example of through one of your listed "first class" data fields, such as `is_pipe_application` as another illustration?

The objective, as I understand it, is to keep two data values in lockstep, e.g:

### 1. `applications.is_pipe_application`: 
A simple property on the `applications` table which can readily be queried and indexed e.g. i) to select PIPE applications for allocation to appropriately qualified assessors, ii) to support indexing by Elasticsearch in preparation for matching against the trio of search dimensions: characteristics, availability and location, etc

### 2. `basic-information.ap-type` in JSON `data`
Deep in the JSON `data` blob , where it represents an answer to a particular questions for an in-progress application. 

### Requirements 
- Update [1] if [2] is changed. This can only be done before the application is submitted and whilst the schema version is unchanged
- Leave [1] unaffected if the schema version changes after the application has been submitted. 

So, does that boil down to needing a way of mapping a list of JSON paths to fields on the `applications` table. And that whenever the `data` blob is saved we want to update those mapped fields on `applications`? And to do that we'll need some code (such as JSONLogic) to map `ap-type == 'PIPE'` to a boolean on `applications.is_pipe_application`?

edavey (commented):
Nice work, but I'm still a bit unclear! I think perhaps we need a proper workshop to hammer this out as a group?

doc/architecture/decisions/0008-add-first-class-citizen-fields-to-applications.md
pezholio: The motivation here is to not have to do the first requirement. Ideally we don't want to have to update multiple fields with the same data, as I think that'll make things brittle. If we update the column type to [JSONB](https://www.compose.com/articles/faster-operations-with-the-jsonb-data-type-in-postgresql/), querying by any arbitary field will be easy and relatively fast, but we won't get the advantage of easy access to fields, but maybe that doesn't matter?

doc/architecture/decisions/0008-add-first-class-citizen-fields-to-applications.md
```diff
@@ -0,0 +1,139 @@
...
+- If an application is for a PIPE placement
+- The target location
+- The desired start date
+
+These are important fields that need to be queried on, and made into first class
+citizens within the database table, without sacrificing flexibility.
+
+## Decision
+
+At the point of submission of a completed application, we will use [JsonLogic](https://jsonlogic.com)
```
brownbearsoftware, line 30: So would `submittedAt` remain a top level property that the frontend sets on the PUT request to trigger a submission - or we would use a JSON Logic query each time a PUT is made to pull out a submittedAt from `data` then continue if it's not null?

doc/architecture/decisions/0008-add-first-class-citizen-fields-to-applications.md
brownbearsoftware: Am I right in thinking that with the updated ADR the answer to this one is that we do not return the first class citizen fields on the API response?

doc/architecture/decisions/0008-add-first-class-citizen-fields-to-applications.md
pezholio: This would only be done once when the final application is submitted. I guess we'll need to have a new endpoint that submits the final application

richpjames:
A meta-comment: I think ADRs are one of the cases where ammending/rebasing commits is less helpful than having a history of what's changed between drafts.

doc/architecture/decisions/0008-add-first-class-citizen-fields-to-applications.md
edavey: yes, "submission" seems like an important milestone -- it's not just another type of persistence, it:

- generates the permanent `document`
- prevents further changes to the application
- advances the application to the assessor allocation stage

so an endpoint like `POST applications/{id}/submission` seems appropriate.

edavey (approved):
Looks good, thanks for working through this. 

Do we want to try out some working code first before merging? 

In case there are some further wrinkles to iron out which would otherwise make the examples less accurate?

pezholio:
@edavey Yeah, I think that makes sense. I'm happy to give it a go on this branch, so let's keep this open (assuming @brownbearsoftware is happy with the general idea)

brownbearsoftware:
Yeah the approach makes sense to me 👍

richpjames (approved):

PR #198, opened 2022-11-01
update service name to be approved premises rather than cas1 so that example seeded data is useful

ian-wells-nimble:
This is a migration script that will update seeded example data so that the service name is 'approved-premises' rather than CAS1. 

By doing so means that example data is more meaningful and useful


src/main/resources/db/migration/local+dev/20220913131811__add_ap_service_name_to_mock_premises.sql
```diff
@@ -0,0 +1,6 @@
+UPDATE premises
```
brownbearsoftware, line 1: Can you change the name of the migration so the timestmap is accurate to today?  e.g.

`20221101124300__add_ap_service_name_to_mock_premises.sql`

the PR description here goes over why: https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/197

brownbearsoftware (approved):
LGTM

PR #200, opened 2022-11-02
Refactor characteristic scope matching

danlivings-dxw:
> See [ticket #505 on the CAS3 Trello board](https://trello.com/c/ikNfvcs7/505-refactor-characteristics-to-support-wildcards).

Currently, the logic to validate a characteristic is the responsibility of the `RoomService`, and upcoming work to add characteristics to premises intends to add very similar validation to the `PremisesService`.

Additionally, when a new room is created, any characteristics it is given are only successfully applied if the model scope is `"room"` and the service scope is equal to the value of the `service` column on the relevant premises's table. Characteristics with wildcard scopes (which should always be considered valid) fail to apply.

This PR creates a single point of access for characteristic matching for entities. To do this, the current characteristic validation logic from `RoomService` has been moved into its own `CharacteristicService` and modified to support the upcoming premises characteristics work. As the canonical implementation of characteristic matching, it also adds support for `"*"` as a wildcard in either the model or service scope (or both).


ian-wells-nimble (approved):
LGTM

PR #203, opened 2022-11-03
Premises characteristics can be saved

ian-wells-nimble:
A premises can have characteristics associated with it such e.g Floor Level Access, etc. For a premises these characteristics need to be persisted in a separate link entity table called '**premises_characteristics**'. 

This is so that it can hold an id reference to the premises and an id reference to the particular characteristic. When a client makes a request to GET a premises then the list of characteristics will be included as it joins the tables together. Any entry present in the premises_characteristics table is implied that it is checked on the UI due to its presence in the table.


brownbearsoftware (approved):
LGTM

PR #204, opened 2022-11-03
Fix for StaffUserDetails deserialization issue

brownbearsoftware:
telephoneNumber is actually nullable


pezholio (approved):
LGTM - ta for the super-quick fix!

PR #206, opened 2022-11-03
819 - Get all assessments

brownbearsoftware:
**Few small changes to the Assessments Open API spec**

- Add a createdByUserId to clarification notes, so we can tell who has created each note in cases where assessment is reallocated or an admin adds a note to an assessment they are not allocated to
- Remove createdAt from NewClarificationNote - this should be determined by the API rather than trusting user input
- Make data optional on Assessment - this would be null until the first section of the assessment has been completed

**Groundwork for Application Clarification Notes**
Adds a table, entity etc. since these need to be returned with the assessment responses

**Implement GET /assessments**
Returns all assessments if the user is a workflow manager, otherwise returns the ones that are allocated to the user.  If any problems are encountered fetching futher info on individual assessments log an error and omit that assessment from the response to prevent a single issue breaking the endpoint entirely.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/AssessmentController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/AssessmentController.kt
@@ -1,0 +1,61 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.controller
+
+import org.slf4j.LoggerFactory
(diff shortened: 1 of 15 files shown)


ian-wells-nimble (approved):
Looks great. Now a paradigm is well and truly established on doing additions/changes like this is getting clearer each time to review these.

brownbearsoftware:
[CRUD endpoints for Assess](https://trello.com/c/mRzWKVmr/819-crud-endpoints-for-assess)

PR #210, opened 2022-11-04
Split out submit endpoint for Applications / Simplify schema version checks

brownbearsoftware:
**Split out submission endpoint in OpenAPI spec**

No longer use the presence of `submittedAt` on a PUT request to indicate a submission - rather split out a dedicated /submission endpoint for this.  This endpoint also accepts the denormalised JSON document of the application.

**Remove submission logic from PUT endpoint**

Remove the code that creates an assessment when updating an application if `submittedAt` is true.

**Simplify schema outdated checks**

Enforce that all applications must have a schema version.  Do not attempt to upgrade schema versions anymore - this can't work because the JSON blob won't ever pass validation until it is 100% complete.  Instead, on GETs return that the schema is outdated if a new schema version has been released.  

Do not allow updates to applications with an outdated schema - return a 400 instead.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ApplicationService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ApplicationService.kt
@@ -71,13 +71,19 @@ class ApplicationService(
   }
 
   fun updateApplication(applicationId: UUID, data: String, isWomensApplication: Boolean?, isPipeApplication: Boolean?, username: String): AuthorisableActionResult<ValidatableActionResult<ApplicationEntity>> {
-    val application = applicationRepository.findByIdOrNull(applicationId)
+    val application = applicationRepository.findByIdOrNull(applicationId)?.let(jsonSchemaService::checkSchemaOutdated)
       ?: return AuthorisableActionResult.NotFound()
 
     val user = userService.getUserForRequest()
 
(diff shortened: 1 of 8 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ApplicationService.kt
```diff
@@ -71,7 +70,7 @@ class ApplicationService(
     return success(createdApplication.apply { schemaUpToDate = true })
   }
 
-  fun updateApplication(applicationId: UUID, data: String, document: String?, isWomensApplication: Boolean?, isPipeApplication: Boolean?, submittedAt: OffsetDateTime?, username: String): AuthorisableActionResult<ValidatableActionResult<ApplicationEntity>> {
+  fun updateApplication(applicationId: UUID, data: String, isWomensApplication: Boolean?, isPipeApplication: Boolean?, username: String): AuthorisableActionResult<ValidatableActionResult<ApplicationEntity>> {
```
pezholio, line 73: Probably not for now, but we probs won't need `isWomensApplication`, `isPipeApplication` etc going forward, as we'll be calculating these on submission

pezholio (approved):
LGTM 👍

brownbearsoftware:
[Implement first-class citizen field work](https://trello.com/c/gzBIhYKC/847-implement-first-class-citizen-field-work)

PR #212, opened 2022-11-07
Implement 'get room' endpoint

danlivings-dxw:
> See [ticket #521 on the CAS3 Trello board](https://trello.com/c/HUy6KtN1/521-a-user-can-see-details-of-a-bedspace).

Currently, there is no way to get details of a single room in the API. This means that the client has to use the `GET /premises/{premisesId}/rooms` endpoint to get all rooms in a premises, and then filter this list to get a particular room. This is an inconvenience for the implementation of the 'view/edit room details' pages.

This PR provides the `GET /premises/{premisesId}/rooms/{roomId}` endpoint to allow the client to just request the details of a single room, which simplifies the client-side implementation for viewing and editing rooms.

### DIFF
--- a/src/main/resources/static/api.yml
+++ b/src/main/resources/static/api.yml
@@ -697,6 +697,42 @@ paths:
         500:
           $ref: '#/components/responses/500Response'
   /premises/{premisesId}/rooms/{roomId}:
+    get:
+      summary: Returns a specific room for a premises
+      parameters:
+        - name: premisesId
+          in: path
+          description: ID of the premises the room is in
+          required: true
+          schema:
+            type: string
+            format: uuid
+        - name: roomId
+          in: path
+          description: ID of the room to get
+          required: true
+          schema:
+            type: string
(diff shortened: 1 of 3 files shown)


src/main/resources/static/api.yml
```diff
@@ -697,6 +697,42 @@ paths:
...
+        - name: premisesId
+          in: path
+          description: ID of the premises the room is in
+          required: true
+          schema:
+            type: string
+            format: uuid
+        - name: roomId
+          in: path
+          description: ID of the room to update
```
brownbearsoftware, line 712: think this should be `ID of the room to get`

brownbearsoftware (approved):
LGTM

brownbearsoftware (approved):

bryangaledxw:
Given this a try with https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/12, looks good!

PR #215, opened 2022-11-08
Fix for Prison API docker container

brownbearsoftware:
Because Prison API is now using a replica datasource for certain read-only requests we need to point this at the persisted nomis-db directory too

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -72,6 +72,9 @@ services:
       - SPRING_DATASOURCE_URL=jdbc:hsqldb:file:/nomis-db/nomisdb;sql.syntax_ora=true;get_column_name=false;shutdown=false;sql.nulls_first=false;sql.nulls_order=false;hsqldb.lock_file=false
       - SPRING_DATASOURCE_USERNAME=sa
       - SPRING_DATASOURCE_PASSWORD=password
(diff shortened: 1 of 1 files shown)


pezholio (approved):

docker-compose.yml
```diff
@@ -72,6 +72,9 @@ services:
       - SPRING_DATASOURCE_URL=jdbc:hsqldb:file:/nomis-db/nomisdb;sql.syntax_ora=true;get_column_name=false;shutdown=false;sql.nulls_first=false;sql.nulls_order=false;hsqldb.lock_file=false
       - SPRING_DATASOURCE_USERNAME=sa
       - SPRING_DATASOURCE_PASSWORD=password
+      - SPRING_REPLICA_DATASOURCE_URL=jdbc:hsqldb:file:/nomis-db/nomisdb;sql.syntax_ora=true;get_column_name=false;shutdown=false;sql.nulls_first=false;sql.nulls_order=false;hsqldb.lock_file=false
```
edavey, line 75: what's the "spring replica datasource"? Would it be useful to have a bit more narrative here about these missing environment variables?

edavey: oh, I see now from the PR message that there's a change in the Prison API which is introducing this "replica"? A link to that commit on the Prison repo might be worth capturing in the commit message?

PR #222, opened 2022-11-09
822 - ACCT endpoint

brownbearsoftware:
Implement GET ACCT Alerts endpoint

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/client/PrisonsApiClient.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/client/PrisonsApiClient.kt
@@ -7,6 +7,7 @@ import org.springframework.stereotype.Component
 import org.springframework.web.reactive.function.client.WebClient
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.config.IS_NOT_SUCCESSFUL
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.prisonsapi.AdjudicationsPage
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.prisonsapi.Alert
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.prisonsapi.CaseNotesPage
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.prisonsapi.InmateDetail
 import java.time.LocalDate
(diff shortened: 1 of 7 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PeopleController.kt
```diff
@@ -5,6 +5,7 @@ import org.springframework.stereotype.Service
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.PeopleApiDelegate
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Adjudication
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Person
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.PersonAcctAlert
```
edavey, line 8: I'm still finding it confusing that this `PersonAcctAlert` class is a build artefact only (`build/generated/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/api/model/PersonAcctAlert.kt`) and not part of the repo / narrative, but not sure there anything we can do about this?

edavey:
Also, I think [we have a decision](https://trello.com/c/vQVk4Hoi/822-api-work-to-support-exposing-acct-info#comment-636cb42d2185ea01717105f3) on the active/historical question: we do want to show historical (no longer active) ACCT alerts.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/client/PrisonsApiClient.kt
```diff
@@ -34,4 +35,8 @@ class PrisonsApiClient(
 
     path = "/api/offenders/$nomsNumber/adjudications"
   }
+
+  fun getAlerts(nomsNumber: String, alertCode: String) = getRequest<List<Alert>> {
+    path = "/api/offenders/$nomsNumber/alerts/v2?alertCodes=HA"
+  }
```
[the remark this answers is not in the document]
brownbearsoftware: Good shout, have done 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PeopleController.kt
brownbearsoftware: We could probably change the config so that the generated classes are output into a different location in the normal `src` directory and check them into source control.

I feel quite strongly that we shouldn't though, since generating the models from the OpenAPI spec is deterministic I'd argue there's nothing to review so it'd be adding extra LOC to pull requests for no benefit.  i.e. Since we've already reviewed the OpenAPI spec changes, we'd be reviewing the same thing twice with the caveat that we can't really change the generated code anyway.

PR #226, opened 2022-11-11
Add 'status' property to Premises

danlivings-dxw:
> See [ticket #482 on the CAS3 Trello board](https://trello.com/c/ewdT1APV/482-properties-can-have-a-status).

Currently, there is no way to distinguish between premises that are in active service from those that are not yet or no longer active.

This PR modifies the request bodies of the `POST /premises/` and `PUT /premises/{premisesId}/` endpoints to require a `status` field containing one of the following values:
- `"active"`
- `"pending"`
- `"archived"`

The `Premises` schema has also been updated to include this status with the effect that any endpoint that includes a `Premises` in its response body will include the provided status. In the case of premises that already exist on the database, this is given a default value of `"active"`.

### DIFF
--- a/src/main/resources/db/migration/all/20221110102259__add_status_column_to_premises_table.sql
+++ b/src/main/resources/db/migration/all/20221110102259__add_status_column_to_premises_table.sql
@@ -1,0 +1,1 @@
+ALTER TABLE premises ADD COLUMN status TEXT NOT NULL DEFAULT 'active';
--- a/src/main/resources/static/api.yml
+++ b/src/main/resources/static/api.yml
@@ -1635,6 +1635,8 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/Characteristic'
+        status:
+          $ref: '#/components/schemas/PropertyStatus'
       discriminator:
         propertyName: service
         mapping:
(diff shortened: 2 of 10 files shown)


src/main/resources/db/migration/all/20221110102259__add_status_column_to_premises_table.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE premises ADD COLUMN status TEXT;
```
brownbearsoftware, line 1: Since this isn't nullable in the Kotlin JPA model I think it'd be sensible to make it not-nullable in the DB schema too.  Probably sensible to set the Premises that are already in dev to active.

src/main/resources/static/api.yml
```diff
@@ -1635,6 +1635,9 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/Characteristic'
+        status:
```
brownbearsoftware, line 1638: Think this ought to be a `'#/components/schemas/PropertyStatus'` too so it's clear from Swagger which values are acceptable

brownbearsoftware (approved):
LGTM

PR #229, opened 2022-11-15
847 - Split out Json Schema tables

brownbearsoftware:
**Split out json_schemas table into approved_premises_application_json_schema and approved_premises_assessment_json_schema**

As we need to store schema-specific JSON logic fields to extract first class citizen fields from the `data` blob on applications we need multiple tables.  This has been done in the same way we split out `Premises` between CAS1/CAS3.

The table names are qualified with `approved_premises_` in anticipation of there needing to exist a `temporary_accommodation_application_json_schema` at some point in the future.

This PR doesn't actually make use of the JSON logic fields yet, will be done in a subsequent PR.


pezholio (approved):
Nice work - a lot of find/replace going on there too! 😆 Looks good to me 👍

PR #230, opened 2022-11-16
847 - Submit application endpoint

brownbearsoftware:
Use JSON Logic fields to determine first class citizens, store them in the AP Application table and then create/allocate an assessment.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/ApplicationServiceTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/ApplicationServiceTest.kt
@@ -2,11 +2,13 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.unit.service
 
 import io.mockk.every
 import io.mockk.mockk
+import io.mockk.verify
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.jupiter.api.Test
 import org.springframework.data.repository.findByIdOrNull
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.factory.ApplicationEntityFactory
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.factory.ApprovedPremisesApplicationJsonSchemaEntityFactory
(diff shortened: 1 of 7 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/ApplicationServiceTest.kt
```diff
@@ -331,4 +336,149 @@ class ApplicationServiceTest {
...
+
+    val result = applicationService.submitApplication(applicationId, "{}", username)
+
+    assertThat(result is AuthorisableActionResult.Success).isTrue
+    result as AuthorisableActionResult.Success
+
+    assertThat(result.entity is ValidatableActionResult.Success).isTrue
+    val validatableActionResult = result.entity as ValidatableActionResult.Success
+    assertThat(validatableActionResult.entity.isPipeApplication).isTrue
+    assertThat(validatableActionResult.entity.isWomensApplication).isFalse
```
pezholio, line 479: I don't quite follow how these resolve to `True` and `False` here - might be better if we're specific when setting up the schema

pezholio (approved):
LGTM - could do with maybe making one of the tests clearer. Might be me misunderstanding something though!

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/ApplicationServiceTest.kt
brownbearsoftware: the return values are hardcoded on lines 464/465 since `JsonLogicService` is responsible for correctly resolving the answers.  i.e. in this test the json logic rules on the schema that's produced on line 443 don't really have any effect so figured it'd perhaps be misleading to set them up to look realistic there.

We are covering the full flow in the integration test too though (i.e. the rules are actually read from the schema and applied against the application's data)

pezholio: Ah, yeah, right you are. That's fine then 👍  Knew I was missing _something_!

PR #235, opened 2022-11-17
Implement bookings for Temporary Accommodation

danlivings-dxw:
> See [ticket #467 on the CAS3 Trello board](https://trello.com/c/RqgsUkHV/467-a-user-can-create-a-new-allocated-booking-for-a-person).

This PR implements bookings for a bed in a way that is compatible with the premises-level bookings for the Approved Premises team.

Before creating the booking, it validates that:
- The bed exists on the premises
- The departure date is not before the start date
- A booking does not already exist for that bed during the given dates

A clashing booking results in a `409 Conflict` response from the server.


brownbearsoftware (approved):
Lovely stuff

PR #240, opened 2022-11-21
Only trigger UI type generation workflow if api.yml has changed

brownbearsoftware:
Only trigger the generate types workflow in `hmpps-approved-premises-ui` when a change to api.yml was merged


pezholio (approved):

brownbearsoftware:
[Trigger UI type generation when API is updated](https://trello.com/c/Zz0RmxpR/902-trigger-ui-type-generation-when-api-is-updated)

PR #241, opened 2022-11-21
OASys Risk Management endpoint

brownbearsoftware:


brownbearsoftware:
[Implement OASys risk management plan endpoint](https://trello.com/c/U6pUsIum/908-implement-oasys-risk-management-plan-endpoint)

src/main/resources/static/api.yml
```diff
@@ -3243,6 +3243,12 @@ components:
         assessmentId:
           type: object
           $ref: '#/components/schemas/OASysAssessmentId'
+        assessmentState:
+          $ref: '#/components/schemas/OASysAssessmentState'
+        date:
+          type: string
+          format: date-time
+          description: If the assessment is completed this is the date it was completed, otherwise it is the date it was initiated
```
edavey, line 3251: hmm, on consideration I'm not sure this is actually very helpful...

Perhaps it would be clearer to always return both:

- `dateStarted: date`, and
- `dateCompleted: date or null`

That way the UI can display the date information more flexibly?

edavey (approved):
Looks good, though I am not sure my earlier thoughts about "date" were properly thought-through (or communicated?)

Sorry!

src/main/resources/static/api.yml
brownbearsoftware: Have updated 👍 will follow up with PR for same change to offence analysis endpoint

PR #245, opened 2022-11-21
909 - OASys RoSH summary endpoint

brownbearsoftware:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/RoshSummaryTransformer.kt
```diff
@@ -0,0 +1,23 @@
...
+    assessmentId = roshSummary.assessmentId,
+    assessmentState = if (roshSummary.dateCompleted != null) OASysAssessmentState.completed else OASysAssessmentState.incomplete,
+    dateStarted = roshSummary.initiationDate,
+    dateCompleted = roshSummary.dateCompleted,
+    riskOfSeriousHarmSummary = listOf(
+      oASysQuestionWithSingleAnswer("Who at Risk", "R10.1", roshSummary.roshSummary.whoAtRisk),
+      oASysQuestionWithSingleAnswer("Nature of Risk", "R10.2", roshSummary.roshSummary.natureOfRisk),
+      oASysQuestionWithSingleAnswer("Risk Greatest", "R10.3", roshSummary.roshSummary.riskGreatest),
+      oASysQuestionWithSingleAnswer("Risk Increase Likely To", "R10.4", roshSummary.roshSummary.riskIncreaseLikelyTo),
+      oASysQuestionWithSingleAnswer("Risk Reduction Likely To", "R10.5", roshSummary.roshSummary.riskReductionLikelyTo)
```
edavey, line 20: It might be slightly more helpful to use the actual form of words from OASys? They are:

"Who is at risk"
"What is the nature of the risk"
"When is the risk likely to be the greatest"
"What circumstances are likely to increase risk"
"What circumstances are likely to reduce the risk"

edavey (approved):
Nice work!

Just a tiny suggestion on the "labels" -- and I provide some accurate wording from OASys.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/RoshSummaryTransformer.kt
brownbearsoftware: Ah these make a lot more sense 👍

PR #247, opened 2022-11-22
905 - Select index offence OApi spec

brownbearsoftware:
This would alter the flow such that:

- When rendering the confirm identity screen, frontend also makes a call to the `/people/{crn}/offences` endpoint
- If there is more than one active offence, a selector is displayed
- When creating the application, the `convictionId`, `eventNumber` & `offenceId` for the selected entry are passed back to the API - these can then be stored on the `applications` table and used by the backend later when generating events, finding documents etc.

**If we merge this, it will break creating an application on the frontend until values are provided for convictionId, eventNumber, offenceId**

Perhaps we could provide some placeholder values until the designs are done to allow me to continue with the backend?

### DIFF
--- a/src/main/resources/static/api.yml
+++ b/src/main/resources/static/api.yml
@@ -1167,6 +1167,37 @@ paths:
                 $ref: '#/components/schemas/Problem'
         500:
           $ref: '#/components/responses/500Response'
+  /people/{crn}/offences:
+    get:
+      summary: Returns the active Offences for a Person
+      parameters:
+        - name: crn
+          in: path
+          description: CRN of the Person to fetch active Offences for
+          required: true
+          schema:
+            type: string
+      responses:
+        200:
+          description: successful operation
+          content:
+            'application/json':
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/ActiveOffence'
+        401:
+          $ref: '#/components/responses/401Response'
+        403:
+          $ref: '#/components/responses/403Response'
+        404:
+          description: invalid CRN
+          content:
+            'application/json':
+              schema:
+                $ref: '#/components/schemas/Problem'
+        500:
+          $ref: '#/components/responses/500Response'
   /premises/{premisesId}/lost-beds:
     post:
       tags:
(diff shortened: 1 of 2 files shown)


brownbearsoftware:
[Record DeliusEventNumber (index offence)](https://trello.com/c/JHG1hZqx/839-record-deliuseventnumber-index-offence)

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ApplicationTest.kt
```diff
@@ -474,7 +474,10 @@ class ApplicationTest : IntegrationTestBase() {
       .header("Authorization", "Bearer $jwt")
       .bodyValue(
         NewApplication(
-          crn = crn
+          crn = crn,
+          convictionId = 123,
+          eventNumber = "123",
```
edavey, line 479: `eventNumber = "5"`

would be a more realistic value

src/main/resources/static/api.yml
```diff
@@ -2805,8 +2805,17 @@ components:
       properties:
         crn:
           type: string
+        convictionId:
+          type: long
+        eventNumber:
+          type: string
+        offenceId:
+          type: string
```
edavey, line 2813: would it be helpful to include example values here?

e.g.

```
convictionId:
  type: long
  example: 1502724704
eventNumber:
  type: string
  example: "7"
offenceId:
  type: string
  example: "M1502750438"
```

edavey (commented):
Do we need to implement [service-specific applications](https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/main/doc/architecture/decisions/0008-service-specific-applications.md) before we can make this breaking change?

jitJoe:
@edavey that's a good question!  @tahb / @danlivings-dxw do either of you know if you'll need this for CAS3 too?  The uses we have for this so far are:
- Emitting domain events to add NSI/Contacts in Delius (although it seems to be tentative if we'll actually end up index number from our call earlier)
- Fetching Delius/Alfresco documents from Community API (since in some cases there can be more than one active offence and we only want the documents for one that the PP selects)

brownbearsoftware:
[@edavey](https://github.com/edavey) that's a good question! [@tahb](https://github.com/tahb) / [@danlivings-dxw](https://github.com/danlivings-dxw) do either of you know if you'll need this for CAS3 too? The uses we have for this so far are:

- Emitting domain events to add NSI/Contacts in Delius (although it seems to be tentative if we'll actually end up index number from our call earlier)
- Fetching Delius/Alfresco documents from Community API (since in some cases there can be more than one active offence and we only want the documents for one that the PP selects)

brownbearsoftware:
Closing this as the relevant bits (+ the suggestions) have been included in other PRs

PR #253, opened 2022-11-25
Add Temporary Accommodation support to arrivals on a booking

danlivings-dxw:
> See [ticket #536 on the CAS3 Trello board](https://trello.com/c/GC1EQ1p0/536-a-user-can-mark-a-confirmed-booking-as-active).

This PR is part of the work to support the Temporary Accommodation booking flow. It allows a booking on a Temporary Accommodation premises to record an arrival.

Changes in this PR:
- The `NewArrival.keyWorkerStaffCode` property has been made optional. Validation has been added to ensure that an Approved Premises arrival still contains this value, and returns a `400 Bad Request` response if not.
- An arrival to a Temporary Accommodation premises will validate the actual arrival date and the expected departure date to make sure that (in case they are different to the booking) there are no conflicts with other bookings for the same bedspace.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
@@ -32,6 +32,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.UpdatePremises
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.UpdateRoom
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApprovedPremisesEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.BookingEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.PremisesEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.ValidationErrors
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.BadRequestProblem
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.ConflictProblem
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
```diff
@@ -261,24 +262,12 @@ class PremisesController(
     }
 
     if (body.departureDate.isBefore(body.arrivalDate)) {
-      validationErrors["$.departureDate"] = "departureBeforeArrival"
+      validationErrors["$.departureDate"] = "beforeBookingArrivalDate"
```
brownbearsoftware, line 265: I believe there's a map in the frontends for these keys - does this need a corresponding change in `hmpps-approved-premises-ui`?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
```diff
@@ -334,6 +323,16 @@ class PremisesController(
   ): ResponseEntity<Arrival> {
     val booking = getBookingForPremisesOrThrow(premisesId, bookingId)
 
+    if (booking.service == ServiceName.temporaryAccommodation.value) {
+      // TODO: Arrivals will likely need to check for overlaps once bed-level bookings are implemented for AP
+      val bedId = when (booking.bed) {
```
brownbearsoftware, line 328: Bikeshedding so feel free to ignore, but this might be a bit easier to read?

```
val bedId = booking.bed?.id 
      ?: throw InternalServerErrorProblem("No bed ID present on Temporary Accommodation booking: $bookingId")
```

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
danlivings-dxw: This check was introduced specifically for Temporary Accommodation a little while back - it's just being changed to be consistent with the key used when creating an arrival. I just did a check and `"departureBeforeArrival"` doesn't show up anywhere in `hmpps-approved-premises-ui` so it shouldn't cause any issues there.

PR #258, opened 2022-11-29
Add service scopes to departure reasons and move-on categories

danlivings-dxw:
> See ticket [#537 on the CAS3 Trello board](https://trello.com/c/6ey0Kwp2/537-a-user-can-mark-an-occupied-bed-as-closed).

This PR is part of the work to support the Temporary Accommodation booking flow. It allows departure reasons and move-on categories to be service-specific.

Changes in this PR:
- The `DepartureReason` and `MoveOnCategory` OpenAPI schemas have had a `serviceScope` property added, to be used in a similar way to `Characteristic.serviceScope`.
- A `service_scope` column has been added to the `departure_reasons` and `move_on_categories` tables. Existing rows in these tables have been assigned either a scope of `"approved-premises"` or `"*"` (the wildcard scope) depending on whether they are specific to AP or will also be used by TA. New rows have been assigned a scope of `"temporary-accommodation"`. ~~**NB: Some departure reasons are still TBC, so these are subject to change.**~~ *We've made the decision on the CAS3 team to reuse AP's departure reasons for 'bed withdrawn' and 'breach/recall' provisionally and to update these as necessary in the future.*
- The `GET /reference-data/departure-reasons` and `GET /reference-data/move-on-categories` endpoints now accept `X-Service-Name` as a header, which if provided, provides only the items which match the given scope. The behaviour if no header is supplied is to return all items to maintain backwards compatibility.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ReferenceDataController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ReferenceDataController.kt
@@ -60,14 +60,20 @@ class ReferenceDataController(
     return ResponseEntity.ok(localAuthorities.map(localAuthorityAreaTransformer::transformJpaToApi))
   }
 
-  override fun referenceDataDepartureReasonsGet(): ResponseEntity<List<DepartureReason>> {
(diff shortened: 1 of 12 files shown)


brownbearsoftware (approved):
LGTM

PR #264, opened 2022-11-30
Remove individual OASys endpoints/Implement unified endpoint

brownbearsoftware:
This removes the individual OASys section endpoints and replaces them with a single endpoint.

Will follow up with another PR to add the call to the upstream health endpoint as described in the last few comments on the Trello ticket.


brownbearsoftware:
[Implement new oasys/sections endpoint](https://trello.com/c/xiQd1Mu7/923-implement-new-oasys-sections-endpoint)

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/PersonOASysSectionsTest.kt
```diff
@@ -0,0 +1,215 @@
...
+    )
+
+    val jwt = jwtAuthHelper.createAuthorizationCodeJwt(
+      subject = "username",
+      authSource = "delius",
+      roles = listOf("ROLE_PROBATION")
+    )
+
+    webTestClient.get()
+      .uri("/people/CRN/oasys/sections?selected-sections=11&selected-sections=12")
```
edavey, line 196: it's interesting how Spring / OpenAPI handle this repeated url param and know to accumulate the values into a list...

edavey (approved):
This looks correct and is easy to follow.

I've made a bunch of minor comments:

- lots of tweaks to labels
- a couple of offence section fields to drop

src/main/resources/static/api.yml
```diff
@@ -2947,182 +2986,166 @@ components:
...
+        label:
+          type: string
+        sectionNumber:
+          type: integer
+        questionNumber:
+          type: string
+        linkedToHarm:
+          type: boolean
+        linkedToReOffending:
+          type: boolean
```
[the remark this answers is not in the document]
brownbearsoftware: Whether the properties of the generated kotlin models are nullable is determined by whether the property here appears in the `required` list which it doesn't in this case - so already typed as `Boolean?` in the kotlin models 👍

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/PersonOASysSectionsTest.kt
brownbearsoftware: I tested this in Postman too - it seems happy with either:
- `key=value1&key=value2`
- `key=value1,value2

But wouldn't take `key=value1|value2`

PR #267, opened 2022-12-01
Pass LocalDateTime instead of LocalDate to Offender Case Notes Service

brownbearsoftware:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/client/CaseNotesClient.kt
```diff
@@ -6,13 +6,16 @@ import org.springframework.stereotype.Component
...
+import java.time.LocalTime
 
 @Component
 class CaseNotesClient(
   @Qualifier("caseNotesWebClient") webClient: WebClient,
   objectMapper: ObjectMapper
 ) : BaseHMPPSClient(webClient, objectMapper) {
   fun getCaseNotesPage(nomsNumber: String, from: LocalDate, page: Int, pageSize: Int) = getRequest<CaseNotesPage> {
-    path = "/case-notes/$nomsNumber?startDate=$from&page=$page&size=$pageSize"
+    val fromLocalDateTime = LocalDateTime.of(from, LocalTime.MIN)
```
edavey, line 18: interesting!

edavey (approved):
👍

PR #269, opened 2022-12-01
Use probation regions when creating or updating a premises

danlivings-dxw:
> See [ticket #529 on the CAS3 Trello board](https://trello.com/c/Jnqzc0pE/529-users-can-assign-a-property-to-a-region).

Currently, when creating or updating a premises, a mock probation region is used that does not correspond to any real location. This PR allows for the probation region a premises is in to be specified when creating or updating it.

Changes in this PR:
- A `GET /reference-data/probation-regions` endpoint has been implemented to provide a list of all available probation regions.
- The `NewPremises` and `UpdatePremises` schemas in the OpenAPI specification have been updated to require a UUID for the `probationRegionId` property.
- The `PremisesService` validates that the provided `probationRegionId` corresponds to an actual probation region in the database.


brownbearsoftware (approved):
LGTM

PR #270, opened 2022-12-02
915 - Switch to OASys for RoSH ratings

brownbearsoftware:
**Use the AP OASys Context API for RoSH ratings**

**Remove ARNS scaffolding**
This was the last endpoint from ARNS we were using, so remove all of the scaffolding too


pezholio (approved):
LGTM - just a couple of smallish things

PR #271, opened 2022-12-02
Update PIPE JsonLogic rule

pezholio:
The original rule was invalid JSON


brownbearsoftware (approved):
LGTM, cheers for sorting!

PR #273, opened 2022-12-05
905 - List Active Offences endpoint

brownbearsoftware:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/community/Conviction.kt
```diff
@@ -0,0 +1,45 @@
...
+  val offences: List<Offence>?
+)
+
+data class Offence(
+  val offenceId: String,
+  val mainOffence: Boolean,
+  val detail: OffenceDetail,
+  val offenceDate: LocalDateTime,
+  val offenceCount: Long,
+  val tics: Long,
```
edavey, line 25: I think we'll need to mark `tics` as optional as I don't see it in responses I've obtained...

brownbearsoftware:

src/main/resources/static/api.yml
```diff
@@ -3225,3 +3256,22 @@ components:
...
         - Completed
         - Incomplete
+    ActiveOffence:
+      type: object
+      properties:
+        deliusEventNumber:
+          type: string
+          example: "7"
+        offenceDescription:
+          type: string
```
edavey, line 3266: perhaps an example value would be helpful here? e.g.

`"Wounding or inflicting grievous bodily harm (inflicting bodily injury with or without weapon) (S20) - 00801"`

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/ConvictionTransformer.kt
```diff
@@ -0,0 +1,17 @@
...
+import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ActiveOffence
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.community.Conviction
+
+@Component
+class ConvictionTransformer {
+  fun transformToApi(conviction: Conviction) = conviction.offences?.map {
+    ActiveOffence(
+      deliusEventNumber = conviction.index,
+      offenceDescription = "${it.detail.mainCategoryDescription} - ${it.detail.subCategoryDescription}",
```
edavey, line 12: 👍 this makes sense as sometimes (though not usually) the sub category description provides additional detail which may be important. Perhaps a "nice to have" would be a function which only includes the sub-category description if it is different from the main-category description...

edavey (approved):
👍 I've left a couple of minor comments.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/community/Conviction.kt
brownbearsoftware: Ah cheers for this!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/ConvictionTransformer.kt
brownbearsoftware: Good shout - have added a check to see if they're the same

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/transformer/ConvictionTransformerTest.kt
```diff
@@ -0,0 +1,75 @@
...
+      )
+      .produce()
+
+    assertThat(convictionTransformer.transformToApi(conviction)).containsExactlyInAnyOrder(
+      ActiveOffence(
+        deliusEventNumber = "5",
+        offenceDescription = "A Description",
+        offenceId = "1",
+        convictionId = 12345
+      )
```
edavey, line 72: nice, thanks for implementing!

edavey (commented):
🚀

edavey (approved):

PR #274, opened 2022-12-05
905 - Split out applications

brownbearsoftware:
Split Applications into service specific tables, entities and responses.

Previously isWomensApplication and isPipeApplication existed on the applications table but these properties are not relevant to CAS3.  We have a need to add more CAS1 specific properties, namely selected offence and RoSH data so this is a prerequisite to that work.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApplicationEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApplicationEntity.kt
@@ -2,23 +2,32 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity
 
 import org.hibernate.annotations.Type
 import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.data.jpa.repository.Query
 import org.springframework.stereotype.Repository
 import java.time.OffsetDateTime
 import java.util.UUID
+import javax.persistence.DiscriminatorColumn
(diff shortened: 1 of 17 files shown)


brownbearsoftware:

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApplicationEntity.kt
```diff
@@ -18,7 +23,9 @@ interface ApplicationRepository : JpaRepository<ApplicationEntity, UUID> {
```
danlivings-dxw, line 23: If there's a situation where a user can move from CAS1 to CAS3 (or vice versa), it might be worth updating the `ApplicationRepository.findAllByCreatedByUser_Id` method to also filter by the application entity class to avoid applications from the wrong service from showing up:

```kotlin
@Query("SELECT a FROM ApplicationEntity a WHERE TYPE(a) = :type AND a.createdByUser.id == :id")
fun <T : ApplicationEntity> findAllByCreatedByUser_Id(id: UUID, type: Class<T>): List<ApplicationEntity>
```

danlivings-dxw (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApplicationEntity.kt
brownbearsoftware: Good spot, have added in 👍

PR #276, opened 2022-12-06
905 - Provide index offence

brownbearsoftware:
Adds convictionId, eventNumber and offenceId properties to the create application endpoint.

The FE can retrieve these from the get active offences endpoint then pass back the correct one when creating the application.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ApplicationsController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ApplicationsController.kt
@@ -67,7 +67,7 @@ class ApplicationsController(
 
     val (offender, inmate) = getPersonDetail(body.crn)
 
-    val application = when (val applicationResult = applicationService.createApplication(body.crn, username, serviceName)) {
+    val application = when (val applicationResult = applicationService.createApplication(body.crn, username, serviceName, body.convictionId, body.deliusEventNumber, body.offenceId)) {
       is ValidatableActionResult.GeneralValidationError -> throw BadRequestProblem(errorDetail = applicationResult.message)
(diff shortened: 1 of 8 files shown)


brownbearsoftware:

src/main/resources/db/migration/all/20221205160719__add_index_offence_fields_to_ap_applications.sql
```diff
@@ -0,0 +1,5 @@
+TRUNCATE TABLE applications CASCADE;
```
pezholio, line 1: This makes me a bit nervous. Can we just set them to a non-sensible default? I just worry about unintended consequences.

pezholio (approved):
LGTM - just one minor concern about the migrations.

src/main/resources/db/migration/all/20221205160719__add_index_offence_fields_to_ap_applications.sql
brownbearsoftware: Sure thing, have updated

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ApplicationsController.kt
```diff
@@ -67,7 +67,7 @@ class ApplicationsController(
 
     val (offender, inmate) = getPersonDetail(body.crn)
 
-    val application = when (val applicationResult = applicationService.createApplication(body.crn, username, serviceName)) {
+    val application = when (val applicationResult = applicationService.createApplication(body.crn, username, serviceName, body.convictionId, body.eventNumber, body.offenceId)) {
```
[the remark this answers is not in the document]
brownbearsoftware: Have updated 👍

PR #281, opened 2022-12-06
Add mocks for Alfresco

brownbearsoftware:
To support the Documents work, we need Community API to return files when requested.  It delegates to a service called Alfresco for this - this adds mock responses for the endpoints used to always return a pdf which contains `This is a mock file`


brownbearsoftware:
[Fetch documents from Delius](https://trello.com/c/2EtClS4K/897-fetch-documents-from-delius)

edavey (approved):
Nice sleuthing 🕵️

PR #282, opened 2022-12-06
Add applicationUrl property to Domain Events

edavey:
It's important that Domain Events provide a link to the UI of the Approved Premises service, to a resource which will allow a human being to explore the "application" for accommodation and its associated resources e.g.:

- the assessment of the application (including clarification notes)
- any bookings related to the application

For example, when a `booking.made` event is consumed by the Delius integration we expect a Delius "contact" to be created. This contact will be of the new (booking) type `EAPB` and will include the `applicationUrl` e.g.

`https://approved-premises/applications/484b8b5e-6c3b-4400-b200-425bbe410713`

Probation officers using Delius will be able to follow this link and view the (possibly multiple) bookings which have been made and explore the AP "application" in full according to their needs.

### DIFF
--- a/src/main/resources/static/domain-events-api.yml
+++ b/src/main/resources/static/domain-events-api.yml
@@ -428,6 +428,11 @@ components:
       type: string
       example: 484b8b5e-6c3b-4400-b200-425bbe410713
 
+    ApplicationUrl:
+      description: The URL on the Approved Premises service at which a user can view a representation of an AP application and related resources, including bookings
+      type: string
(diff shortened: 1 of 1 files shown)


src/main/resources/static/domain-events-api.yml
```diff
@@ -428,6 +428,11 @@ components:
       type: string
       example: 484b8b5e-6c3b-4400-b200-425bbe410713
 
+    ApplicationUrl:
```
brownbearsoftware, line 431: Think the 'A' should be lowercase?

brownbearsoftware (approved):
LGTM

PR #283, opened 2022-12-07
897 - List Documents endpoint

brownbearsoftware:
Adds a `/applications/{applicationId}/documents` endpoint that returns meta data about the documents associated to:
 - The Offender the Application is for
 - The Conviction the Application is associated with

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/community/GroupedDocuments.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/community/GroupedDocuments.kt
@@ -1,0 +1,41 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.model.community
+
+import java.time.LocalDateTime
+import java.util.UUID
+
+data class GroupedDocuments(
+  val documents: List<OffenderLevelDocument>,
+  val convictions: List<ConvictionDocuments>
+)
+
+data class OffenderLevelDocument(
+  val id: UUID,
+  val documentName: String,
+  val author: String,
+  val type: DocumentType,
+  val extendedDescription: String?,
+  val createdAt: LocalDateTime,
+  val lastModifiedAt: LocalDateTime?,
+  val parentPrimaryKeyId: Long?
+)
+
+data class ConvictionDocuments(
+  val convictionId: String,
+  val documents: List<ConvictionLevelDocument>
+)
+
+data class ConvictionLevelDocument(
+  val id: UUID,
+  val documentName: String,
+  val author: String,
+  val type: DocumentType,
+  val extendedDescription: String?,
+  val lastModifiedAt: LocalDateTime?,
+  val createdAt: LocalDateTime,
+  val parentPrimaryKeyId: Long?
+)
+
+data class DocumentType(
+  val code: String,
+  val description: String
+)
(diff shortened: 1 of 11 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/community/GroupedDocuments.kt
```diff
@@ -0,0 +1,40 @@
...
+  val documents: List<OffenderLevelDocument>,
+  val convictions: List<ConvictionDocuments>
+)
+
+data class OffenderLevelDocument(
+  val id: UUID,
+  val documentName: String,
+  val author: String,
+  val type: DocumentType,
+  val extendedDescription: String,
```
edavey, line 16: I think `extendedDescription` needs to be optional as it is often missing...

edavey: I think we also need an optional `parentPrimaryKeyId` on the `OffenderLevelDocument` as this can be present here as well as within the conviction docs

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/factory/OffenderLevelDocumentFactory.kt
```diff
@@ -0,0 +1,66 @@
...
+
+class OffenderLevelDocumentFactory : Factory<OffenderLevelDocument> {
+  private var id: Yielded<UUID> = { UUID.randomUUID() }
+  private var documentName: Yielded<String> = { "${randomStringMultiCaseWithNumbers(5)}.pdf" }
+  private var author: Yielded<String> = { randomStringMultiCaseWithNumbers(4) }
+  private var typeCode: Yielded<String> = { randomStringMultiCaseWithNumbers(4) }
+  private var typeDescription: Yielded<String> = { randomStringMultiCaseWithNumbers(4) }
+  private var extendedDescription: Yielded<String> = { randomStringMultiCaseWithNumbers(10) }
+  private var lastModifiedAt: Yielded<LocalDateTime?> = { null }
+  private var createdAt: Yielded<LocalDateTime> = { LocalDateTime.now().randomDateTimeBefore(5) }
```
edavey, line 20: see above for `parentPrimaryKeyId` which can be present

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/DocumentTransformer.kt
```diff
@@ -0,0 +1,41 @@
...
+          fileName = it.documentName,
+          createdAt = it.createdAt.atOffset(ZoneOffset.UTC),
+          typeCode = it.type.code,
+          typeDescription = it.type.description,
+          description = it.extendedDescription
+        )
+      } ?: emptyList()
+
+    return offenderDocuments + convictionDocuments
+  }
```
edavey, line 40: very neat and clear

edavey (approved):
Nice work, just a couple of tiny tweaks needed to optional properties found in the upstream `documents/grouped` response, I think...

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/community/GroupedDocuments.kt
brownbearsoftware: sorted 👍

PR #287, opened 2022-12-09
Add Temporary Accommodation support to extensions on a booking

danlivings-dxw:
> See [ticket #608 on the CAS3 Trello board](https://trello.com/c/C3QYguIW/608-a-user-can-mark-a-booking-as-extended).

This PR is part of the work to support the Temporary Accommodation booking flow. It allows a booking on a Temporary Accommodation premises to record an extension.

Changes in this PR:
- Bookings have been updated to provide original arrival and departure dates, which are set when the booking is created and never changed. This is distinct from the current arrival and departure dates, which can be changed as a booking changes states. These original dates are required by CAS3 for MI reporting.
- Arrival and departure dates are overwritten for all state transitions that can change these values (i.e. arrivals, departures, and extensions). This behaviour is only applied to TA bookings to preserve the current behaviour for AP bookings.
- Extensions for TA bookings can be used to move a departure date to be earlier as well as later.
- Extensions for TA bookings will check for conflicts with other bookings to ensure that an extension can't cause two bookings to overlap.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
@@ -315,6 +315,8 @@ class PremisesController(
         premises = premises,
         bed = bed,
         service = body.serviceName.value,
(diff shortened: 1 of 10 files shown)


brownbearsoftware (approved):
LGTM

PR #290, opened 2022-12-13
Include filename in Content-Disposition header for Document downloads

brownbearsoftware:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/DocumentsController.kt
```diff
@@ -38,14 +39,16 @@ class DocumentsController(
...
-    }
+    val documentMetaData = documentsMetaData.findDocument(documentId)
+      ?: throw NotFoundProblem(documentId, "Document")
 
     return ResponseEntity(
       StreamingResponseBody { outputStream ->
         offenderService.getDocument(crn, documentId, outputStream)
       },
+      HttpHeaders().apply {
+        put("Content-Disposition", listOf("attachment; filename=\"${documentMetaData.documentName}\""))
```
pezholio, line 50: Why is this a `listOf`?

pezholio (approved):
LGTM  - ta for the quick turnaround! 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/DocumentsController.kt
brownbearsoftware: Any HTTP header that can be represented as a comma seperated list can also be sent as multiple key-value pairs, e.g.

```
Cache-Control: no-cache, no-store
```

is the same as:

```
Cache-Control: no-cache
Cache-Control: no-store
```

So Spring types HttpHeaders as a Multi-Value map, i.e. `Map<String, List<String>>` to support the second case

brownbearsoftware:
[Fetch documents from Delius](https://trello.com/c/2EtClS4K/897-fetch-documents-from-delius)

PR #297, opened 2022-12-15
Refactor OpenAPI specification for Assessments

brownbearsoftware:
Break out Assessment into two subtypes so that the frontend can use the correct one since polymorphism on the `application` property isn't working with the OpenAPI->TS generator.


pezholio (approved):
LGTM 👍  This is what I tried, but didn't think to look at the transformer

PR #301, opened 2022-12-16
903 - Accept seeding jobs

brownbearsoftware:
This adds an endpoint, POST to /seed which can only be called from within the kubernetes pod running the app.  The endpoint a post body with the type of seeding and a filename.  These are passed to a background thread and the call returns a 202 to indicate the work has been accepted.


brownbearsoftware:

PR #304, opened 2022-12-19
903 - Seeding scaffolding

brownbearsoftware:
This lays the foundations for the seeding work.

- Adds an endpoint POST /seed which accepts the type of seeding and a filename
- This endpoint returns 202 Accepted then starts the processing in a background thread
- We check that the entire file can be parsed and deserialized first (if not, collecting errors where possible)
- Then row by row, process each file

This doesn't actually do any processing with the Approved Premises CSV yet as this is more focussed around the reusable parts.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/ApprovedPremisesSeedJob.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/ApprovedPremisesSeedJob.kt
@@ -1,0 +1,49 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.seed
+
+import org.slf4j.LoggerFactory
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.PropertyStatus
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.PremisesRepository
+import java.util.UUID
+
+class ApprovedPremisesSeedJob(
+  fileName: String,
+  private val premisesRepository: PremisesRepository
+) : SeedJob<ApprovedPremisesSeedCsvRow>(
+  id = UUID.randomUUID(),
+  fileName = fileName,
+  requiredColumns = 12
+) {
+  private val log = LoggerFactory.getLogger(this::class.java)
+
+  override fun deserializeRow(columns: Map<String, String>) = ApprovedPremisesSeedCsvRow(
+    id = UUID.fromString(columns["id"]!!),
+    name = columns["name"]!!,
+    addressLine1 = columns["addressLine1"]!!,
+    postcode = columns["postcode"]!!,
+    totalBeds = Integer.parseInt(columns["totalBeds"]!!),
+    notes = columns["notes"]!!,
+    probationRegionId = UUID.fromString(columns["probationRegionId"]!!),
+    localAuthorityAreaId = UUID.fromString(columns["localAuthorityAreaId"]),
+    characteristicIds = columns["characteristicIds"]!!.split(",").map { UUID.fromString(it.trim()) },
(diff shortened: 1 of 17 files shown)


brownbearsoftware:

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/ApprovedPremisesSeedJob.kt
```diff
@@ -0,0 +1,49 @@
...
+    addressLine1 = columns[2],
+    postcode = columns[3],
+    totalBeds = Integer.parseInt(columns[4]),
+    notes = columns[4],
+    probationRegionId = UUID.fromString(columns[5]),
+    localAuthorityAreaId = UUID.fromString(columns[6]),
+    characteristicIds = columns[7].split(",").map { UUID.fromString(it.trim()) },
+    status = PropertyStatus.valueOf(columns[8]),
+    apCode = columns[9],
+    qCode = columns[10]
```
edavey, line 30: Is it possible to address the columns using the column heading e.g. `columns.get("q_code")` rather than `columns[10]`? I have a bit of a horror of the ["connascence of position"](https://connascence.io/position.html) as I find it fragile to maintain and tiring mentally! I saw a [good talk on this years ago by Jim Weirich](https://www.youtube.com/watch?v=HQXVKHoUQxY)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/SeedJob.kt
```diff
@@ -0,0 +1,18 @@
...
+
+import java.util.UUID
+
+abstract class SeedJob <RowType> (
+  val id: UUID = UUID.randomUUID(),
+  val fileName: String,
+  val requiredColumns: Int
+) {
+  init {
+    if (fileName.contains("/") || fileName.contains("\\") || fileName.contains(".")) {
```
edavey, line 11: Did you consider using a filesystem function to verify whether a file of the given name (and expected mime type) exists in the expected seed directory? Perhaps at this point of accepting or rejecting the async job it's not appropriate / possible  to take any external action?

edavey (dismissed):
This seems very clear and re-usable. Thanks for laying it out so clearly. I've made a couple of suggestions / comments.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/ApprovedPremisesSeedJob.kt
brownbearsoftware: Hadn't heard of connascence before but I'll definitely be adding that one to the lexicon.

Interestingly, I had gone for indexes because I've had issues in the past with CSV uploads where the end user downloads the template (with headings correctly named) then decides to change the headers to something slightly different.

In this case though, since we'll be the ones uploading the CSV I'm with you on the index point - have adjusted accordingly.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/SeedJob.kt
brownbearsoftware: I had considered doing the does-file-exist check before returning the 202 but opted against it since we'd then have a situation where some failures are reported synchronously and some aren't (e.g. if the file then happens to be malformed, you'd have to check the logs.)   Can add this anyway though if you feel the pros outweigh the cons?

edavey: right, that makes sense -- I thought it might be in consideration of the async processes 🚀

edavey (approved):

PR #306, opened 2022-12-21
Enable log-client-credentials-jwt-info on preprod

brownbearsoftware:
This setting causes the app to log the client_id, authorities, scopes and expiry when it retrieves a client_credentials JWT for service->service call authentication.  Turning this on in pre-prod to debug an issue with calling Offender Case Notes Service (we need to ensure we have the correct authorities.)  This does not log the JWT signature so is safe to enable.


edavey (approved):
👍

PR #316, opened 2023-01-04
903 user role seeding

brownbearsoftware:
This can be triggered by placing a CSV file in the /seed directory (in dev etc.) or ./seed when running locally with the following format:
```
deliusUsername,roles,qualifications
SomeOne,"ASSESSOR,WORKFLOW_MANAGER","WOMENS,PIPE"
...
```

The job is then started by sending a POST to /seed from within the container (or just from postman if running locally) with the following body (where our CSV file is named user-upload.csv):
```
{
    "seedType": "user",
    "fileName": "user-upload"
}
```
This will return a 202 Accepted to indicate the job has been started. You can then monitor the progress of the job from the API logs.

An initial pass is made to ensure each row can be deserialized. If any rows fails this stage then no processing will happen and an error will be logged explaining the issue that needs to be remediated.

For each user, if they exist in Delius but not yet in our users table, then they are fetchedd from Delius.  If we already have them in our table then just the roles are set.  Roles are set to exactly what is provided in the CSV so any existing role assignments are removed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -75,5 +75,5 @@ sonar-project.properties
 
 nomis-db/
 
-test-seed-csvs/
-seed/
+test-seed-csvs/**
+seed/**
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/WebClientConfiguration.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/WebClientConfiguration.kt
@@ -3,6 +3,11 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.config
 import org.springframework.beans.factory.annotation.Value
 import org.springframework.context.annotation.Bean
 import org.springframework.context.annotation.Configuration
+import org.springframework.security.oauth2.client.AuthorizedClientServiceOAuth2AuthorizedClientManager
+import org.springframework.security.oauth2.client.InMemoryOAuth2AuthorizedClientService
+import org.springframework.security.oauth2.client.OAuth2AuthorizedClientManager
+import org.springframework.security.oauth2.client.OAuth2AuthorizedClientProviderBuilder
+import org.springframework.security.oauth2.client.OAuth2AuthorizedClientService
 import org.springframework.security.oauth2.client.registration.ClientRegistrationRepository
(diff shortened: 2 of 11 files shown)


brownbearsoftware:

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/SeedUserRoleAssignmentsTest.kt
```diff
@@ -0,0 +1,160 @@
...
+    assertThat(persistedUser!!.deliusStaffIdentifier).isEqualTo(6789)
+    assertThat(persistedUser.roles.map(UserRoleAssignmentEntity::role)).containsExactlyInAnyOrder(
+      UserRole.ASSESSOR,
+      UserRole.WORKFLOW_MANAGER
+    )
+    assertThat(persistedUser.qualifications.map(UserQualificationAssignmentEntity::qualification)).containsExactlyInAnyOrder(
+      UserQualification.PIPE
+    )
+  }
+
```
edavey, line 83: Do we also need a spec to describe the behaviour if the seed file contains roles or qualifications which don't exist?

I'm not sure whether:

- the process blows up on `fun addQualificationToUser(user: UserEntity, qualification: UserQualification)` in the `UserService`? 
- or does it blow up in the deserialisation test `roles = columns["roles"]!!.split(",").filter(String::isNotBlank).map(UserRole::valueOf)`?

I'm interested to understand the feedback given to the seeder. Something along the lines of "The role 'WORKSLOW_MANGER' does not exist" would be helpful...

edavey (approved):
This looks really good to me -- but I'd like to understand (and have it documented in specs) how errors in role and qualification names are communicated.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/SeedUserRoleAssignmentsTest.kt
brownbearsoftware: Would have been the latter, good point on conveying this to the user though.  Have altered to make it friendlier and added a few tests.

PR #318, opened 2023-01-04
Provide user reports of all bookings/bookings for a region

danlivings-dxw:
> See [ticket #639 on the CAS3 Trello board](https://trello.com/c/z61TtnJY/639-reporting-for-bookings).

This PR enables users to create reports by providing data about bookings in a spreadsheet format.

Changes in this PR:
- The `GET /reports/bookings` endpoint has been introduced. This endpoint has an optional `probationRegionId` query parameter which can be used for filtering bookings by probation region, or all bookings can be returned if this parameter is omitted. A successful request returns a response with `Content-Type: application/vnd.openxmlformats-officedocument.spreadsheetml.sheet` (i.e. an XLSX file) and the generated spreadsheet in the body.
- The concept of a report generator has been created. The `ReportGenerator` base class can be inherited to provide specific report functionality, such as the `BookingsReportGenerator`. A `ReportGenerator` implementation must specify its input, output, and configuration types. The report generators use the [Kotlin DataFrame](https://kotlin.github.io/dataframe/overview.html) library to provide the transformation and Excel serialization functionality.
- The `ReportService` class has been added to provide functionality to the API endpoint. `ReportService` methods are intended to use `ReportGenerator` instances to filter and transform the data, and are responsible for serializing the generated report to an `OutputStream`.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -43,6 +43,8 @@ dependencies {
 
   implementation("com.github.doyaaaaaken:kotlin-csv-jvm:1.7.0")
 
+  implementation("org.jetbrains.kotlinx:dataframe:0.8.1")
+
   testImplementation("io.github.bluegroundltd:kfactory:1.0.0")
   testImplementation("io.mockk:mockk:1.13.3")
   testImplementation("io.jsonwebtoken:jjwt-api:0.11.5")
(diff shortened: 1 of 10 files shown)


brownbearsoftware (approved):
LGTM - nice and reusable

brownbearsoftware:
One potential thing that just came to mind - it might be worth reading from the Postgres read-replica in prod for this to avoid running a report impacting normal use

danlivings-dxw:
@brownbearsoftware Do we have any examples I can use as a reference? I haven't encountered anything referencing read-replicas so far in the API.

brownbearsoftware:
> @brownbearsoftware Do we have any examples I can use as a reference? I haven't encountered anything referencing read-replicas so far in the API.

Yeah it's not something we're using anywhere at the moment so don't have an example.  Perhaps one for the future if you don't have time to investigate right now.

PR #319, opened 2023-01-04
Seed test environment with data

pezholio:
This allows us to run [repeatable migrations](https://flywaydb.org/documentation/tutorials/repeatable) in the test environment, so we can be sure there's predictable data for user testing etc, and the environment doesn't get clogged up with test data. 

There's also a PR in the Wiremock repo https://github.com/ministryofjustice/hmpps-community-accommodation-wiremock/pull/7 that will allow us to generate these migrations in a predictable and consistent way.


edavey (approved):
I've given this a good test locally and this seems to work well. Nice!

I'm interested to look at the "tools" repo now to understand how maintainable this is going to be!

edavey (approved):

PR #320, opened 2023-01-05
Add test migrations for Temporary Accommodation

pezholio:
This follows on from #319 and creates a bunch of rooms, as well as some additional bookings for the Temporary Accommodation service in the test environment.


danlivings-dxw (approved):
LGTM

PR #322, opened 2023-01-09
(873) Update clarification notes

pezholio:
This adds some new columns to the Clarification Notes object, as well as allowing users to update a note with both a response, and when the response was received.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AssessmentService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AssessmentService.kt
@@ -227,7 +227,8 @@ class AssessmentService(
         assessment = assessment,
         createdByUser = user,
         createdAt = OffsetDateTime.now(),
-        text = text
+        query = text,
+        response = null
       )
     )
 
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AssessmentService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AssessmentService.kt
@@ -254,6 +254,12 @@ class AssessmentService(
       return AuthorisableActionResult.Unauthorised()
     }
 
+    if (clarificationNoteEntity.response !== null) {
+      return AuthorisableActionResult.Success(
+        ValidatableActionResult.GeneralValidationError("A response has already been added to this note")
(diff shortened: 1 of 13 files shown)


pezholio:
[Confirm if there's enough information in assessment](https://trello.com/c/Jd5xAwWR/873-confirm-if-theres-enough-information-in-assessment)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AssessmentService.kt
```diff
@@ -235,6 +235,32 @@ class AssessmentService(
...
+      is AuthorisableActionResult.NotFound -> return AuthorisableActionResult.NotFound()
+    }
+
+    var clarificationNoteEntity = assessmentClarificationNoteRepository.findByAssessmentIdAndId(assessment.id, id)
+
+    if (clarificationNoteEntity === null) {
+      return AuthorisableActionResult.NotFound()
+    }
+
+    if (clarificationNoteEntity.createdByUser !== user) {
```
brownbearsoftware, line 253: This is a bit of a subtle one if you've still got your JS/TS hat on - in Kotlin `==` means `use the equals method on the class to compare two instances` and `===` means `check if these are the same object/location in memory.`  Because `UserEntity` is a data class it gets an equals method generated by the compiler that compares the values of its properties - so we probably want to use that here in case we end up with more two different objects representing the same user entry in the DB.

For more info: https://kotlinlang.org/docs/equality.html

src/main/resources/static/api.yml
```diff
@@ -1818,6 +1818,46 @@ paths:
...
+      requestBody:
+        description: Clarification note
+        content:
+          'application/json':
+            schema:
+              $ref: '#/components/schemas/UpdatedClarificationNote'
+        required: true
+      responses:
+        201:
+          description: successfully created a clarification note
```
brownbearsoftware, line 1850: should this be: `successfully updated a clarification note`

brownbearsoftware (approved):
LGTM bar a small comment on equality checks.

Only other thing that comes to mind is around auditing - should we only be able to update a clarification note once to add the answer?  Otherwise I could potentially come back to an assessment later and modify the response to a question.  Not sure if this is a risk or not though?

pezholio:
That's fair enough. I'll add a commit to do that as well 👍

brownbearsoftware (approved):

PR #326, opened 2023-01-09
Simple implementation of PDU for Temporary Accommodation premises

danlivings-dxw:
> See [ticket #570 on the CAS3 Trello board](https://trello.com/c/TON49LwD/570-users-can-assign-a-property-to-a-region-and-pdu).

This PR introduces PDUs to Temporary Accommodation premises. This has been done in a minimal and unopinionated way as a temporary measure to avoid conflicting with CAS1's planned changes to how geographic entities are modelled in the database, and will be revisited once those changes have been made.

This minimal PDU support is implemented by:
- Adding a `pdu` column to the `temporary_accommodation_premises` table. This column is just stored as text - while a more robust implementation would involve another table and a foreign key relationship, introducing such a table might complicate the previously mentioned model changes.
- Updating the `TemporaryAccommodationPremises` API schema with a required `pdu` property, and the `NewPremises` and `UpdatePremises` schemas with an optional `pdu` property. Validation logic in the service layer enforces that the PDU is provided when creating or updating a Temporary Accommodation premises. The `pdu` property has no effect on an Approved Premises.

Other things to note:
- No reference data endpoint for PDUs has been implemented. To avoid unnecessary breaking changes, the existence of this endpoint relies on the changes to the geographic model being largely in place. As a consequence, the Temporary Accommodation UI must maintain its own internal set of valid PDUs.
- There is no canonical list of PDUs within the API itself. This is closely related to the above point, but as a consequence means that the API currently must accept it as an arbitrary string from the client. The risk from this is that a value that does not correspond to a real PDU is sent to the API, complicating future migrations that change how PDUs are stored without a catch-all mapping. However, the Temporary Accommodation UI will present its internally-stored set of PDUs as a dropdown list without free text entry, and the PDU is not stored for an Approved Premises, so the probability is acceptably low.

### DIFF
--- a/src/main/resources/db/migration/all/20230104163132__add_pdu_column_to_premises.sql
+++ b/src/main/resources/db/migration/all/20230104163132__add_pdu_column_to_premises.sql
@@ -1,0 +1,17 @@
+ALTER TABLE temporary_accommodation_premises ADD COLUMN pdu TEXT;
+
+-- Some TA premises may have existing rows on the table.
+UPDATE temporary_accommodation_premises
+SET pdu = 'Not specified';
+
+ALTER TABLE temporary_accommodation_premises ALTER COLUMN pdu SET NOT NULL;
+
+-- Some TA premises do not have existing rows on the table, and will need them.
+INSERT INTO temporary_accommodation_premises(premises_id, pdu)
+SELECT id, 'Not specified'
+FROM premises
+WHERE premises.service = 'temporary-accommodation'
+AND premises.id NOT IN (
+    SELECT premises_id
+    FROM temporary_accommodation_premises
+);
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
@@ -97,7 +97,8 @@ class PremisesController(
         body.probationRegionId,
         body.characteristicIds,
         body.notes,
-        body.status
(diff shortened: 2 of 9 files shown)


src/main/resources/db/migration/all/20230104163132__add_pdu_column_to_premises.sql
```diff
@@ -0,0 +1,13 @@
+ALTER TABLE temporary_accommodation_premises ADD COLUMN pdu TEXT;
+
+-- Some TA premises may have existing rows on the table.
+UPDATE temporary_accommodation_premises
+SET pdu = 'Not specified';
+
+ALTER TABLE temporary_accommodation_premises ALTER COLUMN pdu SET NOT NULL;
+
+-- Some TA premises do not have existing rows on the table, and will need them.
+INSERT INTO temporary_accommodation_premises(premises_id, pdu)
```
brownbearsoftware, line 10: Will this cause a duplicate PK failure if there are some entries already in temporary_accommodation_premises?

brownbearsoftware (approved):
One small Q around the migration, otherwise LGTM

src/main/resources/db/migration/all/20230104163132__add_pdu_column_to_premises.sql
danlivings-dxw: I've updated the WHERE clause to:
```sql
WHERE premises.service = 'temporary-accommodation'
AND premises.id NOT IN (
    SELECT premises_id
    FROM temporary_accommodation_premises
)
```
So this should no longer be an issue.

PR #331, opened 2023-01-10
Enforce caseload Application restriction

brownbearsoftware:
Only allow a user to create an application for a CRN if the CRN is for an Offender managed by someone in their Delius team(s).


brownbearsoftware:
[Enforce "applicant" role via Delius](https://trello.com/c/5DTyGxcg/853-enforce-applicant-role-via-delius)

edavey (approved):
Thanks Joe, this is clear. 

We can enforce a similar restriction on _viewing_ applications in a separate PR/Ticket to follow.

brownbearsoftware:

PR #339, opened 2023-01-11
Add name to user response

pezholio:
We also need the name of the user who created an application, so adding this in


pezholio:
[Confirm if there's enough information in assessment](https://trello.com/c/Jd5xAwWR/873-confirm-if-theres-enough-information-in-assessment)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
```diff
@@ -39,7 +39,10 @@ class UserService(
       is ClientResult.Failure -> staffUserDetailsResponse.throwException()
     }
 
-    if (deliusUser.email !== user.email || deliusUser.telephoneNumber !== user.telephoneNumber) {
+    val name = "${deliusUser.staff.forenames} ${deliusUser.staff.surname}"
+
+    if (deliusUser.email !== user.email || deliusUser.telephoneNumber !== user.telephoneNumber || name != user.name) {
```
edavey, line 44: Would the purpose of this perhaps be clearer if this became a predicate function, e.g. `hasUserChanged()`? I think that's what this is doing: updating our user record if we detect a change at Delius?

edavey (approved):
👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
pezholio: Good idea! I've made that change 👍

PR #341, opened 2023-01-12
Chore/deploy to prod

tahb:
https://trello.com/c/EnjCxzGs/12-set-up-production

Add the Helm deploy charts to configure the deployed app and CircleCI to allow us to deploy.

This should not be merged until we're happy to deploy the app to production. This is currently pending the CAS1 side of the pentest.

We need to confirm the urls for at least 2 of the production services that have comments. From what I've heard the integrations service might not be in production so I wonder what the impact is if that app isn't available, are we blocked?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -58,21 +58,20 @@ workflows:
             - hmpps-approved-premises-api-preprod
           requires:
             - request-preprod-approval
-
-#      - request-prod-approval:
-#          type: approval
-#          requires:
-#            - deploy_preprod
-#      - hmpps/deploy_env:
-#          name: deploy_prod
-#          env: "prod"
-#          slack_notification: true
-#          slack_channel_name: << pipeline.parameters.releases-slack-channel >>
-#          context:
-#            - hmpps-common-vars
-#            - approved-premises-api-prod
-#          requires:
-#            - request-prod-approval
+      - request-prod-approval:
+          type: approval
+          requires:
+            - deploy_preprod
+      - hmpps/deploy_env:
+          name: deploy_prod
+          env: "prod"
+          slack_notification: true
+          slack_channel_name: << pipeline.parameters.releases-slack-channel >>
+          context:
+            - hmpps-common-vars
+            - approved-premises-api-prod
+          requires:
+            - request-prod-approval
 
   security:
     triggers:
--- a/helm_deploy/values-prod.yaml
+++ b/helm_deploy/values-prod.yaml
@@ -4,11 +4,21 @@
 generic-service:
   ingress:
     host: approved-premises-api.hmpps.service.justice.gov.uk
+    tlsSecretName: hmpps-approved-premises-api-prod-cert
 
   env:
+    SPRING_PROFILES_ACTIVE: prod
+    APPLICATIONINSIGHTS_CONFIGURATION_FILE: applicationinsights.json
+    HMPPS_AUTH_URL: https://sign-in.hmpps.service.justice.gov.uk/auth
+    SERVICES_COMMUNITY-API_BASE-URL: https://community-api-secure.probation.service.justice.gov.uk
+    SERVICES_AP-DELIUS-CONTEXT-API_BASE-URL: https://approved-premises-and-delius.delius.probation.hmpps.dsd.io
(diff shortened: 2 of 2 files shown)


helm_deploy/values-prod.yaml
```diff
@@ -4,6 +4,23 @@
 generic-service:
   ingress:
     host: approved-premises-api.hmpps.service.justice.gov.uk
+    tlsSecretName: hmpps-approved-premises-api-prod-cert
+
+  env:
+    SPRING_PROFILES_ACTIVE: prod
+    APPLICATIONINSIGHTS_CONFIGURATION_FILE: applicationinsights.json
+    HMPPS_AUTH_URL: https://sign-in.hmpps.service.justice.gov.uk/auth
+    SERVICES_COMMUNITY-API_BASE-URL: https://community-api.delius.probation.hmpps.dsd.io # assumed value
```
tahb, line 13: @edavey When you have a moment I wonder if you could do another check on these assumed urls for us please. They aren't in the cloud-platform-environments repo so I'm not sure where to check to confirm these.

edavey: I think Community API is incorrect. See https://github.com/ministryofjustice/make-recall-decision-api/blob/60ae20013a5bd039374c83b918490831f100a1b5/helm_deploy/values-prod.yaml#L12

tahb: Thanks Ed. That has been updated and I'm assuming `SERVICES_PRISONS-API_BASE-URL` and `SERVICES_AP-OASYS-CONTEXT-API_BASE-URL` were both correct.

tahb:
CAS1 now have [authority to operate](https://trello.com/c/MU5j9F5L) and so are ready to deploy the API.

.circleci/config.yml
```diff
@@ -58,21 +58,20 @@ workflows:
...
+      - hmpps/deploy_env:
+          name: deploy_prod
+          env: "prod"
+          slack_notification: true
+          slack_channel_name: << pipeline.parameters.releases-slack-channel >>
+          context:
+            - hmpps-common-vars
+            - approved-premises-api-prod
+          requires:
+            - request-prod-approval
```
edavey, line 74: perhaps we want to see failures in Slack? We could copy the config from MRD:

https://github.com/ministryofjustice/make-recall-decision-api/blob/41ebdcc8d4c618caa6e99af65821f8ec0ac23e95/.circleci/config.yml#L18-L46

helm_deploy/values-prod.yaml
```diff
@@ -4,6 +4,23 @@
...
+    SPRING_PROFILES_ACTIVE: prod
+    APPLICATIONINSIGHTS_CONFIGURATION_FILE: applicationinsights.json
+    HMPPS_AUTH_URL: https://sign-in.hmpps.service.justice.gov.uk/auth
+    SERVICES_COMMUNITY-API_BASE-URL: https://community-api-secure.probation.service.justice.gov.uk
+    SERVICES_AP-DELIUS-CONTEXT-API_BASE-URL: https://approved-premises-and-delius.delius.probation.hmpps.dsd.io
+    SERVICES_ASSESS-RISKS-AND-NEEDS-API_BASE-URL: https://assess-risks-and-needs.hmpps.service.justice.gov.uk
+    SERVICES_HMPPS-TIER_BASE-URL: https://hmpps-tier.hmpps.service.justice.gov.uk
+    SERVICES_PRISONS-API_BASE-URL: https://api.prison.service.justice.gov.uk
+    SERVICES_CASE-NOTES_BASE-URL: https://offender-case-notes.service.justice.gov.uk
+    SERVICES_AP-OASYS-CONTEXT-API_BASE-URL: https://approved-premises-and-oasys.hmpps.service.justice.gov.uk
```
edavey, line 19: I think only Community API and Prisons API need valid cred and the others SERVICES can have placeholder values for now as CAS3 won't be asking the API to fetch anything from them.

edavey (approved):
this looks correct to me.

.circleci/config.yml
tahb: I hope [this standard config](https://github.com/ministryofjustice/hmpps-template-typescript/blob/main/.circleci/config.yml#L177) does send failures to Slack. We could definitely copy if that isn't what happens or want to add more custom messages 👍

PR #351, opened 2023-01-16
993 - Support reallocating Assessments

brownbearsoftware:
This PR implements the POST `/applications/{id}/allocations` endpoint spec.  This allows users with `WORKFLOW_MANAGER` to reallocate an assessment provided the user they are assigning to has the `ASSESSOR` role and where applicable the relevant qualifications.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AssessmentService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AssessmentService.kt
@@ -3,6 +3,7 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.service
 import org.springframework.data.repository.findByIdOrNull
 import org.springframework.stereotype.Service
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApplicationEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApplicationRepository
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApprovedPremisesApplicationEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApprovedPremisesAssessmentJsonSchemaEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.AssessmentClarificationNoteEntity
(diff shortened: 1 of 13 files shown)


brownbearsoftware:

src/main/resources/db/migration/all/20230116143807__add_deallocated_at_to_assessments.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE assessments ADD COLUMN deallocated_at TIMESTAMP WITH TIME ZONE;
```
edavey, line 1: did you consider using  `reallocated_at` rather than `deallocated_at`?

I'm guessing your choice is in recognition of the fact that it's not always possible to reallocate a task?

I wonder if it would clearer (more natural, more intention-revealing) to use `reallocated_at` and only set it once it **has** been possible to identify and set a suitable assessor?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AssessmentService.kt
```diff
@@ -233,6 +236,78 @@ class AssessmentService(
...
+      )
+    }
+
+    if (! assigneeUser.hasAllQualifications(requiredQualifications)) {
+      return AuthorisableActionResult.Success(
+        ValidatableActionResult.FieldValidationError(ValidationErrors().apply { this["$.userId"] = "lackingQualifications" })
+      )
+    }
+
+    currentAssessment.deallocatedAt = OffsetDateTime.now()
```
edavey, line 280: would `currentAssessment.reallocatedAt = ...` be clearer?

as I suggested earlier, It feels like "re-allocation" is what we're all about here, rather than "de-allocation"?

edavey (approved):
This is really clear, thanks!

I think that perhaps this would clearer if we used `reallocatedAt` rather than `deallocatedAt`

I guess we'd need another test scenario to describe how, if reallocation fails for some reason, the `reallocated_at` field remains null after the attempt.

src/main/resources/db/migration/all/20230116143807__add_deallocated_at_to_assessments.sql
brownbearsoftware: So I went with deallocated because we're reallocating the Application which results in:
- The existing Assessment being rendered disabled (deallocated?)
- A new Assessment being created which is allocated to the other user

So I think from the perspective of the original Assessment it has been deallocated, or perhaps `disabled` or `superseded`?

brownbearsoftware: Just noticed I missed this point:

> I'm guessing your choice is in recognition of the fact that it's not always possible to reallocate a task?

As it stands, if it's not possible to allocate to the provided user the current Assessment remains active.  The whole thing is also transactional so any database exceptions should result in keeping the existing state.

edavey: OK, that behaviour makes sense. 👍  So it sounds like it's purely a (more or less "bike-shedding") question of naming: de-allocate versus re-allocate.

But, do you think we should have a test to make it clear that when reallocation fails the assessment does not go into the deallocated state?

brownbearsoftware: Have renamed to reallocated and added an integration test to ensure the operation is amotic

PR #353, opened 2023-01-17
Add repeatable migrations for dev

pezholio:
This ensures we return to a predictable state on each deploy. I’ve kept the old migrations so we don’t get any errors about missing migrations.

I’ve also added an ADR to document this decision, as well as the previous decision to use repeatable migrations in test.

### DIFF
--- a/doc/architecture/decisions/0010-use-repeatable-migrations-for-dev-and-test.md
+++ b/doc/architecture/decisions/0010-use-repeatable-migrations-for-dev-and-test.md
@@ -1,0 +1,88 @@
+# 10. use_repeatable_migrations_for_dev_and_test
+
+Date: 2023-01-17
+
+## Status
+
+Accepted
+
+## Context
+
+We need to ensure that our development and test environments have
+predictable data, so we can log in, run tests and carry out usability
+testing sessions with users. As we run automated tests on dev, it's also
+important that the environment doesn't get clogged with automatically
+created data.
+
+## Decision
+
+We will use [repeatable Flyway migrations](https://flywaydb.org/blog/flyway-timestampsandrepeatables)
+to tear down and set up the data to a known state.
+
+These migrations will be in the format:
+
+```text
+R__$NUM_$NAME_OF_MIGRATION.sql
+```
+
+Where `$NUM` is the order we want the migration to be run in, and
+`$NAME_OF_MIGRATION` is a description of the migration to be run.
+
+Each migration will also have a comment at the top like so:
+
+```sql
+-- ${flyway:timestamp}
+```
+
+This changes the checksum for the repeatable migrations, meaning they
+get executed each time we migrate.
+
+In dev, we will have SQL migrations manually written to drop the
+tables and insert the data previously written in the previous data
+migrations. We will also drop the bookings table to remove any
+bookings created in the e2e tests.
+
+For test, we have a [Wiremock repo](https://github.com/ministryofjustice/hmpps-community-accommodation-wiremock),
+which mocks away external services.
+This repo also has a helper script, which uses mock data to generate /
+re-generate migrations based on any changes made to the mocks.
+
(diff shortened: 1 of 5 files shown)


brownbearsoftware (approved):

src/main/resources/db/migration/local+dev/R__2_create_premises.sql
```diff
@@ -0,0 +1,12 @@
+-- ${flyway:timestamp}
+TRUNCATE TABLE premises CASCADE;
+
+insert into "premises" ("address_line1", "id", "local_authority_area_id", "name", "notes", "postcode", "probation_region_id", "service", "status", "total_beds") values ('Address 1', 'd33006b7-55d9-4a8e-b722-5e18093dbcdf', 'd75ce5b8-fc07-494b-8950-a46a63ac377e', 'Something House', NULL, 'LA9 1DS', 'a02b7727-63aa-46f2-80f1-e0b05b31903c', 'temporary-accommodation', 'active', 30);
```
tahb, line 4: @danlivings-dxw how does TA use total_beds? Do we set it to 0 and not use it or do we update it after the creation of a bedspace? We should set these seed values to something realistic for TA.

doc/architecture/decisions/0010-use-repeatable-migrations-for-dev-and-test.md
```diff
@@ -0,0 +1,86 @@
...
+bookings created in the e2e tests.
+
+For test, we have a Wiremock repo, which mocks away external services.
+This repo also has a helper script, which uses mock data to generate /
+re-generate migrations based on any changes made to the mocks.
+
+When data changes in the mocks, we will run:
+
+```bash
+script/generate_migrations /path/to/api/directory
```
tahb, line 52: Small typo on "generate_migration"

doc/architecture/decisions/0010-use-repeatable-migrations-for-dev-and-test.md
```diff
@@ -0,0 +1,86 @@
...
+data in both of those environments to be ephemeral, unless it is
+specifially added to a repeatable migration.
+
+As these migrations are destructive, it is important that these
+migrations are not run in prod or pre-prod. We see the fact that
+the running of these migrations is controlled by a environment
+variable change, which will be approved by another developer, to
+be an appropriate safeguard against this.
+
+We will also need to be aware that in some rare cases, a
```
tahb, line 83: It would be nice to make it clear when this happens so we can comfortably differentiate real errors against database tables disappearing but I can't think of a way.

Sadly I don't think there's a way to make separate CircleCi projects aware of one another so that the api and frontend deploys are mutually exclusive.

tahb (approved):
LGTM, thanks for looping me in.

Small typo and particular on our use of bed_count for accuracy but nothing blocking.

We're planning to give the test environment to users who are approaching onboarding as a place they can practice. One of our needs was to wipe what's there which this can help us with. @danlivings-dxw It seems like we could use this and add some more migrations for test (and perhaps dev too) that focus on some other tables CAS3 are using such as rooms, beds and a compliment of bookings in different states. I'll drop a card in.

doc/architecture/decisions/0010-use-repeatable-migrations-for-dev-and-test.md
pezholio: Ah, sorry, this command is in the Wiremock repo. Have updated the copy to be clearer.

src/main/resources/db/migration/local+dev/R__2_create_premises.sql
danlivings-dxw: @tahb It just gets set to 0 when the premises is created - the value is calculated by counting the beds in rooms associated with a premises.

danlivings-dxw (approved):

PR #358, opened 2023-01-18
993 - Mini manage authorisation changes

brownbearsoftware:


brownbearsoftware:

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/IntegrationTestBase.kt
```diff
@@ -264,6 +264,10 @@ abstract class IntegrationTestBase {
   fun beforeEach() {
     TimeZone.setDefault(TimeZone.getTimeZone("UTC"))
 
+    webTestClient = webTestClient.mutate()
+      .responseTimeout(Duration.ofMinutes(20))
```
edavey, line 268: hopefully long enough!

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/BookingTest.kt
```diff
@@ -1569,6 +1569,8 @@ class BookingTest : IntegrationTestBase() {
 
   @Test
   fun `Create Confirmation on Booking returns OK with correct body`() {
```
edavey, line 1611: I was expecting to see the familiar set up for the parameterised test here:

```kt
  @ParameterizedTest
  @EnumSource(value = UserRole::class, names = [ "MANAGER", "MATCHER" ])
```

but perhaps the user with those roles was already defined?

edavey (approved):
This looks good to me. 

Yes it would be good to remove some of this test boilerplate -- without making the tests too magical! 

You've had to retro-fit role-based auth onto a lot of endpoints so a fair part of the pain here is the volume of work I think?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/BookingTest.kt
brownbearsoftware: This is because the test is against a TA booking so works without roles - I've amended the test name to make this clear and added another one for Approved Premises that has the Parameterized setup

PR #360, opened 2023-01-19
Add instructions on accessing the remote database

pezholio:
This adds a `remote_shell` command and some instructions on how to install and run jaqy to access the database.


edavey (approved):
Thanks Stu, This is handy!

pezholio:
I'd like to have a one-shot command, but I think it pays to not make it _too_ easy!

PR #363, opened 2023-01-19
Remove Caseload applications restriction

brownbearsoftware:
Due to unforeseen size of some team caseloads this approach is no longer feasible.  We will replace this with a different approach that offers the same strength of authorisation but for now, revert to relying on who created an Application.


edavey (approved):
Thanks for reverting this feature.

PR #367, opened 2023-01-23
Add unique constraint over delius username

brownbearsoftware:
There should always be a 1:1 relationship between an entry in the "users" table and a Delius user.


edavey (approved):
👍 thks

edavey:
@brownbearsoftware Should we add a unit test to verify that this includes attempts to have similar names using different cases? e.g. 

- `temporary-accommodation-e2e-tester` in Wales
- `TEMPORARY-ACCOMMODATION-E2E-TESTER` in Kent

See https://trello.com/c/pb8e911I/1263-delius-usernames-are-not-required-to-be-unique-casewise

PR #369, opened 2023-01-23
Add ADR to control applicant access using Delius caseload

edavey:


doc/architecture/decisions/0011-control-applicant-access-using-delius-team-caseload.md
```diff
@@ -0,0 +1,168 @@
...
+
+Search for applications created by the logged-in user (`applications.created_by_user_id`). 
+Actually viewing the application could be restricted to people who are currently a member 
+of a team managing the CRN.
+
+## Consequences
+
+There are a couple of open questions:
+
+1. what do we do in the case of applicants who are members of more than one team? Do we:
```
brownbearsoftware, line 157: I don't think this is as much of an issue anymore as we will know which of the user's teams manages the Offender from the response from `https://approved-premises-and-delius/{crn}/managingTeams/{delius-staffcode}`.

The edge case is reduced to: the user is a member of more than one team where more than one of those teams has the Offender in its caseload.  Ideally I think we would ask the user to select the relevant team but given time constraints perhaps just storing all of the teams is the way to go here?

doc/architecture/decisions/0011-control-applicant-access-using-delius-team-caseload.md
```diff
@@ -0,0 +1,168 @@
...
+
+
+
+### Overview of the service's roles
+
+#### Applicant
+**Role**: `APPLICANT`  
+**Description**: probation officer making application  
+**Authorised to**:
+- create an application subject to the restrictions on their Delius user account, with
```
pezholio, line 27: Should we add something about continuing applications that they have made? Or do we want them to be able to continue applications their teammates have made? Though this could get messy.

doc/architecture/decisions/0011-control-applicant-access-using-delius-team-caseload.md
```diff
@@ -0,0 +1,168 @@
...
+- respond to requests for information raised during the assessment process (not MVP)
+
+#### Workflow manager
+**Role**: `WORKFLOW_MANAGER`   
+**Description**: CRU staff ensuring that assessment and allocation tasks are completed  
+**Authorised to**:
+- view all applications and their allocations to staff members
+- reassign applications to different staff members as necessary
+- mark application as withdrawn (e.g. if person has died)
+- ask applicant for further info, independent of the assessment itself (in MVP this is
```
pezholio, line 42: Just flagging: This isn't functionality we support yet

doc/architecture/decisions/0011-control-applicant-access-using-delius-team-caseload.md
```diff
@@ -0,0 +1,168 @@
...
+- mark application as withdrawn (e.g. if person has died)
+- ask applicant for further info, independent of the assessment itself (in MVP this is 
+  limited to adding notes based on off-line communication)
+
+#### Assessor
+**Role**: `ASSESSOR`  
+**Description**: Staff member who is tasked with assessing specific applications  
+**Authorised to**:
+- assess a particular application
+- add additional assessment content (recommendations, clarifications etc) to an application
```
pezholio, line 50: Assessors _will_ be able to ask an applicant for further info too

doc/architecture/decisions/0011-control-applicant-access-using-delius-team-caseload.md
```diff
@@ -0,0 +1,168 @@
...
+exception: the "applicant" role which is not modelled explicitly within the service, but
+is derived from caseload information from Delius.
+
+The Approved Premises service exposes personal data from NOMIS and from OASys and so
+it's necessary to restrict creating and viewing "applications" to probation officers with an
+appropriate caseload.
+
+
+
+### Overview of the service's roles
```
pezholio, line 21: Should this be part of the decision?

doc/architecture/decisions/0011-control-applicant-access-using-delius-team-caseload.md
```diff
@@ -0,0 +1,168 @@
...
+
+### 1. Creating an application
+
+To achieve [1] we're asking the probation-integration (PI) team to give us a new endpoint on the
+[approved-premises-and-delius](https://github.com/ministryofjustice/hmpps-probation-integration-services/tree/main/projects/approved-premises-and-delius) API which they've built.
+
+For example, something like:
+
+```sh
+curl "https://approved-premises-and-delius/{crn}/managingTeams/{delius-staffcode}"
```
pezholio, line 111: I feel like a more RESTful approach would be something like:

https://approved-premises-and-delius/teams/search

With a JSON payload eg:

```json
{ "crn": "{crn}", "staffCode": "{delius-staffcode}" }
```

Though I guess the implementation is up to the integrations team

doc/architecture/decisions/0011-control-applicant-access-using-delius-team-caseload.md
pezholio: I think we'll need to understand if there is ever a case where an offender is in more than one caseload first? It seems to me like it'd be a 1:1 relationship right?

brownbearsoftware: I think this depends on the answer to point 2.  i.e. an Offender can appear in the caseload for multiple teams at more granular levels so it depends which level of granulairty the integrations team endpoint operates at.  If I recall correctly this was examplified as part of their probation terms might be the responsibilit of a different team to the one that bares overall responsibility for them

doc/architecture/decisions/0011-control-applicant-access-using-delius-team-caseload.md
marcus-bcl: Another option would be to use query parameters as filters:
```
GET /teams?managesCrn={crn}&hasStaffMember={delius-staffcode}
{
    "teams": ["N01ABC"]
}
```

But we're happy to use whichever implementation - does your team have a preference?

edavey: Thanks @marcus-bcl 

@brownbearsoftware Would our OpenAPI code generator have trouble with any of these approaches? 

If not then probably it makes sense for you to decide on a design which is consistent within your mono-repo?

brownbearsoftware: We'd be consuming this endpoint so it wouldn't need to touch OpenAPI Generate so any approach will work for us 👍

edavey: good point! I forget which side of the fence I'm on!

marcus-bcl: We ended up going with:
```
GET /teams/managingCase/{crn}?staffCode={staffCode}
{
  "teamCodes": ["N01ABC"]
}
```

Hope that works for you.  The endpoint is up for review now, so should be available to test soon.

doc/architecture/decisions/0011-control-applicant-access-using-delius-team-caseload.md
edavey: Yes I believe that more than one team can have a person in their caseload so the endpoint may return more than one team. And it sounds like storing a list of team codes against each application (e.g. `application.managing_team_codes`) would be a good solution.

doc/architecture/decisions/0011-control-applicant-access-using-delius-team-caseload.md
edavey: I think we've generally agreed that:

- the creator of a application can edit it until it's been submitted and then it's read-only -- apart from the ability to add clarification notes (the form of these post-submission exchanges between assessor and applicant is not finalised)
- collaborating on an application is out of scope, i.e. if I am struck by lightning halfway through creating an application then a colleague will need to create a new application from scratch

doc/architecture/decisions/0011-control-applicant-access-using-delius-team-caseload.md
edavey: hmm, I thought it was helpful to include in the "Context" section as background, but on reflection I think I agree that "less is more" and that it's context enough to say:

- we've got some internally managed rolew
- we've got one special default role whose capabilities are defined by team caseloads managed within Delius

doc/architecture/decisions/0011-control-applicant-access-using-delius-team-caseload.md
edavey: removed all this details from the ADR

doc/architecture/decisions/0011-control-applicant-access-using-delius-team-caseload.md
edavey: Removed this level of detail

doc/architecture/decisions/0011-control-applicant-access-using-delius-team-caseload.md
edavey: this looks good, thanks

doc/architecture/decisions/0011-control-applicant-access-using-delius-team-caseload.md
edavey: I've added this to the draft

edavey:
@pezholio @brownbearsoftware 
I've updated this draft to:

- remove some background detail of dubious relevance
- include details of the actual endpoint implementation
- incorporate the latest thinking on handling multiple team codes
- a note about inter-PDU applications

pezholio (dismissed):

edavey:
@pezholio can I get a re-approval please as this went stale before I could merge!

PR #370, opened 2023-01-24
Fix logback config for pre-prod & prod

brownbearsoftware:


edavey:
Thanks: will this have broader effects on logging detail in the pre-prod and prod environments? Beyond the `seedLogger`?

If I understand correctly this is turning on the:  

`NOPAppender` and 
`ConsoleAppender`

in those environments? And that this will provide the output we need in Spring Boot:

```
    <logger name="org.springframework.boot" additivity="false" level="INFO">
        <appender-ref ref="logAppender"/>
        <appender-ref ref="consoleAppender"/>
    </logger>
```

But that's just my guess. It would be good for this to be detailed in the commit.

My concern is that we don't expose PII by increasing logging levels in areas other than seeding in pre-prod and prod...

brownbearsoftware:
@edavey that's correct - this will turn on logging

PR #371, opened 2023-01-24
Reduce healthcheck AppInsights percentage to 2% on non-prod environments

brownbearsoftware:
We're currently hitting 402 errors, hopefully massively reducing the amount of traces sent for the healthcheck endpoint will alleviate this.


edavey (approved):

edavey (approved):
Nice, that should help "future us".

edavey:
@brownbearsoftware  is there any reason why we aren't merging this?

PR #375, opened 2023-01-24
Additional seed data for Temporary Accommodation rooms/bookings in the test env

danlivings-dxw:
This is a rework of #366.

This PR achieves the same result, but empties out the `R__0_clear_bookings.sql` instead of renaming it (avoiding breaking the migrations) and uses `ON CONFLICT(id) DO NOTHING` for the booking seeding.


edavey (approved):
That looks good to me. I haven't run it locally, but I'm sure you have!

PR #376, opened 2023-01-24
1038 - Associate booking with Application

brownbearsoftware:
**Introduce the concept of an Offline Application**

A new table `offline_applications` stores info on offline/pre-digital service applications.  There is currently no way of inserting data into this table, this will be a seeding task in a subsequent PR.

**Add a reference to Application or Offline Application from Booking**

Allow a Booking to link to an Application or Offline Application.  The reference is not currently populated but will be in a subsequent PR.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BookingEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BookingEntity.kt
@@ -43,6 +43,12 @@ data class BookingEntity(
   var cancellation: CancellationEntity?,
   @OneToOne(mappedBy = "booking")
   var confirmation: ConfirmationEntity?,
+  @OneToOne
+  @JoinColumn(name = "application_id")
+  var application: ApplicationEntity?,
+  @OneToOne
+  @JoinColumn(name = "offline_application_id")
+  var offlineApplication: OfflineApplicationEntity?,
   @OneToMany(mappedBy = "booking")
   var extensions: MutableList<ExtensionEntity>,
   @ManyToOne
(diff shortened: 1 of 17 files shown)


brownbearsoftware:
[Associate booking with application](https://trello.com/c/AImdImUa/1038-associate-booking-with-application)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BookingEntity.kt
```diff
@@ -43,6 +43,12 @@ data class BookingEntity(
   var cancellation: CancellationEntity?,
   @OneToOne(mappedBy = "booking")
   var confirmation: ConfirmationEntity?,
+  @OneToOne
+  @JoinColumn(name = "application_id")
+  var application: ApplicationEntity?,
+  @OneToOne
+  @JoinColumn(name = "offline_application_id")
+  var offlineApplication: OfflineApplicationEntity?,
```
edavey, line 51: are we able to ask a booking for its application in a consistent way e.g. `booking.application` which retrieves both varieties of application?

Or are there two "association" methods now? `booking.application` and `booking.offlineApplication`?

edavey (approved):
This looks really good, though I'm a bit concerned **if** we don't have an "easy" way of consistently referencing a booking's application e.g. as `booking.application`. When writing domain events and generating KPI reports, that association is going to be in constant use as the application (representing the "point in time" risk assessment) is the key reference point of all related activity.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BookingEntity.kt
brownbearsoftware: Unfortunately not without making OfflineApplication derive from Application (so option A or B from the diagram we discussed the other day)

I think in practice though this won't make much difference as you would need to do a type check anyway, e.g.

```
if (booking.application != null) {
   // do something
} else if (booking.offlineApplication != null) {
   // do something else
} else {
   throw RuntimeException("Missing application for booking ${booking.id}")
}
```

versus

```
when (booking.application) {
  is OnlineApplicationEntity -> //do something
  is OfflineApplicationEntity -> //do something else
  null -> throw RuntimeException("Missing application for booking ${booking.id}")
}
```

PR #377, opened 2023-01-25
Log row level errors when seeding, and continue

edavey:
If there's an error during the `ensureCsvCanBeDeserialized()` test we continue to abort the entire seeding.

However, if there's an error processing an individual row (`job.processRow()`) we now catch the error and:

- log the full exception
- accumulate the exception message in a list which we log at the end of the seeding job

In this way we can attempt to seed 100 users, have 95 successful seedings and see a list of 5 problem rows to take away and remediate, e.g.

```
  The following row-level errors were raised:
  Error on row 1: Could not get user FREDBLOGGS
  Error on row 21: Could not get user JOENINETY
  Error on row 32: Could not get user CINDYSHERMAN
  Error on row 76: Could not get user ANDYPANDY
  Error on row 95: Could not get user OLIVEOIL

```


brownbearsoftware (approved):
LGTM

PR #379, opened 2023-01-25
Allow risks to be skipped with a flag set

pezholio:
This allows us to set a flag to skip the creation of risks when creating an application. This will allow us to go live without the reliance on the OASys integration, which may not be ready for beta. If we do use this, we’ll need to backfill any applications without risks added, which may be a manual process.


brownbearsoftware (approved):

pezholio:
[Allow us to skip Oasys integration with a feature flag](https://trello.com/c/khI0VQrF/1057-allow-us-to-skip-oasys-integration-with-a-feature-flag)

PR #382, opened 2023-01-25
Automatically associate bookings with an Application

brownbearsoftware:
When creating a Booking, where possible associate it with the latest submitted Application or Offline Application.

For AP Bookings, failure to do this will return a 400.  TA Bookings are still creatable without an Application.


brownbearsoftware:
[Associate booking with application](https://trello.com/c/AImdImUa/1038-associate-booking-with-application)

brownbearsoftware:
@danlivings-dxw we don't need this for now after some release targets have changed so no need to review

danlivings-dxw (approved):

PR #383, opened 2023-01-25
Integration Test refactoring

brownbearsoftware:
Integration Test Refactoring

- Move common setups for Users and Offenders into `Given a User` and `Given an Offender` constructs.  These setup the required entities in the database and mock upstream calls.  These generally work without any arguments but have several optional params where required (e.g. for adding roles to a User or overriding the details of an Offender)
- Move upstream HTTP mocks into file-per-service.  In the httpmocks directory there is now a file for each upstream service which contain extension methods for IntegrationTestBase.  Using any of these will automatically mock the call to HMPPS Auth to get a client_credentials token so we no longer have to include `mockClientCredentialsJwtRequest()` calls in most tests


danlivings-dxw:
I think this is a good approach. It seems like it would be easy to extend to add specific details (e.g. adding a probation region to a user). It might be nice to have a `GivenBuilder` to avoid deep nesting when multiple givens are used, although I'm not sure how that would be implemented.

brownbearsoftware:
[Refactor Integration Tests](https://trello.com/c/OMU9Py0F/1071-refactor-integration-tests)

edavey (approved):
🏆 less code and more clarity at the cost of a bit of "clever" and somewhat complex code in the new "givens" and persisted factory class.

I haven't scrutinised or understood every line but as this refactors tests only with no changes to app behaviour, 🚀 

One question: do you know what impact this has, if any, on the time the test suite takes to run?

brownbearsoftware:
@edavey it looks like it's faster from the Actions runs (where the first succesful run was just one test changed to this style versus the latest which is all of them) https://github.com/ministryofjustice/hmpps-approved-premises-api/actions?query=branch%3A1071-integration-tests-refactor-poc++

However, I'm not sure that's enough runs with all other things being equal to really draw conclusions.  I would have expected it to have a negligble speed impact.

PR #384, opened 2023-01-26
Expose users and roles to ROLE_ADMIN only

edavey:
The list is sorted by name.

Our plan is that first phase of the Role Admin area will comprise a list
of users and their role. Adjustments to the role assignment will be managed
through the existing user seeding mechanism. See [`UsersSeedJob`](https://github.com/ministryofjustice/approved-premises-api/blob/expose-users-and-roles-to-role-admin/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/UsersSeedJob.kt)

The facility to add and remove roles through the UI will follow.

In due course we'll need to paginate this list. For the initial release
we expect about 250 user with roles to be managed.

Perhaps we'll want some filtering, e.g.

- only show users with internally managed roles
- show users with a particular role

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/UsersController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/UsersController.kt
@@ -6,12 +6,15 @@ import org.springframework.stereotype.Service
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.UsersApiDelegate
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ServiceName
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.User
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.UserQualification
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.UserRole
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.ForbiddenProblem
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.NotFoundProblem
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.results.AuthorisableActionResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.UserService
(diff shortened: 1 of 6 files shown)


src/main/resources/static/api.yml
```diff
@@ -3403,7 +3403,7 @@ components:
         - manager
         - workflow_manager
         - applicant
-        - admin
+        - role_admin
```
brownbearsoftware, line 3406: How come this has been given a prefix of `role_` but the others haven't?

edavey: The facility granted to holders of this role is managing the roles of other users. So the role is "administrator of roles" or "role admin". As per https://trello.com/c/0CGdMDiu/524-draft-roles-for-the-ap-service#comment-633d9bca5897d900de2ca81b

Clearly this naming is confusing. Though it does seem to be have been adopted and understood as  intended by the CRU staff. It's used on the `Master list - N54 and N55 User List Revised 230123` list used for the seeding.

Can you think of a better name for this role?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/UsersController.kt
```diff
@@ -27,4 +29,16 @@ class UsersController(
 
     return ResponseEntity(userTransformer.transformJpaToApi(userEntity), HttpStatus.OK)
   }
+
+  override fun usersGet(roles: List<UserRole>?, qualifications: List<UserQualification>?): ResponseEntity<List<User>> {
+    val user = userService.getUserForRequest()
+    if (!user.hasAnyRole(uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserRole.ROLE_ADMIN)) {
```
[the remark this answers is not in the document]
edavey: thks, I will try to remember this "import alias" mechanism

src/main/resources/static/api.yml
edavey: maybe `ROLES_ADMIN` would be a (slight) improvement?

brownbearsoftware: I see!  Think it's fine as-is now that I know, not sure `roles_admin` is any more obvious without the explanation.

brownbearsoftware (approved):

pezholio (approved):
LGTM 👍

PR #385, opened 2023-01-26
Add probation region to users

danlivings-dxw:
> See [ticket #714 on the CAS3 Trello board](https://trello.com/c/p0UPjpJn/714-api-can-retrieve-regional-information-about-a-member-of-staff).

This PR introduces probation regions for users. This is to enable upcoming work to restrict users' access to premises depending on their region.

Changes in this PR:
- The `GET users/{id}` and `GET /profile` endpoints now expect an `X-Service-Name` header with a value of either `temporary-accommodation` or `approved-premises`. This returns either a complete user object for Approved Premises, or a stripped-down user object for Temporary Accommodation (as TA does not have a requirement for personal information from this endpoint).
- The above endpoints also include the ID and name of the probation region for the user in their response.
- The `UserService.getUserForUsername` method now inspects the value of `probationArea.code` from the Community API response, and uses this to retrieve the appropriate probation region from the database to assign to the user.
- The `IntegrationTestBase.mockStaffUserInfoCommunityApiCall` method now accepts an optional boolean parameter, `createProbationRegionForStaffAreaCode`, which can be used to ensure a probation region exists with the same Delius code as the returned response for this mocked Community API call. The parameter's default value is `true` but can be set to `false` in situations where other parts of the test create an appropriate probation region.
- Many tests require a user entity to be created to pass. All such tests have been updated to ensure that they have a suitable probation region for the user. This has been made into its own commit to make reviewing easier.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ProfileTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ProfileTest.kt
@@ -1,9 +1,13 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.integration
 
 import org.junit.jupiter.api.Test
-import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.User
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ApprovedPremisesUser
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ProbationRegion
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ServiceName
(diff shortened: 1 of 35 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ProfileTest.kt
```diff
@@ -44,7 +48,59 @@ class ProfileTest : IntegrationTestBase() {
...
+    }
+
+    webTestClient.get()
+      .uri("/profile")
+      .header("Authorization", "Bearer $jwt")
+      .exchange()
+      .expectStatus()
+      .is4xxClientError
+      .expectBody()
+      .also { it.consumeWith { println(it.responseBody?.let { String(it) }) } }
```
brownbearsoftware, line 79: leftover debugging code?

PR #392, opened 2023-01-30
Give users in local, dev, and test environments real probation regions

danlivings-dxw:
Currently the user `JimSnowLdap` does not exist in our local and dev environments until it is created by a call to `UserService.getUserForUsername`. This calls the Community API, which in the dev environment returns a response with a probation area code of `"GCS"`, which does not correspond to any probation region. This is causing the error `Unknown probation region code 'GCS' for user 'JIMSNOWLDAP'` to appear in the dev environment.

This also appears to be the case for users used for end-to-end tests, as well as personal accounts on the dev and test environments.

It seems that in production-like environments, the Community API will always return codes corresponding to real probation regions, and that this is just a consequence of using dummy data.

This PR adds two mechanisms to avoid this problem in lower environments:
1. Some user accounts used for testing have been added to the repeatable migrations for `local+dev` so that when `UserService.getUserForUsername` is called, it can return early as the user will exist in the database, rather than trying to determine the probation region and failing when the code doesn't match any existing regions.
2. An `assign-default-region-to-users-with-unknown-region` feature flag has been introduced, with a default value of `false` but set to `true` for local, dev, and test environments. This flag instructs the `UserService.getUserForUsername` method to use the North West probation region as a default when the lookup using the Community API's probation area code fails, as well as emit a warning to the logs that the default value is being used.

### DIFF
--- a/helm_deploy/values-dev.yaml
+++ b/helm_deploy/values-dev.yaml
@@ -20,6 +20,7 @@ generic-service:
     SERVICES_AP-OASYS-CONTEXT-API_BASE-URL: https://approved-premises-and-oasys-dev.hmpps.service.justice.gov.uk
     SPRING_FLYWAY_LOCATIONS: classpath:db/migration/all,classpath:db/migration/local+dev
     LOG-CLIENT-CREDENTIALS-JWT-INFO: true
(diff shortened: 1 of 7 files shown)


danlivings-dxw:
There's some additional users that should be added so I've changed this to a draft

brownbearsoftware (approved):

PR #393, opened 2023-01-31
Fix migration to give existing users a default region

danlivings-dxw:
The `20230123154608__add_probation_region_to_users.sql` migration has been updated to ensure that any existing users have been assigned a default probation region (North West) before enforcing non-nullability on the `probation_region_id` column.

This should allow the migration to succeed in the preprod and test environments.

### DIFF
--- a/src/main/resources/db/migration/all/20230123154608__add_probation_region_to_users.sql
+++ b/src/main/resources/db/migration/all/20230123154608__add_probation_region_to_users.sql
@@ -1,2 +1,13 @@
-ALTER TABLE "users" ADD COLUMN probation_region_id UUID NOT NULL;
+ALTER TABLE "users" ADD COLUMN probation_region_id UUID;
 ALTER TABLE "users" ADD FOREIGN KEY (probation_region_id) REFERENCES probation_regions(id);
+
+-- Give existing users a default region.
+UPDATE "users"
+SET probation_region_id = (
+    SELECT id
+    FROM probation_regions
+    WHERE delius_code = 'N51' -- North West
+    LIMIT 1
+);
+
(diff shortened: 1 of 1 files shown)


src/main/resources/db/migration/all/20230123154608__add_probation_region_to_users.sql
```diff
@@ -1,2 +1,8 @@
-ALTER TABLE "users" ADD COLUMN probation_region_id UUID NOT NULL;
+ALTER TABLE "users" ADD COLUMN probation_region_id UUID;
 ALTER TABLE "users" ADD FOREIGN KEY (probation_region_id) REFERENCES probation_regions(id);
+
+-- Give existing users a default region.
+UPDATE "users"
+SET probation_region_id = 'a02b7727-63aa-46f2-80f1-e0b05b31903c'; -- North West
```
pezholio, line 6: Could we make this more rigorous with a subquery, rather than a UUID, i.e:

```suggestion
SET probation_region_id = (
        SELECT
          id
        FROM
          probation_regions
        WHERE
          delius_code = 'N51'
        LIMIT 1
      );
```

pezholio (approved):
LGTM - I guess this won't cause issues with the migration table as it didn't run cleanly, so won't have updated the migrations table right?

danlivings-dxw:
> LGTM - I guess this won't cause issues with the migration table as it didn't run cleanly, so won't have updated the migrations table right?

I've removed the last few rows in the `flyway_schema_history` table (as well as dropping the `probation_region_id` column) in dev to allow this to run again there, but it never successfully ran on preprod or test so it shouldn't cause problems there.

PR #395, opened 2023-02-01
(1059) Add a status to applications

pezholio:
This fetches one of three statuses:

- `inProgress` - if the application is not submitted
- `submitted` - if the application is submitted and all clarificationNotes have responses
- `requestedFurtherInformation` - if the application’s latest assessment has any clarificationNotes without responses

As part of this, we also need to load in assessements for an application, so I've added a new variable to the Application entity, together with a helper to fetch the latest entity.

The transformer tests are a little more long winded than I'd like, but I was getting too bogged down with trying to DRY them up!


pezholio:
[Submitted applications stay in the open application screen.](https://trello.com/c/mJbwwETZ/1059-submitted-applications-stay-in-the-open-application-screen)

brownbearsoftware (approved):
LGTM

PR #396, opened 2023-02-01
Domain events scaffolding

brownbearsoftware:
This PR adds the groundwork for Domain Events. 

A `DomainEventService` provides a mechanism for saving and retrieving the events.  They are stored in a single database table with the bulk of the event being kept in a JSON column to avoid having to create and maintain a table structure for each type of event.

Perhaps controversially: the data stored in the table is serialized from/deserializes to the same model as is returned by the endpoints to avoid the maintenance overhead of keeping an identical "domain" version of each event and transforming between them.

### DIFF
--- a/src/main/resources/db/migration/all/20230130160546__domain_events.sql
+++ b/src/main/resources/db/migration/all/20230130160546__domain_events.sql
@@ -1,0 +1,10 @@
+CREATE TABLE domain_events (
+    id UUID NOT NULL,
+    application_id UUID NOT NULL,
+    crn TEXT NOT NULL,
+    type TEXT NOT NULL,
+    occurred_at TIMESTAMP WITH TIME ZONE NOT NULL,
+    created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
+    data JSON NOT NULL,
+    PRIMARY KEY (id)
+);
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/factory/events/ApplicationSubmittedFactory.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/factory/events/ApplicationSubmittedFactory.kt
@@ -1,0 +1,135 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.factory.events
+
+import io.github.bluegroundltd.kfactory.Factory
+import io.github.bluegroundltd.kfactory.Yielded
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.ApplicationSubmitted
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.ApplicationSubmittedSubmittedBy
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.Cru
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.Ldu
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.PersonReference
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.ProbationArea
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.Region
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.StaffMember
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.Team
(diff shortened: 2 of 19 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/factory/events/ApplicationSubmittedFactory.kt
```diff
@@ -0,0 +1,135 @@
...
+    submittedAt = this.submittedAt(),
+    submittedBy = ApplicationSubmittedSubmittedBy(
+      staffMember = this.submittedByStaffMember(),
+      probationArea = this.submittedByProbationArea(),
+      cru = this.submittedByCru(),
+      team = this.submittedByTeam(),
+      ldu = this.submittedByLdu(),
+      region = this.submittedByRegion()
+    ),
+    sentenceLengthInMonths = this.sentenceLengthInMonths()
```
edavey, line 133: when we reviewed `application.submitted` we identified a few changes but didn't actually write them up. I think what we agreed was:

- `mappaCategories`: this will comprise both mappa level and mappa category in a single string, e.g. `"CAT C3/LEVEL L2"`, so we could rename this to simply `mappa`?
- `offenceDescription`: we don't save the offence description, so let's return the `offenceId` instead and let the consumer lookup the human readable name
- `releaseType`: we noted that we'll be storing a type such as `["hdc", "licence", "pss", "rotl"]` rather than a string like `"Release on Temporary Licence (ROTL)"` So as long as we coerce that type list into a string (e.g. JSON) then we don't need to do anything beyond updating the "example" in OpenAPI spec? See [JSON schema in the UI](https://github.com/ministryofjustice/hmpps-approved-premises-ui/blob/d3816238499ed282fedcda46bdb0687e81690f1c/server/form-pages/apply/schema.json#L40)

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/DomainEventServiceTest.kt
```diff
@@ -0,0 +1,126 @@
...
+        data = ApplicationSubmittedEnvelope(
+          id = id,
+          timestamp = occurredAt,
+          eventType = "approved-premises.application.submitted",
+          eventDetails = ApplicationSubmittedFactory().produce()
+        )
+      )
+    )
+
+    verify(exactly = 1) { domainEventRespositoryMock.save(any()) }
```
edavey, line 124: Should we flesh out this "verification" to demonstrate for example that the repository receives a `DomainEventEntity` where:

-  the `data` attribute is a JSON version of `application.submitted` and 
- that `type` is the expected `DomainEventType.APPROVED_PREMISES_APPLICATION_SUBMITTED`

Or will that come later / be covered elsewhere?

edavey (approved):
lovely to see this plan coming together! Very clear and easy to follow. Just 2 comments/suggestions:

- should we implement the changes (now)  to `application.submitted` which we discussed recently - and which I've written up here?
- should the unit test for the domain event service go a bit further in describing the save message it sends to the repository?

src/main/resources/db/migration/all/20230130160546__domain_events.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE domain_events (
+    id UUID NOT NULL,
+    application_id UUID NOT NULL,
+    crn TEXT NOT NULL,
+    type TEXT NOT NULL,
+    occurred_at TIMESTAMP WITH TIME ZONE NOT NULL,
+    data JSON NOT NULL,
+    PRIMARY KEY (id)
```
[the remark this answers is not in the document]
brownbearsoftware: Seems sensible 👍  Have added this in

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/factory/events/ApplicationSubmittedFactory.kt
brownbearsoftware: Good shout - have made these changes

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/DomainEventServiceTest.kt
brownbearsoftware: Good point!  Have updated the verification.

brownbearsoftware:
[Get the 'application submitted' domain event created, stored & exposed](https://trello.com/c/LhNuCu4q/991-get-the-application-submitted-domain-event-created-stored-exposed)

PR #398, opened 2023-02-03
991 - Application submitted event

brownbearsoftware:
Create an Application Submitted Domain Event when application is succesfully submitted.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ApplicationService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ApplicationService.kt
@@ -1,8 +1,20 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.service
 
+import org.springframework.beans.factory.annotation.Value
 import org.springframework.data.repository.findByIdOrNull
 import org.springframework.stereotype.Service
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.ApplicationSubmitted
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.ApplicationSubmittedEnvelope
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.ApplicationSubmittedSubmittedBy
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.Ldu
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.PersonReference
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.ProbationArea
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.Region
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.StaffMember
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.Team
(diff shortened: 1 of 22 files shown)


brownbearsoftware:
[Get the 'application submitted' domain event created, stored & exposed](https://trello.com/c/LhNuCu4q/991-get-the-application-submitted-domain-event-created-stored-exposed)

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/ApplicationServiceTest.kt
```diff
@@ -693,6 +763,56 @@ class ApplicationServiceTest {
...
+              code = staffUserDetails.probationArea.code,
+              name = staffUserDetails.probationArea.description
+            )
+          ) &&
+            data.mappa == risks.mappa.value!!.level &&
+            data.sentenceLengthInMonths == null &&
+            data.offenceId == application.offenceId
+        }
+      )
+    }
```
edavey, line 815: Bravo for this epic piece of mocking! 

I think building the event requires more API calls than it should because we are having to obtain some pieces of information for our Domain Event which ought to be part of the data/document -- as they are also needed for displaying the read-only "point-in-time" application:

- Community API: `offenderService.getRiskByCrn`:
	- mappa value 

- Community API: `offenderService.getOffenderByCrn`:
	- age
	- NOMS number
	- gender

- sentence length (not currently displayed)

Perhaps the right approach to fixing this deficiency would be to 

1. have the calling `ApplicationService.submitApplication()` function make these API calls to Community API and
2.  apply the "missing" first class attributes to the application before 
3.  passing it through to this `createApplicationSubmittedEvent()` function

edavey (approved):
Great work, thank you. This is very clear.

However, I think there are some important properties which we need to store as first class attributes of the `application` as well as storing them in the `application.submitted` domain event. e.g. by moving a couple of calls to Community API out of the event builder function up into the `submitApplication` function

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ApplicationService.kt
```diff
@@ -254,8 +274,108 @@ class ApplicationService(
...
+    val mappaLevel = risks.mappa.value?.level ?: throw RuntimeException("Mappa not present on Risks when creating Application Submitted Domain Event")
+
+    val staffDetailsResult = communityApiClient.getStaffUserDetails(username)
+    val staffDetails = when (staffDetailsResult) {
+      is ClientResult.Success -> staffDetailsResult.body
+      is ClientResult.Failure -> staffDetailsResult.throwException()
+    }
+
+    // TODO: The integrations team have agreed to build us an endpoint to determine which team's caseload an Offender belongs to
+    //       once that is done we should use that here to select the right team when the user is a memnber of more than one
```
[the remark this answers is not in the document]
brownbearsoftware: Yeah I think so, this would come with the benefit of making the unit tests a bit more paletable too

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/ApplicationServiceTest.kt
brownbearsoftware: As discussed on Slack - will leave as-is for now as this info should probably be placed into the `data` blob by the frontend and displayed on check your answers.

PR #400, opened 2023-02-03
Implement seed jobs for Temporary Accommodation premises and bedspaces

danlivings-dxw:
> See [ticket #560 on the CAS3 Trello board](https://trello.com/c/jhDScZ8w/560-technical-seeding-of-property-data).

This PR extends the `POST /seed` endpoint by introducing two new `seedType`s: `temporary_accommodation_premises` and `temporary_accommodation_bedspaces`. These correspond to two tabs on the spreadsheet template that has been distributed to each of the probation regions to allow us to populate the live Temporary Accommodation service with real data when each region is onboarded.


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/SeedTemporaryAccommodationPremisesTest.kt
```diff
@@ -0,0 +1,270 @@
...
+    assertThat(persistedPremises.postcode).isEqualTo(csvRow.postcode)
+    assertThat(persistedPremises.probationRegion.name).isEqualTo(csvRow.probationRegion)
+    assertThat(persistedPremises.localAuthorityArea!!.name).isEqualTo(csvRow.localAuthorityArea)
+    assertThat(persistedPremises.pdu).isEqualTo(csvRow.pdu)
+    assertThat(persistedPremises.characteristics.map { it.name }).isEqualTo(csvRow.characteristics)
+    assertThat(persistedPremises.notes).isEqualTo(csvRow.notes)
+  }
+
+  @Test
+  fun `Updating an existing Temporary Accommodation Premises persists correctly`() {
```
tahb, line 105: Good to see we're testing for idempotentcy 👍 My main concern with this type of process is what happens if an error is thrown half way through or we need to re-run this.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/SeedTemporaryAccommodationPremisesTest.kt
```diff
@@ -0,0 +1,270 @@
...
+      withApArea(apAreaEntityFactory.produceAndPersist())
+    }
+
+    val updatedProbationRegion = probationRegionEntityFactory.produceAndPersist {
+      withApArea(apAreaEntityFactory.produceAndPersist())
+    }
+
+    val originalLocalAuthorityArea = localAuthorityEntityFactory.produceAndPersist()
+
+    val updatedLocalAuthorityArea = localAuthorityEntityFactory.produceAndPersist()
```
tahb, line 116: It looks like we're using region and local authority as the fields we check to see if our updates were made on an existing record.

Do `probationRegionEntityFactory.produceAndPersist` and  `localAuthorityEntityFactory.produceAndPersist()` always return different values?

Might it be better to check for updates on a more straightforward field such as `notes` or `postcode`?

tahb (commented):
This looks generally sound. Generally I'd like to run this locally to see all the parts work together, I'm not set up to do that but have you and are happy with it?

I found the second commit to be a challenge to read due to the size of it and number of things going on. In future I'd definitely like to see us getting back to smaller commits. 

For example I found it hard to see if SeedUtils had any unit tests and there was no space in the commit to document why these might not be needed, since it had everything else in it too.

I'll leave the approval to @brownbearsoftware

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/SeedTemporaryAccommodationPremisesTest.kt
danlivings-dxw: This test adds a characteristic that isn't originally present on the premises ("Floor level access") and tests for its presence after the update.

I believe that `*Factory.produceAndPersist` does produce a different object each time as otherwise it wouldn't be able to insert it into the database as they'd have the same value for the primary key.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/SeedTemporaryAccommodationPremisesTest.kt
danlivings-dxw: The seeding process is set up so that if an error occurs within a row, the error is logged and the row is skipped so that following rows can be processed. Correcting the CSV file and rerunning the job should work fine, as I've done this multiple times during my manual testing.

danlivings-dxw:
> This looks generally sound. Generally I'd like to run this locally to see all the parts work together, I'm not set up to do that but have you and are happy with it?

Yes, I've done this with both the dummy data and the East of England spreadsheet we've received from the HPT team.

> I found the second commit to be a challenge to read due to the size of it and number of things going on. In future I'd definitely like to see us getting back to smaller commits.

I would have preferred to have split it up into smaller commits but I couldn't really figure out how to do so in a logical way. I can give it another attempt if necessary.

> For example I found it hard to see if SeedUtils had any unit tests and there was no space in the commit to document why these might not be needed, since it had everything else in it too.

`SeedUtils` is mostly a couple of lookup tables to handle some weirdnesses in our spreadsheets (i.e. the region "Kent, Surrey and Sussex" being presented as two separate regions due a split on the comma, and about half a dozen local authorities with a similar issue) as well as to make an attempt at protecting against simple variations caused by people typing in values rather than selecting them (for example, typing "southend on sea" instead of "Southend-on-Sea").

These lookup tables could be trimmed down if we make the assumption that only mistakes caused by us will show up, but I chose to err on the side of caution here. I can change this if the above assumption is acceptable to make.

The other functions there I would say are fairly trivial and I don't see a particular benefit of writing dedicated tests for them, but I can do so if there's some value there.

tahb:
Thanks for those replies @danlivings-dxw.

I know you packed this in on Friday to help get it finished for the team today so I completely understand 👍 Usually I would suggest we go back and break this down but given the urgency I think it's ok not to - provided @brownbearsoftware doesn't have a hard time reading through it.

In terms of testing coverage we [should](https://playbook.dxw.com/tech/how-we-do-development/#4-develop-the-code-and-the-tests-side-by-side) be taking the time to write automated tests for all the code we write. With the exception of other peoples code, prototypes and perhaps private methods. The line of reasoning in my mind is that if we depend on this working as we expect in production, it's worth testing to document and to catch any regressions and that's our standard. I recall something about automated test coverage checkers being in place to help with this but I might be wrong(?). In terms of next steps for testing here, my question is are we confident we have enough in place to catch regressions. 

I have some ideas for chunking this up into commits if that's helpful but as I said I don't think it's essential this time around.

brownbearsoftware (approved):
LGTM.

I would take the view that the canonicalize work in the utils shouldn't be the responsibility of the service and be done when prepping the upload CSV but given time constraints I don't think this is enough of an issue to delay on.

PR #402, opened 2023-02-06
Don't sample health endpoints with Sentry

brownbearsoftware:
The /health/** endpoints are currently using up too much of our Sentry quota for Distributed Tracing.  This prevents them from being sampled.

https://docs.sentry.io/platforms/java/guides/spring-boot/configuration/filtering/#using-platformidentifier-nametraces-sampler-


brownbearsoftware:
[Improve Sentry config (instrumentation)](https://trello.com/c/CZgrT05u/1036-improve-sentry-config-instrumentation)

edavey (dismissed):
Thanks Joe, This looks great. 

Is there a link to docs on this which we could include in the commit message? Seems like it's slightly magical how the callback is registered or the `sample` function overwritten?

PR #403, opened 2023-02-06
Don't sample health endpoints with Sentry

brownbearsoftware:
The /health/** endpoints are currently using up too much of our Sentry quota for Distributed Tracing. This prevents them from being sampled.

https://docs.sentry.io/platforms/java/guides/spring-boot/configuration/filtering/#using-platformidentifier-nametraces-sampler-


edavey (approved):
thanks 🚀

PR #405, opened 2023-02-07
Filter premises by user region

danlivings-dxw:
> See [ticket #765 on the CAS3 Trello board](https://trello.com/c/xv01P30x/765-minimal-api-authorisation-work-for-regional-filtering).

Currently, the API does not provide a way to restrict users to only be able to access premises within their probation region. This is needed in order to onboard the Greater Manchester team so that they do not interact with East of England premises (and vice versa).

This PR introduces the ability for filtering premises by region in the `GET /premises` endpoint. This filtering is enabled by setting the value of the `X-User-Region` header to the ID of the probation region for the current user. The existing behaviour (i.e. no filtering) is preserved if this header is not supplied to avoid a breaking change.

Currently, the API trusts that the client (in this case, the Temporary Accommodation UI) supplies the correct probation region ID for the user, and does not check for correctness. Further work to prevent circumventing the region restriction is planned soon (see: [ticket #770 on the CAS3 Trello board](https://trello.com/c/mDxlIttG/770-protect-against-users-circumventing-property-filtering)), but these extra precautions are not considered to be blocking the onboarding of a second regional team to the TA service.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
@@ -121,7 +121,7 @@ class PremisesController(
     return ResponseEntity.ok(premisesTransformer.transformJpaToApi(updatedPremises, updatedPremises.totalBeds))
   }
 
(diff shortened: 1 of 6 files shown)


brownbearsoftware (dismissed):
LGTM provided the risk around trusting the client mentioned in the description have been accepted by relevant party

brownbearsoftware (approved):

PR #406, opened 2023-02-07
Add logging to JsonSchemaService validation

brownbearsoftware:
When failing to validate a schema, log the JSON path and the validation message


edavey (approved):
👍

PR #409, opened 2023-02-08
Update AP Applications JSON Schema

brownbearsoftware:
This is an update to the last inserted schema rather than a new one because it addresses issues with the schema itself rather than representing a change to the model the schema validates.


pezholio (approved):
LGTM - would be good to remove that errant merge commit though 👍

pezholio (approved):

PR #410, opened 2023-02-08
1135 - Publish domain events to SNS topic

brownbearsoftware:
Publish to the HMPPS domain events SNS topic when an Application Submitted event occurs


brownbearsoftware:
[Publish domain events to SQS](https://trello.com/c/B9qoYr0h/1135-publish-domain-events-to-sqs)

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/DomainEventServiceTest.kt
```diff
@@ -132,5 +128,22 @@ class DomainEventServiceTest {
...
+            deserializedMessage.version == 1 &&
+            deserializedMessage.description == "An application has been submitted for an Approved Premises placement" &&
+            deserializedMessage.detailUrl == domainEventToSave.data.eventDetails.applicationUrl &&
+            deserializedMessage.occurredAt == domainEventToSave.occurredAt &&
+            deserializedMessage.additionalInformation.applicationId == applicationId &&
+            deserializedMessage.personReference.identifiers.any { it.type == "CRN" && it.value == domainEventToSave.data.eventDetails.personReference.crn } &&
+            deserializedMessage.personReference.identifiers.any { it.type == "NOMS" && it.value == domainEventToSave.data.eventDetails.personReference.noms }
+        }
+      )
+    }
```
edavey, line 147: I'm thinking that it's probably quite important to demonstrate that the Domain Event Service does NOT emit an event if it's unable to `save()` the event? 

Should we have an additional scenario where `domainEventRespositoryMock.save()` returns something unhappy or raises an error and then assert that `mockHmppsTopic.snsClient.publish()` received no messages?

edavey (dismissed):
This looks really good! I'm just thinking that we ought to describe the behaviour (not publishing) in the scenario where we can't save the event?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/DomainEventServiceTest.kt
brownbearsoftware: Good shout, I've added a test for this case too

edavey (approved):
👍 thanks

PR #413, opened 2023-02-08
Update delius integrations service hostname

tahb:
The value we are choosing was supplied more recently than the value ending .dsd.io [1]. 

We don't think this is causing an issue yet but it may when the integrations service comes to be used.

[1] https://github.com/ministryofjustice/hmpps-approved-premises-api/commit/8e4676c99386a425b35bea24166efa55de950693#diff-bb4e80ba0a6c43663498b44c19e42c074dcceac3d9564b49801202073b5c87ff


edavey (approved):
yes this is correct and more up to date since the move to the Cloud Platform

PR #414, opened 2023-02-09
1226 - Record managing team

brownbearsoftware:
Record Delius team codes when creating an AP Application.  Only allow creation if at least one team is present for the calling user.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/UserEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/UserEntity.kt
@@ -41,7 +41,8 @@ data class UserEntity(
   val id: UUID,
   var name: String,
   val deliusUsername: String,
-  val deliusStaffIdentifier: Long,
+  var deliusStaffCode: String?,
+  var deliusStaffIdentifier: Long,
   var email: String?,
   var telephoneNumber: String?,
   @OneToMany(mappedBy = "createdByUser")
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ApplicationService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ApplicationService.kt
@@ -13,15 +13,19 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.Region
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.StaffMember
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.Team
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ServiceName
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.ApDeliusContextApiClient
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.ClientResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.CommunityApiClient
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApplicationEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApplicationRepository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApplicationTeamCodeEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApplicationTeamCodeRepository
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApprovedPremisesApplicationEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApprovedPremisesApplicationJsonSchemaEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.OfflineApplicationEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.OfflineApplicationRepository
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.TemporaryAccommodationApplicationEntity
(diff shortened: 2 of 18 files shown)


brownbearsoftware:

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/UserEntity.kt
```diff
@@ -41,7 +41,8 @@ data class UserEntity(
   val id: UUID,
   var name: String,
   val deliusUsername: String,
-  val deliusStaffIdentifier: Long,
+  var deliusStaffCode: String?,
```
edavey, line 44: so is it might worth saying explicitly in the commit message and PR that once all pre-existing records have been updated we intend to make this a mandatory property?

And does that make the case for some kind of data migration process to attempt to update all at the same time rather than having to wait an undefined period for users to log in and be re-fetched individually? 

If I'm understanding properly, that is?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
```diff
@@ -102,6 +104,25 @@ class UserService(
...
+      name = "${staffUserDetails.staff.forenames} ${staffUserDetails.staff.surname}"
+      deliusStaffIdentifier = staffUserDetails.staffIdentifier
+      deliusStaffCode = staffUserDetails.staffCode
+      email = staffUserDetails.email
+      telephoneNumber = staffUserDetails.telephoneNumber
+    }
+
+    return userRepository.save(user)
+  }
+
```
edavey, line 125: hmm,

I think we may be replicating some similar logic within the `getUserForId()` function?

https://github.com/ministryofjustice/approved-premises-api/blob/4023d11c9a5ce2a1b32824dacda85c3cc9bc4bc8/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt#L47-L65

perhaps we need to refactor so that both `getUserForUsername()` and `getUserForId()` use a common `updateUserIfChanged()` function?

At any rate I think we'll need to ensure that the `staffCode` is updated in both pathways?

src/main/resources/db/migration/all/20230209144819__team_codes_for_ap_applications.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE approved_premises_application_team_codes (
+    id UUID NOT NULL,
+    application_id UUID NOT NULL,
+    team_code TEXT NOT NULL,
+    PRIMARY KEY (id),
+    FOREIGN KEY (application_id) REFERENCES applications(id),
+    UNIQUE (application_id, team_code)
```
edavey, line 7: seems like a good design. 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ApplicationService.kt
```diff
@@ -182,6 +198,16 @@ class ApplicationService(
...
 
+    managingTeamCodes.forEach {
+      createdApplication.teamCodes += applicationTeamCodeRepository.save(
+        ApplicationTeamCodeEntity(
+          id = UUID.randomUUID(),
+          application = createdApplication,
+          teamCode = it
+        )
+      )
+    }
```
edavey, line 209: I'm curious about the state we end up in if there's a failure in storing the team codes on the previously created application. It doesn't look to me like these are wrapped up in a transaction?

Would the application be rolled back if the team codes weren't created?

If not, is it possible to send the whole package of db entities for creation to JPA together e.g. something like

```kt
val createdApplication = applicationRepository.save(
      ApprovedPremisesApplicationEntity(
        id = UUID.randomUUID(),
        crn = crn,
        createdByUser = user,
        data = null,
        document = null,
        ...
        teamCodes = teamCodes.map { buildTeamCode(it) } // something like this???
      )
    )
```

edavey (commented):
Very nice work, super clear.

I have a question about the application creation and the team code creation appearing to be 2 separate transactions. Perhaps I'm wrong about how this works, but I don't think we'd want to make it technically possible for an application to exists without its team codes...

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/UserEntity.kt
brownbearsoftware: > And does that make the case for some kind of data migration process to attempt to update all at the same time rather than having to wait an undefined period for users to log in and be re-fetched individually?

Agreed, will do this in a subsequent PR as will need to do some scaffolding similar to seeding.  Will update the commit message to reflect this.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
brownbearsoftware: Good point, I'm hesitant to combine these methods as getUserForId calls Community API every time whereas getUserForUsername just uses what's stored in the DB (and is called much more prolifically.)

Have updated getUserForId to also update staffCode though.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ApplicationService.kt
brownbearsoftware: There's a `@Transactional` annotation on `ApplicationsController::applicatonsPost` so an exception inserting the team codes will cause the Application insert to be rolled back.

Have added an integration test (`Create new application returns 500 and does not create Application without team codes when write to team code table fails`) to prove this.

edavey: ah, interesting, thank you!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
edavey: 👍 sounds good, thks!

edavey (approved):
👍 thks, looking good

PR #417, opened 2023-02-13
(1216) Fetch assessment for application

pezholio:
This adds a new `applications/{id}/assessment` endpoint to allow us to get an application's correct assessment. This will allow us to easily fetch the latest assessment for an application together with the application data, which we need for allocations.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/AssessmentEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/AssessmentEntity.kt
@@ -19,7 +19,7 @@ import javax.persistence.Table
 @Repository
 interface AssessmentRepository : JpaRepository<AssessmentEntity, UUID> {
   fun findAllByAllocatedToUser_Id(userId: UUID): List<AssessmentEntity>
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/AssessmentEntity.kt
```diff
@@ -20,6 +21,7 @@ import javax.persistence.Table
 interface AssessmentRepository : JpaRepository<AssessmentEntity, UUID> {
   fun findAllByAllocatedToUser_Id(userId: UUID): List<AssessmentEntity>
   fun findByApplication_IdAndReallocatedAtNull(applicationId: UUID): AssessmentEntity
+  fun findByApplication_Id(applicationId: UUID): AssessmentEntity?
```
brownbearsoftware, line 24: This will throw an exception when more than 1 Assessment exists for an Application which would be the case after a reallocation.  The method above `findByApplication_IdAndReallocatedAtNull` will do what you need here though as reallocating is transactional so there can only ever be one Assessment per Application where reallocatedAt is null.

brownbearsoftware (dismissed):

brownbearsoftware (approved):

PR #418, opened 2023-02-13
1226 - Migration process to update users from Community API

brownbearsoftware:
This adds scaffolding for running ad-hoc data migation tasks that require kotlin code (e.g. where calls to upstream services need to be made.)

These can be called in a similar way to the seeding jobs - via a POST to `/migration-job` with the type of migration.  This does not require a JWT but can only be done locally.

Implemented is the first such task which updates all of the users in our `users` table from Community API.  The need for which is shown here: https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/414#discussion_r1101781763

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
@@ -45,7 +45,7 @@ class UserService(
     return userRepository.findAll(hasQualificationsAndRoles(qualifications, roles), Sort.by(Sort.Direction.ASC, "name"))
   }
 
-  fun getUserForId(id: java.util.UUID): AuthorisableActionResult<UserEntity> {
+  fun updateUserFromCommunityApiById(id: UUID): AuthorisableActionResult<UserEntity> {
     var user = userRepository.findByIdOrNull(id) ?: return AuthorisableActionResult.NotFound()
     val staffUserDetailsResponse = communityApiClient.getStaffUserDetails(user.deliusUsername)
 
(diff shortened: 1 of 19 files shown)


brownbearsoftware:

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
```diff
@@ -67,7 +67,7 @@ class UserService(
 
   fun getUserForUsername(username: String): UserEntity {
     val existingUser = userRepository.findByDeliusUsername(username)
-    if (existingUser != null && existingUser.deliusStaffCode == null) return updateUserFromCommunityApi(existingUser)
+    if (existingUser != null && existingUser.deliusStaffCode == null) return updateUserFromCommunityApiById(existingUser)
     if (existingUser != null) return existingUser
```
edavey, line 71: thanks for tackling this area of potential confusion. Maybe it's just me but I find it confusing that we now have 2 functions with the same name but different signatures:

1. `fun updateUserFromCommunityApiById(id: UUID)` and
2. `fun updateUserFromCommunityApiById(user: UserEntity)`

I read the function names `*ById` as indicating that they take an ID as the main argument -- which isn't the case with the second private function. Did you mean to call the private function `updateUserFromCommunityApiByUser()` perhaps?

edavey (approved):
👍 this looks ideal -- I just had one query on the `UserService` refactoring, where possibly a typo has confused me?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
brownbearsoftware: Not too sure what happened here!  Have put this method back to its original name.

edavey: probably IntelliJ getting carried away!

PR #422, opened 2023-02-14
Increase initial delay on liveness probe to 60s

brownbearsoftware:
Deployments are often timing out because pods are being restarted before they can become ready.  This should hopefully give the pods enough time to startup.


pezholio (approved):
LGTM 👍

PR #423, opened 2023-02-15
1135 - Publish Domain Events

brownbearsoftware:
This is the same work as in https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/421 with some tweaks to the config to hopefully resolve the binding issues that caused it to be rolled back.

Namely:
 - remove dash from topic name, I'm not sure this should matter but the docs say lowercase only
 - only mount secrets in dev (as this is the only place they currently exist)


edavey (approved):
🚀

PR #424, opened 2023-02-15
(1242) Only return un-reallocated assessments on list

pezholio:
When an assessment has been reallocated to a new user, the assessments are treated as readonly, so we should not return them in the list of assessments.


pezholio:
[Hide reallocated Assessments](https://trello.com/c/YSOwkNh8/1242-hide-reallocated-assessments)

brownbearsoftware (approved):

PR #427, opened 2023-02-15
Log for failed requests to registrations endpoint

brownbearsoftware:
This should help us investigate an issue in dev where a CRN with registrations is returning as error in the enveloped response.


pezholio (approved):

edavey (approved):
👍 it would be good to see this accompanied by a unit test describing the offender service's exception handling and logging behaviour in this scenario where registrations are missing from Delius.

PR #431, opened 2023-02-16
Domain Events Fixes

brownbearsoftware:
Correct detailUrl for Application Submitted domain events - was previously set to the frontend application URL rather than the event route on the API.

Log SNS receipt info when emitting events - this should help us with debugging etc.


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ApplicationTest.kt
```diff
@@ -1116,7 +1116,7 @@ class ApplicationTest : IntegrationTestBase() {
 
           assertThat(emittedMessage.eventType).isEqualTo("approved-premises.application.submitted")
           assertThat(emittedMessage.description).isEqualTo("An application has been submitted for an Approved Premises placement")
-          assertThat(emittedMessage.detailUrl).isEqualTo("http://frontend/applications/$applicationId")
+          assertThat(emittedMessage.detailUrl).startsWith("http://frontend/events/application-submitted/")
```
pezholio, line 1119: Could we maybe match against a regex here? I'm guessting your doing `startsWith` because we don't know the ID of the event at this point, but it'd be good to have confidence that a UUID is included in the payload

pezholio (approved):

PR #432, opened 2023-02-16
Move and add to infrastructure documentation

tahb:
Includes notes on the way accessing API logs is different from the frontend. If there wasn't this difference and potential for further divergences I would have chosen to add this in once in Confluence. We may still want to do this.

[We've added a similar bit of documentation to the TA frontend recently](https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/155).

Includes some automatic markdown linting which I noticed my editor found as an optional contribution that's riding on this proposals coat tails.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,8 +2,10 @@
 
 This is the shared backend for the Community Accommodation services, currently:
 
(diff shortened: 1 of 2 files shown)


tahb:
Changed my mind, I think the majority of this belongs in Confluence.

PR #433, opened 2023-02-16
Make staffcode non nullable

brownbearsoftware:
This should always be present a user, it was previously nullable to allow us to do a migration.  Now that's complete, we can set this to not nullable.


brownbearsoftware:

edavey (commented):
Did we already verify with @danlivings-dxw that this constraint won't invalidate any existing CAS3 users or cause problems creating new ones?

danlivings-dxw:
@edavey I don't think this change would cause us problems as I believe all of our users must have a Delius account to use the service. Might be worth double checking with @tahb?

brownbearsoftware:
All users definitely have Delius accounts or they wouldn't be able to use the API at all as it stands since we're checking for `ROLE_PROBATION` which you can only get on Delius accounts - so I'm confident this will be fine.

danlivings-dxw (approved):

PR #434, opened 2023-02-17
1225 - Limit application access by caseload

brownbearsoftware:
Allow users to view an Application if they are a member of a team that the CRN was managed by when the Application was created.


brownbearsoftware:
[Limit applicant access to CRNs using membership of "managing teams"](https://trello.com/c/cigy4nXP/1225-limit-applicant-access-to-crns-using-membership-of-managing-teams)

edavey (approved):
neatly done, thanks!

PR #436, opened 2023-02-20
Adding CircleCI job to trigger UI tests on deployment

adamhughes86:


adamhughes86:
I've ran a local test in a different environment and this _can_ work. It's setup a bit different for this project though so the trigger may not work on this first pass. I'll monitor the build when it's merged though to see what happens

.circleci/config.yml
```diff
@@ -72,6 +95,9 @@ workflows:
             - hmpps-approved-premises-api-prod
           requires:
             - request-prod-approval
+      - trigger-ui-tests:
+          requires:
+            - request-prod-approval
```
edavey, line 100: I'm not clear why these tests require prod approval? My understanding was that they run in the dev environment.

adamhughes86: I was thinking that the api deploys to prod and then the UI tests are run to validate the latest

adamhughes86: I've changed this to a deploy_dev requirement now. I've tested it by adding the feature branch to the filtered branches list and it triggers the UI builds once the api is deployed to dev

edavey (approved):
thanks a lot for figuring this out!

PR #437, opened 2023-02-20
847 - Hard Delete Booking

brownbearsoftware:
This adds an endpoint at DELETE `/internal/booking/{booking id}` that can only be called from localhost.

This will delete the Booking and all dependent entities such as arrivals, nonarrivals, extensions etc.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
@@ -318,6 +318,35 @@ class BookingService(
     return GetBookingForPremisesResult.Success(booking)
   }
 
+  @Transactional
+  fun deleteBooking(booking: BookingEntity) {
+    if (booking.arrival != null) {
+      arrivalRepository.delete(booking.arrival!!)
+    }
+
+    if (booking.departure != null) {
+      departureRepository.delete(booking.departure!!)
+    }
+
+    if (booking.nonArrival != null) {
+      nonArrivalRepository.delete(booking.nonArrival!!)
+    }
+
+    if (booking.cancellation != null) {
+      cancellationRepository.delete(booking.cancellation!!)
+    }
+
+    if (booking.confirmation != null) {
+      confirmationRepository.delete(booking.confirmation!!)
+    }
+
+    booking.extensions.forEach {
+      extensionRepository.delete(it)
+    }
+
+    bookingRepository.delete(booking)
+  }
+
   private fun serviceScopeMatches(scope: String, booking: BookingEntity): Boolean {
     return when (scope) {
       "*" -> true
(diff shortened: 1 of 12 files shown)


brownbearsoftware:
[Script to hard delete a booking](https://trello.com/c/HGhDgiyT/847-script-to-hard-delete-a-booking)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
```diff
@@ -318,6 +318,35 @@ class BookingService(
     return GetBookingForPremisesResult.Success(booking)
   }
 
+  @Transactional
```
tahb, line 321: 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
```diff
@@ -318,6 +318,35 @@ class BookingService(
...
+  fun deleteBooking(booking: BookingEntity) {
```
tahb, line 322: I wondered if it was possible to factor away the extra knowledge this method has to have about all the possible booking associations. Doing so would reduce the repetition and therefore make the code easier to change. Any new association or change wouldn't have to be weaved through multiple places. 

[Rails leverages SQL to do this type of thing](https://apidock.com/rails/ActiveRecord/Base/destroy) for us in a DRY way:

```
# File activerecord/lib/active_record/base.rb, line 2614
      def destroy
        unless new_record?
          connection.delete(
            "DELETE FROM #{self.class.quoted_table_name} " +
            "WHERE #{connection.quote_column_name(self.class.primary_key)} = #{quoted_id}",
            "#{self.class.name} Destroy"
          )
        end

        @destroyed = true
        freeze
      end
```

I don't think we need to take action on this, just sharing some idle thoughts on how it might be even better incase it's of interest 🙂

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/DeleteBookingTest.kt
```diff
@@ -0,0 +1,151 @@
...
+import org.springframework.data.repository.findByIdOrNull
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ExtensionRepository
+
+class DeleteBookingTest : IntegrationTestBase() {
+  @SpykBean
+  lateinit var realExtensionRepository: ExtensionRepository
+
+  @Test
+  fun `Deleting a Booking successfully deletes all related entities and deletes the Booking itself`() {
+    val premises = approvedPremisesEntityFactory.produceAndPersist {
```
tahb, line 16: I think we just need premises to be set up for us to enable the rest of the bookings to be set up? If so, I'd probably reach to putting this into a reusable helper. It could give others the steer and encourage a single definition and help the reader skip to the essential bits. Just an idea, no action needed from me 👍

tahb (approved):
Wonderful PR! Thanks for jumping onto it. I've left some passing thoughts just incase it's of interest but nothing essential from me.

PR #439, opened 2023-02-21
Create service-specific tables and JPA entities for lost beds

danlivings-dxw:
>  See [ticket #806 on the CAS3 Trello board](https://trello.com/c/RDiktZxl/806-introduce-service-specific-lost-bed-tables-entities).

This PR is part of the work to implement support for lost beds within the Temporary Accommodation service. It will enable service-specific behaviour such as validation for endpoints concerning lost beds.

Changes in this PR:
- The `approved_premises_lost_beds` and `temporary_accommodation_lost_beds` tables have been introduced to provide service-specific data and the `lost_beds` table now contains common data shared between both services.
- The `LostBedsEntity` now has `ApprovedPremisesLostBedsEntity` and `TemporaryAccommodationLostBedEntity` as subclasses to do likewise within the API itself.
- The existing API logic has been updated to support these changes while preserving the current behaviour.


brownbearsoftware (approved):
Nice 🚀

PR #440, opened 2023-02-22
Don't return Exception messages from unhandled Exceptions in problem responses

brownbearsoftware:


brownbearsoftware:
[Don't return exception message in problem response for unhandled exceptions](https://trello.com/c/XShreKsw/1271-dont-return-exception-message-in-problem-response-for-unhandled-exceptions)

danlivings-dxw (approved):

PR #441, opened 2023-02-22
1259 - Emit Booking Made Domain event

brownbearsoftware:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
```diff
@@ -194,10 +193,7 @@ class BookingService(
                 name = approvedPremises.name,
                 apCode = approvedPremises.apCode,
                 legacyApCode = approvedPremises.qCode,
-                probationArea = ProbationArea(
-                  code = staffDetails.probationArea.code,
-                  name = staffDetails.probationArea.description
-                )
+                localAuthorityAreaName = approvedPremises.localAuthorityArea!!.name
```
edavey, line 196: The probation area here was intended to be the probation area of the AP itself. But on further investigation I realise that a premises isn't associated with a probation area. It is associated with:

- **local authority area** e.g. Kirklees, Bradford
- **probation region** e.g. Yorkshire & The Humber, North East
- **ap area** e.g. North East (through probation region)

For reporting purposes I think local authority area is the most useful information to include, but will check this with AJB / performance hub. 

This change is going to break [the "integration" which consumes this event](https://github.com/ministryofjustice/hmpps-probation-integration-services/blob/98701178578ed616323657da2b4ebd2979fc4322/projects/approved-premises-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ApprovedPremisesService.kt#L107-L119) as the premises' probation area is being used to set the `team` on the created Delius contact. 

Perhaps the probation integrations team could derive the correct team from the `premises.legacyApCode` instead? That's the Delius "Qcode" e.g. `Q057`. 

☝️ @marcus-bcl @anthony-britton-moj

edavey (commented):
Thanks Joe.

I've tagged Marcus and Anthony from Probation integrations regarding the proposed removal of `premises.probationArea`...

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
```diff
@@ -366,12 +387,80 @@ class PremisesController(
...
         originalArrivalDate = body.arrivalDate,
         originalDepartureDate = body.departureDate,
-        createdAt = OffsetDateTime.now(),
+        createdAt = bookingCreatedAt,
         application = if (associateWithOnlineApplication) newestSubmittedOnlineApplication else null,
         offlineApplication = if (associateWithOfflineApplication) newestOfflineApplication else null
       )
     )
 
+    if (body is NewApprovedPremisesBooking && booking.application != null) {
```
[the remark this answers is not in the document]
brownbearsoftware: It's no longer possibe to create a CAS1 Booking without an associated Application, there might still be some in dev/preprod that this will guard against.  The main reason this check is here though is that we need an associated Online Application to produce a Domain Event since we need the Delius Event Number which we don't know for Offline Applications.  Have made this a bit clearer in the refactored version: https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/441/files#diff-6c0f9c23780af629762d73597aa46737ce4787a4616d75563fcf92d23039827fR142

brownbearsoftware:
[booking.made Domain Event](https://trello.com/c/0p0YSIGN/1259-bookingmade-domain-event)

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/CruService.kt
```diff
@@ -0,0 +1,19 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.unit.service
+
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.CruService
+
+class CruService {
```
edavey, line 7: I think the file and class name ought to include "Test"? Not sure if the test runner relies on that naming or not?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/BookingServiceTest.kt
```diff
@@ -1144,4 +1180,370 @@ class BookingServiceTest {
...
+            localAuthorityAreaName = premises.localAuthorityArea!!.name
+          )
+        }
+      )
+    }
+  }
+
+  @ParameterizedTest
+  @EnumSource(value = UserRole::class, names = [ "MANAGER", "MATCHER" ])
+  fun `createApprovedPremisesBooking saves Booking but does not create Domain Event when associated Application is an Offline Application`(role: UserRole) {
```
edavey, line 1412: Perhaps it will be helpful to document here **why** we can't produce a Domain Event for offline applications? e.g.

`createApprovedPremisesBooking saves Booking but does not create Domain Event when associated Application is an Offline Application as Event Number is not present`

I think this is important to highlight as communicating arrivals and departures of people whose applications were made under the old regime (outside the new service) is an open problem which we need to explore and resolve.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
edavey: ✅

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/CruService.kt
brownbearsoftware: Good spot, have amended.  The `Test` postfix is just convention, actual tests are discovered by JUnit via reflection looking for the @Test annotation.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/BookingServiceTest.kt
brownbearsoftware: Fair point, have changed the test name to reflect

edavey (approved):
Thanks for the multiple revisions!

PR #442, opened 2023-02-23
Introduce service scope for lost bed reasons

danlivings-dxw:
> See [ticket #809 on the CAS3 Trello board](https://trello.com/c/BckxYLXE/809-make-lost-bed-reasons-service-specific).

This PR is part of the work to implement support for lost beds within the Temporary Accommodation service. It introduces a service scope to lost bed reasons to allow each service to provide a different set of reasons for lost beds.

Changes in this PR:
- Existing lost bed reasons have been given a service scope of `"approved-premises"`, except "Other", which has a scope of `"*"`
- The list of reasons for Temporary Accommodation has been seeded using a Flyway migration
- The `GET /reference-data/lost-bed-reasons` endpoint now accepts an `X-Service-Name` header, which allows for filtering of the returned list of reasons if provided
- The `PremisesService.createLostBeds` method returns `"incorrectLostBedReasonServiceScope"` as a validation error if the provided lost bed reason does not have a valid scope for the premises.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ReferenceDataController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ReferenceDataController.kt
@@ -98,8 +98,11 @@ class ReferenceDataController(
     return ResponseEntity.ok(cancellationReasons.map(cancellationReasonTransformer::transformJpaToApi))
   }
 
-  override fun referenceDataLostBedReasonsGet(): ResponseEntity<List<LostBedReason>> {
-    val lostBedReasons = lostBedReasonRepository.findAll()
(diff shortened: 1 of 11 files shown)


brownbearsoftware (approved):
LGTM - not sure if we need a change on the AP frontend to start sending the service header or if it's sent on all requests automatically @richpjames ?

danlivings-dxw:
@brownbearsoftware I had a quick poke around and it looks like the `X-Service-Name` header is automatically sent to the API.

- [`RestClient` creates a default header if the config specifies the service name](https://github.com/ministryofjustice/hmpps-approved-premises-ui/blob/14b93d68a6cc9320de6be01bada035126137c891/server/data/restClient.ts#L50-L53)
- For subclasses of `RestClient` it seems that they get the config from [here](https://github.com/ministryofjustice/hmpps-approved-premises-ui/blob/14b93d68a6cc9320de6be01bada035126137c891/server/config.ts#L78-L86), which provides a service name

PR #445, opened 2023-02-24
Always return Drug and Alcohol OASys sections

brownbearsoftware:
Don't offer Drug & Alcohol on the selection endpoint, always return them on the sections endpoint.

The selection endpoint is used to build the list of OASys sections that are auto-imported or optionally imported in the UI.  Removing these sections from the selection endpoint will result in them no longer appearing in any of the checkbox lists.

The section endpoint currently only returns the drug and alcohol sections if they are specifically requested, this changes the endpoint to always return the Drug and Alcohol sections.


brownbearsoftware:
[OASys Import Screen - remove drug and alcohol from checklist](https://trello.com/c/27zZbIRv/1119-oasys-import-screen-remove-drug-and-alcohol-from-checklist)

danlivings-dxw (approved):

PR #457, opened 2023-02-28
Fix for application submitted on in Person Arrived Domain Event

brownbearsoftware:
Integrations team were expecting an OffsetDateTime & we were also sending the wrong value

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
@@ -367,7 +367,7 @@ class BookingService(
                 legacyApCode = approvedPremises.qCode,
                 localAuthorityAreaName = approvedPremises.localAuthorityArea!!.name
               ),
-              applicationSubmittedOn = occurredAt.toLocalDate(),
+              applicationSubmittedOn = application.submittedAt!!,
               keyWorker = StaffMember(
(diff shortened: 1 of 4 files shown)


brownbearsoftware:
[Person.Arrived Domain Event](https://trello.com/c/8jzwXQHb/1279-personarrived-domain-event)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
```diff
@@ -367,7 +367,7 @@ class BookingService(
                 legacyApCode = approvedPremises.qCode,
                 localAuthorityAreaName = approvedPremises.localAuthorityArea!!.name
               ),
-              applicationSubmittedOn = occurredAt.toLocalDate(),
+              applicationSubmittedOn = application.submittedAt!!,
```
edavey, line 370: My understanding is that this was added (see https://github.com/ministryofjustice/hmpps-approved-premises-api/commit/684c5ff243ec1d170d8d495cffb4426403803fa6) so that the **date** of the submission could be added to the NSI when it's made at Delius as part of the processing of `person.arrived`.

Perhaps it's worth confirming with PI whether it's a date or datetime which is actually needed on the NSI? If the former then maybe the integration ought to change?

However, if the domain event changes at our end as proposed here, then I think we will need to rename the property to `applicationSubmittedAt` rather than `applicationSubmittedOn` -- so that we stick to our convention that dates end with `*On` and datetimes with `*At`...

PR #458, opened 2023-03-01
Fix for application submitted on in Person Arrived Domain Event

brownbearsoftware:
We were sending the wrong value for this


edavey (approved):
well spotted! thanks

PR #460, opened 2023-03-01
(1276) Optionally check caseload on user search

pezholio:
This adds a `checkCaseload` option to the /people/search endpoint. If this is provided, then we also call the `getTeamsManagingCase` endpoint to check if the offender is part of the logged in user’s caseload. If they aren’t then we return a 403 response.


pezholio:
[API returning 400 rather than 403 when applicant not authorised to create application](https://trello.com/c/gwUTNw2m/1276-api-returning-400-rather-than-403-when-applicant-not-authorised-to-create-application)

brownbearsoftware (approved):

PR #461, opened 2023-03-01
1287 - Person departed domain event

brownbearsoftware:


edavey (approved):
This all looks correct. Thanks for seeding the departure reasons and move-on categories .

PR #462, opened 2023-03-02
Implement cancellations for lost beds

danlivings-dxw:
> See [ticket #812 on the CAS3 Trello board](https://trello.com/c/7BTZ0rpr/812-a-user-can-cancel-a-void-booking).

This PR is part of the work to implement support for lost beds within the Temporary Accommodation service. It allows users of the API to cancel lost beds.

Changes in this PR:
- A `lost_bed_cancellations` table has been created that holds the cancelled lost bed, the timestamp for when it was cancelled, and any notes about the cancellation.
- The OpenAPI specification has been updated:
  - A `POST /premises/{premisesId}/lost-beds/{lostBedId}/cancellations` endpoint has been created to allow users to cancel a lost bed.
  - The `LostBedCancellation` and `NewLostBedCancellation` schemas have been introduced to support the above endpoint.
  - The `LostBed` schema now contains `status` and `cancellation` properties. The `status` is either `"active"` or `"cancelled"`, and the `cancellation` is either `null` or a `LostBedCancellation` instance.
- The `PremisesService.cancelLostBed` method has been introduced to create and persist a lost bed cancellation.
- The `BookingTest.withinSeconds` utility method has been moved into its own utility file (`src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/util/Matchers.kt`) so that it can be reused in `LostBedsTest`.


brownbearsoftware:
Does BookingService::getAvailabilityForRange also need to be updated?

brownbearsoftware (approved):

PR #463, opened 2023-03-02
1288 Seed Job for premises and room characteristics

edavey:
This is a preliminary to seeding the actual premises, rooms and beds in the CAS1 initial area (North East).

This allows us to create and update `characteristics` which can be scoped to:

- service: `approved-premises` or `temporary-accommodation`
- model: `premises` or `room`

For example, a CSV in this format
```
| characteristic_property_name | characteristic_name                           | service_scope     | model_scope |
| ---------------------------- | --------------------------------------------- | ----------------- | ----------- |
| hasWideDoor                  | Is the door to this room at least 900mm wide? | approved-premises | room        |
| hasFixedMobilityAids         | Are there fixed mobility aids in this room?   | approved-premises | room        |
| isIap                        | Is this an IAP?                               | approved-premises | premises    |
| isPipe                       | Is this AP a PIPE?                            | approved-premises | premises    |
```
will:

-  create characteristics where property name, service scope and domain scope
don't match existing records.

- update existing characteristics with the given name

These characteristics aren't all going to be surfaced or used in matching
people to suitable accommodation, but it's appropriate for all important
accommodation characteristics to be mastered in the new service.

Note that:

- a new `characteristics.property_name` attribute is added to provide a lasting key to the characteristic whilst allowing the `name` to be a changeable human-friendly string
- existing test/dev migration which seeds some characteristics no longer drops the table on each migration run
- we introduce `enum` control on permitted characteristic `serviceScope` and `modelScope` values
- the seed files (not sensitive) to run this seed job `premises_characteristics.csv` and `room_characteristics.csv` are attached to [Trello ticket 1288](https://trello.com/c/iFbjTR9P/1288-seed-ne-premises-beds-and-rooms)
- I apologise that this PR isn't broken into smaller commits to ease reviewing. My Kotlin skills are too low to give me much foresight! I can go back and break it apart if necessary.


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/SeedCharacteristicsTest.kt
```diff
@@ -0,0 +1,154 @@
...
+import org.assertj.core.api.Assertions
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.TestInstance
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.SeedFileType
+
+@TestInstance(TestInstance.Lifecycle.PER_METHOD)
+class SeedCharacteristicsTest : SeedTestBase() {
+  @Test
+  fun `Attempting to seed characteristic with missing characteristic_name field fails and logs error`() {
+    characteristicRepository.deleteAll() // can't seem to do this in a BeforeEach setUp function
```
brownbearsoftware, line 12: I deleted all of these and added the following which worked on my machine:

```
@BeforeEach
  fun removeDefaultCharacteristicsFromDatabaseMigrations() {
    characteristicRepository.deleteAll()
  }
```

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/SeedCharacteristicsTest.kt
```diff
@@ -0,0 +1,154 @@
...
+    withCsv(
+      "invalid-characteristics-missing-name",
+      "characteristic_property_name,characteristic_name,service_scope,model_scope\n" +
+        "hasWideDoor,Is the door to this room at least 900mm wide?,approved-premises,room\n" +
+        "hasWideDoor,,temporary-accommodation,room\n"
+    )
+
+    seedService.seedData(SeedFileType.characteristics, "invalid-characteristics-missing-name")
+
+    Assertions.assertThat(characteristicRepository.count()).isEqualTo(0)
```
brownbearsoftware, line 23: Would import `assertThat` rather than `Assertions` so these read a bit more naturally:

- Add `import org.assertj.core.api.Assertions.assertThat` to the imports at the top of the file
- Remove all of the `Assertions.`

brownbearsoftware (approved):
LGTM.  The checks are failing because of a ktLint issue - won't help this time since you'll need to rebase the fixes in so precommit hooks won't run but if you run this gradle task you should be ok moving forwards.

image

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/SeedCharacteristicsTest.kt
edavey: yes, that does work! Thanks. Not sure exactly what was going wrong before...

edavey:
> The checks are failing because of a ktLint issue - won't help this time since you'll need to rebase the fixes in so precommit hooks won't run but if you run this gradle task you should be ok moving forwards.
 
Thanks, I've added that `KtlintFormat` pre-commit hook now.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/SeedCharacteristicsTest.kt
edavey: thanks, I've done that.

PR #464, opened 2023-03-06
Postcode Districts

brownbearsoftware:
Add Postcode District data, this is required for scoping out search


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PostCodeDistrictEntity.kt
```diff
@@ -0,0 +1,17 @@
...
+import javax.persistence.Table
+
+@Entity
+@Table(name = "postcode_districts")
+data class PostCodeDistrictEntity(
+  @Id
+  val id: UUID,
+  val outcode: String,
+  val latitude: BigDecimal,
+  val longitude: BigDecimal
```
edavey, line 16: In our experiments in alpha we used the `geometry(point, 4326)` type for PostGIS. Are you thinking of adding that later once we get into querying the data? I do think it would probably make sense to store the coordinates separately as lat and lng also, at least initially...

edavey (approved):
This looks good -- I'm not sure where we ought to embrace the `geometry(point, 4326)` column type at this point or later?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PostCodeDistrictEntity.kt
brownbearsoftware: I hadn't thought that far ahead if I'm honest but I think I'll merge this in as-is then look at PostGIS as will need to do some work to install it if we go that way

PR #466, opened 2023-03-07
Seed coordinates in Premises seeding

edavey:
We add `latitude` and `longitude` columns to the generic `premises` table, using the "precision double" type with 15 decimal digits precision.

https://www.postgresql.org/docs/current/datatype-numeric.html#DATATYPE-NUMERIC-DECIMAL

We expect that once we begin to do geospatial queries using this information we'll migrate both coordinates to a PostGIS `geometry(point, 4326)` data type.

We've added the latitude and longitude columns to the shared premises table as we believe that geospatial information is common and core to all varieties of accommodation.


edavey:
@danlivings-dxw Could you please review as I've assumed that CA3 would be happy to have (currently nullable) geospatial data on the main `premises` table. If there's a better / more considerate way for us to all this information, please let me know!

brownbearsoftware (approved):
LGTM

danlivings-dxw (approved):
@edavey I don't expect this to cause us any issues. As of now it might be cleaner to have to have the latitude/longitude fields on the AP-specific table, but thinking ahead there's the possibility that we'd want some kind of spatial search. I don't know what the timeline for that is though.

PR #472, opened 2023-03-08
Add all roles to all dev users

pezholio:
This adds all the available roles to the development users. I’ve also renamed the Delius username for `JIMSNOWLDAP` as it comes back from Delius as all uppercase.

### DIFF
--- a/src/main/resources/db/migration/local+dev/R__1_create_dev_users.sql
+++ b/src/main/resources/db/migration/local+dev/R__1_create_dev_users.sql
@@ -3,10 +3,203 @@
 --These are randomly generated names
 
 INSERT INTO "users" (id, name, delius_username, delius_staff_identifier, probation_region_id, delius_staff_code) VALUES
-    ('aa30f20a-84e3-4baa-bef0-3c9bd51879ad', 'Default User', 'JimSnowLdap', 2500041001, '43606be0-9836-441d-9bc1-5586de9ac931', 'STAFFCODE'), -- South West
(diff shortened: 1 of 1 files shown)


brownbearsoftware (approved):

src/main/resources/db/migration/local+dev/R__1_create_dev_users.sql
```diff
@@ -3,10 +3,203 @@
...
+    'aa30f20a-84e3-4baa-bef0-3c9bd51879ad'
+  ),
+  (
+    '0354ff61-295a-4d06-b9e7-ae9fd53eba12',
+    'WORKFLOW_MANAGER',
+    'aa30f20a-84e3-4baa-bef0-3c9bd51879ad'
+  ),
+  (
+    'dd949e2b-5f9b-4560-b3bc-c3ad6793fb28',
+    'APPLICANT',
```
edavey, line 46: `APPLICANT` is not a role, so will cause the enum to blow up...

pezholio: Is it not? I was going from UserRoles as per https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/4928a7455cbb35274d19adbd0563c24301a1f230/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/UserEntity.kt#L79

edavey: Interesting! I stand corrected -- it won't blow up as the `APPLICANT` enum is indeed present. 

However, APPLICANT is not an internally managed role. It's a pseudo-role. It refers to the capabilities which a logged in user defaults to when they don't have any of the internally managed roles. i.e as an applicant you can make and view applications for people who are managed by teams of which you are a member...

So I think it ought to be removed from the enum...?

PR #476, opened 2023-03-08
1288: Seed premises characteristics for individual premises

edavey:
### Set characteristics from individual columns within the `SeedApprovedPremisesJob`.

i.e. this makes entries in the join table `premises_characteristics` for each characteristic of
each premises, as opposed to the available characteristics which are handled 
by the `SeedCharacteristicsJob` (see https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/463)

Previously these were seeded from a comma-delimited list in a single `characteristics`
column.

The characteristic columns which are now required (with a  case-insensitive `yes|no` value)
as part of the CSV passed to the `SeedApprovedPremisesJob` are:

- `isIAP`
- `isPIPE`
- `isESAP`
- `isSemiSpecialistMentalHealth`
- `isRecoveryFocussed`
- `isSuitableForVulnerable`
- `acceptsSexOffenders`
- `acceptsChildSexOffenders`
- `acceptsNonSexualChildOffenders`
- `acceptsHateCrimeOffenders`
- `isCatered`
- `hasWideStepFreeAccess`
- `hasWideAccessToCommunalAreas`
- `hasStepFreeAccessToCommunalAreas`
- `hasWheelChairAccessibleBathrooms`
- `hasLift`
- `hasTactileFlooring`
- `hasBrailleSignage`
- `hasHearingLoop`

We now verify that each given characteristic is scoped to both:

- the service "approved-premises", and
- the model "premises"

from the `processRow()` function rather than from `deserializeRow()`.
This is because building the list of "positive" characteristics (those with
a YES value) is a little different and now uses a `CharacteristicValue`
data class to map the `propertyName`s and `value`s.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/ApprovedPremisesSeedJob.kt
```diff
@@ -36,6 +36,25 @@ class ApprovedPremisesSeedJob(
     probationRegion = columns["probationRegion"]!!,
     localAuthorityArea = columns["localAuthorityArea"]!!,
     characteristics = columns["characteristics"]!!.split(",").filter { it.isNotBlank() }.map { it.trim() },
+    isIAP = parseBooleanStringOrThrow(columns["isIAP"]!!, "isIAP"),
```
brownbearsoftware, line 39: Could we not type these as Boolean in ApprovedPremisesSeedCsvRow and return a Boolean from `parseBooleanStringOrThrow` ?

brownbearsoftware (approved):
LGTM, few minor comments.

The only thing that's nagging me a bit is that without running the seed job for the Characteristic types we now have code that assumes certain characteristic types exist when they might not.  Not a huge deal but could be a bit of a headscratcher down the line for someone.

src/main/resources/db/migration/all/20230310000608__update_approved_premises_characteristics.sql
```diff
@@ -0,0 +1,44 @@
...
+INSERT INTO characteristics (id, property_name, name, service_scope, model_scope) VALUES ('151e21e2-cad0-4338-84ee-404edbacf07c', 'isSemiSpecialistMentalHealth', 'Is this AP semi specialist Mental Health?', 'approved-premises', 'premises') ON CONFLICT (property_name, service_scope, model_scope) DO NOTHING;
+INSERT INTO characteristics (id, property_name, name, service_scope, model_scope) VALUES ('c238722f-624c-4d91-8e62-3559374a259a', 'isRecoveryFocussed', 'Is this a Recovery Focussed AP?', 'approved-premises', 'premises') ON CONFLICT (property_name, service_scope, model_scope) DO NOTHING;
+INSERT INTO characteristics (id, property_name, name, service_scope, model_scope) VALUES ('0dd72992-0e04-4d69-8858-7471cbcb7c8e', 'isSuitableForVulnerable', 'Is this AP suitable for people at risk of criminal exploitation? N.B Enhanced Security sites answer No, other APs answer Yes.', 'approved-premises', 'premises') ON CONFLICT (property_name, service_scope, model_scope) DO NOTHING;
```
pezholio, line 9: I think we need to model these characteristics as statements, rather than questions. My current thinking is that we'll need to pull these characteristics in at the matching information stage of an assessment (see https://miro.com/app/board/uXjVO1BpVA8=/?moveToWidget=3458764547717803302&cot=14), so a user will need to see, for example `Suitable for people at risk of criminal exploitation`. We'll also need a way to differentiate between practical placement requirements (e.g adaptations, disability etc) and offence and risk information (e.g suitable for sex offenders, arson rooms etc).

edavey: thanks Stu:

1. Changing the description in `Characteristic.name` can be done as we need via the `CharactertisticsSeedJob` -- this migration here is to ensure the presence of the mandatory set of `Charactertic.propertyName`s which are permanent keys to the required characteristics and which are needed by the `ApprovedPremisesSeedJob`
2. Possibly we can handle this grouping of characteristics with a `Characteristic.category` attribute which is either `need` or `risk`?

PR #478, opened 2023-03-09
Postgres search spike

brownbearsoftware:
This adds a POST `/beds/search` endpoint that allows searching for beds that are free for a given period matching given Premises and Room characteristics.

- PostGIS will need to be installed on the dev/test/preprod/prod Postgres databases before this can be merged
- Tearing down the database then remigrating causes issues with PostGIS with pooled connections so I've had to disable connection pooling for integration tests, this sems to have made the tests slower.  There's potentially room for improvement here by pooling for the duration of each test and clearing the pool at the end
- This assumes the work to use Rooms/Beds is done (there's currently a placeholder commit for this) for AP - since the models aren't actually split I didn't need to do this as a prerequisite but to actually use this outside of tests the changes would need to be made
- I opted for NamedParameterJdbcTemplate with a handwritten query over trying to achieve this with JPA Specifications.  Given the complexity of the search, if this is possible with the Specification API (unsure because of PostGIS etc.) I can't see it being as understandable as the raw SQL query.
- The majority of the line diff comes from a postcode distrcits SQL seed file - this is also potentially slowing down the integration tests, might make sense to not run this migration for the integration tests since they don't rely on it anyway.
- The work to make AP Lost Beds point at Room/Bed ids will need to be broken up into a few releases for a no-errors deployment.  This branch in general probably needs some careful breaking up and sequencing but the majority of the work here should be usable.


brownbearsoftware:
[Spike: investigate matching strategies](https://trello.com/c/M215fjGC/1302-spike-investigate-matching-strategies)

PR #479, opened 2023-03-09
(1315) Allow a task to be allocated to a user

pezholio:
This updates the reallocate endpoint to support multiple tasks. I’ve updated the body to accept a NewRellocation object with a Task Type and user ID. Rather than returning the Assessment object (which we don’t need), the endpoint now returns a Reallocation object with the Task Type and the user who the task has been reallocated to.

As we don’t currently support any task type other than assessements, we return a 405 method not allowed response. I’ve also added some tests for all the other types to serve as placeholders, which can be replaced as we add support for the other task types.


pezholio:
[Allow a Task to be allocated to a user](https://trello.com/c/tqUoLdLV/1315-allow-a-task-to-be-allocated-to-a-user)

brownbearsoftware (approved):
LGTM.

Only bikeshed would be: the response for unsupported task types feels more like a [501 Not Implemented](https://httpwg.org/specs/rfc9110.html#status.501) to me.

pezholio:
That was my initial instinct, but then I cam across this Stack Overflow question: https://stackoverflow.com/questions/32889983/is-http-501-appropriate-for-an-unimplemented-api

TLDR: 501 is when the server doesn't recognise the request, and 405 is when the server recognises the request but can't carry it out

PR #482, opened 2023-03-10
(1318) Create and allocate a placement request on assessment approval

pezholio:
This PR adds the concept of placement requests. These objects are created when an assessment has been accepted, and allocated to a Matcher. The matching "algorithm" (such as it is) is pretty naive, but enough to be going on with.

As it stands, the criteria that a placement request has does not directly map to room characteristics, but we've judged that this is OK for now, as we won't (yet) allow searching by the characteristics, and leave it up to the matcher to use their own judgement to find the most applicable place.


brownbearsoftware:
> As it stands, the criteria that a placement request has does not directly map to room characteristics, but we've judged that this is OK for now, as we won't (yet) allow searching by the characteristics, and leave it up to the matcher to use their own judgement to find the most applicable place.

Are we sure on this?  Adam's doc from the call this morning has "Preloading of Application" section which would suggest prepopulating a search is MVP?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/UserEntity.kt
```diff
@@ -33,6 +33,20 @@ interface UserRepository : JpaRepository<UserEntity, UUID>, JpaSpecificationExec
...
+    FROM "users"  u
+	    LEFT JOIN user_role_assignments ura ON ura.user_id = u.id 
+    WHERE ura.role = 'MATCHER'
+    ORDER BY 
+      (SELECT COUNT(1) FROM placement_requests a WHERE a.allocated_to_user_id = u.id AND a.booking_id IS NULL) ASC 
+    LIMIT 1
+    """,
+    nativeQuery = true
+  )
+  fun findMatcherWithLeastPendingPlacementRequests(): UserEntity?
```
edavey, line 49: Would random allocation be a fairer strategy, in that it wouldn't over-allocate to those matchers who are doing their work quickest?

edavey (approved):
Really clear 👍

edavey (approved):
looking good!

PR #483, opened 2023-03-13
Set order of characteristics for a 'deterministic' assertion

edavey:
This assertion was failing inconsistently due to changes in the ordering of characteristics. By sorting the lists prior to comparison we make the assertion "deterministic" or reliably repeatable.


pezholio (approved):
LGTM 👍

PR #486, opened 2023-03-14
Identify existing APs by ApCode rather than UUID in seeding

edavey:
ApCodes are unique so we can use these to identify existing APs which will be updated as distinct from new APs which need to be created.

No longer having to manage a list of UUIDs reduces opportunities for errors.


pezholio (approved):
This is much tidier 👍

PR #493, opened 2023-03-15
Resolve N+1 queries problem for GET /premises endpoint

danlivings-dxw:
> See [ticket #863 on the CAS3 Trello board](https://trello.com/c/7DZThdtS/863-investigate-api-performance-concerns).

This PR serves as a proof of concept for a solution to the performance issues we've seen in various parts of the API caused by the N+1 queries problem. It focuses on alleviating this problem for the `GET /premises` API endpoint.

Changes in this PR:
- The `PremisesRepository` has new query methods to load dependent entities named along the lines of `loadPremises*`, which accepts a `List<PremisesEntity>` and returns the same list with the dependent entities attached to each premises (e.g. `loadPremisesProbationRegions`, which attaches the `premises.probationRegion` property).
  A few other repositories have also introduced these methods where required to enable `GET /premises` endpoint.
- Properties on entity classes with a `@*ToOne` annotation that can be loaded using one of the above methods have been updated to use lazy fetching rather than eager fetching (which will perform one query per entity with the relevant annotation).
- Properties on entity classes that can be loaded using one of the above methods now have a corresponding `*Loaded` boolean property that defaults to `false`. This enables methods like `PremisesService.getAvailabilityForRange` to determine whether relevant dependent entities have already been loaded and perform a separate dedicated query to fetch them if they have not been, avoiding either an empty collection or a `LazyInitializationException` being thrown.
- The `PremisesService.loadDependentEntities` method has been introduced as a convenience, which accepts a `List<PremisesEntity>` and returns the same list. It accepts several optional boolean parameters which can be used to configure which dependent entities to load onto the given list. If a dependent entity is configured to be loaded, it will execute the relevant `load` repository query method and set the relevant `*Loaded` property to `true`.
- A new `log-sql` Spring profile has been introduced to enable testing that the N+1 problem has been resolved. It is used by `NPlus1QueriesTest`, which hooks into the logger for `org.hibernate.SQL` to count the number of SQL queries dispatched by Hibernate while serving a request to `GET /premises`.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PremisesEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PremisesEntity.kt
@@ -2,6 +2,7 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity
 
 import org.springframework.data.jpa.repository.JpaRepository
 import org.springframework.data.jpa.repository.Query
+import org.springframework.data.jpa.repository.QueryHints
 import org.springframework.stereotype.Repository
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.PropertyStatus
 import java.util.UUID
@@ -10,6 +11,7 @@ import javax.persistence.DiscriminatorValue
 import javax.persistence.Entity
 import javax.persistence.EnumType
 import javax.persistence.Enumerated
+import javax.persistence.FetchType
 import javax.persistence.Id
 import javax.persistence.Inheritance
 import javax.persistence.InheritanceType
@@ -19,7 +21,9 @@ import javax.persistence.ManyToMany
 import javax.persistence.ManyToOne
 import javax.persistence.OneToMany
 import javax.persistence.PrimaryKeyJoinColumn
+import javax.persistence.QueryHint
 import javax.persistence.Table
+import javax.persistence.Transient
 
 @Repository
 interface PremisesRepository : JpaRepository<PremisesEntity, UUID> {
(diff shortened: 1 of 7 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PremisesEntity.kt
```diff
@@ -31,6 +35,30 @@ interface PremisesRepository : JpaRepository<PremisesEntity, UUID> {
   @Query("SELECT p FROM PremisesEntity p WHERE p.probationRegion.id = :probationRegionId AND TYPE(p) = :type")
   fun <T : PremisesEntity> findAllByProbationRegion_IdAndType(probationRegionId: UUID, type: Class<T>): List<PremisesEntity>
 
+  @Query("SELECT DISTINCT p FROM PremisesEntity p LEFT JOIN FETCH p.probationRegion WHERE p IN (:premises)")
```
tahb, line 38: I would expect the ORM to abstract all of this complexity. As we change our associations this is going to be hard to maintain.

tahb (commented):
Clever stuff! I'm worried about the complexity this introduces. Did you experiment with configuring Hibernate to eager load in the end?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PremisesEntity.kt
pezholio: Yeah, that was my first instinct too. It's very neat and clever, but I'm surprised there's not an in-built way to do this stuff

danlivings-dxw: The builtin way to do this I think would be [`EntityGraph`s](https://www.baeldung.com/jpa-entity-graph), but to my mind doing it that way ends up being of similar complexity, and may not avoid the [Cartesian product problem](https://allaroundjava.com/hibernate-cartesian-product-problem/) where we have multiple to-many relations on the same entity.

brownbearsoftware: Spent some time looking at this today too and agree with @danlivings-dxw that this is probably the most straight forward/efficient way of getting a PremisesEntity loaded into memory with its Rooms fully loaded which in turn have their Beds fully loaded.

However, I wonder if the real issue is the amount of data we're returning from the GET /premises endpoint?  For AP the /premises route on the UI shows a table with only:
- Name
- Code
- Number of Beds

Perhaps the API's `/premises` endpoint should just return some core information about each Premises with the full detail being available at the `/premises/{id}` endpoint?

As an example here, one big offender is the `bedCount` property on the Premises response - not returning this for TA dropped the number of queries issued from 4608 to 1003 for me when experimenting with Dan's Integration Test against main.  Only returning core results would be easier to optimise as it could be done with a purpose built JPQL query which would have the benefit of not leaking this complexity out into the rest of the codebase.

tahb: Sadly I don't think CAS3 can easily drop the need for `bedCount` on the premises listing page. 

The idea of simplifying the premises response is interesting. The data CAS3 needs is: `address`, `pdu`, `bed count` and `status`. Subject to change of course.

My SQL is rusty and I can't quickly test it out but is there a way we could get just these fields with something like this?

```
SELECT premises.address_line_1, premises.pdu, premises.status, COUNT(beds.bed_id) AS bed_count
FROM premises
JOIN rooms ON premises.id = rooms.premises_id
JOIN beds ON rooms.room_id = beds.room_id
GROUP BY premises.id;
```

And use a query param to indicate to the `/premises` endpoint that we want a subset of fields in the response? Such as `/premises?summary=true` ?

brownbearsoftware: There's a bit of complication because the `pdu` property is on `temporary_accomodation_premises` rather than `premises` but it's doable I think.  This query seems to work locally for me:

```
SELECT
    p.id,
    p.name,
    p.address_line1,
    p.address_line2,
    p.postcode,
    tap.pdu,
    p.status,
    (SELECT count(1) FROM beds b WHERE b.room_id IN (SELECT id FROM rooms r WHERE r.premises_id = p.id)) AS bed_count
FROM premises p
LEFT JOIN temporary_accommodation_premises tap on p.id = tap.premises_id
WHERE p.service = 'temporary-accommodation' -- OR approved-premises ofc
;
```

> And use a query param to indicate to the /premises endpoint that we want a subset of fields in the response? Such as /premises?summary=true ? That feels restful and gives us a way to release it as a non-breaking change.

I think we'd struggle to get this to work with OpenAPI Generate because we'd need 2 layers of heirarachy (full vs core then ta & ap off each) which hasn't worked for me when I've tried a similar setup previously.  Perhaps a new endpoint `/premises/summary`, then move the UI over to that, then replace the original route, move the UI again, then delete the summary one?

edavey: we could consider moving PDU onto the main `premises` table if that would simplify? PDU is a sensible attribute to CAS1 to implement also...

tahb: > Perhaps a new endpoint /premises/summary, then move the UI over to that, then replace the original route, move the UI again, then delete the summary one?

Sounds good to me 👍 

@danlivings-dxw what are your thoughts on this option of a new endpoint with less querying?

tahb: Spoke to Dan. We're going to try and explore this new endpoint and see how far we get!

PR #498, opened 2023-03-16
(1320) Add an API endpoint to list all placement requests for a logged in user

pezholio:
This adds an API endpoint to list all placement requests that are allocated to a logged in user, in a similar way to how we handle assessments. I've also added a couple of fields to the placement requests, as we'll need the person detail and tier when we list on the frontend.


pezholio:
[Add an API endpoint to list all Placement Requests for a logged in user](https://trello.com/c/dQd6Waq0/1320-add-an-api-endpoint-to-list-all-placement-requests-for-a-logged-in-user)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/PlacementRequestService.kt
```diff
@@ -49,6 +49,7 @@ class PlacementRequestService(
           mentalHealthSupport = requirements.mentalHealthSupport,
           createdAt = OffsetDateTime.now(),
           application = assessment.application,
+          application = assessment.application as ApprovedPremisesApplicationEntity,
```
brownbearsoftware, line 56: I'd perhaps pull this out into a variable and throw an exception with a useful message for non-AP applications as seems likely someone in TA will bump into this in the future, e.g.

```
val application = (assessment.application as? ApprovedPremisesApplicationEntity) ?: throw RuntimeException("Only Approved Premises Assessments are currently supported for Placement Requests")
```

brownbearsoftware (approved):
LGTM

PR #500, opened 2023-03-17
Update to Spring 3

brownbearsoftware:
Seems there's a script that does a lot of this too, think there are some bits missing: 
https://github.com/ministryofjustice/hmpps-tech-docs/blob/main/scripts/upgrade-spring-boot-3.bash

Need to fix the JMS listener in the tests to progress this.


pezholio:
@brownbearsoftware Can this be closed now?

PR #502, opened 2023-03-17
Change AP Bookings & Lost Beds to be based on Bed IDs

brownbearsoftware:
This removes the AP/TA model and table split for Lost Beds so that they can both be based on a bed reference rather than AP using total beds count.


brownbearsoftware:
[Change Manage to use rooms and beds - Back end](https://trello.com/c/xJ3aFbGV/1323-change-manage-to-use-rooms-and-beds-back-end)

danlivings-dxw (approved):
This might cause some rework on my end for #509 but I'm happy for this to go in.

PR #509, opened 2023-03-20
Add checks for conflicting lost beds when creating or updating a lost bed or booking in Temporary Accommodation

danlivings-dxw:
> See [ticket #906 on the CAS3 Trello board](https://trello.com/c/PhHRtAyn/906-a-user-cannot-create-a-booking-or-another-void-if-an-existing-void-is-occupying-any-of-those-dates).

This PR extends the existing validation for Temporary Accommodation bookings that prevent double booking a room to lost beds. This allows for the following situations to be prevented:
- Double booking a lost bed
- Creating a lost bed to take place while a bed is occupied
- Creating a booking on a bed to take place while the bed is out of commission
- Changing the dates of a lost bed such that it overlaps with another lost bed or a booking
- Changing the dates of a booking (either through an arrival or an extension) such that it overlaps with a lost bed


brownbearsoftware (approved):
LGTM

PR #511, opened 2023-03-21
1288: seed AP rooms and beds

edavey:
This PR adds an `ApprovedPremisesRoomsSeedJob` which seeds rooms, their characteristics, and beds via the seeding script run on remote pod (or local environment) as:

```sh
./script/run_seed_job approved_premises_rooms rooms_with_characteristics
```

The CSV provided must have the following columns:

- `apCode`
- `bedCode`
- `roomNumber`
- `bedCount`
- `isSingle`
- `isGroundFloor`
- `isFullyFm`
- `hasCrib7Bedding`
- `hasSmokeDetector`
- `isTopFloorVulnerable`
- `isGroundFloorNrOffice`
- `hasNearbySprinkler`
- `isArsonSuitable`
- `isArsonDesignated`
- `hasArsonInsuranceConditions`
- `isSuitedForSexOffenders`
- `hasEnSuite`
- `isWheelchairAccessible`
- `hasWideDoor`
- `hasStepFreeAccess`
- `hasFixedMobilityAids`
- `hasTurningSpace`
- `hasCallForAssistance`
- `isWheelchairDesignated`
- `isStepFreeDesignated`
- `notes`


brownbearsoftware:
Feels a bit strange to me that Bed and Room names would need to be unique across the entire CAS1/CAS2/CAS3 combined estate?  e.g. Shouldn't `Bed 1` be a valid name for a bed in every Room of a given a Premises?  Similarly, shouldn't `Room 1` be a valid name for a Room in every AP?

brownbearsoftware:
@edavey seems sensible 👍 might need to double check which Postgres version we're on and how it handles NULLs wrt UNIQUE constraints - can we havve more than one entry in a table with a NULL value on a unique column?

edavey:
> might need to double check which Postgres version we're on and how it handles NULLs wrt UNIQUE constraints - can we havve more than one entry in a table with a NULL value on a unique column?

@brownbearsoftware we're on `PostgreSQL 15.2`. I believe the default behaviour is to treat `null`s as distinct, so we can have multiple `room` records with a null `code`. Postgres 15 introduces a `NULLS NOT DISTINCT` option, which would make that impossible -- not what we want here...

See https://pganalyze.com/blog/5mins-postgres-unique-constraint-null-parallel-distinct

brownbearsoftware (approved):

PR #512, opened 2023-03-21
1325 - Add point to Premises

brownbearsoftware:
Add nullable point to Approved Premises which will be used for Bed Search.

One this value has been populated in all environments another PR will make it not-nullable.


brownbearsoftware:
[Create a search endpoint for Match](https://trello.com/c/LtxpyraG/1325-create-a-search-endpoint-for-match)

edavey (approved):
thanks, this is very timely!

PR #516, opened 2023-03-21
Fixes for Bed Search OpenAPI spec

brownbearsoftware:
- Camel case property names instead of snake case
- Use service name enum
- Use wrapper type as repsonse body rather than array of results


bryangaledxw (approved):
LGTM - one last thing I want to check, is `probationDeliveryUnit` intended to be required on `TemporaryAccommodationBedSearchParameters`? The TypeScript that gets generated doesn't make it obvious if that's the intent

brownbearsoftware:
[Create a search endpoint for Match](https://trello.com/c/LtxpyraG/1325-create-a-search-endpoint-for-match)

PR #519, opened 2023-03-22
(1367) Add an API endpoint to get a task

pezholio:
This adds an endpoint to get a task of a particular type for an Application. Currently only Assessments or Placement Requests are supported

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/util/PlacementRequestUtils.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/util/PlacementRequestUtils.kt
@@ -1,0 +1,35 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.util
+
+import org.slf4j.Logger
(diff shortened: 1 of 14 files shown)


pezholio:
[Add an API endpoint to get a task for a user](https://trello.com/c/AMmePWli/1367-add-an-api-endpoint-to-get-a-task-for-a-user)

richpjames (dismissed):

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/util/PlacementRequestUtils.kt
brownbearsoftware, line 1: Unless I'm missing something, should be able to replace all of the `Any` with a type here to avoid casting at the call site?

pezholio: Oh nice!

brownbearsoftware (approved):

PR #525, opened 2023-03-23
Allow Temporary Accommodation user to edit cancelled or closed bookings

danlivings-dxw:
> See [ticket #801 on the CAS3 Trello board](https://trello.com/c/TJfk6Ccd/801-make-a-cancelled-or-closed-booking-editable-so-users-can-add-more-information-after-the-action-has-happened-in-the-system).

This PR implements the ability for users of the Temporary Accommodation service to edit departures or cancellations on bookings if new information becomes available or to correct an error.

Changes in this PR:
- The `@OneToOne` requirement on the `BookingEntity` has been relaxed to `@OneToMany` for departures and cancellations. This allows for the history of changes to be preserved, rather than updating departures/cancellations in place.
- `BookingEntity.departure` and `BookingEntity.cancellation` have been converted into getter properties which return the most recently created departure or booking respectively (if it exists). This allows existing code that reads from these fields to continue to function as expected without changes at those locations.
- The OpenAPI specification for the `Booking` schema has been updated to also provide `departures` and `cancellations` properties containing the full change history. The original `departure` and `cancellation` properties are kept for backwards compatibility, as well as to provide a useful shortcut to the most recent version of the departure/cancellation for situations in which the full history is not needed.
- The validation enforcing that a request to the `POST /premises/{premisesId}/bookings/{bookingId}/departures` or `POST /premises/{premisesId}/bookings/{bookingId}/cancellations` endpoints can only be called once on any given booking has been relaxed to only apply to Approved Premises.


brownbearsoftware (approved):
LGTM - Nicely done wrt no breaking changes

PR #530, opened 2023-03-28
Use DbExtension to manage database for tests to improve performance

danlivings-dxw:
This PR aims to reduce the time taken to run unit and integration tests.

It introduces a JPA extension (`DbExtension`) that performs the following optimisations:
- Run Flyway migrations once at the start of the test run
- Store data seeded into the database at the start of the test run
- Truncate all tables and reinsert seeded data before each test to ensure the database is in a known state

It also improves the performance of PostGIS on ARM64 machines (such as newer Macs) by providing Docker Compose configurations using a community PostGIS image that run natively, rather than emulating AMD64, bringing the time taken to run the tests in line with the CI pipeline. The `test_database` and `development_database` scripts will automatically use these configurations on an ARM64 environment.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/util/DbExtension.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/util/DbExtension.kt
@@ -1,0 +1,160 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.util
+
+import org.flywaydb.core.Flyway
+import org.hibernate.internal.SessionFactoryImpl
+import org.junit.jupiter.api.extension.BeforeAllCallback
+import org.junit.jupiter.api.extension.BeforeEachCallback
+import org.junit.jupiter.api.extension.ExtensionContext
+import org.postgresql.ds.PGSimpleDataSource
+import org.slf4j.LoggerFactory
+import org.springframework.boot.autoconfigure.jdbc.DataSourceProperties
+import org.springframework.boot.context.properties.bind.Bindable
+import org.springframework.boot.context.properties.bind.Binder
+import org.springframework.context.ApplicationContext
+import org.springframework.jdbc.core.JdbcTemplate
+import org.springframework.test.context.junit.jupiter.SpringExtension
+import java.sql.Connection
+import java.sql.SQLException
+import java.util.concurrent.atomic.AtomicBoolean
+import javax.persistence.EntityManagerFactory
+import javax.persistence.metamodel.EntityType
+import javax.sql.DataSource
+
+class DbExtension : BeforeAllCallback, BeforeEachCallback {
+  companion object {
+    private val STARTED = AtomicBoolean(false)
+    private val LOGGER = LoggerFactory.getLogger(DbExtension::class.java)
+
+    private val TABLES_TO_IGNORE = listOf(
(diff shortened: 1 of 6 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/util/DbExtension.kt
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  private fun loadTablesToClean(connection: Connection): List<TableData> {
+    val databaseMetaData = connection.metaData
+    val resultSet = databaseMetaData.getTables(connection.catalog, null, null, arrayOf("TABLE"))
+
+    val tablesToClean = mutableListOf<TableData>()
+    while (resultSet.next()) {
+      val table = TableData(
+        schema = resultSet.getString("TABLE_SCHEM"),
```
adamhughes86, line 134: @danlivings-dxw - Is `TABLE_SCHEM` correct?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/util/DbExtension.kt
```diff
@@ -0,0 +1,160 @@
...
+    val sessionFactory = entityManagerFactory.unwrap(SessionFactoryImpl::class.java)
+
+    sessionFactory.openSession().use { session ->
+      INITIAL_DB_STATE = session.metamodel.entities.associateWith {
+        val query = session.criteriaBuilder.createQuery()
+        val root = query.from(it)
+        query.select(root)
+        val results = session.createQuery(query).resultList
+
+        println("Found ${results.count()} ${it.name} instances")
```
brownbearsoftware, line 74: leftover debug log?

brownbearsoftware (approved):
Nice 👍

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/util/DbExtension.kt
danlivings-dxw: The documentation for [`DatabaseMetaData.getTables`](https://docs.oracle.com/javase/7/docs/api/java/sql/DatabaseMetaData.html#getTables(java.lang.String,%20java.lang.String,%20java.lang.String,%20java.lang.String[])) specifies `"TABLE_SCHEM"` to retrieve the schema name.

danlivings-dxw:
@brownbearsoftware @tahb @pezholio I've managed to get some extra time shaved off on my machine by using a community PostGIS image, which I've added to this PR as an extra commit. There are some potential security implications of using an unofficial image, but I believe these would only be used locally. Is this an acceptable risk?

pezholio:
@danlivings-dxw That seems like a fair enough trade-off given the fact that it’s only local 👍🏻

tahb:
Happy to use the community postgis image locally for faster feedback however for high parity I think we should continue to have CI test against the image we deploy with. Given that is running at 15 minutes that seems relatively acceptable. 

I think we can also dry up these docker-compose files so that we that changes to shared resources can be made in one place. I think we can create a docker-compose.arm64.yml file with just ammendments and then use docker-compose to handle the merging with `docker compose -f docker-compose.test.yml -f docker-compose.test.arm64.yml up`

PR #532, opened 2023-03-29
1328 - Create Booking from Placement Request

brownbearsoftware:


brownbearsoftware:
[Add an endpoint to create a booking from a placement request](https://trello.com/c/SjH2GSPI/1328-add-an-endpoint-to-create-a-booking-from-a-placement-request)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/results/AuthorisableActionResult.kt
```diff
@@ -3,5 +3,5 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.results
 sealed interface AuthorisableActionResult<EntityType> {
   data class Success<EntityType>(val entity: EntityType) : AuthorisableActionResult<EntityType>
   class Unauthorised<EntityType> : AuthorisableActionResult<EntityType>
-  class NotFound<EntityType> : AuthorisableActionResult<EntityType>
+  class NotFound<EntityType>(val entityType: String? = null, val id: String? = null) : AuthorisableActionResult<EntityType>
```
pezholio, line 6: 🙌 I wanted to do this for something else, but I lacked the technical nous!

src/main/resources/static/api.yml
```diff
@@ -2099,6 +2099,39 @@ paths:
           $ref: '#/components/responses/403Response'
         500:
           $ref: '#/components/responses/500Response'
+  /placement-requests/{id}/booking:
```
pezholio, line 2102: Thinking about this - I wonder if it makes more sense to relate the booking to the Application, as this is the thing we _really_ care about. The PlacementRequest is more a record to aid with matching

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
```diff
@@ -82,12 +82,11 @@ class BookingService(
   private val cancellationReasonRepository: CancellationReasonRepository,
   @Value("\${application-url-template}") private val applicationUrlTemplate: String
 ) {
-  fun createBooking(bookingEntity: BookingEntity): BookingEntity = bookingRepository.save(bookingEntity)
   fun updateBooking(bookingEntity: BookingEntity): BookingEntity = bookingRepository.save(bookingEntity)
   fun getBooking(id: UUID) = bookingRepository.findByIdOrNull(id)
 
   @Transactional
-  fun createApprovedPremisesBooking(
+  fun createApprovedPremisesAdHocBooking(
```
pezholio, line 166: Good naming - we might want to bin this off at some point I guess

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
```diff
@@ -82,11 +83,85 @@ class BookingService(
...
   private val cancellationReasonRepository: CancellationReasonRepository,
   private val bedRepository: BedRepository,
+  private val placementRequestRepository: PlacementRequestRepository,
   @Value("\${application-url-template}") private val applicationUrlTemplate: String
 ) {
   fun updateBooking(bookingEntity: BookingEntity): BookingEntity = bookingRepository.save(bookingEntity)
   fun getBooking(id: UUID) = bookingRepository.findByIdOrNull(id)
 
+  @Transactional
+  fun createApprovedPremisesBookingFromPlacementRequest(
```
pezholio, line 93: Again, sorry for the misdirection here, but I reckon the `createApprovedPremisesBookingFromApplication` is the thing to do here

pezholio (commented):
All in all, this looks great, but (sorry!) I think an Application might be the best first-class relation to use

src/main/resources/static/api.yml
pezholio: On reflection (and chatting to @brownbearsoftware) I realised that Applications can have multiple PlacementRequests, so linking through the request is the thing to do here

PR #534, opened 2023-03-30
1328 - Prevent double booking of AP beds

brownbearsoftware:
This prevents a Booking being made on an AP Bed if there is a conflicting Booking or Lost Bed for the date range.

This also moves the detecting conflicts for Bookings and Lost Beds to a JPQL query to avoid performance issues further down the road with iterating over all of the Bookings for a Premises


brownbearsoftware:
[Add an endpoint to create a booking from a placement request](https://trello.com/c/SjH2GSPI/1328-add-an-endpoint-to-create-a-booking-from-a-placement-request)

pezholio (approved):
LGTM 👍

PR #538, opened 2023-04-04
Use Instant for API response models instead of OffsetDateTime

brownbearsoftware:
The default behaviour of the kotlin-spring OpenAPI Generator is to type OpenAPI `date-time` properties as `OffsetDateTime` in the Kotlin models.

Two `OffsetDateTime`s are only considered equal if they have the same DateTime and Offset components.  This can cause issues for integration testing when round-tripping date-times via the database (postgres does not store the offset so always returns in the database timezone which is usually UTC.)  We then end up comparing a +01:00 `OffsetDateTime` with a +00:00 `OffsetDateTime` which represent the same point in time but are not considered equal.

This PR changes OpenAPI Generator config to output `java.time.Instant` instead which does not store Offset information - this is fine because it's assumed to be UTC which is what happens when you roundtrip via Postgres anyway.

Instant more accurately represents what we're trying to model (a point in time, rather than a specific way of describing that point in time) and what we want to compare.

The serialized version of the Instants is still ISO-8601 with an offset so there are no compatibility issues for the frontend.


brownbearsoftware:
[Add an endpoint to record the fact that a match cannot be found](https://trello.com/c/R42iyWyU/1331-add-an-endpoint-to-record-the-fact-that-a-match-cannot-be-found)

build.gradle.kts
```diff
@@ -104,7 +104,10 @@ openApiGenerate {
     put("gradleBuildFile", "false")
     put("exceptionHandler", "false")
     put("useBeanValidation", "false")
+    put("dateLibrary", "custom")
   }
+  typeMappings.put("DateTime", "Instant")
+  importMappings.put("Instant", "java.time.Instant")
```
pezholio, line 110: Oh, this is neat!

pezholio (approved):

PR #539, opened 2023-04-04
1331 - Record Booking Not Made

brownbearsoftware:
Endpoint to record a Booking Not Made against a Placement Request.

Also emits an `approved-premises.booking.not-made` Domain Event


brownbearsoftware:
[Add an endpoint to record the fact that a match cannot be found](https://trello.com/c/R42iyWyU/1331-add-an-endpoint-to-record-the-fact-that-a-match-cannot-be-found)

pezholio (approved):
LGTM - this commit history seems to be a bit wonky though, could use a manual rebase/force push

PR #541, opened 2023-04-05
Implement 'booking search' endpoint

danlivings-dxw:
> See [ticket #971 on the CAS3 Trello board](https://trello.com/c/QeqfZz0s/971-new-api-endpoint-to-get-all-bookings-filtered-by-status).

This PR implements the `GET /bookings/search` endpoint introduced by the OpenAPI specification changes from #533. This enables the delivery of the "find a booking" feature in the Temporary Accommodation service.

Changes in this PR:
- The `X-Service-Name` header in the `GET /bookings/search` endpoint has been added as a required header. This is a formality from the frontend client's point of view as it's already provided for all API requests by default, but allows the API to filter bookings by service type, avoiding the possibility of reporting bookings for the wrong service.
- The `BookingSearchRepository` class has been introduced to perform the actual query against the database for bookings. It allows for optionally filtering the results by booking status and probation region, but is not responsible for sorting the results (see below).
- The `BookingSearchService` class has been introduced. The service determines the user's probation region if the calling service is Temporary Accommodation and queries the repository using the provided or derived filters. It then iterates over the results to retrieve the name of the person the booking is for by calling the Community API using their CRN. Finally, it sorts according to the specified sort field and order - which must happen at this stage rather than in the SQL query as the person's name is required to be one of the sort fields, and is not available any earlier than this.
- The `BookingSearchResultTransformer` class has been introduced to map the results of the `BookingSearchService` to the API model.
- The `BookingSearchController` class has been introduced to handle the API endpoint. It provides default values for the search order and field if they are not provided by the client before delegating to the `BookingSearchService`.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/BookingSearchController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/BookingSearchController.kt
@@ -1,0 +1,51 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.controller
+
+import org.springframework.http.ResponseEntity
+import org.springframework.stereotype.Service
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.BookingsApiDelegate
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.BookingSearchResults
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.BookingSearchSortField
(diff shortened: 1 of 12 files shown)


brownbearsoftware (approved):
LGTM

brownbearsoftware:
Might be worth adding indexes on status and probation region?

danlivings-dxw:
@brownbearsoftware I've requested a re-review as there's been some rework following a discussion in the CAS3 team about using the same statuses in the search as are used in the bookings themselves. I've also added the suggested indices to the tables in a Flyway migration.

PR #544, opened 2023-04-05
Add premises summary endpoint for CAS1 premises

tahb:
- New `/premises/summary` endpoint
- Currently we just return TA premises but more work is needed to extend this out further for AP premises
- We include the n+1 load test but exclude it from the default test suite

Screenshot 2023-04-05 at 18 06 45

Does this need a unit test for the service? When writing it, it felt identical to the integration test. I noted there wasn't a unit test for the getPremises method. The unit tests might be useful if we wanted to test more variances in behaviour such as empty results, getting AP premises, etc.


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/PremisesSummaryTest.kt
```diff
@@ -0,0 +1,57 @@
...
+
+      webTestClient.get()
+        .uri("/premises/summary")
+        .header("Authorization", "Bearer $jwt")
+        .header("X-Service-Name", ServiceName.temporaryAccommodation.value)
+        .exchange()
+        .expectStatus()
+        .isOk
+        .expectBody()
+        .jsonPath("$..[0]id").isEqualTo(uuid.toString())
```
tahb, line 48: There might be a nicer, less fragile way to do this assertion?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/NPlus1QueriesTest.kt
```diff
@@ -0,0 +1,78 @@
...
+  @BeforeEach
+  fun setup() {
+    val logger = LoggerFactory.getLogger("org.hibernate.SQL") as ch.qos.logback.classic.Logger
+    logger.addAppender(appender)
+
+    every { appender.doAppend(capture(capturedLogs)) } returns Unit
+  }
+
+  @Test
+  @Disabled("Load tests are not run by default")
```
tahb, line 35: Key line here that skips it in the normal test runs.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/PremisesService.kt
```diff
@@ -44,6 +55,23 @@ class PremisesService(
 
   fun getAllPremises(): List<PremisesEntity> = premisesRepository.findAll()
 
+  fun getAllPremisesSummary(serviceName: ServiceName): List<PremisesSummary> {
```
tahb, line 58: So far the api spec and controller set the expectation that this can be used for both services. However this is the point where we pause and focus just on returning temporary-accommodation premises. This is mostly so I can check what I'm doing will work before I keep going and invest in the AP side of things.

src/main/resources/static/api.yml
```diff
@@ -76,6 +76,40 @@ paths:
...
       x-codegen-request-body-name: body
+  /premises/summary:
+    get:
+      tags:
+        - Premises
+      summary: Returns a list of premises
+      parameters:
+        - name: X-Service-Name
+          in: header
+          required: false
```
danlivings-dxw, line 87: I think this should be `required: true` as both the frontend clients provide this header by default and it was made optional for backwards compatibility at the time.

src/main/resources/static/api.yml
```diff
@@ -76,6 +76,40 @@ paths:
...
+          description: If given, only premises for this service will be returned
+          schema:
+            $ref: '#/components/schemas/ServiceName'
+        - name: X-User-Region
+          in: header
+          required: false
+          description: If given, only premises within this region will be returned
+          schema:
+            type: string
+            format: uuid
```
danlivings-dxw, line 97: I don't think this header is needed as it's now possible to get the user's probation region without it.

src/main/resources/static/api.yml
```diff
@@ -2616,6 +2650,46 @@ components:
...
+          example: LS1 3AD
+        bedCount:
+          type: integer
+          example: 22
+        status:
+          type: string
+          enum:
+            - pending
+            - active
+            - archived
```
danlivings-dxw, line 2679: These lines will be ignored when using `$ref`.

danlivings-dxw (approved):

PR #547, opened 2023-04-05
Don't seed actual postcode districts for integration tests

brownbearsoftware:


pezholio:
Oh wow! This is a huge improvement! I wonder if this speaks to something like missing indexes on the Postcode Districts table though?

brownbearsoftware:
13 mins 46 sec on my machine

brownbearsoftware:
> I wonder if this speaks to something like missing indexes on the Postcode Districts table though?

There is a UNIQUE constraint over `outcode` on `postcode_districts` (which is the field we query against) which automatically creates an index.  So I think it's just the overhead of inserting then updating ~2800 rows.

PR #549, opened 2023-04-06
Fix for latitude/longitude being wrong way around for PostGIS point in AP Seed Job

brownbearsoftware:


pezholio (approved):
LGTM - I can run the seed job once this is merged

PR #551, opened 2023-04-07
Fix tasksGet endpoint

pezholio:
Before this was fetching assessments regardless of if they’d been submitted or not, as well as only fetching placement requests that had been allocated to the logged in user, which is not what we want.

This adds a `getAllReallocatable` service method to the Assessment service, which only fetches unsubmitted and unreallocated assessments and a `getAllReallocatable` service method to the PlacementRequest service, which only fetches unreallocated Placement Requests without an associated booking.


pezholio:
[Fix allocations index view](https://trello.com/c/42gDZQXN/1436-fix-allocations-index-view)

PR #555, opened 2023-04-11
1403 - Don't create a Placement Request when Application doesn't have a release date

brownbearsoftware:
- Adds a new release date first class field to Applications which the frontend will need to start sending when submitting an Application (or null if not known yet)
- Only create a Placement Request when an Assessment is approved if the Application to which it pertains has a release date


brownbearsoftware:
[Don’t create a placement request when approving an application without a release date](https://trello.com/c/h5zDccaN/1403-dont-create-a-placement-request-when-approving-an-application-without-a-release-date)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/AssessmentController.kt
```diff
@@ -155,9 +156,11 @@ class AssessmentController(
       is ValidatableActionResult.Success -> assessmentValidationResult.entity
     }
 
-    extractResultEntityOrThrow(
-      placementRequestService.createPlacementRequest(assessment, assessmentAcceptance.requirements)
-    )
+    if ((assessment.application as? ApprovedPremisesApplicationEntity)?.releaseDate != null) {
```
pezholio, line 159: I think we can use `assessmentAcceptance.requirements.expectedArrival` here, as this will be set to the release date. In fact, it may be worth us not sending any requirements if there is no arrival date.

pezholio (commented):
I think we can do this using the

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/AssessmentController.kt
brownbearsoftware: I think we need to persist this so we can use it to determine the status for the Application for 1405 too as we won't have the placement request data at that point? https://trello.com/c/pZz38nxr/1405-add-a-pending-status-to-an-application-assessment

PR #556, opened 2023-04-11
1403 - Don't create a Placement Request without requirements

brownbearsoftware:
When the frontend does not send requirements (which it will do when there is no known release date) do not create a Placement Request.


brownbearsoftware:
[Don’t create a placement request when approving an application without a release date](https://trello.com/c/h5zDccaN/1403-dont-create-a-placement-request-when-approving-an-application-without-a-release-date)

pezholio (approved):
This is much neater - thanks for bearing with me! 👍

PR #557, opened 2023-04-11
1405 - Add additional statuses to Applications

brownbearsoftware:
If an application has a submitted, rejected assessment its status should be rejected
If an application has a submitted, approved assessment but no placement requests it should be pending
If an application has a submitted, approved assessment with a placement request that does not have a booking it should be awaiting placement
If an application has a submitted, approved assessment with a placement request that does have a booking it should be placed


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/ApplicationsTransformer.kt
```diff
@@ -58,9 +59,25 @@ class ApplicationsTransformer(
...
-    entity.submittedAt !== null -> ApplicationStatus.submitted
-    else -> ApplicationStatus.inProgress
+  private fun getStatus(entity: ApplicationEntity): ApplicationStatus {
+    val latestAssessment = entity.getLatestAssessment()
+
+    if (entity is ApprovedPremisesApplicationEntity) {
+      return when {
+        latestAssessment?.submittedAt != null && latestAssessment.decision == AssessmentDecision.REJECTED -> ApplicationStatus.rejected
+        latestAssessment?.submittedAt != null && latestAssessment.decision == AssessmentDecision.ACCEPTED && entity.getLatestPlacementRequest() == null -> ApplicationStatus.pending
+        latestAssessment?.submittedAt != null && latestAssessment.decision == AssessmentDecision.ACCEPTED && entity.getLatestPlacementRequest()!!.booking == null -> ApplicationStatus.awaitingPlacement
```
pezholio, line 69: Could we have a `getBooking()` function on the Application? I imagine it'd be useful elswhere too.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/transformer/ApplicationsTransformerTest.kt
```diff
@@ -179,6 +183,98 @@ class ApplicationsTransformerTest {
     assertThat(result.status).isEqualTo(ApplicationStatus.submitted)
   }
 
+  @Test
+  fun `transformJpaToApi transforms an Approved Premises application with a rejected Assessment correctly`() {
```
pezholio, line 187: I feel like the names of these tests would be better with something along with lines of the following rather than just saying "transforms correctly":

```
transformJpaToApi sets an Approved Premises application's status to "rejected" if the associated assessment has a decision of "rejected"
```

pezholio (approved):

PR #569, opened 2023-04-13
Add/update departure reasons and move-on categories for Temporary Accommodation

danlivings-dxw:
> See [ticket #1004 on the CAS3 Trello board](https://trello.com/c/MTRCoSe1/1004-update-new-move-on-category-and-departure-reasons-when-closing-a-booking).

The CAS3 team has been provided with a updated list of departure reasons and move-on categories to be used within the Temporary Accommodation service.

In many cases these are changes in wording to existing entries that are currently shared across both Temporary Accommodation and Approved Premises, so in these cases the shared entries have been retired (setting `is_active` to `false`) and new editions have been created with the original wording for Approved Premises and the changed wording for Temporary Accommodation. Similarly, for entries that are no longer used by Temporary Accommodation, the shared entry has been retired and replaced with a new entry specific to Approved Premises.
Changes in wording to existing entries that are specific to Temporary Accommodation have been made in place.

This strategy should be backwards compatible in that existing departed bookings can reference the retired departure reasons/move-on categories, but new departures will be created using the new entries.


danlivings-dxw:
@brownbearsoftware I've requested a re-review as the Temporary Accommodation frontend doesn't filter inactive entries so API changes were needed to avoid presenting them to the users of the service.

brownbearsoftware (approved):

PR #571, opened 2023-04-14
Set selfDiagnosis level to DEBUG on AppInsights in dev to ascertain available JMX metrics

brownbearsoftware:
Ideally we will collect and ship to AppInsights:
- Max Heap
- Heap Committed
- Heap Used
- GCs

This property causes AppInsights to log available JMX metrics so we can progress this.

### DIFF
--- a/applicationinsights.nonprod.json
+++ b/applicationinsights.nonprod.json
@@ -11,7 +11,8 @@
     }
   },
   "selfDiagnostics": {
-    "destination": "console"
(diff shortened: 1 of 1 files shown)


richpjames (approved):

tahb:
https://trello.com/c/RBe2uS1z/1451-out-of-memory-loading-applications-for-some-users-in-dev

PR #572, opened 2023-04-14
Revert setting selfDiagnosis level to DEBUG on AppInsights in dev

brownbearsoftware:


richpjames (approved):

tahb:
@pezholio @richpjames Do we remember why this was reverted? Currently looking at options for getting more insight into performance.

richpjames:
@tahb I think Joe put it in to increase logging temporarily and then reverted once he had what he needed

PR #577, opened 2023-04-18
Implement support for either PDU name or ID when creating or updating a Temporary Accommodation premises

danlivings-dxw:
> See [ticket #1015 on the CAS3 Trello board](https://trello.com/c/5Jhi2OZF/1015-re-associate-existing-premises-to-the-new-pdu-records).

This PR is the second part of the work to migrate Temporary Accommodation premises away from using opaque strings for probation delivery units (PDUs). It follows on from #550, which introduced a reference data endpoint for PDUs to the API and provided the infrastructure to perform this migration.

This PR updates the `POST /premises` and `PUT /premises/{premisesId}` endpoints to allow an ID representing a PDU in the database to be provided.

To maintain backwards compatibility while transitioning to modelling premises PDUs in a structured way, when creating or updating a premises, the `PremisesService` now accepts an `Either<String, UUID>` that represents both the legacy PDU name as well as the ID of a PDU that exists in the database. The `PremisesService.tryGetProbationDeliveryUnit` method is used to perform the validation for the legacy `pdu` name field or the `probationDeliveryUnitId` UUID (as appropriate) and uses the relevant repository method to obtain a `ProbationDeliveryUnitEntity` instance.

From this point, the use of `probationDeliveryUnitId` is preferred, which is reflected in the use of `UUID` as the second type argument of `Either` (which is right-biased) and the validation error `"empty"` being reported on that property instead of `pdu` when neither has a value.

The PDU is still stored in the premises as a string name rather than as an ID to a foreign key table, but performing the work to resolve the name or ID now will reduce the work needed to migrate premises away from string PDU names.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -46,6 +46,8 @@ dependencies {
 
   implementation("org.jetbrains.kotlinx:dataframe:0.10.0")
 
+  implementation("io.arrow-kt:arrow-core:1.1.5")
+
   testImplementation("io.github.bluegroundltd:kfactory:1.0.0")
   testImplementation("io.mockk:mockk:1.13.5")
   testImplementation("io.jsonwebtoken:jjwt-api:0.11.5")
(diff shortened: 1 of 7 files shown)


brownbearsoftware (approved):

bryangaledxw:
https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/297 should be merged and deployed ahead of this

bryangaledxw:
Required forward-compat UI change is now deployed

PR #583, opened 2023-04-19
1451 - Applications Summary

brownbearsoftware:
The `GET /applications` endpoint in dev has been taking a long time to respond and often hitting out of memory exceptions.  This is because:
- the `data` and `document` fields on Applications are huge
- we don't have pagination on this endpoint
- we do upstream requests to Community API for offender info to construct the `person` part of the response - when these are a cache miss this takes a toll
- we do upstream requests to Community API for each user-offender pair to establish whether they're a limited access offender and if the user can view their records if so

To try and alleviate:
- changed `PUT /applications/{id}` to allow sending the first class properties for AP Applications
- made release date a first class property (this is the only piece of data needed for the table in the frontend that's only in `data` at the moment)
- removed `data` and `document` from the responses (to reduce response size)
- changed to a bespoke query to fetch these Application "Summaries" which avoids JPA loading the `data` and `document` fields into API memory

Improvements to response size:
image

Improvements to response times:
image

Both users have the `WORKFLOW_MANAGER` role meaning they can see all Applications.
- The first call from user 1 has to make upstream requests to Community API and Prison API (which are then cached in Redis) AND calls to Community API to check if the user is allowed to access each CRN (as the Offender may be an LAO offender)
- The second call from user 1 has cache hits for all upstream calls
- The first call from user 1 has cache hits for the Offender and Inmate detail upstream requests but cache misses for the LAO calls

Whilst this change will alleviate OOM problems, we're still not where we could be for response times.  Further optimisations might be:
- Preemptive caching for Offender and Inmate details requests, so cache misses for these calls will be rare
- Make LAO check calls concurrently
- or: A bulk endpoint from PI team for LAO checks, e.g. we post them a Delius username and an array of CRNs and they respond with whether that user is allowed to view each CRN
- Refactor such that we're looking at the `currentRestriction`, `currentExclusion` entries from the Community API Offender Details response.  If these are both `false` which will be the case for the majority of CRNs there is no need to make the User Access call for that CRN at all.  The difficulty here is that the Controller makes those calls after the user access call is made by the service layer.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ApplicationsController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ApplicationsController.kt
@@ -137,7 +137,7 @@ class ApplicationsController(
 
     val serializedData = objectMapper.writeValueAsString(body.data)
 
-    val applicationResult = when(body) {
+    val applicationResult = when (body) {
       is UpdateApprovedPremisesApplication -> applicationService.updateApprovedPremisesApplication(
         applicationId = applicationId,
         data = serializedData,
@@ -347,10 +347,10 @@ class ApplicationsController(
     return applicationsTransformer.transformJpaToApi(application, offender, inmate)
   }
 
-  private fun getPersonDetailAndTransformToSummary(application: ApplicationEntity): ApplicationSummary {
-    val (offender, inmate) = getPersonDetail(application.crn)
+  private fun getPersonDetailAndTransformToSummary(application: uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApplicationSummary): ApplicationSummary {
+    val (offender, inmate) = getPersonDetail(application.getCrn())
 
-    return applicationsTransformer.transformJpaToApiSummary(application, offender, inmate)
+    return applicationsTransformer.transformDomainToApiSummary(application, offender, inmate)
   }
 
   private fun getPersonDetailAndTransform(offlineApplication: OfflineApplicationEntity): Application {
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ApplicationsController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ApplicationsController.kt
@@ -18,6 +18,8 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Task
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.TaskType
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.TaskWrapper
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.UpdateApplication
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.UpdateApprovedPremisesApplication
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.UpdateTemporaryAccommodationApplication
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.User
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.convert.EnumConverterFactory
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApplicationEntity
@@ -50,6 +52,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.getPersonDetailsFor
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.kebabCaseToPascalCase
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.transformAssessment
 import java.net.URI
+import java.time.ZoneOffset
 import java.util.UUID
 import javax.transaction.Transactional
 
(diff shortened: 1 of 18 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApplicationEntity.kt
```diff
@@ -40,6 +41,78 @@ interface ApplicationRepository : JpaRepository<ApplicationEntity, UUID> {
...
+    (SELECT COUNT(1) FROM assessment_clarification_notes acn WHERE acn.assessment_id = ass.id AND acn.response IS NULL) > 0 as latestAssessmentHasClarificationNotesWithoutResponse,
+    (SELECT COUNT(1) FROM bookings b WHERE b.application_id = taa.id) > 0 as hasBooking
+FROM temporary_accommodation_applications taa 
+LEFT JOIN applications a ON a.id = taa.id 
+LEFT JOIN assessments ass ON ass.application_id = taa.id AND ass.reallocated_at IS NULL 
+WHERE a.created_by_user_id = :userId
+""",
+    nativeQuery = true
+  )
+  fun findAllTemporaryAccommodationSummariesCreatedByUser(userId: UUID): List<TemporaryAccommodationApplicationSummary>
```
pezholio, line 115: Lovely!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ApplicationsController.kt
```diff
@@ -347,10 +347,10 @@ class ApplicationsController(
     return applicationsTransformer.transformJpaToApi(application, offender, inmate)
   }
 
-  private fun getPersonDetailAndTransformToSummary(application: ApplicationEntity): ApplicationSummary {
-    val (offender, inmate) = getPersonDetail(application.crn)
+  private fun getPersonDetailAndTransformToSummary(application: uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApplicationSummary): ApplicationSummary {
```
pezholio, line 350: Do we need `uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApplicationSummary` here? Can we not import it with an alias?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/ApplicationsTransformer.kt
```diff
@@ -116,4 +125,24 @@ class ApplicationsTransformer(
...
+  private fun getStatusFromSummary(entity: DomainApplicationSummary): ApplicationStatus {
+    if (entity is DomainApprovedPremisesApplicationSummary) {
+      return when {
+        entity.getSubmittedAt() != null && entity.getLatestAssessmentDecision() == AssessmentDecision.REJECTED -> ApplicationStatus.rejected
+        entity.getSubmittedAt() != null && entity.getLatestAssessmentDecision() == AssessmentDecision.ACCEPTED && !entity.getHasPlacementRequest() -> ApplicationStatus.pending
+        entity.getSubmittedAt() != null && entity.getLatestAssessmentDecision() == AssessmentDecision.ACCEPTED && !entity.getHasBooking() -> ApplicationStatus.awaitingPlacement
+        entity.getSubmittedAt() != null && entity.getLatestAssessmentDecision() == AssessmentDecision.ACCEPTED && entity.getHasBooking() -> ApplicationStatus.placed
+        entity.getLatestAssessmentHasClarificationNotesWithoutResponse() -> ApplicationStatus.requestedFurtherInformation
+        entity.getSubmittedAt() !== null -> ApplicationStatus.submitted
+        else -> ApplicationStatus.inProgress
```
pezholio, line 138: Probs not for now, but I'm wondering if all this stuff here has a perfomance impact? Could we manage statuses via a state machine or similar and record them directly in the database?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ApplicationsController.kt
```diff
@@ -133,7 +136,23 @@ class ApplicationsController(
 
     val serializedData = objectMapper.writeValueAsString(body.data)
 
-    val applicationResult = applicationService.updateApplication(applicationId, serializedData, username)
+    val applicationResult = when(body) {
+      is UpdateApprovedPremisesApplication -> applicationService.updateApprovedPremisesApplication(
```
pezholio, line 141: Are all these required? We won't necessarily have all these things at the start of the application

pezholio (commented):
Just a quick drive by - couple of questions

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/ApplicationsTransformer.kt
brownbearsoftware: Having a `status` column on the applications table would definitely make reading simpler and more efficient.  It is of course a trade off as if we want to change the status logic (as we did recently to add more statuses) at some point we have to figure out a migration process too.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ApplicationsController.kt
brownbearsoftware: `isWomensApplication`, `isPipeApplication`, `releaseType` and `arivalDate` are all optional in the OpenAPI spec and in the service layer so the FE is fine to just send whatever it has

pezholio: Cool cool 👍

pezholio (approved):
LGTM 👍

src/main/resources/static/api.yml
```diff
@@ -3847,6 +3850,71 @@ components:
...
+        person:
+          $ref: '#/components/schemas/Person'
+        createdAt:
+          type: string
+          format: date-time
+        submittedAt:
+          type: string
+          format: date-time
+      discriminator:
+        propertyName: type
```
pezholio, line 3868: Where does this `propertyName` get sent in the API? I'm trying to build the frontend changes now and I get a 400 when trying to send as `UpdateApprovedPremisesApplication` object

PR #586, opened 2023-04-20
Fix for Placement Request failure sequencing

brownbearsoftware:
This fixes the issue where a failure to create a Placement Request after an `application.assessed` Domain Event has been emitted causes the Domain Event to be removed from the database when the transaction rolls back.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AssessmentService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AssessmentService.kt
@@ -10,6 +10,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.Cru
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.PersonReference
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.ProbationArea
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.StaffMember
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.PlacementRequirements
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.ClientResult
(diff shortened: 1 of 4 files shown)


brownbearsoftware:
[Placement Request failure after emitting Application Assessed Domain Event issue](https://trello.com/c/P0uA1h89/1452-placement-request-failure-after-emitting-application-assessed-domain-event-issue)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AssessmentService.kt
```diff
@@ -223,6 +225,15 @@ class AssessmentService(
 
     val savedAssessment = assessmentRepository.save(assessment)
 
+    if (placementRequirements != null) {
+      val placementRequestValidationResult = placementRequestService.createPlacementRequest(assessment, placementRequirements)
+
+      when (placementRequestValidationResult) {
+        is ValidatableActionResult.Success -> Unit
+        else -> return AuthorisableActionResult.Success(placementRequestValidationResult.translateError())
```
pezholio, line 233: This threw me a bit, should be we returning `Success` if it's an error?

pezholio: I feel like an `if` would be better here, then we don't have to return a `Unit`

pezholio (approved):
LGTM - just one question about possibly improving the service.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AssessmentService.kt
brownbearsoftware: > should be we returning Success if it's an error?

This is because the return type for the function is `AuthorisableActionResult<ValidatableActionResult<AssessmentEntity>>` so here we're returning `yes this user is authorised to make this change but there is a validation error with the parameters provided`

Agreed on the `if` point - have updated

PR #587, opened 2023-04-20
Fix for Wiremock port race condition when running tests

brownbearsoftware:
Stopping Wiremock in the @AfterEach gives a period of time before it is restarted in @BeforeEach in which another process can claim the port.  Switching to resetting instead should resolve this.


pezholio (approved):
Happy with this. Will fix the flaxiness too, so let's get it in!

PR #592, opened 2023-04-24
Update repeatable Flyway migration to create premises in dev environment to use PDU IDs for Temporary Accommodation

danlivings-dxw:
This PR updates the `R__2_create_premises.sql` migration in the `dev` environment to replace the string `temporary_accommodation_premises.pdu` column (which was dropped in #589) with the new `temporary_accommodation_premises.probation_delivery_unit_id` column in the `INSERT` statements. This change will allow the API to deploy and start successfully in dev and therefore allow it to be deployed in higher environments.


pezholio (approved):
LGTM 👍

PR #593, opened 2023-04-24
(1472) Set an application as inapplicable

pezholio:
This adds a new `isInapplicable` property that can be sent when an application is updated. If it is set to true, then the status returned is `inapplicable` and the application is filtered out when requesting all applications.


brownbearsoftware (approved):
LGTM

PR #595, opened 2023-04-24
(Fix) Update and submit applications accepts date

pezholio:
We’re passing the arrival dates from the frontend as dates, not datetimes, so we need to update the service to accept dates. The underlying data model stores the `arrivalDate` as a datetime, but I think it’s OK to keep them like this to give us flexibility.


richpjames (approved):
Makes sense to me but worth taking my approval with a pinch of salt

brownbearsoftware (approved):

PR #599, opened 2023-04-27
Add turnaroundWorkingDayCount to TemporaryAccommodationPremises

alvaro:
### Task
https://trello.com/c/iWShWKBw/1016-premises-can-record-the-turnaround-day-count

### Changes
This PR builds on https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/597 and includes changes to:

- Add `turnaround_working_day_count` field to `temporary_accommodation_premises` table
- Expose/handle the new field when using CRUD operations on TemporaryAccommodationPremises entities

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PremisesEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PremisesEntity.kt
@@ -171,6 +171,7 @@ class TemporaryAccommodationPremisesEntity(
   @ManyToOne
   @JoinColumn(name = "probation_delivery_unit_id")
   var probationDeliveryUnit: ProbationDeliveryUnitEntity?,
+  var turnaroundWorkingDayCount: Int
 ) : PremisesEntity(
   id,
   name,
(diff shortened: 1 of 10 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PremisesEntity.kt
```diff
@@ -171,6 +171,7 @@ class TemporaryAccommodationPremisesEntity(
   @ManyToOne
   @JoinColumn(name = "probation_delivery_unit_id")
   var probationDeliveryUnit: ProbationDeliveryUnitEntity?,
+  var turnaroundWorkingDayCount: Int?
```
brownbearsoftware, line 174: Since the database column is `NOT NULL` I think this should also be non-nullable

alvaro: Makes sense. I have made changes to make to the field non-nullable

brownbearsoftware (approved):
LGTM

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/factory/TemporaryAccommodationPremisesEntityFactory.kt
```diff
@@ -94,6 +95,10 @@ class TemporaryAccommodationPremisesEntityFactory : Factory<TemporaryAccommodati
     this.probationDeliveryUnit = probationDeliveryUnit
   }
 
+  fun withTurnaroundWorkingDayCount(turnaroundWorkingDayCount: Yielded<Int>) = apply {
+    this.turnaroundWorkingDayCount = turnaroundWorkingDayCount
+  }
```
danlivings-dxw, line 100: I think this would be more consistent as:

```suggestion
  fun withTurnaroundWorkingDayCount(turnaroundWorkingDayCount: Int) = apply {
    this.turnaroundWorkingDayCount = { turnaroundWorkingDayCount }
  }
```

danlivings-dxw (approved):

PR #604, opened 2023-04-27
Seed CAS3 rooms and beds for local and dev

tahb:
It would be useful to have these seeded onto the dev environment so we have a sample set of data to use when testing features or demonstrating.

Following the precedant for seeding these for the test environment https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/main/src/main/resources/db/migration/test/R__2_create_rooms.sql

I'm not 100% sure why these don't already exist. It might be because we had e2e tests running on the dev environment which were creating realistic looking data. Since the recent move back to truncating the Premises table on deploy we only have the old premises seeds which are no longer enough?

Next steps are to seed bookings for these beds.

### DIFF
--- a/src/main/resources/db/migration/local+dev/R__6_create_rooms_and_beds.sql
+++ b/src/main/resources/db/migration/local+dev/R__6_create_rooms_and_beds.sql
@@ -1,0 +1,83 @@
+-- ${flyway:timestamp}
+
+TRUNCATE TABLE rooms CASCADE;
+TRUNCATE TABLE beds CASCADE;
+
+-- approved_premises (placeholder)
+
+-- temporary_accommodation_premises
+--      Premises 1 (2x rooms, 1x bed each)
+  insert into
+    rooms ("id", "name", "notes", "premises_id")
+  values
+    (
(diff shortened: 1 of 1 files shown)


src/main/resources/db/migration/local+dev/R__6_create_rooms_and_beds.sql
```diff
@@ -0,0 +1,83 @@
...
+-- temporary_accommodation_premises
+--      Premises 1 (2x rooms, 1x bed each)
+  insert into
+    rooms ("id", "name", "notes", "premises_id")
+  values
+    (
+      '14c0911e-6296-4b3f-ad00-5a2cf6f23a08',
+      'ROOM1',
+      NULL,
+      'd6447105-4bfe-4f1e-add7-4668e1ca28b0'
```
tahb, line 17: I should say that these premises ids come from the TA seed file that runs before hand https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/main/src/main/resources/db/migration/local%2Bdev/R__2_create_premises.sql#L31

brownbearsoftware (approved):
LGTM

PR #609, opened 2023-04-28
Add 'create turnaround' endpoint

alvaro:
> See [ticket #1025 on the CAS3 Trello board](https://trello.com/c/ERqViWY2/1025-new-endpoint-for-editing-a-booking-turnaround).

This PR introduces the `POST /premises/{premisesId}/bookings/{bookingId}/turnarounds` endpoint, which can be used to update the turnaround period for the specified booking.

This is implemented by creating a new `TurnaroundEntity` and adding it to the `BookingEntity.turnarounds` list.

The endpoint validates that:
- The `workingDays` value is a positive integer.
- The updated turnaround time does not conflict with another booking or a lost bed on the bed that the booking is for.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
brownbearsoftware, line 1: Guessing you've got autoformat turned on in the ktLint intelliJ plugin?  Has added a load of unrelated trailing commas to the changeset because the rules it uses are different than the gradle plugin.

PR #607, opened 2023-04-28
Create `turnarounds` table and JPA entity

danlivings-dxw:
> See [ticket #1027 on the CAS3 Trello board](https://trello.com/c/k0gQjnra/1027-new-bookings-automatically-create-a-new-turnaround-record).

This PR is the first part of the work to introduce and automatically create turnarounds for bookings.

Changes in this PR:
- A Flyway migration has been introduced to create the `turnarounds` table and another migration creates a turnaround for each existing Temporary Accommodation booking, using a value of zero working days.
- The `TurnaroundEntity` JPA entity has been introduced and the `BookingEntity` has been updated to reference its turnarounds.

PR #606, opened 2023-04-27
Set up an integration to gov uk bank holidays

alvaro:
### Task
https://trello.com/c/LNcyCgA9/1057-set-up-an-integration-to-govuk-bank-holidays
### Changes
This PR include changes to add:
-   An integration for gov uk bank holidays endpoint
-  A working day count service to return working day count between two dates (both inclusive)

PR #611, opened 2023-05-02
Automatically create a turnaround when creating a new Temporary Accommodation booking

danlivings-dxw:
> See ticket #1027 on the CAS3 Trello board](https://trello.com/c/k0gQjnra/1027-new-bookings-automatically-create-a-new-turnaround-record).

This PR is the second part of the work to introduce and automatically create turnarounds for bookings. It follows on from #607, which created the `turnarounds` table in the database and introduced the associated `TurnaroundEntity`.

This PR contains the following changes:
- A turnaround is automatically created for a booking when the booking itself is created. The working day count is set to the value as specified on the premises that the booking is within.
- The Redis configuration has been updated to support the GOV.UK Bank Holidays API, which was introduced as a dependency in #606, and configured with a cache expiry of 1 day.
- The booking conflict check has been updated to incorporate the turnaround time, which requires the calculation of a given number of working days (i.e. Monday-Friday and not a bank holiday) from a given date.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
@@ -69,6 +69,7 @@ class BookingService(
   private val domainEventService: DomainEventService,
   private val cruService: CruService,
   private val applicationService: ApplicationService,
+  private val workingDayCountService: WorkingDayCountService,
   private val communityApiClient: CommunityApiClient,
   private val bookingRepository: BookingRepository,
   private val arrivalRepository: ArrivalRepository,
@@ -346,11 +347,12 @@ class BookingService(
     enableTurnarounds: Boolean,
   ): AuthorisableActionResult<ValidatableActionResult<BookingEntity>> {
     val validationResult = validated {
-      getBookingWithConflictingDates(arrivalDate, departureDate, null, bedId)?.let {
-        return@validated it.id hasConflictError "A Booking already exists for dates from ${it.arrivalDate} to ${it.departureDate} which overlaps with the desired dates"
+      val expectedLastUnavailableDate = workingDayCountService.addWorkingDays(departureDate, premises.turnaroundWorkingDayCount)
+      getBookingWithConflictingDates(arrivalDate, expectedLastUnavailableDate, null, bedId)?.let {
+        return@validated it.id hasConflictError "A Booking already exists for dates from ${it.arrivalDate} to ${it.lastUnavailableDate} which overlaps with the desired dates"
       }
 
-      getLostBedWithConflictingDates(arrivalDate, departureDate, null, bedId)?.let {
+      getLostBedWithConflictingDates(arrivalDate, expectedLastUnavailableDate, null, bedId)?.let {
         return@validated it.id hasConflictError "A Lost Bed already exists for dates from ${it.startDate} to ${it.endDate} which overlaps with the desired dates"
       }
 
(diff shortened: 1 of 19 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/BookingTransformer.kt
```diff
@@ -44,6 +45,8 @@ class BookingTransformer(
     originalArrivalDate = jpa.originalArrivalDate,
     originalDepartureDate = jpa.originalDepartureDate,
     createdAt = jpa.createdAt.toInstant(),
+    turnaround = jpa.turnaround?.let { turnaroundTransformer.transformJpaToApi(it) },
```
brownbearsoftware, line 48: No big deal, but believe you can do a method reference here too like you have on the next line?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
```diff
@@ -346,11 +347,12 @@ class BookingService(
     enableTurnarounds: Boolean,
   ): AuthorisableActionResult<ValidatableActionResult<BookingEntity>> {
     val validationResult = validated {
-      getBookingWithConflictingDates(arrivalDate, departureDate, null, bedId)?.let {
-        return@validated it.id hasConflictError "A Booking already exists for dates from ${it.arrivalDate} to ${it.departureDate} which overlaps with the desired dates"
+      val expectedClosedDate = workingDayCountService.addWorkingDays(departureDate, premises.turnaroundWorkingDayCount)
```
brownbearsoftware, line 350: should this be `departureDate.plusDays(1)` ?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
```diff
@@ -839,6 +841,14 @@ class BookingService(
     newDepartureDate: LocalDate,
     notes: String?
   ) = validated<ExtensionEntity> {
+    val expectedClosedDate = workingDayCountService.addWorkingDays(newDepartureDate, booking.turnaround?.workingDayCount ?: 0)
```
brownbearsoftware, line 844: as above with plusDays(1)?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
```diff
@@ -938,6 +947,9 @@ class BookingService(
     endDate,
     thisEntityId
   ).firstOrNull()
+
+  val BookingEntity.closedDate: LocalDate
+    get() = workingDayCountService.addWorkingDays(this.departureDate, this.turnaround?.workingDayCount ?: 0)
```
brownbearsoftware, line 952: as above with plusDays(1)?

brownbearsoftware:
Other than the `plusDays(1)` questions, looks good to me.

There's a bit of duplication with my reports PRs (TurnaroundEntityFactory, the bank holidays wiremock stuff and a method to calculate end date given a start date and number of working days) but happy to rebase after you've merged this as this seems more critical path.

danlivings-dxw:
@brownbearsoftware I think as the checks for overlapping dates include a date being equal on either side that the `plusDays(1)` is not needed in those places:

- A booking with today as the departure date and a turnaround of zero days should not prevent a new booking from being made for an arrival in the same room tomorrow, which `today.plusDays(1)` would do.
- The same booking but with a turnaround of one day should prevent a booking being made for tomorrow, but not Friday.

Perhaps this is an issue of overloading the meaning of 'closed date'? I believe that it's supposed to mean the date on which the room is available again, but here I'm using it to mean the last day for which it's still unavailable because of the turnaround. Perhaps renaming it to something like `lastUnavailableDate` would be better?

brownbearsoftware (approved):
LGTM

PR #616, opened 2023-05-04
Add seed Approved Premises for the dev envrionemnt

pezholio:
This adds a bunch of Approved Premises and their corresponding beds/rooms. This is based on the North East premises with the names, addresses/locations etc anonymised. This will ensure we have a realistic corpus of data in dev, without having to add potentially confidential data to the repo


brownbearsoftware (approved):
Looks really realistic, nice

brownbearsoftware (approved):

PR #623, opened 2023-05-10
1037 - Cache clear mechanism

brownbearsoftware:
This adds a `DELETE /cache/{cache name}/clear` endpoint that can only be accessed locally which clears the named cache.

### DIFF
--- a/src/main/resources/static/api.yml
+++ b/src/main/resources/static/api.yml
@@ -2628,6 +2628,24 @@ paths:
           $ref: '#/components/responses/403Response'
         500:
           $ref: '#/components/responses/500Response'
+  /cache/{cacheName}:
+    delete:
+      summary: Clears the given cache, can only be called from a local connection
+      parameters:
+        - in: path
+          name: cacheName
+          required: true
+          schema:
+            $ref: '#/components/schemas/CacheType'
+      responses:
+        200:
+          description: successfully cleared cache
+        401:
+          $ref: '#/components/responses/401Response'
+        403:
+          $ref: '#/components/responses/403Response'
+        500:
+          $ref: '#/components/responses/500Response'
   /migration-job:
     post:
       summary: Starts a migration job (process for data migrations that can't be achieved solely via SQL migrations), can only be called from a local connection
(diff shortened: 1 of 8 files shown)


brownbearsoftware:
[Add mechanism for clearing API caches without releasing a new build](https://trello.com/c/GjkGZYO2/1037-add-mechanism-for-clearing-api-caches-without-releasing-a-new-build)

src/main/resources/static/api.yml
```diff
@@ -2628,6 +2628,24 @@ paths:
           $ref: '#/components/responses/403Response'
         500:
           $ref: '#/components/responses/500Response'
+  /cache/{cacheName}/clear:
```
pezholio, line 2631: With my REST hat on, we should just call this `/cache/{cacheName}`, as `cacheName` is the thing we're `DELETE`ing

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/CacheClearTest.kt
```diff
@@ -0,0 +1,139 @@
...
+
+      // Call our endpoint again
+      webTestClient.get()
+        .uri("/premises/$premisesId/staff")
+        .header("Authorization", "Bearer $jwt")
+        .exchange()
+        .expectStatus()
+        .isOk
+
+      // Validate upstream request was made again, proving the cache was cleared
```
pezholio, line 51: Loving these comments - makes it so much easier to follow! Might be worth following the convention we use in the UI and use Cucumber-like syntax, ie:

- Given I call the qCode endpoint twice
- Then the upstream request should only have been made once
- When I clear the qCode cache
- And I call the endpoint again
- Then the upstream request should be made again

pezholio (approved):
LGTM - Just thinking about how to make those comments even more useful. Might be worth copying for future integration tests too. Often there's a lot going on, so cucumber-like syntax makes it easy to follow 👍

src/main/resources/static/api.yml
brownbearsoftware: Good point, have switched it

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/CacheClearTest.kt
brownbearsoftware: Done for that one and the other test too

danlivings-dxw (approved):
LGTM

PR #627, opened 2023-05-11
Use CSVs for seeding data in `dev` environment instead of Flyway migrations

danlivings-dxw:
> See [ticket #1500 on the CAS1 Trello board](https://trello.com/c/LaFasPQ1/1500-improve-seeding).

This PR introduces the ability to use CSV files to seed data in local, dev, and test environments. This functionality already exists (and is used to seed actual user data into production), but requires manual intervention. Therefore, this PR introduces a mechanism to reuse this existing functionality in an automated way.

This functionality is disabled by default - the configuration property `seed.auto.enabled` must be set to `true`, either in `application.yml` or through the `SEED_AUTO_ENABLED` environment variable in the Helm chart. When enabled, every folder specified in `seed.auto.file-prefixes` will be scanned for CSV files. Any files with a name matching that of a job (e.g. `approved_premises_rooms.csv`) will execute that job using that file as input. The `seed.auto.file-prefixes` configuration property works much like `spring.flyway.locations`, in that it is a comma-separated list that supports classpath resources (such as `classpath:db/seed/env1,classpath:db/seed/env2`). Any folders that don't exist or files that aren't named after a known job type are skipped, with an appropriate warning logged.

It's worth noting that this method of seeding is slower than the current approach of using SQL files to seed data, but as this runs only at startup (and is only intended to be used in lower environments) this shouldn't be too much of an issue.


brownbearsoftware (approved):
LGTM

PR #631, opened 2023-05-15
1443 - Gov Notify integration

brownbearsoftware:
This adds a Gov Notify integration and uses it to send an email when an Assessment for an Application is allocated or reallocated.

The `notify.mode` property has 3 options:
- DISABLED: no interaction with the Notify service actually takes place
- TEST_AND_GUEST_LIST: use a Notify Guest List key for certain users, use a different key for other users
- ENABLED: use a single Notify key for all users

There are 3 types of Notify key:
- Test: Doesn't actually send emails but the API responds as though it has
- Guest List: Does send emails to verified email addresses but responds with an error for other email addresses
- Live: Sends emails to any email address

In dev we can combine the guest list and test modes by having a list on our side (which is the new table `notify_guest_list_users`) which we first check to see which key to use.  This allows us to actually send emails to the product team and not error for other emails.

The email templates are setup on the Notify web app - we are awaiting copy for these so I have improvised for now but since the feature isn't enabled in any environment as part of this PR no emails will actually be sent yet.

Also refactored the existing URL templates into a more structured form e.g. `url-templates.api.application-submitted-event-detail` instead of `application-submitted-detail-url-template`.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AssessmentService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AssessmentService.kt
@@ -13,6 +13,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.StaffMe
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.PlacementRequirements
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.ClientResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.CommunityApiClient
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.config.NotifyConfig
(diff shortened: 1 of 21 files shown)


brownbearsoftware:
[Start implementing Emails](https://trello.com/c/IXtIl6hL/1443-start-implementing-emails)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AssessmentService.kt
```diff
@@ -446,6 +459,15 @@ class AssessmentService(
       )
     )
 
+    emailNotificationService.sendEmail(
+      user = assigneeUser,
+      templateId = "f3d78814-383f-4b5f-a681-9bd3ab912888",
```
pezholio, line 464: Same here

pezholio (approved):
Looks generally OK, but wondering if we should make the template ID an env var rather than hardcoded.

PR #632, opened 2023-05-15
1446 - Preemptive caching

brownbearsoftware:
This is the same original set of commits as in: https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/596

With some additional changes based on the attempted deployment of that branch:

- Fix for RedLock connection, now handles TLS correctly
- Exception handling for RedLock issues - previously would kill the thread but now handled and loop continues
- Increase delay between attempts to get lock to reduce impact on Redis
- Store metadata (response status, soft TTL etc.) in different Redis entry than actual response data to avoid transferring entire response body from Redis->API during refresh process


brownbearsoftware:
[Performance Improvements - preemptive caching of Community API and Prison API](https://trello.com/c/S2z8BXtQ/1446-performance-improvements-preemptive-caching-of-community-api-and-prison-api)

danlivings-dxw (approved):

PR #637, opened 2023-05-16
Fix/#1515 add missing fields to assessment summaries

bell-pa:
Extra bits for card #1515 'assessments to only return necessary information to prevent OOM issues'
Added these fields to each returned assessment summary.
* `arrivalDate`, optional.
* `decision`, optional.  'accepted', 'rejected' or null.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/AssessmentSummaryQueryTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/AssessmentSummaryQueryTest.kt
@@ -9,6 +9,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.integration.givens.`Give
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.integration.givens.`Given an Assessment for Temporary Accommodation`
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApprovedPremisesApplicationEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.AssessmentClarificationNoteEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.AssessmentDecision
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.AssessmentEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.AssessmentRepository
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.DomainAssessmentSummary
@@ -41,6 +42,7 @@ class AssessmentSummaryQueryTest : IntegrationTestBase() {
               withAllocatedToUser(user2)
               withAssessmentSchema(apAssessment.schemaVersion)
               withApplication(application)
+              withDecision(AssessmentDecision.ACCEPTED)
             }
 
             val results: List<DomainAssessmentSummary> = realAssessmentRepository.findAllAssessmentSummariesNotReallocated()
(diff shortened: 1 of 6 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/AssessmentSummaryQueryTest.kt
```diff
@@ -68,18 +83,18 @@ class AssessmentSummaryQueryTest : IntegrationTestBase() {
     }
   }
 
-  private fun assertForAssessmentSummary(summary: DomainAssessmentSummary, assessment: AssessmentEntity, dateOfInfoRequest: OffsetDateTime) {
+  private fun assertForAssessmentSummary(summary: DomainAssessmentSummary, assessment: AssessmentEntity, dateOfInfoRequest: OffsetDateTime?) {
```
pezholio, line 88: Big fan of DRYing up the expectations like this 👍

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/AssessmentSummaryQueryTest.kt
```diff
@@ -23,22 +23,37 @@ class AssessmentSummaryQueryTest : IntegrationTestBase() {
...
+    `Given a User` { user1, _ ->
+      `Given an Assessment for Approved Premises`(user1, user1, reallocated = true) { _, _ -> }
+      `Given an Assessment for Approved Premises`(user1, user1) { apAssessment, _ ->
+
+        val expectedApAssessmentNote = earliestUnansweredClarificationNote(apAssessment, user1)
+        `Given an Assessment for Temporary Accommodation`(user1, user1) { taAssessment, _ ->
+          val expectedTaAssessmentNote = earliestUnansweredClarificationNote(taAssessment, user1)
+          `Given a User` { user2, _ ->
+
+            val u2Assessment = assessmentEntityFactory.produceAndPersist {
```
pezholio, line 36: Not for now, but I was talking to @brownbearsoftware the other day about adding cucumber-like comments to the integration tests to make them easier to follow. This one is quite small, and the Fixture stuff gives a lot of the `Given`-type context, but sometimes there can be a lot of code running through these tests and it can get quite tricky to grok

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/transformer/AssessmentTransformerTest.kt
```diff
@@ -239,4 +248,13 @@ class AssessmentTransformerTest {
     assertThat(apiSummary.risks).isEqualTo(risksTransformer.transformDomainToApi(personRisks, domainSummary.crn))
     assertThat(apiSummary.person).isNotNull
   }
+
+  companion object {
+    @JvmStatic
+    fun assessmentDecisionPairs(): Stream<Arguments> = Stream.of(
```
pezholio, line 254: This might be better put before the tests (assuming there's not a Kotlin/Java reason you can't) just to save the misdirection of having to look at the bottom of the file, especially as the `value` is set to a string value, so it's not immediately obvious it's referring to a function.

pezholio (approved):
LGTM - just some minor non-blocking points. Feel free to ignore / push back!

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/transformer/AssessmentTransformerTest.kt
bell-pa: Done

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/AssessmentSummaryQueryTest.kt
bell-pa: Kotest BehaviourSpec might help, although the given/when/then blocks have to be nested.
https://kotest.io/docs/framework/testing-styles.html#behavior-spec

Kotest assertions are much more readable the junit / assertj assertions (in my opinion):
https://kotest.io/docs/assertions/assertions.html

pezholio: Yeah, too much nesting can make things much much worse! I was just thinking comments are a light touch solution. I like Kotest assertions though - much more like Rspec matchers, which, as a Rubyist, makes me very happy! Probably too late to switch out wholesale though.

pezholio (approved):

PR #641, opened 2023-05-17
Add support for creating Temporary Accommodation applications

danlivings-dxw:
> See [ticket #1104 on the CAS3 Trello board](https://trello.com/c/nonHg12u/1104-a-user-can-create-a-temporary-accommodation-application).

This PR allows clients to create an application for the Temporary Accommodation service. The process is similar to creating an Approved Premises application, but no checks are made using the `apDeliusContextApiClient.getTeamsManagingCase` function. Temporary Accommodation will be using probation regions to determine access to applications at this stage.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ApplicationTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ApplicationTest.kt
@@ -25,7 +25,9 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ApprovedPremis
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.NewApplication
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.OfflineApplication
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ReleaseTypeOption
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ServiceName
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.SubmitApplication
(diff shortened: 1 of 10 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ApplicationTest.kt
```diff
@@ -977,6 +979,44 @@ class ApplicationTest : IntegrationTestBase() {
...
+        }
+
+        val result = webTestClient.post()
+          .uri("/applications")
+          .header("Authorization", "Bearer $jwt")
+          .header("X-Service-Name", ServiceName.temporaryAccommodation.value)
+          .bodyValue(
+            NewApplication(
+              crn = offenderDetails.otherIds.crn,
+              convictionId = 123,
```
tahb, line 998: `convictionId` and `deliusEventNumber` are new to me. I don't suppose you know what they are as presumably they are required?

tahb, line 998: `convictionId` and `deliusEventNumber` are new to me. I don't suppose you know what they are as presumably they are required?

tahb (commented):
LGTM. I like the way the commits step through this change, it made it really easy to review. CI is failing but otherwise :+1:

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ApplicationTest.kt
danlivings-dxw: They come in through the API. I'm not 100% sure what they're for but I suspect that they're related to the index offence. It corresponds to [this file in the frontend](https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/blob/main/server/%40types/shared/models/ActiveOffence.ts) which seems to be used in various places.

tahb (approved):

PR #646, opened 2023-05-19
Add support for submitting Temporary Accommodation applications

danlivings-dxw:
> See [ticket #1105 on the CAS3 Trello board](https://trello.com/c/P7SY195s/1105-a-user-can-submit-a-temporary-accommodation-application).

This PR allows clients to submit an application for the Temporary Accommodation service. The process is similar to submitting an Approved Premises application, but with the following differences:
- An assessment object is not created when the application is submitted. As of now there is no requirement to assign submitted applications to a particular user based on qualifications, so this step has been skipped.
- A domain event is not fired when the application is submitted. This might be a requirement in the future, but at the moment it's unknown what data would be needed to be provided, so this step has also been skipped.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ApplicationService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ApplicationService.kt
@@ -489,8 +489,62 @@ class ApplicationService(
     applicationId: UUID,
     submitApplication: SubmitTemporaryAccommodationApplication
   ): AuthorisableActionResult<ValidatableActionResult<ApplicationEntity>> {
+    var application = applicationRepository.findByIdOrNullWithWriteLock(applicationId)?.let(jsonSchemaService::checkSchemaOutdated)
+      ?: return AuthorisableActionResult.NotFound()
+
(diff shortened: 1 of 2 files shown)


brownbearsoftware (approved):

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ApplicationTest.kt
```diff
@@ -1339,6 +1340,69 @@ class ApplicationTest : IntegrationTestBase() {
...
+    `Given a User`(
+      staffUserDetailsConfigBlock = {
+        withTeams(
+          listOf(
+            StaffUserTeamMembershipFactory().produce()
+          )
+        )
+      }
+    ) { submittingUser, jwt ->
+      `Given a User`(roles = listOf(UserRole.ASSESSOR), qualifications = listOf(UserQualification.PIPE, UserQualification.WOMENS)) { assessorUser, _ ->
```
tahb, line 1354: When we submit a TA application does it matter what the user role is? Are we tied into the CAS1 stuff for the moment?

Qualifications aren't applicable to us and won't be for the foreseeable. Not sure if we can remove this setup.

tahb (approved):
LGTM

PR #648, opened 2023-05-22
(1536) Add an endpoint to list a summary of all beds in a premises

pezholio:
This add a `/premises/$premises.id/beds` endpoint to return a summary of all the beds in a premises. I borrow heavily from @bell-pa's work on Assessment summaries to carry out a JPQL query that returns booleans for if a room is booked or out of service, and then transform that into a status response.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BedEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BedEntity.kt
@@ -1,18 +1,70 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity
 
 import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.data.jpa.repository.Query
 import org.springframework.stereotype.Repository
 import java.util.UUID
+import javax.persistence.ColumnResult
+import javax.persistence.ConstructorResult
 import javax.persistence.Entity
 import javax.persistence.Id
 import javax.persistence.JoinColumn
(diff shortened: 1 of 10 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BedEntity.kt
```diff
@@ -1,19 +1,91 @@
...
+                  and lost_bed.start_date <= CURRENT_DATE
+                  and lost_bed.end_date >= CURRENT_DATE
+          ) > 0 
+          then 
+            true 
+          else 
+            false
+          end
+        ) as boolean
+      ) as bedOutOfService
```
danlivings-dxw, line 65: I don't think the `case`/`when` or the `cast` are necessary for these subqueries:

```suggestion
      (
        select count(booking.id)
        from bookings booking
        where booking.bed_id = b.id
          and booking.arrival_date <= CURRENT_DATE
          and booking.departure_date >= CURRENT_DATE
      ) > 0 as bedBooked,
      (
        select count(lost_bed.id)
        from lost_beds lost_bed
        where lost_bed.bed_id = b.id
          and lost_bed.start_date <= CURRENT_DATE
          and lost_bed.end_date >= CURRENT_DATE
      ) > 0 as bedOutOfService
```

pezholio: Ah, of course! That's much nicer 👍

danlivings-dxw (approved):

PR #649, opened 2023-05-22
Include risk data in API responses returning a Temporary Accommodation application

danlivings-dxw:
> See [ticket #1120 on the CAS3 Trello board](https://trello.com/c/5FEf051U/1120-include-risk-information-in-the-api-responses-for-application-endpoints).

The frontend logic that was copied across from AP includes the RoSH/MAPPA risks widget. This information is stored in the database for a Temporary Accommodation application but is not returned by the API.

This should be returned for now as although we don’t know if we need it in the future, it’s fairly trivial to implement, compared to the effort involved in removing it from the frontend.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/transformer/ApplicationsTransformerTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/transformer/ApplicationsTransformerTest.kt
@@ -14,6 +14,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ApprovedPremis
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ApprovedPremisesApplicationSummary
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Person
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.TemporaryAccommodationApplication
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.TemporaryAccommodationApplicationSummary
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.factory.ApAreaEntityFactory
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.factory.ApprovedPremisesApplicationEntityFactory
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.factory.ApprovedPremisesEntityFactory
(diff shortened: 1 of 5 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/transformer/ApplicationsTransformerTest.kt
```diff
@@ -581,4 +584,50 @@ class ApplicationsTransformerTest {
...
+      override fun getLatestAssessmentHasClarificationNotesWithoutResponse() = false
+      override fun getHasBooking() = false
+    }
+
+    val result = applicationsTransformer.transformDomainToApiSummary(application, mockk(), mockk()) as TemporaryAccommodationApplicationSummary
+
+    assertThat(result.id).isEqualTo(application.getId())
+    assertThat(result.createdByUserId).isEqualTo(application.getCreatedByUserId())
+    assertThat(result.status).isEqualTo(ApplicationStatus.inProgress)
+    assertThat(result.risks).isNotNull
```
tahb, line 608: Is this enough of a test? It looks like the `applicationTransformer` is using an `objectMapper` to load risk data for a person, if present it's also sent to the `risksTransformer` to transform it again into `PersonRisks`. In our test we are only asserting something is present so presumably trusting the typing to make sure whatever it is, is of type `PersonRisks`. Should we be mocking these other classes and testing the interactions?

tahb (approved):
👍 

It seems like we're reasonably following the steer from CAS1 on how to get and set the risk data. Do we understand in what scenarios we might not have any?

I presume everyone on probation has some form of risk rating whether that be high or low? Whenever we see no risk ratings on an application can we presume that they exist but there was a problem finding them upstream? It seems like an important bit of data to be optional.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/transformer/ApplicationsTransformerTest.kt
danlivings-dxw: The main `TemporaryAccommodationApplicationEntity` class defines its `riskRatings` property to be of type `PersonRisks`. While theoretically the row in the database could contain arbitrary JSON data, I think in practice the only way this would happen is with a manual `INSERT` SQL query.

There doesn't seem to be a huge amount of testing around the risk data in general (there are no unit tests for `PersonRisksTransformer`, for example). It'd be a good idea to tighten up some of these tests, but this PR is a blocker for the frontend work so I haven't invested a huge amount of time in putting the additional infrastructure in place.

danlivings-dxw:
> 👍
> 
> It seems like we're reasonably following the steer from CAS1 on how to get and set the risk data. Do we understand in what scenarios we might not have any?
> 
> I presume everyone on probation has some form of risk rating whether that be high or low? Whenever we see no risk ratings on an application can we presume that they exist but there was a problem finding them upstream? It seems like an important bit of data to be optional.

One way is that the 'create application' endpoint can be called with the query string `?createWithRisks=false`, which we've inherited from AP. I don't know the reason this query parameter exists, and it might be the case that we never want to be able to create an application without the risk information. If that's the case, we can throw an error or ignore it, but I didn't want to be too opinionated about what we do given that it seems like there's some unknowns regarding how Apply works in TA.

PR #650, opened 2023-05-23
(1536) Add endpoint to show bed details

pezholio:
This builds on #648 to add an endpoint to show all the details of a given bed, together with the characteristics that are tied to that bed's room. This will allow us to build out the bed view as shown in https://approved-premises-prototype-main.apps.live.cloud-platform.service.justice.gov.uk/manage/bed-information


brownbearsoftware (approved):
I would perhaps have gone for the more traditional approach of getting a BedEntity and traversing the properties for this one since it's a get-single-thing endpoint just because it's easier to follow and there should be less performance worries than with a list-all-things endpoint.  Should work fine though and looks more performant than doing it that way so LGTM

PR #658, opened 2023-05-25
Fix for auto seeding when running from JAR

brownbearsoftware:
When running tests & locally through IntelliJ seed files can be loaded because they physically exist on the Filesystem.  When running from a JAR they are embedded within the JAR so first need to be copied to a real file before the rest of the seeding process can read them.


pezholio (approved):
Ah nice - I did wonder if it was something like that. Bloomin' compiled languages!

danlivings-dxw (approved):

PR #659, opened 2023-05-25
Add user/pass delimiter to rediss urls

brownbearsoftware:
The format we were outputting originally (only a password) is valid but JedisURIHelper that's used by RedLock assumes username and password always appear together

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/RedisConfiguration.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/RedisConfiguration.kt
@@ -68,7 +68,7 @@ class RedisConfiguration {
     @Value("\${spring.redis.ssl}") ssl: Boolean,
   ): RedLock {
     val scheme = if (ssl) "rediss" else "redis"
-    val passwordString = if (password.isNotEmpty()) "$password@" else ""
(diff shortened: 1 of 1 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/RedisConfiguration.kt
```diff
@@ -68,7 +68,7 @@ class RedisConfiguration {
     @Value("\${spring.redis.ssl}") ssl: Boolean,
   ): RedLock {
     val scheme = if (ssl) "rediss" else "redis"
-    val passwordString = if (password.isNotEmpty()) "$password@" else ""
+    val passwordString = if (password.isNotEmpty()) ":$password@" else ""
```
pezholio, line 71: This confused me at first - I'm assuming there's no user, but it needs to be in the format `user:password` right?

pezholio (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/RedisConfiguration.kt
brownbearsoftware: that's right yeah

PR #661, opened 2023-05-26
(171) Allow placement requests to be created without a start date

pezholio:
This splits the Placement Requests into two objects, the Placement Request itself (which is used to match with a placement) and Placement Requirements, which contain all the requirements for a placement. If an Application doesn't have a date, we create the Requirements, but not the request. These will be used by the PP to request placment request(s) when the dates are known.

I've tried to break this up a bit, but it's been tricky as there are so many moving parts everywhere. Let me know if there's anything I can do to make it easier to review 🙏


brownbearsoftware (approved):
LGTM

PR #670, opened 2023-05-30
POST and GET CAS-2 applications

Harriethw:
# Context

[Trello card](https://trello.com/c/gYeRYOSP/214-save-an-application)
As part of our work on CAS-2, we need to be able to create and retrieve blank applications in a CAS-2 specific manner.

# Changes in this PR

We are mostly following the [pattern already set out by CAS-3](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/641). When POSTing to `/applications` we check the `ServiceName` and diverge to saving our own application. When GETing applications we check the service name in the `ApplicationService`.

### feedback/questions

* naming: the CAS-2 repo is currently called `supported-accommodation` but this name has now been rejected - it looks like we will be using CAS-2 as a primary name for users, with a second name TBA. After chatting with other devs it seems like there is a future plan to bring everything under the CAS umbrella, so we feel _ok_ using CAS2 going forward  in code, although it doesn't feel great sitting next to `temporaryAccommodation`/`approvedPremises` and [maybe doesn't fit with moj naming guidelines](https://technical-guidance.service.justice.gov.uk/documentation/standards/naming-things.html#principles). Any feedback or suggestions welcome. 
* We are not familiar with Kotlin best practices at all, so please feedback on whether we can cut down on any repetition or tidy the code in any way. 
* Have we added everything necessary in `api.yml`?
* We are currently re-using the `TemporaryAccommodation` JSON schema, as we have not got to the stage of saving our own application form, is that ok for now?
* There were a couple of things we weren't sure how to test in `ApplicationService` -> `createCas2Application` , which we copied from the CAS3 method
  * what is the `validationErrors.any()` check doing? is that if there are any errors on the `ApplicationEntity`, based on what fields?
  * there is a check for `offenderDetails.otherIds.nomsNumber == null` - how would we recreate that in a unit test?

# Release checklist

[Release process documentation](https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/edit-v2/4247847062?draftShareId=a1c360ab-bd31-4db1-aae3-7cc002761de9).

## Pre-merge checklist

- [ ] Are any changes required to dependent services for this change to work?
  (eg. CAS API)
- [ ] Have they been released to production already?

## Post-merge checklist

- [ ] [Manually approve](https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/4247847062/Release+process#Manual-releases)
  release to test
- [ ] [Manually approve](https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/4247847062/Release+process#Manual-releases)
  release to preprod
- [ ] [Manually approve](https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/4247847062/Release+process#Manual-releases)
  release to prod

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ApplicationsController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ApplicationsController.kt
@@ -123,6 +123,7 @@ class ApplicationsController(
 
     val applicationResult = when (xServiceName ?: ServiceName.approvedPremises) {
       ServiceName.approvedPremises -> applicationService.createApprovedPremisesApplication(body.crn, user, deliusPrincipal.token.tokenValue, body.convictionId, body.deliusEventNumber, body.offenceId, createWithRisks)
+      ServiceName.cas2 -> applicationService.createCas2Application(body.crn, user, deliusPrincipal.token.tokenValue)
       ServiceName.temporaryAccommodation -> applicationService.createTemporaryAccommodationApplication(body.crn, user, deliusPrincipal.token.tokenValue, body.convictionId, body.deliusEventNumber, body.offenceId, createWithRisks)
     }
 
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
@@ -115,6 +115,7 @@ class PremisesController(
 
         summaries.map(premisesSummaryTransformer::transformDomainToApi)
       }
+      ServiceName.cas2 -> throw RuntimeException("CAS2 not supported")
       ServiceName.temporaryAccommodation -> {
         val user = usersService.getUserForRequest()
         val summaries = premisesService.getAllTemporaryAccommodationPremisesSummaries(user.probationRegion.id)
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApplicationEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApplicationEntity.kt
@@ -131,9 +131,9 @@ SELECT
     CAST(a.created_by_user_id AS TEXT) as createdByUserId,
     a.created_at as createdAt,
     a.submitted_at as submittedAt,
-    CAST(apa.risk_ratings AS TEXT) as riskRatings
-FROM cas_2_applications apa
-LEFT JOIN applications a ON a.id = apa.id
+    CAST(c2a.risk_ratings AS TEXT) as riskRatings
+FROM cas_2_applications c2a
+LEFT JOIN applications a ON a.id = c2a.id
 """,
     nativeQuery = true,
   )
@@ -257,6 +257,41 @@ data class ApplicationTeamCodeEntity(
 )
 
 @Entity
+@DiscriminatorValue("cas-2")
+@Table(name = "cas_2_applications")
+@PrimaryKeyJoinColumn(name = "id")
+class Cas2ApplicationEntity(
+  id: UUID,
+  crn: String,
+  createdByUser: UserEntity,
+  data: String?,
+  document: String?,
+  schemaVersion: JsonSchemaEntity,
+  createdAt: OffsetDateTime,
+  submittedAt: OffsetDateTime?,
+  schemaUpToDate: Boolean,
+  assessments: MutableList<AssessmentEntity>,
+  nomsNumber: String,
+  @Type(type = "com.vladmihalcea.hibernate.type.json.JsonType")
+  @Convert(disableConversion = true)
+  val riskRatings: PersonRisks?,
+) : ApplicationEntity(
+  id,
+  crn,
+  createdByUser,
+  data,
+  document,
+  schemaVersion,
+  createdAt,
+  submittedAt,
+  schemaUpToDate,
+  assessments,
+  nomsNumber,
+) {
+  override fun getRequiredQualifications(): List<UserQualification> = emptyList()
+}
+
+@Entity
 @DiscriminatorValue("temporary-accommodation")
 @Table(name = "temporary_accommodation_applications")
 @PrimaryKeyJoinColumn(name = "id")
@@ -318,5 +353,9 @@ interface ApprovedPremisesApplicationSummary : ApplicationSummary {
 }
 
 interface TemporaryAccommodationApplicationSummary : ApplicationSummary {
+  fun getRiskRatings(): String?
+}
+
+interface Cas2ApplicationSummary : ApplicationSummary {
   fun getRiskRatings(): String?
 }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApplicationEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApplicationEntity.kt
@@ -122,6 +122,22 @@ WHERE a.created_by_user_id = :userId
 
   @Query("SELECT DISTINCT(a.nomsNumber) FROM ApplicationEntity a")
   fun getDistinctNomsNumbers(): List<String>
+
+  @Query(
+    """
+SELECT
+    CAST(a.id AS TEXT) as id,
+    a.crn,
+    CAST(a.created_by_user_id AS TEXT) as createdByUserId,
+    a.created_at as createdAt,
+    a.submitted_at as submittedAt,
+    CAST(apa.risk_ratings AS TEXT) as riskRatings
+FROM cas_2_applications apa
+LEFT JOIN applications a ON a.id = apa.id
+""",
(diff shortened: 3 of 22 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
```diff
@@ -115,6 +115,11 @@ class PremisesController(
 
         summaries.map(premisesSummaryTransformer::transformDomainToApi)
       }
+      ServiceName.cas2 -> {
```
Harriethw, line 118: we can delete this as we won't have premises/don't have them yet

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApplicationEntity.kt
```diff
@@ -122,6 +122,22 @@ WHERE a.created_by_user_id = :userId
...
+    CAST(a.created_by_user_id AS TEXT) as createdByUserId,
+    a.created_at as createdAt,
+    a.submitted_at as submittedAt,
+    CAST(apa.risk_ratings AS TEXT) as riskRatings
+FROM cas_2_applications apa
+LEFT JOIN applications a ON a.id = apa.id
+""",
+    nativeQuery = true,
+  )
+  fun findAllSupportedAccommodationSummaries(): List<Cas2ApplicationSummary>
```
Harriethw, line 140:
```suggestion
  fun findAllCas2Summaries(): List<Cas2ApplicationSummary>
```

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApplicationEntity.kt
```diff
@@ -122,6 +122,22 @@ WHERE a.created_by_user_id = :userId
...
+  @Query(
+    """
+SELECT
+    CAST(a.id AS TEXT) as id,
+    a.crn,
+    CAST(a.created_by_user_id AS TEXT) as createdByUserId,
+    a.created_at as createdAt,
+    a.submitted_at as submittedAt,
+    CAST(apa.risk_ratings AS TEXT) as riskRatings
+FROM cas_2_applications apa
```
brownbearsoftware, line 135: fairly minor but maybe change the alias from `apa` to `c2a` or something along those lines

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
```diff
@@ -121,6 +121,7 @@ class PremisesController(
 
         summaries.map(premisesSummaryTransformer::transformDomainToApi)
       }
+      else -> throw RuntimeException("Could not determine service")
```
brownbearsoftware, line 124: Would prefer just adding the missing enum value rather than an else so it's easy to see where changes potentially need to be made if another service is ever added (I know there's only CAS1/2/3 so this is unlikely but still), i.e.

```
ServiceName.cas2 -> throw RuntimeException("cas2 not supported")
```

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/reporting/generator/ReportGenerator.kt
```diff
@@ -42,5 +42,6 @@ abstract class ReportGenerator<Input : Any, Output : Any, Properties>(private va
     when (serviceName) {
       ServiceName.approvedPremises -> premisesEntity is ApprovedPremisesEntity
       ServiceName.temporaryAccommodation -> premisesEntity is TemporaryAccommodationPremisesEntity
+      else -> throw RuntimeException("Could not determine service")
```
brownbearsoftware, line 45: as above with regard to adding the missing enum branch instead of an `else`

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ApplicationService.kt
```diff
@@ -251,6 +251,54 @@ class ApplicationService(
...
+    }
+
+    val createdApplication = applicationRepository.save(
+      Cas2ApplicationEntity(
+        id = UUID.randomUUID(),
+        crn = crn,
+        createdByUser = user,
+        data = null,
+        document = null,
+        // TODO: replace this with a CAS-2 specific JSON schema
```
brownbearsoftware, line 288: I'll leave a more detailed general comment about the JSON Schema stuff but just a note here that you should do this now rather than later

src/main/resources/db/migration/all/20230525131850__add_risk_fields_to_cas_2_applications.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE cas_2_applications ADD COLUMN risk_ratings JSON;
```
brownbearsoftware, line 1: Would just include this in the CREATE TABLE DDL in `20230525120902__cas_2_applications.sql`

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ApplicationsController.kt
```diff
@@ -120,6 +120,7 @@ class ApplicationsController(
 
     val applicationResult = when (xServiceName ?: ServiceName.approvedPremises) {
       ServiceName.approvedPremises -> applicationService.createApprovedPremisesApplication(body.crn, user, deliusPrincipal.token.tokenValue, body.convictionId, body.deliusEventNumber, body.offenceId, createWithRisks)
+      ServiceName.cas2 -> applicationService.createCas2Application(body.crn, user, deliusPrincipal.token.tokenValue)
```
brownbearsoftware, line 123: Definitely a nitpick but I'd say this is a change rather than a refactor since we're adding functionality

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ApplicationService.kt
```diff
@@ -70,7 +71,6 @@ class ApplicationService(
   fun getAllApplicationsForUsername(userDistinguishedName: String, serviceName: ServiceName): List<ApplicationSummary> {
     val userEntity = userRepository.findByDeliusUsername(userDistinguishedName)
       ?: return emptyList()
-
```
brownbearsoftware, line 73: Has something gone astray during a rebase here as this commit doesn't seem to do much?

brownbearsoftware (approved):
Some minor comments but overall looks good!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
Harriethw: good shout on handling the missing enum value 👍 

I don't _think_ we'll have the concept of 'premises' in CAS-2, at least for foreseeable future - a third party called NACRO will be handling everything after the apply stage, we've got no plans to show premises at the moment.

brownbearsoftware:
On Json Schemas:

Initially we had planned to keep a full JSON schema for Applications/Assessments etc. and validate against it at submission time.  We moved away from this idea because we realised incompattible changes to the frontend would potentially result in the frontend breaking for an Application before submission point.

We intend to remove the JSON schema checking but for the time being are just using an empty JSON schema that any valid JSON will work with.

Instead we now do a version check on the schema that an Application was created with versus the latest one - if they are not the same then an Application cannot be edited anymore and the Application must be restarted.  We suspect this will be infrequent post-launch so the potential user disruption is an accepted risk.  This is why we also have the `document` field which is a denormalised question-answer format which the frontend will always be able to render.

This version checking is why I'd recommend you create your own JSON schema for Applications as part of this PR as if CAS2 decide to bump their version because they've made incompattible changes in the frontend then all of your Applications will be classed as out of date even if they aren't.

brownbearsoftware:
> what is the validationErrors.any() check doing? is that if there are any errors on the ApplicationEntity, based on what fields?

In `ValidationErrors.kt` there's a few bits which make up a DSL around validating things in the service layer.  Inside of a `validated<T> { .. }` block, doing something like:

```
"$.propertyName" hasValidationError "empty"
```

will cause `validationErrors.any()` to evaluate to false.  In this case you could actually remove that check as it will never be true.

brownbearsoftware:
> there is a check for offenderDetails.otherIds.nomsNumber == null - how would we recreate that in a unit test?

Something like:

```
val offenderDetails = OffenderDetailsSummaryFactory()
   .withCrn(crn)
   .withNomsNumber(null)
   .produce()

every { mockOffenderService.getOffenderByCrn(crn, username) } returns AuthorisableActionResult.Success(
      offenderDetails,
    )
```

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ApplicationService.kt
patrickjfl: yes good point, these changes ended up getting rolled into the first commit

PR #673, opened 2023-05-31
Enable auto seeding of premises in test

pezholio:
We need to seed them in the test environment too


helm_deploy/values-test.yaml
```diff
@@ -25,6 +25,8 @@ generic-service:
     SENTRY_ENVIRONMENT: test
     CACHES_STAFFMEMBERS_EXPIRY-SECONDS: 20
     DOMAIN-EVENTS_EMIT-ENABLED: false
+    SEED_AUTO_ENABLED: true
+    SEED_AUTO_FILE-PREFIXES: classpath:db/seed/local+dev
```
brownbearsoftware, line 29: Perhaps we should rename this directory to `local+dev+test` or it might trip someone up later

brownbearsoftware (approved):

PR #675, opened 2023-05-31
Run tests in CircleCI

pezholio:
This moves the running of the tests to CircleCI with seperate build, lint and test jobs. The build step builds the code, and persists the built code to a workspace, which is then used by the link and test jobs. The test job uses [parellelism](https://circleci.com/docs/parallelism-faster-jobs/) to split the tests and run them over four seperate containers, speeding up the build.

On initial inspection, this seems to shave at least 5 minutes off the total build time!


pezholio:
This feels like a 🐰 🕳️ . Parking for now

brownbearsoftware (approved):

PR #676, opened 2023-05-31
183 - Booking Made Notify Emails

brownbearsoftware:
Send an email to the probation practitioner that created the original Application when a Booking is created


pezholio (approved):
LGTM 👍

PR #679, opened 2023-06-01
Add turnarounds to Temporary Accommodation bookings where the turnaround is missing

danlivings-dxw:
The work to implement automatic creation of turnarounds was split over two PRs: #607 and #611. These PRs were merged and deployed about three working days apart. The first PR backfilled the `turnarounds` table for all existing bookings, but before the second PR any new bookings would not have a turnaround automatically created.

This is accounted for in the API, which treats a missing turnaround and a zero-day turnaround as equivalent. However, an [error was reported in Sentry](https://ministryofjustice.sentry.io/issues/4218927270/?project=4504129156218880) for the frontend, which received a booking with a `"turnaround": null` property.

An investigation determined that [this issue only applies to a single booking](https://mojdt.slack.com/archives/C048BJS7S2F/p1685550709483009), but this PR performs the migration in a generic way.


tahb (approved):
LGTM

PR #680, opened 2023-06-01
66 - Seed offline applications

brownbearsoftware:
- Remove `submittedAt` from OfflineApplications as we won't actually have this data for legacy applications
- Add a Seed Job that creates an Offline Application for each CRN in a CSV

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/ApprovedPremisesOfflineApplicationsSeedJob.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/ApprovedPremisesOfflineApplicationsSeedJob.kt
@@ -1,0 +1,55 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.seed
+
+import org.slf4j.LoggerFactory
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ServiceName
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.OfflineApplicationEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.OfflineApplicationRepository
+import java.time.OffsetDateTime
+import java.util.UUID
+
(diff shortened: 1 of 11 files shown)


brownbearsoftware:
[Seeding existing bookings](https://trello.com/c/x7XayjUT/66-seeding-existing-bookings)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/ApprovedPremisesOfflineApplicationsSeedJob.kt
```diff
@@ -0,0 +1,55 @@
...
+
+  override fun deserializeRow(columns: Map<String, String>) = OfflineApplicationsSeedCsvRow(
+    crn = columns["crn"]!!,
+  )
+
+  override fun processRow(row: OfflineApplicationsSeedCsvRow) {
+    val existingOfflineApplications = offlineApplicationRepository.findAllByServiceAndCrn(ServiceName.approvedPremises.value, row.crn)
+
+    if (existingOfflineApplications.none()) {
+      offlineApplicationRepository.save(
```
pezholio, line 35: This is going to be potentially quite slow with a lot of CRNs, could we get all the CRNs with existing bookings with an `IN` query, diff the array of CRNs, create an array of `OfflineApplicationEntity`s and then save all in one?

Some psuedocode:

```
val crns = content_of_csv.map { it.crn }
val existingBookings = offlineApplicationRepository.findAllWhereCrnIn(crns)
val crnsWithExistingBookings = existingBookings.map { it.crn }
val crnsToCreateBookingsFor = crns - crnsWithExistingBookings
val applications = crnsToCreateBookingsFor.map {
  OfflineApplicationEntity(
    id = UUID.randomUUID(),
    crn = it,
    service = ServiceName.approvedPremises.value,
  )
}
offlineApplicationRepository.save(applications)
```

pezholio: Ah, course. Nah, I think this approach is fine in that case

PR #681, opened 2023-06-01
(173) Show Placement Applications as tasks

pezholio:
This allows Placement Applications to show up as tasks and be reallocated, in a similar way to assessments and placement requests. These were originally going to be called `PlacementRequestReview`s, but the thinking has evolved, so I've renamed them to `PlacementApplications`


pezholio:
[Show Placement Requests as Tasks](https://trello.com/c/hthY4oTV/173-show-placement-requests-as-tasks)

PR #687, opened 2023-06-02
(174) Add first class entities to placement applications

pezholio:
This adds some first-class entities to placement applications, so they are easier to fetch when listing / viewing / using to create a placement request. The first is a placement type, and the second is a many to one relation of PlacementDates. This will give us the flexibility for PPs to request mutliple placements as shown in the Figma design here:

V3_ Add another date

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PlacementDateEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PlacementDateEntity.kt
@@ -1,0 +1,37 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity
+
+import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.data.jpa.repository.Query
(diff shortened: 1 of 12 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PlacementDateEntity.kt
```diff
@@ -0,0 +1,36 @@
...
+  @Id
+  val id: UUID,
+
+  val createdAt: OffsetDateTime,
+
+  @OneToOne
+  @JoinColumn(name = "placement_application_id")
+  val placementApplication: PlacementApplicationEntity,
+
+  val expectedArrival: OffsetDateTime,
```
brownbearsoftware, line 33: Think this should be a `LocalDate` based on the postgres type being `DATE`?

pezholio: Oh, good point

brownbearsoftware (approved):

PR #690, opened 2023-06-05
(174) Allow a matcher to see in review placement applications

pezholio:
This updates the Tasks to return different values depending on the task type. It also updates the `/tasks` endpoint to return tasks that are allocated to the logged-in user. At the moment, this only works for matchers, but we can easily  extend this to return assessments, and any other type of tasks we choose to add in the future.

The existing `/tasks` endpoint has been moved to `/tasks/reallocatable`, which fetches all reallocatable tasks for a workflow manager. We'll need to update the UI to call this endpoint when this work is in.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/TaskTransformer.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/TaskTransformer.kt
@@ -18,6 +18,8 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.prisonsapi.InmateD
 class TaskTransformer(
   private val personTransformer: PersonTransformer,
   private val userTransformer: UserTransformer,
+  private val risksTransformer: RisksTransformer,
+  private val placementRequestTransformer: PlacementRequestTransformer,
 ) {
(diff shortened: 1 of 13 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/TaskTransformer.kt
```diff
@@ -45,14 +48,31 @@ class TaskTransformer(
...
+        expectedArrival = it.expectedArrival,
+        duration = it.duration,
+      )
+    },
+    releaseType = placementRequestTransformer.getReleaseType(placementApplication.application.releaseType)!!,
+    placementType = getPlacementType(placementApplication.placementType!!),
   )
 
+  private fun getPlacementType(placementType: JpaPlacementType): ApiPlacementType = when {
+    placementType === JpaPlacementType.ROTL -> ApiPlacementType.rotl
```
brownbearsoftware, line 70: Go for double equals over triple.  In this case though you could move placementType into the `when`:

```
when (placementType) {
   JpaPlacementType.ROTL -> ApiPlacementType.rotl
   etc.
}
```

pezholio: Argh, that's such a bad habit of mine! 😆 I blame Typescript

brownbearsoftware (approved):

PR #693, opened 2023-06-05
Introduce internal-only 'hard delete premises' API endpoint

danlivings-dxw:
> See [ticket #1186 on the CAS3 Trello board](https://trello.com/c/NpfQMjTV/1186-add-script-to-hard-delete-a-property).

This PR introduces the `DELETE /internal/premises/{premisesId}` endpoint that can be used to delete a premises.
This endpoint acts in a similar fashion to the `DELETE /internal/booking/{bookingId}` endpoint introduced in #437, in that:
- It can only be accessed from `localhost`, i.e. within the API's pod.
- It will remove the premises from the database instead of marking it as archived/deleted.
- It will remove entities that depend on the premises (i.e. rooms and beds) from the database.
- All deletes are contained within a single transaction, so if any dependent entity fails to delete, the premises will be restored to the exact state it was in prior to the API call.

In addition, it's important not to hard-delete premises where operational data would be lost, as this endpoint is only intended to be used when a premises is created in error. Therefore, the API will return a `409 Conflict` response if the premises has any bookings associated with it. If it's critically important that a premises with bookings must be deleted, this can be achieved by hard-deleting all associated bookings beforehand using `DELETE /internal/booking/{bookingId}`.

To use this endpoint in a convenient way, the `script/delete_booking` script has been genericised and renamed to `script/hard_delete`. It's now used like so:
```sh
# For a booking
script/hard_delete booking {booking id}
# For a premises
script/hard_delete premises {premises id}
```

The `doc/how-to/delete_booking_remotely.md` has been renamed and updated to reflect the change in usage of this script.


brownbearsoftware (approved):

doc/how-to/delete_booking_or_premises_remotely.md
```diff
@@ -0,0 +1,13 @@
+# How to delete a Booking or Premises remotely
```
tahb, line 1: I initially overlooked voids in this equation and if we should prevent delete the premises if they exist.

Thinking about it now I think think not. Given this script will be used in edge cases to catch admin errors and the check for bookings would be the crucial record of accommodation.

Perhaps worth noting the extra behaviour here in our docs though?

- If the premises has at least one booking, it won't delete
- If the premises has any amount of voids, they will be included in the deletion

tahb (approved):
Really easy to read 👍 LGTM

PR #695, opened 2023-06-06
66 - Bookings seed

brownbearsoftware:


pezholio (approved):
Looks really good, and a super-clear series of tests too! I wonder if we can try it out in dev with a corpus of data once this is deployed?

brownbearsoftware:
> I wonder if we can try it out in dev with a corpus of data once this is deployed?

Definitely 👍

PR #696, opened 2023-06-06
(175) Add an endpoint to approve a placement application

pezholio:
This adds a new endpoint to allow a matcher to approve a placement application. When it has been approved, we can then create a placement request and assign it to a matcher in the same way as when an Assessment with date information has been completed.


pezholio:
[Add an endpoint to approve a placement application](https://trello.com/c/suzaFi2d/175-add-an-endpoint-to-approve-a-placement-application)

src/main/resources/static/api.yml
```diff
@@ -5462,6 +5501,26 @@ components:
...
+      required:
+        - decision
+        - summaryOfChanges
+        - decisionSummary
+    PlacementApplicationDecision:
+      type: string
+      enum:
+        - accepted
+        - rejected
+        - withdraw
```
brownbearsoftware, line 5522: What's the difference between `withdrawn` and `withdrawn`?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/PlacementRequestService.kt
```diff
@@ -110,6 +115,28 @@ class PlacementRequestService(
...
   }
 
+  fun createPlacementRequestsFromPlacementApplication(placementApplicationEntity: PlacementApplicationEntity, notes: String?): AuthorisableActionResult<List<PlacementRequestEntity>> {
+    val placementRequirements = placementRequirementsRepository.findByApplication(
+      placementApplicationEntity.application,
+    ) ?: return AuthorisableActionResult.NotFound("Placement Requirements", placementApplicationEntity.application.id.toString())
+
+    val placementDateEntities = placementDateRepository.findAllByPlacementApplication(placementApplicationEntity)
+
+    if (placementDateEntities.size === 0) {
```
brownbearsoftware, line 125: Triple equals

brownbearsoftware: `.isEmpty()` perhaps more readable though

src/main/resources/static/api.yml
pezholio: Withdraw is if a new application is required, and withdrawn is if the PP has withdrawn the application. Will rename to be clearer 👍

brownbearsoftware (approved):

PR #701, opened 2023-06-07
212 - Allow Double Booking AP Beds / Automatically create Offline Applications

brownbearsoftware:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
```diff
@@ -229,11 +230,18 @@ class BookingService(
       val newestSubmittedOnlineApplication = applicationService.getApplicationsForCrn(crn, ServiceName.approvedPremises)
         .filter { it.submittedAt != null }
         .maxByOrNull { it.submittedAt!! }
-      val newestOfflineApplication = applicationService.getOfflineApplicationsForCrn(crn, ServiceName.approvedPremises)
+      var newestOfflineApplication = applicationService.getOfflineApplicationsForCrn(crn, ServiceName.approvedPremises)
         .maxByOrNull { it.createdAt }
 
       if (newestSubmittedOnlineApplication == null && newestOfflineApplication == null) {
-        validationErrors["$.crn"] = "doesNotHaveApplication"
+        newestOfflineApplication = applicationService.createOfflineApplication(
```
pezholio, line 237: Is this what we want to do? I forget now! 😆

brownbearsoftware: Yeah when we discussed yesterday in the huddle the takeaways were:
 - Offline Applications don't really do what we initially intended them to do (carry information about the legacy or offline process that's taken place) because we don't actually have the info to put in them
 - Removing Offline Applications is probably a bit too much trouble for right now (and this doesn't make it any harder to remove them if/when we do)
 - If we don't do this if anyone wants to make an ad-hoc Booking for a CRN that doesn't already have an Application they'll have to contact us to seed an Offline Application before they can do so which is going to be a bottleneck

pezholio: Cool, that makes sense 👍

danlivings-dxw (approved):

PR #708, opened 2023-06-08
213 - Return Bookings cancelled due to an appeal to matching

brownbearsoftware:
- Create a new Cancellation Reason "Booking successfully appealed"
 - When this is used to cancel an AP Booking which is linked to a Placement Request, create a new Placement Request with the same criteria


pezholio (approved):
LGTM - just a rogue line of code that's snuck into the wrong commit

PR #711, opened 2023-06-12
66 - Add artificial delay to Booking seeding

brownbearsoftware:
As we do an upstream call to Community API to get Offender details it makes sense to throttle our requests to be neighbourly


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/ApprovedPremisesBookingSeedJob.kt
```diff
@@ -80,6 +80,8 @@ class ApprovedPremisesBookingSeedJob(
   )
 
   override fun processRow(row: ApprovedPremisesBookingSeedCsvRow) {
+    Thread.sleep(1_000)
```
pezholio, line 83: Just wondering, is there an endpoint to bulk-fetch CRNs from Delius via CommunityAPI?

brownbearsoftware: Doesn't look like it unfortunately: https://github.com/ministryofjustice/community-api/blob/main/src/main/java/uk/gov/justice/digital/delius/controller/secure/OffendersResource.java

pezholio: Oh boo - I might ask the question. Would help massively with our preemptive caching too

pezholio (approved):

PR #712, opened 2023-06-12
125 - Endpoint for calendar view

brownbearsoftware:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/CalendarTransformer.kt
```diff
@@ -0,0 +1,73 @@
...
+            length = it.startDate.getDaysUntilInclusive(it.endDate).size,
+            startDate = it.startDate,
+            endDate = it.endDate,
+          )
+        }
+      }.toMutableList()
+
+      var startOfOpenPeriod: LocalDate? = null
+      var endOfOpenPeriod: LocalDate? = null
+      startDate.getDaysUntilInclusive(endDate).forEach { dateInRange ->
```
pezholio, line 42: I'm finding this block of code a little hard to follow, I'm wondering if we can do it in a different way by getting all the dates for the booked periods (Not sure if this is valid Kotlin!):

```kotlin
val bookedDates = transformedSchedule.map { it.startDate.getDaysUntilInclusive(it.endDate) }
val entirePeriod = startDate.getDaysUntilInclusive(endDate)

// Get all the missing dates in the schedule
val openDates = entirePeriod.minus(bookedDates)

openDates.forEach { i, startDate -> 
  val endDate: LocalDate? = null

  // Loop through all the dates following the current date
  openDates.drop(i + 1).forEach { length, date -> 
    // If the date is not the next day, we set next day to the previous day in the loop
    if (endDate !== startDate.addDays(length + 1)) {
      endDate = openDates[i -1]
    }
  }

 transformedSchedule += BedOccupancyOpenEntry(
   type = BedOccupancyEntryType.open,
   length = startDate!!.getDaysUntilInclusive(endDate!!).size,
   startDate = startDate!!,
   endDate = endDate!!,
 )
}
```

pezholio (commented):
Looks great on the whole, just some suggestions to make the code a bit easier to grok

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/repository/CalendarRepository.kt
```diff
@@ -0,0 +1,130 @@
...
+      bookingsForPremisesQuery,
+      params,
+      ResultSetExtractor { resultSet ->
+        while (resultSet.next()) {
+          val bedKey = CalendarBedInfo(
+            bedId = UUID.fromString(resultSet.getString("bed_id")),
+            bedName = resultSet.getString("bed_name"),
+          )
+
+          ensureResultsMapContainsBed(bedKey)
```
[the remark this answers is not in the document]
brownbearsoftware: Unfortunately there's no ternary operator in Kotlin, closest would probably be:

```
resultsMap[bedKey] = if (resultsMap[bedKey] == null) resultsMap[bedKey]!! else mutableListOf()
```

But it doesn't feel particularly idiomatic - this did prompt me to remember that putIfAbsent is a thing though so I've updated to use that

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/CalendarTransformer.kt
brownbearsoftware: Wouldn't this add multiple entries for the same open period with the start date moving up by a day for each entry?

brownbearsoftware:
[API endpoint for calendar view](https://trello.com/c/lIBoyJVa/125-api-endpoint-for-calendar-view)

pezholio (approved):

PR #713, opened 2023-06-13
Notify content changes

brownbearsoftware:
**Add variables to existing templates:**

- crn for all templates
- startDate/endDate for Booking Made template

**Add redundant CAS1-only checks around sending emails because copy is AP specific.**

Most of these are at call-sites that can only currently be hit for AP Applications but this will add another layer of protection against future changes to allow CAS2/3 to use these inadvertedly sending AP-specific emails to CAS2/3 users.

**Enable Notify in dev**

Using the `TEST_AND_GUEST_LIST` will send real emails to people on the Notify Guest List (and in the `notify_guest_list_users` table our side) otherwise it will not send.


pezholio (approved):
LGTM 👍

brownbearsoftware:
@pezholio added some changes to the Booking Made event so requested another look

pezholio (approved):

PR #716, opened 2023-06-14
Fix missing Notify dependency

brownbearsoftware:


pezholio:
@brownbearsoftware This can be closed now right?

PR #722, opened 2023-06-14
Prevent loss of data between restarts in local environment

danlivings-dxw:
> See [ticket #1164 on the CAS3 Trello board](https://trello.com/c/Vii8PQLT/1164-retro-action-prevent-resetting-local-database-between-restarts).

This PR is intended to resolve an issue identified in a CAS3 retrospective where, when running the API locally or within the `ap-tools` utility, data added or modified is not preserved between restarts. This is caused by some repeatable Flyway migrations in the `local+dev+test` folder containing `TRUNCATE TABLE` statements.

To resolve this, the repeatable migrations containing `TRUNCATE TABLE` statements have been moved to a new `dev+test` folder to preserve the behaviour for the dev and test environments, where it is desired. Similar migrations (without the `TRUNCATE TABLE`) statements have been created in a `local` folder to yield a similar initial set of data upon first run of the API application. If restoring the local database to a known state is desired, the `TRUNCATE TABLE` SQL statements can be executed locally using a tool such as DBeaver, or the Docker volume for the database can be destroyed and recreated.

The application configurations for each of the three affected environments have been updated to ensure that the application locates migrations from the appropriate new folder for its environment.


brownbearsoftware (approved):

pezholio:
Thanks for doing this - it was a pain locally!

PR #725, opened 2023-06-15
Correct recipient of deallocated emails

brownbearsoftware:
These should go to the person from whom the Assessment was deallocated, not the new assignee


brownbearsoftware:
[Gov.Notify technical work](https://trello.com/c/2NtmPNON/183-govnotify-technical-work)

danlivings-dxw (approved):

PR #728, opened 2023-06-16
Add bed and room name to lost bed response

pezholio:
This adds a couple of extra attributes to the API response for lost beds.


Harriethw (approved):
LGTM! thanks

PR #729, opened 2023-06-16
(214) Return cancellations when requesting a placement request

pezholio:
This allows us to fetch any previous cancellations for a placement request's application, so matchers can be made aware of any previous cancellations that have happened, together with the reasoning.


brownbearsoftware (approved):
LGTM.

I would perhaps argue the controller should get the cancellations in a seperate call to the service layer instead of returning a Pair from `getPlacementRequestForUser` but that's more of a stylistic choice.

PR #730, opened 2023-06-16
Return Assessment decision with the application

pezholio:
This allows us to quickly get the decision of an assessment alongside an application


pezholio:
Yeah, they return null by default

PR #719, opened 2023-06-14
Notify content changes - fix

brownbearsoftware:

PR #731, opened 2023-06-16
ADR for internal service naming in favour of `CASX`

tahb:
As promised during a CAS API check-in this change documents the decision to move towards the naming convention of CASX within the codebase.

Rather than Approved Premises and Temporary Accommodation, we change course towards CAS1, CAS2 and CAS3 for internal identification.

### DIFF
--- a/doc/architecture/decisions/0013-transition-internal-service-naming-towards-casx.md
+++ b/doc/architecture/decisions/0013-transition-internal-service-naming-towards-casx.md
@@ -1,0 +1,60 @@
+# 13. transition-internal-service-naming-towards-casx
+
+Date: 2023-06-16
+
+## Status
+
+Accepted
+
+## Context
+
+- Approved Premises (CAS1) was the first project and set their Frontend and API
+  up accordingly
+- Temporary Accommodation (CAS3) joined the space and became the second user of
+  the API
+  - During this transition the team has carried various changes to support
+      different services. In doing so introduced "temporary-accommodation" as
+      the language alongside "approved-premises":
+    - X-Service-Name header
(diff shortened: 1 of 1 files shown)


bryangaledxw:
Is it worth any reference to whether we call this `type`, `service`, or `serviceName` in this ADR? All three our present in our OpenAPI spec

tahb:
Good suggestion thank you. 

`type` is too generic so I'd vote for either `service` or `serviceName`. It looks like they are both used about the same amount.

We do also have the `X-Service-Name` header. 

I don't have a strong opinion but `serviceName` seems like a good choice, what do we think? I'll write it in and if the consensus is something else we can switch it.

pezholio (approved):
Good move - this has tripped me up a couple of times, so would be good to be consistent!

tahb:
One for your eyeballs too @danlivings-dxw @brownbearsoftware before it goes in.

danlivings-dxw (approved):
+1 for using `serviceName` consistently as a discriminator between the different services.

Also, moving to `CAS1`/`CAS2`/`CAS3` as accepted values should clear up situations where the API returns JSON with duplicate keys but different values. I've seen examples of responses like this before:
```
{
    // ...
    "service": "CAS3",
    "service": "temporary-accommodation"
}
```

brownbearsoftware:
> where the API returns JSON with duplicate keys but different values. I've seen examples of responses like this before:
@danlivings-dxw I fixed this (in a pretty horrible way) recently: https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/719/commits/698a2d352897e7a84006e5cfeaf71a4a93df5e07

PR #737, opened 2023-06-19
(3) API endpoint to move beds

pezholio:
This adds an endpoint to move a booking from one bed to another, together with any notes. On submit, the booking's bed is changed, as well as creating a new `BedMoveEntity` to record the change. We don't do anything with this entity yet, but we may do in future.


pezholio:
[Moving beds API](https://trello.com/c/I3MepePw/3-moving-beds-api)

brownbearsoftware (approved):
LGTM

PR #738, opened 2023-06-20
Return 500s when no NOMS Number or OASys present during Application creation

brownbearsoftware:
Return a 500 with detail of `No nomsNumber present for CRN` or `No OASys present for CRN` to be consistent with other endpoints to allow frontend to handle more elegantly


brownbearsoftware:
[Return error at create Application when OASys does not exist](https://trello.com/c/S2tPkxq7/248-return-error-at-create-application-when-oasys-does-not-exist)

tahb:
Checking for critical downstream info early is perfect. I guess there's a frontend side of this work to use the error to steer users into creating an application by other means or getting that missing record created?

brownbearsoftware:
> I guess there's a frontend side of this work to use the error to steer users into creating an application by other means or getting that missing record created?

@tahb that's correct, it'll be the latter in this case I believe.  Adam is looking into whether we'll need to support people without Prison API entries for launch or not which if so will be a bigger piece of work.

pezholio (approved):
LGTM - I'll create a frontend ticket to make sure we handle these errors. Also might need @selina-dxw to come up with some copy so users know what's going on

PR #739, opened 2023-06-21
Fix for PostGIS Point missing

brownbearsoftware:
This fixes an issue where updating an existing Approved Premises via seeding did not update the `latitude`, `longtitude` and `point` properties


brownbearsoftware:
[Updating Premises via seeding does not update PostGIS point](https://trello.com/c/DGeHmclu/250-updating-premises-via-seeding-does-not-update-postgis-point)

pezholio (approved):

PR #742, opened 2023-06-22
Add roles for JimSnowLdap

bryangaledxw:
We seed roles for our local development user, to allow us to use the CAS3 Apply feature end-to-end during local development


danlivings-dxw (approved):
LGTM, a test has failed but it's one that CircleCI has identified as flaky so it should be good to go once the pipeline has been rerun

PR #743, opened 2023-06-22
256 - Make oays requests concurrently

brownbearsoftware:
The eventual upstream (OASys) can be slow to respond, to reduce the risk of bumping into the 10s timeout between our frontend and API we can fetch the sections concurrently.  We're using coroutines dispatched to a blocking IO thread pool to do this to avoid refactoring the entire client layer to be async.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -15,6 +15,7 @@ val springDocVersion = "1.7.0"
 val sentryVersion = "6.23.0"
 
 dependencies {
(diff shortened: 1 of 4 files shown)


brownbearsoftware:
[Make OASys calls concurrently for sections endpoint](https://trello.com/c/sAdW02pC/256-make-oasys-calls-concurrently-for-sections-endpoint)

pezholio (approved):

PR #744, opened 2023-06-22
Introduce Gatling performance tests

danlivings-dxw:
> See [ticket #1244 on the CAS3 Trello board](https://trello.com/c/RVHj2JFc/1244-can-we-spare-1-day-to-set-up-some-load-testing-on-apply-to-de-risk-go-live-for-both-us-and-cas1).

This PR provides a prototype for using Gatling as a tool to run performance tests against the API.
This will help us to identify parts of our application that regress in performance or are vulnerable to high load, which will give us confidence that it will be able to handle the increased traffic from onboarding new groups of users.

Specifically, this PR runs a simulation of the Apply journey in Temporary Accommodation. Each simulated user will perform the following steps:
1. Create a new application.
2. Update the application a random number of times.
3. Submit the application.

Between each step is a pause to simulate user thinking/typing time, although this has been deliberately made small to emulate adverse conditions.

Similarly, the user populations have been chosen to represent adverse conditions:
- For the first two minutes, users begin the simulation at random intervals, at an overall rate of 50 users per second.
- For an additional minute, the number of users added ramps up to a peak value of 5000 concurrent users.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -5,6 +5,7 @@ plugins {
   kotlin("plugin.spring") version "1.8.22"
   id("org.openapi.generator") version "5.4.0"
   id("org.jetbrains.kotlin.plugin.jpa") version "1.8.22"
+  id("io.gatling.gradle") version "3.9.5"
 }
 
 configurations {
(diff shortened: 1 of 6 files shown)


adamhughes86 (commented):
Looks good to me

If @brownbearsoftware is happy with it I'm good

brownbearsoftware (approved):

PR #751, opened 2023-06-26
255 - LAO and auth changes

brownbearsoftware:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
```diff
@@ -45,7 +47,36 @@ class UserService(
...
+  }
+
+  fun getUserForPlacementApplicationAllocation(crn: String): UserEntity {
+    val unsuitableUsers = mutableListOf<UUID>(UUID.randomUUID())
+
+    while (true) {
+      val potentialUser = userRepository.findQualifiedMatcherWithLeastPendingPlacementApplications(unsuitableUsers)
+        ?: throw RuntimeException("No Users that pass LAO could be found")
+
+      if (offenderService.getOffenderByCrn(crn, potentialUser.deliusUsername) !is AuthorisableActionResult.Success) {
```
pezholio, line 76: There's potentially going to be a lot of upstream calls here, not the end of the world given we're racing to a deadline, but might be worth putting a `TODO` comment here to figure out a tidier way of getting users who can access an LAO?

pezholio (approved):
LGTM - one thing I did think - how are we caching CRNs? Do we cache on a user-by-user basis, or do we cache just on a CRN basis? Just concerned about the case if a user has access for a LAO that it'll get cached for users who don't? Also vice versa, if a user does not have access to an LAO, it'll get cached even for users who do. Maybe the solution is to never cache LAOs, as they'll hopefully be in the minority.

brownbearsoftware:
> LGTM - one thing I did think - how are we caching CRNs? Do we cache on a user-by-user basis, or do we cache just on a CRN basis? Just concerned about the case if a user has access for a LAO that it'll get cached for users who don't? Also vice versa, if a user does not have access to an LAO, it'll get cached even for users who do. Maybe the solution is to never cache LAOs, as they'll hopefully be in the minority.

@pezholio we cache the response from `{community api}/secure/offenders/crn/{crn}` which we make using a client credentials JWT rather than the requesting users so Community API doesn't apply the LAO for us - we only store one copy of this in the cache because it's not scoped to a particular user.  Then in `OffenderService::getOffenderByCrn` we take that response (which is probably served from the cache because of the preemptive caching except for when starting a new application or creating an ad-hoc booking) and look at its `currentExclusion` and `currentRestriction` properties which tell us whether we need to do an LAO check.  If we do we call `{community api}/secure/offenders/crn/{crn}/user/{username}/userAccess` which is cached using the normal Spring Cacheable annotation - this uses the method arguments as the cache key which contains the Delius Username so this cache is scoped to each individual user meaning the LAO from earlier users acccessing a CRN won't affect a later user accessing the same CRN.

pezholio:
Cool - was hoping that was the case! 👍

PR #753, opened 2023-06-26
270 - Fix for In Community release type breaking Tasks

brownbearsoftware:
Value actually being stored in the database is `inCommunity` but the transformer was looking for `in_community`


brownbearsoftware:
["In Community" release type breaks Task transformation](https://trello.com/c/zLtQpFLY/270-in-community-release-type-breaks-task-transformation)

pezholio (approved):

pezholio:
This bugged me for the longest time! I assumed it was something to do with my local data / envrionment, so kept changing it back manually in the database, and never got round to investigating properly.

PR #757, opened 2023-06-27
(268) Fix placement application bug

pezholio:
Reallocating a task according to the application ID caused some issues, because applications can have multiple “active” placement applications, so this was raising an error. Rather than have a special case for placement applications, it makes sense to have all tasks act on their own IDs, rather than relying on the applicationID being present. This also involves so changes to the URLs that the UI causes, so this is a breaking UI change, but I'll be working on this in parallel.


pezholio:
[Fix Placement Application bug](https://trello.com/c/kyUCbi82/268-fix-placement-application-bug)

brownbearsoftware (approved):

PR #762, opened 2023-06-28
Improve error message when allocation fails

pezholio:
This should hopefully make debugging easier.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
```diff
@@ -57,8 +57,14 @@ class UserService(
     val qualifications = application.getRequiredQualifications()
 
     while (true) {
+      var errorMessage: String
```
brownbearsoftware, line 60: Might be worth putting the user who created the application too to save us having to go into the DB to figure it out?  `application.createdByUser.deliusUsername`

PR #764, opened 2023-06-28
Add request information to MDC

danlivings-dxw:
This adds the following to the MDC for a given request:
- `request.id`: a random per-request UUID that can be used to group all log messages for a given request.
- `request.uri`: the requested URI, e.g. `/premises/a93602aa-4787-4aa4-a488-a22ba01149fb`.
- `request.method`: the request method, e.g. `GET`.
- `request.pathPattern`: the path pattern that best matches the requested URI in the format that developers talk about it, e.g. `/premises/{premisesId}`.
- `request.params`: all path parameters and query parameters are added as subkeys. For example, the above value of `request.uri` would have a corresponding `request.params.premisesId` MDC key with a value of `a93602aa-4787-4aa4-a488-a22ba01149fb`.
- `request.serviceName`: the service the request is for, derived from the `X-Service-Name` header. If this header does not have a value, it will have a value of `Not specified`.
- `request.user`: the Delius username of the user making the request, e.g. `JIMSNOWLDAP`. If the user could not be retrieved, this will have a value of `Anonymous`.

These values in the MDC will be captured by Application Insights and stored as custom dimensions for log messages.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/MDCHandlerInterceptor.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/MDCHandlerInterceptor.kt
@@ -1,0 +1,52 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.config
+
+import org.slf4j.MDC
+import org.springframework.stereotype.Component
+import org.springframework.web.servlet.HandlerInterceptor
+import org.springframework.web.servlet.HandlerMapping
+import org.springframework.web.servlet.config.annotation.InterceptorRegistry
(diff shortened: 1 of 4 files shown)


danlivings-dxw:
@brownbearsoftware it's worth noting that I don't have access to Application Insights, so while I'm pretty confident that MDC contents will be picked up by Application Insights, it would be a good idea to confirm this in the dev environment once this is deployed.

brownbearsoftware:
Looks great, will this break for endpoints that don't require authentication though (e.g. health)?

danlivings-dxw:
@brownbearsoftware good spot, I've changed it so that in those cases it sets `request.user` to `Anonymous`.

brownbearsoftware (approved):

PR #766, opened 2023-06-29
272 - Allocation and LAO changes

brownbearsoftware:
**Add `isEmergencyApplication` and `isEsapApplication` properties when submitting an Application.**

These are stored in the approved_premises_applications table.

**Add new roles & qualifications**

New qualifications:
`EMERGENCY` and `ESAP` are used to determine who is allocated these types of Assessment
`LAO` is used to determine who Assessments, Placement Requests and Placement Applications for LAO Offenders are allocated to.  This is also used to not enforce the LAO check when retrieving Assessments, Placement Requests and Placement Applications.  i.e. a user does not need to pass the LAO check (both inclusion and exclusion) if they have this qualification in order to access these types of entity.  Additionally, this qualification is now used to build the list of possible users to reallocate tasks to as well as performing the LAO check so the available users are everyone with relevant role/assessment type qualifications AND (this qualification or passes LAO).

New roles: `CAS1_EXCLUDED_FROM_ASSESS_ALLOCATION`, `CAS1_EXCLUDED_FROM_MATCH_ALLOCATION`, `CAS1_EXCLUDED_FROM_PLACEMENT_APPLICATION_ALLOCATION` are used to exclude people from automatic allocation for the three types of task

**Users with qualifications not allocated Assessments that do not require qualifications**
Users with one or more of `LAO`, `PIPE`, `EMERGENCY`, `ESAP` qualifications will not be allocated Assessments that do not require any qualifications.  These will only be allocated to people with the ASSESSOR role but no qualifications.

This is not the case for placement request and placement application allocation - these continue to go to anyone with the MATCHER role but with the added caveat that the LAO qualification must be present for requests/applications relating to an LAO Offender


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/OffenderService.kt
```diff
@@ -114,6 +114,21 @@ class OffenderService(
     return AuthorisableActionResult.Success(offender)
   }
 
+  fun isLao(crn: String): Boolean {
```
pezholio, line 119: 👍

pezholio (approved):
There's a lot going on here, but it's looking really good - happy for this to go in 👍

PR #775, opened 2023-06-30
276 - Application auth changes

brownbearsoftware:
- Create Application: if you pass LAO check
- Application appears in the list view (i.e. /applications in the UI): only if you created it
- Can access Application by its ID (i.e. through direct link to it /applications/{id}): if you pass LAO check or have the LAO qualification

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ApplicationsController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ApplicationsController.kt
@@ -106,7 +106,7 @@ class ApplicationsController(
     val deliusPrincipal = httpAuthService.getDeliusPrincipalOrThrow()
     val user = userService.getUserForRequest()
 
-    val (offender, inmate) = getPersonDetail(body.crn)
+    val (offender, inmate) = getPersonDetail(body.crn, true)
 
     val applicationResult = when (xServiceName ?: ServiceName.approvedPremises) {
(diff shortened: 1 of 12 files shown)


brownbearsoftware:
[Remove caseload check in API/Application Auth changes](https://trello.com/c/Tyo68ATu/276-remove-caseload-check-in-api-application-auth-changes)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ApplicationsController.kt
```diff
@@ -243,10 +243,16 @@ class ApplicationsController(
...
-  private fun getPersonDetail(crn: String): Pair<OffenderDetailSummary, InmateDetail> {
+  private fun getPersonDetail(crn: String, ignoreLaoQualification: Boolean = false): Pair<OffenderDetailSummary, InmateDetail> {
     val user = userService.getUserForRequest()
 
-    val offenderResult = offenderService.getOffenderByCrn(crn, user.deliusUsername, user.hasQualification(UserQualification.LAO))
+    val ignoreLao = if (ignoreLaoQualification) {
+      false
+    } else {
+      user.hasQualification(UserQualification.LAO)
+    }
```
danlivings-dxw, line 253: I found it a bit difficult to understand what the logic was here. I think renaming `ignoreLaoQualification` to something like `forceLaoChecks` would make it clearer what the intent is.

danlivings-dxw (approved):

PR #785, opened 2023-07-07
Hide bookings for LAO offenders from bedspace search endpoint when user does not have LAO access

danlivings-dxw:
> See [ticket #1334 on the CAS3 Trello board](https://trello.com/c/DBJ35e6B/1334-find-a-booking-journey-doesnt-log-the-user-out-if-a-booking-for-a-lao-exists).

Currently, if a user does not have LAO access, when they search for bookings the API will return a `403 Forbidden` response if one of the results is for a LAO offender. This will have the effect of logging the user out of the application on the frontend, making the booking search inaccessible.

Instead, these bookings should be filtered out when the API computes the response so that the user can successfully see bookings for which they have permissions.


tahb (approved):
LGTM!

PR #787, opened 2023-07-10
Handle non-arrivals

pezholio:
Currently, if a booking is recorded as non-arrived, then it still shows in the calendar and blocks the bed for future bookings. This should fix the underlying issue.


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/CalendarRepositoryTest.kt
```diff
@@ -133,6 +133,57 @@ class CalendarRepositoryTest : IntegrationTestBase() {
     assertThat(result[expectedBedKey]).isEmpty()
   }
 
+  @Test
+  fun `Results are correct for a Premises with a Room & Bed and a non-arrived lost bed and booking`() {
```
brownbearsoftware, line 137: Should this read:

`Results are correct for a Premises with a Room & Bed and a lost bed and non-arrived booking`

brownbearsoftware (approved):

PR #788, opened 2023-07-10
1547 - withdraw application

brownbearsoftware:
Allow withdrawin an Application by setting its `isWithdrawn` property to true when updating it.  This then changes the status to `withdrawn` which the FE can use to filter out withdrawn Applications.


brownbearsoftware:
[Withdraw Application](https://trello.com/c/HeRfyOs0/1547-withdraw-application)

pezholio (commented):
I think this is a good start, but at the moment, we have to send the `data` payload with the update request, which doesn't seem right. I wonder whether a `applications/{applicationId}/withdraw` endpoint would be the right thing? 

I think users will also want withdrawn applications to be hidden on the `/applications` view

PR #791, opened 2023-07-10
1547 - Withdraw application endpoint

brownbearsoftware:
`PUT /applications/{id}/withdraw` sets a new `isWithdrawn` property on AP Applications.

Applications that are withdrawn do not appear in the list Application Summaries results.

### DIFF
--- a/src/main/resources/static/api.yml
+++ b/src/main/resources/static/api.yml
@@ -1717,6 +1717,35 @@ paths:
           $ref: '#/components/responses/403Response'
         500:
           $ref: '#/components/responses/500Response'
+  /applications/{applicationId}/withdrawal:
+    post:
+      tags:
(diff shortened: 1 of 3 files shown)


src/main/resources/static/api.yml
```diff
@@ -1717,6 +1717,35 @@ paths:
           $ref: '#/components/responses/403Response'
         500:
           $ref: '#/components/responses/500Response'
+  /applications/{applicationId}/withdraw:
+    put:
```
pezholio, line 1721: I think this should be a `POST` to `/applications/{applicationId}/withdrawl` because we're creating a withdrawl here

pezholio (approved):
LGTM, just one minor RESTy comment

PR #794, opened 2023-07-12
245 - Booking date change endpoint

brownbearsoftware:


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/BookingServiceTest.kt
```diff
@@ -3599,4 +3604,365 @@ class BookingServiceTest {
...
+        .withUnitTestControlTestProbationAreaAndLocalAuthority()
+        .produce()
+
+      val room = RoomEntityFactory()
+        .withPremises(premises)
+        .produce()
+
+      val bed = BedEntityFactory()
+        .withRoom(room)
+        .produce()
```
pezholio, line 3626: Some of these entities can probably be at the class level to DRY things up

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/transformer/DateChangeTransformerTest.kt
```diff
@@ -0,0 +1,68 @@
...
+        newArrivalDate = LocalDate.parse("2023-07-18"),
+        newDepartureDate = LocalDate.parse("2023-07-20"),
+        booking = booking,
+        changedByUser = user,
+      ),
+    )
+
+    assertThat(result).isEqualTo(
+      DateChange(
+        id = UUID.fromString("22931e1d-9a8c-4087-90d9-9fd7638837fa"),
```
pezholio, line 58: Could probably initialise the dateChangeEntity as a `val` here and assert that the ID is equalty to `dateChangeEntity.id` to avoid repeating ourselves

pezholio (approved):
LGTM, just a couple of DRYing up opportunities

PR #796, opened 2023-07-12
Logging changes

brownbearsoftware:
- Fix MDC issues
- Log 400/500 reasons to aid live issue investigations

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/ExceptionHandling.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/ExceptionHandling.kt
@@ -6,6 +6,7 @@ import com.fasterxml.jackson.databind.exc.MismatchedInputException
 import com.fasterxml.jackson.databind.node.ArrayNode
 import com.fasterxml.jackson.databind.node.ObjectNode
 import io.sentry.Sentry
+import org.slf4j.LoggerFactory
 import org.springframework.http.HttpStatus
 import org.springframework.http.ResponseEntity
 import org.springframework.http.converter.HttpMessageNotReadableException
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ApplicationsController.kt
```diff
@@ -108,6 +108,8 @@ class ApplicationsController(
     val deliusPrincipal = httpAuthService.getDeliusPrincipalOrThrow()
     val user = userService.getUserForRequest()
 
+    throw BadRequestProblem(mapOf("key" to "value", "otherkey" to "othervalue"), "some error detail")
```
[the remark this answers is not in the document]
brownbearsoftware: 😅 woops

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/ExceptionHandling.kt
```diff
@@ -130,6 +135,19 @@ class ExceptionHandling(
     return true
   }
 
+  override fun log(throwable: Throwable, problem: Problem, request: NativeWebRequest, status: HttpStatus) {
+    when (problem) {
+      is BadRequestProblem -> logBadRequestProblem(problem)
+      is InternalServerErrorProblem -> logInternalServerErrorProblem(problem)
```
pezholio, line 141: Should we be calling `logInternalServerErrorProblem` somewhere too? (like we do with `logBadRequestProblem` on line 115) Or is there some magic going on here?

brownbearsoftware: this `log` method is from the `ProblemHandling` interface and gets called whenever we throw a Problem from a controller.  We only call logBadRequestProblem explicitly from line 115 because this `log` function never gets called on that particular route since it never reaches a controller.

pezholio: Ah, gotcha

pezholio (approved):

PR #800, opened 2023-07-14
Feature/389 update cas2 application

Harriethw:


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ApplicationTest.kt
```diff
@@ -1442,6 +1443,67 @@ class ApplicationTest : IntegrationTestBase() {
     }
   }
 
+  @Test
+  fun `Update existing CAS2 application returns 200 with correct body`() {
+    `Given a User` { submittingUser, jwt ->
+      `Given a User` { userEntity, _ ->
```
edavey, line 1449: I don't think this second ``Given a User` { userEntity, _ ->` context is used or needed?

PR #803, opened 2023-07-14
Take current workload into account when allocating

pezholio:
The current allocations engine allocates assessments to users who have the lowest number of allocated, but incomplete assessments. This ends up penalising assessors who complete a large number of assessments quickly, meaning they keep getting assessments pile up.

This alters the query to take into account recent applications that have been assessed, so, for example, if two assessors have no pending assessments, but one assessor has recently completed an assessment, it will prioritise the one with no completed assessments in their workload.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/UserEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/UserEntity.kt
@@ -22,54 +22,105 @@ interface UserRepository : JpaRepository<UserEntity, UUID>, JpaSpecificationExec
 
   @Query(
     """
-    SELECT u.*, ura.*, uqa2.* 
+    SELECT u.*, ura.*, uqa2.*,
+     RANK() OVER (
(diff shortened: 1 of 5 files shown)


pezholio:
[Take into acount workload when allocating assessments](https://trello.com/c/DrtjYLiI/313-take-into-acount-workload-when-allocating-assessments)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/UserEntity.kt
brownbearsoftware, line 1: Seems sensible enough - think we should rename the function too as it no longer reflects what the query does - perhaps:

`findQualifiedAssessorWithLeastPendingOrCompletedInLastWeekAssessments`

Do we also need to make this change for the other two allocation queries in this file?:
- findQualifiedMatcherWithLeastPendingPlacementApplications
- findQualifiedMatcherWithLeastPendingPlacementRequests

brownbearsoftware (approved):

brownbearsoftware (approved):

PR #804, opened 2023-07-17
309 - Return less data about a Person in accordance with LAO

brownbearsoftware:
This replaces the `Person` response type with `PersonInfo` which has two subtypes.  If the requesting user is allowed to see everything about the person in question a `FullPersonInfo` is returned which has all of the same fields as the original `Person` model.  If the requesting user is not allowed to see everything about the person in question a `RestrictedPersonInfo` is returned which only contains the CRN.


brownbearsoftware:
[Cannot list bookings from Premises with LAO Offenders 🐞](https://trello.com/c/IdrDyeuM/309-cannot-list-bookings-from-premises-with-lao-offenders-%F0%9F%90%9E)

brownbearsoftware:
[Cannot list bookings from Premises with LAO Offenders 🐞](https://trello.com/c/IdrDyeuM/309-cannot-list-bookings-from-premises-with-lao-offenders-%F0%9F%90%9E)

pezholio (commented):
This all looks pretty good. Before I give it the OK, I'm going to do a bit of a spike into how easy this will be to implement on the frontend. I've got a fairly good idea of how to implement it in a way that will be portable over the CAS3 too.

src/main/resources/static/api.yml
```diff
@@ -3740,42 +3706,59 @@ components:
...
-          type: string
-        status:
-          type: string
-          enum:
-            - InCustody
-            - InCommunity
-            - Unknown
-        prisonName:
+        type:
           type: string
```
pezholio, line 3715: Can this be an enum?

src/main/resources/static/api.yml
```diff
@@ -3740,42 +3706,62 @@ components:
             - extensions
             - departures
             - cancellations
-    Person:
+    PersonInfo:
```
pezholio, line 3709: Does this have to be named `PersonInfo`? Would make life easier in the frontend if we could just call it `Person` 😃

brownbearsoftware: Initially it was so I could change each usage of `Person` in api.yml in a discreet commit.  Have added another commit to rename it back to Person now.

pezholio: Noice one!

pezholio (approved):
LGTM - this is good to merge now https://github.com/ministryofjustice/hmpps-approved-premises-ui/pull/974 is in

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
```diff
@@ -267,17 +267,11 @@ class PremisesController(
...
-        val personDetails = getPersonDetailsForCrn(log, it.crn, user.deliusUsername, offenderService, user.hasQualification(UserQualification.LAO))
+        val personInfo = offenderService.getInfoForPerson(it.crn, user.deliusUsername, user.hasQualification(UserQualification.LAO))
 
-        if (personDetails == null) {
-          log.warn("Unable to get Person via crn: ${it.crn}")
-          return@mapNotNull null
-        }
-
-        val (offenderDetails, inmateDetails) = personDetails
+        if (personInfo !is PersonInfoResult.Success) return@mapNotNull null
```
tahb, line 272: When looking at a bedspace to make a new booking I think it's very important for the user to be able to see basic information about all bookings on that bed and in the rest of the property. 

By omitting this data for any reason we introduce the risk that a user makes an incompatible/unsafe booking because they incorrectly determine that there were no other bookings. When in fact we've hidden it from them because of an LAO or perhaps a connection problem to Delius.

I think we need to change this to always return a booking either in full (if the user has access) or the minimal and safe amount of info: CRN, Arrival date, Departure date, booking status.

tahb:
In an effort to test this locally without figuring out how to see CRNs that act like LAO I tried the following:

Replace the last[ two blocks in this method](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/804/files#diff-4a8c3767d589a6283081bfd38ef72be5ebc2dd5ec9d401127ded27adb5ec015cR446) with `return PersonInfoResult.Success.Restricted(crn, offender.otherIds.nomsNumber)`.

In CAS3 when looking at a bedspace I now get a few errors such as `Error: Invalid Date: undefined` which I'm working through. This may help others debug this locally @pezholio

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
brownbearsoftware: Sorry for the late response to this one, only just spotted the comment.  In the case where the person the Booking is for is LAO and the user doesn't pass the LAO check for them or have the LAO qualification then `personInfo` here will come from this return statement in OffenderService::getInfoForPerson: https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/804/files#diff-4a8c3767d589a6283081bfd38ef72be5ebc2dd5ec9d401127ded27adb5ec015cR429

This will be a `PersonInfoResult.Success.Restricted` which just contains the CRN and NOMS number - which I think is in line with what you're asking to happen?

There is the scenario where one of the upstream calls fails (e.g. the call to check the LAO access or get Offender info) in which case the Booking will just not be shown in the list which isn't great - do you think the preferable behaviour there would be to 500 rather than present partial data?

tahb: > There is the scenario where one of the upstream calls fails (e.g. the call to check the LAO access or get Offender info) in which case the Booking will just not be shown in the list which isn't great

This is my concern 👍 I do think it would be better to fail safely with a 500 rather than let a user make a new booking without a full picture of nearby bookings. If one of the missing bookings represents a person in residence that could cause a risk for a new booking then we have a safety problem.

tahb:
[CAS3 have now merged in some guards based off of this branch so we are compatible with this change.](https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/539)

PR #809, opened 2023-07-18
319 - Emit Booking cancelled domain event

brownbearsoftware:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
```diff
@@ -884,6 +888,67 @@ class BookingService(
       )
     }
 
+    if (booking.service == ServiceName.approvedPremises.value && booking.application != null && user != null) {
+      val dateTime = OffsetDateTime.now()
+
+      val domainEventId = UUID.randomUUID()
+
+      val offenderDetails = when (val offenderDetailsResult = offenderService.getOffenderByCrn(booking.crn, user.deliusUsername, user.hasQualification(UserQualification.LAO))) {
```
pezholio, line 896: Given this is going to an internal service, do we need to worry about if the user has the qualification here?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
```diff
@@ -884,6 +888,67 @@ class BookingService(
...
+        is AuthorisableActionResult.Success -> offenderDetailsResult.entity
+        is AuthorisableActionResult.Unauthorised -> throw RuntimeException("Unable to get Offender Details when creating Booking Cancelled Domain Event: Unauthorised")
+        is AuthorisableActionResult.NotFound -> throw RuntimeException("Unable to get Offender Details when creating Booking Cancelled Domain Event: Not Found")
```
pezholio, line 899: I wonder if we should just return null here if we can't access Delius (and maybe log upstream?), as it's only fetching the NOMS number, which we can get at a later date. Doesn't make sense to block a user action for something which is not necessarily their fault.

pezholio (approved):
LGTM - just a couple of things to think about (which I think affects other domain events too)

danlivings-dxw (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
brownbearsoftware: Good point, have updated this & added another commit to address for the existing events

PR #815, opened 2023-07-19
320 - Booking date changes domain event

brownbearsoftware:
Emit booking changed events when:
 - Extension is created
 - Date change is created


pezholio (approved):
LGTM - just one tiny suggestion

PR #817, opened 2023-07-19
(323) Allow a Booking to be created without a bedID

pezholio:
This updates the `placement-requests/$id/bookings` endpoint to accept either a bedId or premisesId. As we can’t easily do an either/or null in the OpenAPI spec, we check for the presence of either bedId or premisesId, and raise an error if neither is present.

Bed IDs are already optional, so this shouldn’t cause any breakages elsewhere. I’ve had a quick look around and everything seems to work as expected when a Bed ID is missing.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PlacementRequestsController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PlacementRequestsController.kt
@@ -95,6 +95,7 @@ class PlacementRequestsController(
       user = user,
       placementRequestId = id,
       bedId = newPlacementRequestBooking.bedId,
+      premisesId = newPlacementRequestBooking.premisesId,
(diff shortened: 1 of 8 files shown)


pezholio:
[Allow a booking to be created without a bed](https://trello.com/c/XdHQ8cCd/323-allow-a-booking-to-be-created-without-a-bed)

brownbearsoftware:
Will have a more thorough look over this tomorrow - one thing that comes to mind though is that `createExtension` and `createDateChange` both have the following check in which will need to be removed:

```
val bedId = booking.bed?.id
      ?: throw InternalServerErrorProblem("No bed ID present on Booking: ${booking.id}")
```

pezholio:
Yikes, good call!

brownbearsoftware (approved):
LGTM

Only other thing of note is these won't appear in the calendar view as-is but suppose this isn't a big concern at the moment

pezholio:
Yeah, that was my first concern, but I think at the mo, that's fine. I imagine we'll be pivoting to allocating rooms on arrival soon enough anyhoo

PR #818, opened 2023-07-19
318 - Application withdrawn reasons & emit application withdrawn domain event

brownbearsoftware:
Add Application Withdrawal Reasons and accept when withdrawing an Application.

This is optional for now to allow the frontend time to implement.

When present, also emit an application withdrawn domain event.  This part of the work was already reviewed in https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/819


brownbearsoftware:
[Application Withdrawn Domain Event](https://trello.com/c/XEoaYXch/318-application-withdrawn-domain-event)

src/main/resources/db/migration/all/20230719145544__add_application_withdrawn_reasons.sql
```diff
@@ -0,0 +1,18 @@
...
+    id UUID NOT NULL,
+    name TEXT NOT NULL,
+    is_active BOOL NOT NULL,
+    service_scope TEXT NOT NULL,
+    PRIMARY KEY (id),
+    UNIQUE (name)
+);
+
+INSERT INTO withdrawal_reasons (id, name, is_active, service_scope) VALUES
+    ('f41589c9-7af0-475c-822a-8ff6657ea59e', 'Alternative identified - placement no longer required', true, 'approved-premises'),
```
pezholio, line 11: I know we've done this elsewhere, but I'm not a huge fan of having withdrawl reasons in the database - we'll have to make an extra API call, and adding/removing reasons becomes a bit of a chore. Could be make them an enum instead? That way it's easier to keep in line with the frontend.

brownbearsoftware: Since we can only do enums of string values in OpenAPI are you thinking we just hardcode their proper names in the UI and have them like:

```
WithdrawalReason:
   type: string
   enum:
     - alternative_identified_placement_no_longer_required
     - etc.
```

Then hardcode the readable versions in the frontend?

Can do this change if you'd like but I think the database approach is a bit more future proof in the sense that if CAS3 decide to do the bit of work required for them to start using this too they'll need to put this back in the database to support the service scope stuff.

pezholio: Yeah, I guess so. It's just annoying that we have to make an additional API request to get the reasons.

pezholio: And having config data in the database makes my teeth itch.

brownbearsoftware: Have changed to use an enum

danlivings-dxw (approved):

PR #825, opened 2023-07-24
318 - Domain event feature flags

brownbearsoftware:
Setting `arrived-departed-domain-events-disabled` to true prevents Domain Events being emitted for:
 - Arrivals
 - Non Arrivals
 - Departures

Setting `manual-bookings-domain-events-disabled` to true prevents Domain Events being emitted for:
 - Booking Made
 - Booking Changed
 - Booking Cancelled
When the Booking has an Application associated with it but does not have a Placement Request associated with it (i.e. it is a manual Booking made after an Application was submitted for the CRN)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
```diff
@@ -1255,7 +1255,7 @@ class BookingService(
     booking.extensions.add(extension)
     updateBooking(booking)
 
-    if (booking.service == ServiceName.approvedPremises.value && booking.application != null && user != null) {
+    if (booking.service == ServiceName.approvedPremises.value && booking.application != null && user != null && (!manualBookingsDomainEventsDisabled || booking.placementRequest != null)) {
```
pezholio, line 1258: Could we break this conditional into a private function called maybe `shouldCreateDomainEventForBooking` or similar?

pezholio (approved):
Looks generally good, just a coupla minor suggestions 👍

PR #827, opened 2023-07-25
Increase memory in prod

brownbearsoftware:
Previously request of 512Mi, limit of 1Gi
Increased to request of 1Gi, limit of 1.5Gi
Also increase Xmx (max heap size) value to 1000m to allow the JVM to use this extra memory


adamhughes86 (approved):

pezholio (approved):
I feel like a lot of this is probably down to the fact that we have a lot of data now, so we really need to start thinking about pagination. Hopefully this will fix the issue for now though

PR #828, opened 2023-07-25
Add an isParole column to PlacementRequests

pezholio:
This will usually be `false`, but in the case of when a Request is created from a PlacementApplication that has a type of `RELEASE_FOLLOWING_DECISION`, we need to set `isParole` to true, so the assessors know that the startDate is malleable and can be within six weeks of the stated date.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/PlacementRequestService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/PlacementRequestService.kt
@@ -22,6 +22,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.PlacementRequ
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.PlacementRequestRepository
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.PlacementRequirementsEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.PlacementRequirementsRepository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.PlacementType
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserQualification
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserRole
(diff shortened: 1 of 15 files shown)


src/main/resources/db/migration/all/20230725101601__add_is_parole_to_placement_requests.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE placement_requests ADD COLUMN is_parole BOOLEAN NOT NULL DEFAULT FALSE;
```
brownbearsoftware, line 1: Minor point but setting a default in a migration can sometimes lead to unexpected behaviour further down the line if it's missed on an INSERT statement and the default value is inserted instead (which really was just supposed to be a `set this value for all existing rows` at the point of migration).  Prefer:

```
ALTER TABLE placement_requests ADD COLUMN is_parole BOOLEAN;
UPDATE placement_requests SET is_parole = FALSE;
ALTER TABLE placement_requests ALTER COLUMN is_parole SET NOT NULL;
```

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/PlacementRequestService.kt
```diff
@@ -128,13 +129,14 @@ class PlacementRequestService(
...
-      this.createPlacementRequest(placementRequirements, placementDates, notes)
+      val isParole = placementApplicationEntity.placementType == PlacementType.RELEASE_FOLLOWING_DECISION
+      this.createPlacementRequest(placementRequirements, placementDates, notes, isParole)
     }
 
     return AuthorisableActionResult.Success(placementRequests)
   }
 
-  fun createPlacementRequest(placementRequirements: PlacementRequirementsEntity, placementDates: PlacementDates, notes: String?): PlacementRequestEntity {
+  fun createPlacementRequest(placementRequirements: PlacementRequirementsEntity, placementDates: PlacementDates, notes: String?, isParole: Boolean?): PlacementRequestEntity {
```
brownbearsoftware, line 139: Does `isParole` actually need to be nullable?

brownbearsoftware (approved):

src/main/resources/db/migration/all/20230725101601__add_is_parole_to_placement_requests.sql
pezholio: Good call 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/PlacementRequestService.kt
pezholio: No, probably not now

PR #841, opened 2023-07-27
Add assessments to Temporary Accommodation applications

danlivings-dxw:
> See [ticket #1287 on the CAS3 Trello board](https://trello.com/c/72apWgsr/1287-the-frontend-can-get-applications-for-an-hpt).

This PR enables assessments to be used for the Temporary Accommodation service. Assessments for Temporary Accommodation have differences to those for Approved Premises, namely:
- They are not automatically allocated to an assessor, and must be marked as explicitly being in progress.
- The list of assessments visible to a user can be sorted in various ways, or filtered by the status.
- Temporary Accommodation assessments require a different set of statuses from Approved Premises assessments, but this has not been implemented within this PR (it will be implemented as part of the upcoming [ticket #1308](https://trello.com/c/X2ymPmWq/1308-application-state-can-be-updated-by-the-frontend)).

PR #849, opened 2023-07-31
Implement service-specific statuses for assessments

danlivings-dxw:
> See [ticket #1308 on the CAS3 Trello board](https://trello.com/c/X2ymPmWq/1308-application-state-can-be-updated-by-the-frontend).

This PR introduces a set of statuses specific to Temporary Accommodation assessments. This was mentioned but not implemented in #841 to keep the scope of that PR small; the described change is here.

Specifically, in Temporary Accommodation, the following states are required:
State transitions for Temporary Accommodation assessments

### DIFF
--- a/src/main/resources/static/api.yml
+++ b/src/main/resources/static/api.yml
@@ -5141,8 +5141,6 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/ClarificationNote'
-        status:
-          $ref: '#/components/schemas/AssessmentStatus'
       discriminator:
         propertyName: service
         mapping:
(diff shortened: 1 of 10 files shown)


src/main/resources/static/api.yml
```diff
@@ -5211,16 +5243,13 @@ components:
...
             allocatedToStaffMember:
               $ref: '#/components/schemas/TemporaryAccommodationUser'
+            status:
+              $ref: '#/components/schemas/TemporaryAccommodationAssessmentStatus'
           required:
             - application
     AssessmentSummary:
       type: object
       properties:
-        type:
```
brownbearsoftware, line 5219: I think you need to keep this along with the discriminator for the frontend's OpenAPI tooling to work properly

danlivings-dxw: I've restored this property, but I had to remove the `enum` part of it to get it to compile.

brownbearsoftware (approved):
LGTM - the actual JSON being returned should be compattible with what's deployed already but will need frontend changes to get the Typescript to compile again so maybe worth mentioning to the FE folk before merging?

PR #853, opened 2023-07-31
(371) Fix issue with lost bed dates in the calendar not always displaying correctly

pezholio:
We’ve noticed an issue where if a booking/lost bed finishes on the first day of the search period, it won’t show up. I’ve looked into this, and it seems that the Postgres `OVERLAPS` functionality is not inclusive. After a bit of Googling, it seems that using `tsrange` and the `&&` overlaps operator seems to do the trick.

I’ve also added a test to demonstrate and fix the regression.


pezholio:
[Fix issue with lost bed dates in the calendar not always displaying correctly](https://trello.com/c/R5TSTwce/371-fix-issue-with-lost-bed-dates-in-the-calendar-not-always-displaying-correctly)

brownbearsoftware (approved):

PR #855, opened 2023-08-01
Add an ADR for Pagination

pezholio:
The AP team are preparing to add pagination / filtering to an endpoint.  Ahead of this, I wanted to get everyone’s eyes on a possible approach, so we can make sure we’re all aligned.

### DIFF
--- a/doc/architecture/decisions/0014-paginating-large-data-sets.md
+++ b/doc/architecture/decisions/0014-paginating-large-data-sets.md
@@ -1,0 +1,161 @@
+# 14. Paginating large data sets
+
+Date: 2023-08-01
+
+## Status
+
+Accepted
+
+## Context
+
+As the service evolves, and more data is being added to the application, a lot of our "list-type"
+pages are becoming hard to navigate, as well as causing performance issues in the API and the 
+frontend.
+
+As such we need to start looking at using pagination to break up these datasets into smaller, 
+more manageable chunks, as well as returning metadata to communicate to the client how many 
+pages there are, what the current page is, and the number of the next page etc.
+
+Before we start this work, we should ensure that we agree a common approach to be rolled out
+across all endpoints and get agreement from everyone.
+
+## Decision
+
+We will implement "page-based pagination", starting first with the `/placement-requests/dashboard`
+endpoint. The page query will look as follows:
+
+```
+/placement-requests/dashboard?page=$pageNumber
+```
+
+Where `$pageNumber` is the page number the client wants to return (starting at 1). If the 
+`$pageNumber` is not present, the API returns all results. This can be tweaked as the 
+service evolve, and we have pagination on all endpoints.
+
+This will then be passed into the service to determine which page of results to return.
+
+We could possibly use a [`PagingAndSortingRepository`][1] for this.
+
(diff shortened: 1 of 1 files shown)


brownbearsoftware (commented):
Including the page info in response headers makes me a little bit uneasy but I can see the pragmatic side of this too.

I would suggest that maybe instead of the pageNumber query parameter being absent meaning page 1 it should mean return everything to preserve current behaviour or we'll encounter weird behaviour when trying to roll this out.  We could then change this to mean page 1 once the frontend is fully deployed.

richpjames (commented):
Sounds good, I don't really have any concerns but I have some questions to help my understanding: 
 1. Why go for headers over a query string? 
 2. What does returning the next page give the UI (presumably so we don't have to duplicate the `if (currentPage < totalPages) { currentPage + 1 } else { null }` logic? )

pezholio:
@brownbearsoftware 

> I would suggest that maybe instead of the pageNumber query parameter being absent meaning page 1 it should mean return everything to preserve current behaviour or we'll encounter weird behaviour when trying to roll this out. We could then change this to mean page 1 once the frontend is fully deployed.

Great idea!

doc/architecture/decisions/0014-paginating-large-data-sets.md
```diff
@@ -0,0 +1,129 @@
...
+## Decision
+
+We will implement "page-based pagination", starting first with the `/placement-requests/dashboard`
+endpoint. The page query will look as follows:
+
+```
+/placement-requests/dashboard?page=$pageNumber
+```
+
+Where `$pageNumber` is the page number the client wants to return. If the `$pageNumber` is not]
```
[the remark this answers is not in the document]
pezholio: I think it makes sense to be 1-indexed. Willl update the ADR

doc/architecture/decisions/0014-paginating-large-data-sets.md
```diff
@@ -0,0 +1,161 @@
...
+                type: integer
+              description: The total number of pages
+            X-Pagination-TotalResults:
+              schema:
+                type: integer
+              description: The total number of results
+            X-Pagination-PageSize:
+              schema:
+                type: integer
+              description: The total number of results
```
edavey, line 81: "The number of results per page" might be clearer / more accurate?

doc/architecture/decisions/0014-paginating-large-data-sets.md
```diff
@@ -0,0 +1,161 @@
...
+        401:
+          $ref: '#/components/responses/401Response'
+        403:
+          $ref: '#/components/responses/403Response'
+        500:
+          $ref: '#/components/responses/500Response'
+```
+
+Again, using the `PagingAndSortingRepository`, we can get this metadata from the response and return
+it alongside the actual data itself, perhaps using a `Pair`, e.g:
```
edavey, line 97: is it worth including a link to some docs on `Pair` for the benefit of Kotlin newbies? 

e.g. https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-pair/

edavey (approved):
Looks excellent, Stu thks

pezholio:
Closing this in favour of #859, which has the implementation too.

PR #860, opened 2023-08-02
Seed CAS3 bookings into the test environment for 2 regions

tahb:
Since turning off wiremock[1] we have to make sure we have seeded information. The previous seeding[2] contained a useful set of bookings in differing states. We reduce the amount of bookings in the same state to 1 and then tailor them further to relate them to our CAS3 premises in KSS and EofE[3].

We replace the CRN that would have been seeded by Wiremock and instead use the newly provisioned CRNs[4] to give us further variety. We have bookings for people with a single offence, multiple offence, missing oasys data, limited access offenders.

I've copied these migrations into my local seed directory and given it a run and it succeeds but because only JimSnow is available locally (and they don't belong to EofE or KSS) I can't eyeball the records within the service.

[1] https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/854
[2] https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/854/commits/234f235c8ca8cf60c711802e30a859b39e3d49a1#diff-9f4b088410aaca34010d146e0d1aee07bbc8ca7b69306600fe7ed82d8ad09095
[3] https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/b03b809d93d17cdbb872854ed21c6a947eff48f4/src/main/resources/db/migration/dev%2Btest/R__1_create_premises.sql#L10
[4] https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/4467949648/CRNs+for+Testing+in+Development+Environment


danlivings-dxw (approved):
LGTM

danlivings-dxw:
I'm not sure whether the ordering is stable with two `R__5` migrations but in this case I don't think it really matters.

PR #862, opened 2023-08-03
Implement API endpoints to change the state of a Temporary Accommodation assessment

danlivings-dxw:
> See [ticket #1308 on the CAS3 Trello board](https://trello.com/c/X2ymPmWq/1308-application-state-can-be-updated-by-the-frontend).

This PR modifies existing endpoints and introduces new endpoints to allow changing the state of a Temporary Accommodation assessment.

Specifically, the list of states a Temporary Accommodation assessment can be in and the corresponding API endpoint are as follows:
- `unallocated`: This is the default state when the assessment is created, but can be reached using the `DELETE /tasks/assessment/{assessmentId}/allocations` endpoint if a user needs to unallocate an assessment from themself.
- `rejected`: The `POST /assessments/{assessmentId}/rejection` endpoint is used to reject an application.
- `closed`: The `POST /assessments/{assessmentId}/closure` endpoint is used to close an assessment once an assessment has been successfully placed.
- `ready_to_place`: The `POST /assessments/{assessmentId}/acceptance` endpoint is used to accept an application and allow a booking to be made. This has been modified to make the `requirements` optional, as it is not currently needed for Temorary Accommodation. Validation logic has been added to enforce the field's non-nullability for other services.
- `in_review`: The `POST /tasks/assessment/{assessmentId}/allocations` endpoint is used to assign an assessment to the current user. This has been modified to make the request body optional, as in Temporary Accommodation the assessment should always be assigned to the current user when they mark it as being in review. Validation logic has been added to enforce that the request body is provided for other services.

These endpoints should correspond to the state transitions given in the diagram below:
State transitions for Temporary Accommodation assessments

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/AssessmentController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/AssessmentController.kt
@@ -186,6 +186,25 @@ class AssessmentController(
     return ResponseEntity(HttpStatus.OK)
   }
 
+  override fun assessmentsAssessmentIdClosurePost(assessmentId: UUID): ResponseEntity<Unit> {
+    val user = userService.getUserForRequest()
+
+    val assessmentValidationResult = when (val assessmentAuthResult = assessmentService.closeAssessment(user, assessmentId)) {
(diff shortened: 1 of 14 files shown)


nsmj63-dxw (approved):
Changes look good. Testing looks particularly robust.
That said, I wouldn't mind another look from someone who's been on the project a bit longer than me!

PR #863, opened 2023-08-03
Omit Bookings from `/premises/{premises id}/bookings` where user does pass LAO check

brownbearsoftware:
This is a stopgap measure to unblock people at several APs that can no longer view any Bookings until https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/804 is merged.


pezholio (approved):

tahb:
We had spotted and were worried about not returning bookings in these situations. Not showing a booking could incorrectly imply availability to the person booking. It could also impact place where someone is looking for a booking in the same residence to check for compatibility. 

I think we need to get to having a slim booking returned asap which just has the CRN and the dates asap. You mentioned this is a stop gap, is this type of thing where you think we're headed towards?

pezholio:
I think we can close this now right @tahb ?

tahb:
@pezholio yes I think so. We don't want to omit a booking and I think we're both working okay without this.

PR #864, opened 2023-08-03
Various improvements to the booking seed

tahb:
- Fix previous seeds by setting the arrivalDateTime for Arrivals for all arrivals. We do this in a single statement rather than worrying about creating a matching timestamp on each insert.
- We add a void to each region
- We add a confirmed and departed booking
- We spread the existing bookings out across more beds to make the timeline look more realistic

I've given this a test locally


nsmj63-dxw (approved):
Given that this is just editing of seed data in the database migrations, and there's proof of it working locally, I'm reasonably confident to rubber-stamp this despite my limited understanding of the API itself.

PR #869, opened 2023-08-07
(373) Filter placement requests by status

pezholio:
This updates the placement requests dashboard to be able to filter by its status, rather than the parole status.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/PlacementRequestsTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/PlacementRequestsTest.kt
@@ -4,7 +4,6 @@ import org.assertj.core.api.Assertions.assertThat
 import org.junit.jupiter.api.Nested
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.params.ParameterizedTest
-import org.junit.jupiter.params.provider.CsvSource
 import org.junit.jupiter.params.provider.EnumSource
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.data.repository.findByIdOrNull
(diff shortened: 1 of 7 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/PlacementRequestsTest.kt
```diff
@@ -161,154 +165,110 @@ class PlacementRequestsTest : IntegrationTestBase() {
...
     @CsvSource(
-      "/placement-requests/dashboard,nonParole",
-      "/placement-requests/dashboard?isParole=1,parole",
-      "/placement-requests/dashboard?isParole=0,nonParole",
+      "/placement-requests/dashboard,notMatched",
+      "/placement-requests/dashboard?status=notMatched,notMatched",
+      "/placement-requests/dashboard?status=matched,matched",
+      "/placement-requests/dashboard?status=unableToMatch,unableToMatch",
     )
     fun `It returns all the placement requests when the user is a manager`(url: String, type: String) {
```
danlivings-dxw, line 173: I think this might be better to have this as separate tests, as there's no interdependency between the different types of placement request and there's a lot of nesting.

danlivings-dxw (commented):
The change to the sort field enum to snake case has caused a couple of tests to fail. I suspect that the `Sort.by` in the mocks for those tests also needs to be made snake case to make the tests pass.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/PlacementRequestsTest.kt
pezholio: Good call - makes it a bit tider too 👍

PR #870, opened 2023-08-07
(373) Return booking summary in placement request list

pezholio:
This allows us to get the premises name for a placement requests’s booking when listing the placement requests.


danlivings-dxw (approved):
LGTM

PR #872, opened 2023-08-08
373 - Placement requests search

brownbearsoftware:
Change Placement Request dashboard endpoint to return all Placement Requests instead of defaulting to filtering on not matched when status not provided.

Add crn, tier, arrivalDateStart, arrivalDateEnd optional filters.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PlacementRequestEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PlacementRequestEntity.kt
@@ -61,11 +61,13 @@ interface PlacementRequestRepository : JpaRepository<PlacementRequestEntity, UUI
         END
       ) = :#{#status?.toString()})
       AND (:crn IS NULL OR (SELECT COUNT(1) FROM applications a WHERE a.id = pq.application_id AND a.crn = UPPER(:crn)) = 1)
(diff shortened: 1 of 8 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PlacementRequestEntity.kt
```diff
@@ -61,10 +61,11 @@ interface PlacementRequestRepository : JpaRepository<PlacementRequestEntity, UUI
         END
       ) = :#{#status?.toString()})
       AND (:crn IS NULL OR (SELECT COUNT(1) FROM applications a WHERE a.id = pq.application_id AND a.crn = UPPER(:crn)) = 1)
+      AND (:tier IS NULL OR (SELECT COUNT(1) FROM approved_premises_applications apa WHERE apa.id = pq.application_id AND apa.risk_ratings -> 'tier' -> 'value' ->> 'level' = :tier) = 1)
```
pezholio: Unless I'm missing something, there doesn't appear to be any tests for the tier search? Would be good to get something in, especially as we're querying over JSON here.

pezholio (approved):
LGTM - just once concern over missing tests that would be good to get in 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PlacementRequestEntity.kt
brownbearsoftware: Good spot, have added a test for this

PR #873, opened 2023-08-08
Link bookings to applications in Temporary Accommodation

danlivings-dxw:
> See [ticket #1381 on the CAS3 Trello board](https://trello.com/c/K6AjM2OT/1381-a-user-can-also-link-to-a-referral-as-part-of-creating-a-provisional-booking).

This PR introduces the ability to link bookings to applications within the Temporary Accommodation service. This is done by passing in an `assessmentId` parameter when sending a request to the `POST /premises/{premisesId}/bookings` endpoint.

This new field is optional, so the existing behaviour is preserved if this ID is not provided.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
@@ -32,6 +32,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApprovedPremi
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApprovedPremisesEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ArrivalEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ArrivalRepository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.AssessmentRepository
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.BedEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.BedMoveEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.BedMoveRepository
@@ -112,6 +113,7 @@ class BookingService(
   private val turnaroundRepository: TurnaroundRepository,
   private val bedMoveRepository: BedMoveRepository,
   private val premisesRepository: PremisesRepository,
+  private val assessmentRepository: AssessmentRepository,
   private val notifyConfig: NotifyConfig,
   @Value("\${url-templates.frontend.application}") private val applicationUrlTemplate: String,
   @Value("\${url-templates.frontend.booking}") private val bookingUrlTemplate: String,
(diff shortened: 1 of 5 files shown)


bryangaledxw:
Tested UI manually against API. Looks good

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/BookingTest.kt
```diff
@@ -540,6 +540,88 @@ class BookingTest : IntegrationTestBase() {
 
   @Test
   fun `Create Temporary Accommodation Booking returns OK with correct body`() {
+    `Given a User`(roles = listOf(UserRole.CAS3_ASSESSOR)) { userEntity, jwt ->
+      `Given an Offender` { offenderDetails, inmateDetails ->
+        val premises = temporaryAccommodationPremisesEntityFactory.produceAndPersist {
```
tahb, line 545: I wonder if there's a small opportunity to move this property+bed test set up into a helper function given it's not the focus of the test. That might then be reusable and make this and other tests like it more concise.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
```diff
@@ -594,6 +597,17 @@ class BookingService(
         "$.departureDate" hasValidationError "beforeBookingArrivalDate"
       }
 
+      val application = when (assessmentId) {
+        null -> null
+        else -> {
+          val result = assessmentRepository.findByIdOrNull(assessmentId)
+          if (result == null) {
+            "$.assessmentId" hasValidationError "doesNotExist"
```
tahb, line 605: It looks to set an error message on the object that's then validated in the next block which makes sense but I'm curious as to how this line works.

tahb (approved):
LGTM 👍

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/BookingTest.kt
tahb: I see now that this pattern of placing all set up in the test seems to be widely adopted already so any switch away should be more a coordinated effort.

danlivings-dxw: I can put a ticket in the backlog to create a `Given a Premises`/`Given a Bed` helper function and make this switch?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
danlivings-dxw: The `validated` method creates a [`ValidatedScope`](https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/main/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/ValidationErrors.kt), which holds a `MutableMap<String, String>` for validation errors.

The `hasValidationError` function is an [infix function](https://kotlinlang.org/docs/functions.html#infix-notation) which in this case puts `"$.assessmentId" -> "doesNotExist"` into the error map.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/BookingTest.kt
tahb: It sounds good to me if you also think this type of thing would be of value.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
tahb: I didn't know hat an infix function was! So the long hand is `"$.assessmentId".hasValidationError("doesNotExist")`. Thanks :)

PR #876, opened 2023-08-09
373 - Add name to ap applications

brownbearsoftware:
This PR adds a name field to Approved Premises Applications and begins populating it for newly created Applications.

For existing Applications we also introduce a Migration Job which we can trigger which will backfill the names for existing Applications.  Once this has been run we can make the column not-null in a subsequent release (assuming we are able to succesfully backfill every Application with a name)


pezholio (approved):
LGTM - My only concern is what we do if someone changes their name - I know we're only doing this for search / ordering, but it might result in some unexpected behaviour. I'm tempted to leave it though, and see if / when it comes up. Given it's an edge case, we can just rerun the seed job when/if it does come up.

PR #877, opened 2023-08-09
Fix 403 Forbidden from being returned when a user with `CAS3_ASSESSOR` role tries to access assessments

danlivings-dxw:
Currently, the API returns the error response when a user tries to access an assessment that was not created by them.
However, any assessments within the same region should be accessible to a user, given that they have the `CAS3_ASSESSOR` role.
This fixes the authorisation logic to ensure that an assessment can be accessed by a user with this role.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AssessmentService.kt
```diff
@@ -84,7 +84,13 @@ class AssessmentService(
       else -> throw RuntimeException("Assessment type '${assessment::class.qualifiedName}' is not currently supported")
     }
 
-    if (!user.hasRole(UserRole.CAS1_WORKFLOW_MANAGER) && assessment.allocatedToUser != user) {
+    // TODO: This should be delegated to a method in UserAccessService so that the check can be done in a service-aware way
+    //   to prevent (for example) a user with the `CAS1_WORKFLOW_MANAGER` role from being able to access other users'
+    //   assessments in CAS3.
+    //   They should only be able to authorise with the correct role for the service, but this needs to be working for a
```
tahb, line 90: Happily we don't need to rush this for 11 minutes time anymore as we have an account in dev with dual roles that lets us workaround this. How much space do we need to create to do this properly now?

danlivings-dxw: I think doing this properly wouldn't be too complex to implement, but it would be worth having tests to check that the cross-pollination of roles that this change introduces has been fixed. I imagine this would take a day or so to implement with appropriate tests.

tahb: Understood, I don't think we can afford to carve out a whole day this week. I'll put something in the backlog and let's both try to keep this on the agenda for next week.

Perhaps can just remove part of this comment about the "11 minute demo" and then we're good to go?

tahb (approved):

tahb:
Card to capture tidying this up https://trello.com/c/WKfi06nk

PR #879, opened 2023-08-10
1321 - Add human notes to applications

nsmj63-dxw:
> See [ticket #1321 on the CAS3 Trello board.](https://trello.com/c/iIjSVHfm/1321-human-notes-can-be-added-to-applications)

CAS1 originally had the idea of building a generic notes system that allowed for both human-written and system-generated notes to be associated with applications. This PR implements the human side of this system for use in both CAS1 and CAS3.

These notes will be collated together to show the history of a referral; as demonstrated below:
image

The base-level note functionality, and the human side, is implemented here. [The system notes half of this functionality will be added in #1351.](https://trello.com/c/FTcVxbYl/1351-when-a-status-is-changed-a-system-note-is-created)

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/factory/AssessmentReferralHistoryUserNoteEntityFactory.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/factory/AssessmentReferralHistoryUserNoteEntityFactory.kt
@@ -1,17 +1,16 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.factory
 
-import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.AssessmentReferralHistoryUserNoteEntity
-
 import io.github.bluegroundltd.kfactory.Factory
 import io.github.bluegroundltd.kfactory.Yielded
-import java.time.OffsetDateTime
-import java.util.*
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.AssessmentEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.AssessmentReferralHistoryUserNoteEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.randomDateTimeBefore
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.randomStringMultiCaseWithNumbers
+import java.time.OffsetDateTime
+import java.util.UUID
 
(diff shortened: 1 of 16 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/factory/AssessmentReferralHistoryUserNoteEntityFactory.kt
```diff
@@ -0,0 +1,47 @@
...
+
+import io.github.bluegroundltd.kfactory.Factory
+import io.github.bluegroundltd.kfactory.Yielded
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.AssessmentEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.AssessmentReferralHistoryUserNoteEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.randomDateTimeBefore
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.randomStringMultiCaseWithNumbers
+import java.time.OffsetDateTime
+import java.util.*
```
danlivings-dxw, line 11: This wildcard import should be expanded to single imports.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/repository/AssessmentReferralHistoryUserNoteTestRepository.kt
```diff
@@ -0,0 +1,9 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.repository
+
+import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.stereotype.Repository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.AssessmentReferralHistoryUserNoteEntity
+import java.util.*
```
danlivings-dxw, line 6: This wildcard import should be expanded to single imports.

danlivings-dxw (commented):
Generally speaking, it's cleaner to rebase your branch onto `main` and use `git push --force-with-lease` to update your branch rather than merging `main` in (see the [dxw Playbook](https://playbook.dxw.com/tech/how-we-do-development/#6-include-any-upstream-changes)).

Also, rather than committing a ktlint cleanup explicitly, using [the ktlint-gradle helper tasks](https://github.com/JLLeitschuh/ktlint-gradle#additional-helper-tasks) to set up a pre-commit hook will help keep your commits compliant in future. In this case, I think it'd be better to just squash the two cleanup commits into one rather than going through each commit and linting each one individually.

danlivings-dxw (approved):

PR #881, opened 2023-08-11
373 - Search placement request by name

brownbearsoftware:
This adds a `crnOrName` parameter to the dashboard search endpoint.  If provided this will match on CRN exactly (after being uppercased) or to Offender Name as a substring match (e.g. a search for 'John' would return Applications for Offenders 'John Smith' and 'John Jacobs' but a search for 'John Smith' would not return Applications for Offender 'John Rupert Smith')

Part of this PR is to set the `name` column on the `approved_premises_applications` table to not null.  Before this can be released the migration job [that was added in the last PR](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/876) needs to be run on all of the environments - have been unable to do this yet because of the CAS3 pen test.  Instructions on how to run this are here: https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/main/doc/how-to/run_migration_job_remotely.md

After this has run it would be pertinent to check that it has succeeded for all entries in approved_premises_applications in all environments as otherwise the `SET NOT NULL` in the migration for this PR will fail when this work is released.

If there are failing entries in dev, test or preprod then the solution is just to delete the entry in `approved_premises_applications` and associated entry in `applications`.  If any on prod fail then further investigation will be needed.


brownbearsoftware:
[Users need a way to search for placement requests](https://trello.com/c/7MvoORdC/373-users-need-a-way-to-search-for-placement-requests)

pezholio (approved):

PR #882, opened 2023-08-11
NN-5364: get adjudication data from AdjudicationsAPI instead of PrisonAPI

neilmendum:
https://dsdmoj.atlassian.net/browse/NN-5364

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/client/AdjudicationsApiClient.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/client/AdjudicationsApiClient.kt
@@ -1,0 +1,28 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.client
+
+import com.fasterxml.jackson.databind.ObjectMapper
+import org.springframework.beans.factory.annotation.Qualifier
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.data.redis.core.RedisTemplate
+import org.springframework.stereotype.Component
+import org.springframework.web.reactive.function.client.WebClient
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.adjudications.AdjudicationsPage
+
+@Component
+class AdjudicationsApiClient(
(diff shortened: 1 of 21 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/client/AdjudicationsApiClient.kt
```diff
@@ -0,0 +1,28 @@
...
+) : BaseHMPPSClient(webClient, objectMapper, redisTemplate, preemptiveCacheKeyPrefix) {
+
+  fun getAdjudicationsPage(nomsNumber: String, offset: Int?, pageSize: Int) = getRequest<AdjudicationsPage> {
+    withHeader("Page-Limit", pageSize.toString())
+
+    if (offset != null) {
+      withHeader("Page-Offset", offset.toString())
+    }
+
+    path = "/adjudications/$nomsNumber/adjudications"
```
neilmendum, line 26: [Here is this endpoint](https://github.com/ministryofjustice/hmpps-manage-adjudications-api/blob/9e4fb75fe356845eda7c96c66aab998f792aecee/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsmanageadjudicationsapi/controllers/AdjudicationSummaryController.kt#L142-L208) in the AdjudicationsAPI

We may need to add the `VIEW_ADJUDICATIONS` role to the client credentials as the [AdjudicationSummaryController has it configured](https://github.com/ministryofjustice/hmpps-manage-adjudications-api/blob/9e4fb75fe356845eda7c96c66aab998f792aecee/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsmanageadjudicationsapi/controllers/AdjudicationSummaryController.kt#L29), though it may have it already because it's [one of the valid roles for the existing PrisonAPI endpoint](https://github.com/ministryofjustice/prison-api/blob/a46928fa96e554e59b56723c689e6df79031e632/src/main/java/uk/gov/justice/hmpps/prison/api/resource/OffenderResource.java#L384-L385)

PR #888, opened 2023-08-15
Run UI E2E tests after dev deploy

pezholio:
This ensures both frontends are still working as expected after any changes have been made to the API.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,6 +3,8 @@ version: 2.1
 orbs:
   hmpps: ministryofjustice/hmpps@7.3.3
   mem: circleci/rememborb@0.0.2
+  node: circleci/node@4.5.2
+  slack: circleci/slack@4.10.1
 
 parameters:
   alerts-slack-channel:
@@ -38,27 +40,6 @@ executors:
           - DATA_DIR=/tmp/localstack/data
 
 jobs:
-  trigger-ui-tests:
-    docker: 
-      - image: cimg/base:stable
-    resource_class: small
-    steps:
-      - run:
-          name: "Approved Premises UI test"
-          command: | 
-              curl --request POST \
-                --url https://circleci.com/api/v2/project/github/ministryofjustice/hmpps-approved-premises-ui/pipeline \
-                --header "Circle-Token: $CIRCLECI_PIPELINE_TOKEN" \
-                --header "content-type: application/json" \
-                --data '{"branch":"main"}'
-      - run:
-          name: "Temporary Accommodation UI test"
-          command: | 
-              curl --request POST \
-                --url https://circleci.com/api/v2/project/github/ministryofjustice/hmpps-temporary-accommodation-ui/pipeline \
-                --header "Circle-Token: $CIRCLECI_PIPELINE_TOKEN" \
-                --header "content-type: application/json" \
-                --data '{"branch":"main"}'
   generate_trivyignore:
     # Needed to get Trivy to respect the .trivyignore file, as it doesn't use the one that's inside the Docker image.
     docker:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -196,6 +177,68 @@ jobs:
...
+          destination: playwright-report
+      - store_artifacts:
+          path: test-results
+          destination: test-results
+      - slack/notify:
+          event: fail
+          channel: << pipeline.parameters.alerts-slack-channel >>
+          template: basic_fail_1
+  cas3_e2e_tests:
+    parallelism: 4
```
tahb, line 208: Noticed CAS1 doesn't have this and it sounds useful to keep deploys faster. What's the max? Guessing it's the number of cores on whatever sized resource we pay for and so can't be increased?

.circleci/config.yml
```diff
@@ -196,6 +177,68 @@ jobs:
...
+          name: Checkout e2e repo
+          command: |
+            git clone https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui.git .
+      - run:
+          name: install
+          command: apt-get install xz-utils
+      - run:
+          name: Update npm
+          command: 'npm install -g npm@latest'
+      - node/install-packages
```
tahb, line 227: Is this our script or something that we get out of the cypress/browsers:node16.14.2-slim-chrome100-ff99-edge image, what packages does it install that's not covered by npm install? Excuse the silly question.

.circleci/config.yml
```diff
@@ -196,6 +177,68 @@ jobs:
...
+      - run:
+          name: E2E Check - << parameters.environment >>
+          command: |
+            TESTS=$(circleci tests glob "e2e/tests/*.feature" | circleci tests split --split-by=timings | paste -sd ',')
+            npm run test:e2e:ci --\
```
tahb, line 232: I don't expect this to change much in the frontend but remembering to come and update this seems like a very easy thing to miss. Perhaps we could leave a link to this file in [our frontend e2e.env files](https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/blob/main/e2e.local.env) as a pointer?

(I assume it's impossible to load environment variables from another circleci context and that even if we could that might have unintended consequences anyway).

.circleci/config.yml
```diff
@@ -242,6 +285,8 @@ workflows:
       - request-preprod-approval:
           type: approval
           requires:
+            - cas1_e2e_tests
```
tahb, line 288: I wonder if we should release this first without these requires so if there's a problem we don't block the pipeline. Then once it's green we follow-up with a small change to add them back in?

.circleci/config.yml
pezholio: Yeah, I haven't bothered for CAS1, as we have fewer e2e tests and they're only split across two files. We could try increasing it, but it'll be diminishing returns.

.circleci/config.yml
pezholio: It's a CircleCI orb thing I think, it handles caching and the like too, as well as running `npm install`

.circleci/config.yml
pezholio: Good plan. In a seperate push I commented out the requires and it seems to work, but that sounds sensible too

.circleci/config.yml
pezholio: I've added all the shared secrets to a [Context](https://app.circleci.com/settings/organization/github/ministryofjustice/contexts/4a855c71-1cc7-498a-b7bf-4e7d1cfbe1b1?return-to=%2F) here, so we can share them between projects.

tahb (approved):
👍

PR #891, opened 2023-08-16
394-batch-applications-into-manageable-chunks-when-setting-names-on-a…

garethhallberg:
…pplications

394-batch-applications-into-manageable-chunks-when-setting-names-on-applications - removed unused function

There was a problem with running out of memory on prod when migrating as the code used a findAll. The fix has been to use a Pageable. A pagesize is by default 50 but can be passed in as a parameter. The tests all use a pagesize of one.


pezholio (approved):
LGTM - not an issue/blocker for now, but if you have a look at some other commit messages on the project to get an idea of style, that would help you get aligned 👍

PR #892, opened 2023-08-16
NN-5366: get adjudication data from AdjudicationsAPI instead of PrisonAPI 

neilmendum:
With https://dsdmoj.atlassian.net/browse/NN-5366 and the epic https://dsdmoj.atlassian.net/browse/NN-5363 the Adjudications team are switching consumers of PrisonAPI `/adjudications` endpoints to source data from the AdjudicationsAPI instead

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/client/AdjudicationsApiClient.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/client/AdjudicationsApiClient.kt
@@ -16,13 +16,7 @@ class AdjudicationsApiClient(
   @Value("\${preemptive-cache-key-prefix}") preemptiveCacheKeyPrefix: String,
 ) : BaseHMPPSClient(webClient, objectMapper, redisTemplate, preemptiveCacheKeyPrefix) {
 
-  fun getAdjudicationsPage(nomsNumber: String, offset: Int?, pageSize: Int) = getRequest<AdjudicationsPage> {
-    withHeader("Page-Limit", pageSize.toString())
-
-    if (offset != null) {
-      withHeader("Page-Offset", offset.toString())
-    }
-
-    path = "/adjudications/$nomsNumber/adjudications"
+  fun getAdjudicationsPage(nomsNumber: String, page: Int, pageSize: Int) = getRequest<AdjudicationsPage> {
+    path = "/adjudications/$nomsNumber/adjudications?page=$page&size=$pageSize"
   }
 }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/PrisonAdjudicationsConfig.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/PrisonAdjudicationsConfig.kt
@@ -10,5 +10,5 @@ class PrisonAdjudicationsConfigBindingModel {
 }
 
 data class PrisonAdjudicationsConfig(
-  val prisonApiPageSize: Int,
+  val adjudicationsApiPageSize: Int,
 )
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/adjudications/AdjudicationsPage.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/adjudications/AdjudicationsPage.kt
@@ -3,8 +3,12 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.model.adjudications
 import java.time.LocalDateTime
 
 data class AdjudicationsPage(
-  val results: List<Adjudication>,
+  val results: Results,
   val agencies: List<Agency>,
+)
+
+data class Results(
+  val content: List<Adjudication>,
 )
 
 data class Adjudication(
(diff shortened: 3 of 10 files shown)


wiremock/mappings/AdjudicationsApi_GetAdjudicationsPage.json
neilmendum, line 1: Rather than add another service to docker-compose, which seemed like overkill 1 for endpoint, I decided to introduce this as a WireMock request. The response is taken from `https://manage-adjudications-api-dev.hmpps.service.justice.gov.uk/adjudications/A8349DY/adjudications?page=0&size=10`

helm_deploy/hmpps-approved-premises-api/values.yaml
```diff
@@ -50,6 +50,8 @@ generic-service:
       SPRING_SECURITY_OAUTH2_CLIENT_REGISTRATION_HMPPS-TIER_CLIENT-SECRET: "SYSTEM_CLIENT_TIER_SECRET"
       SPRING_SECURITY_OAUTH2_CLIENT_REGISTRATION_PRISONS-API_CLIENT-ID: "SYSTEM_CLIENT_NOMIS_ID"
       SPRING_SECURITY_OAUTH2_CLIENT_REGISTRATION_PRISONS-API_CLIENT-SECRET: "SYSTEM_CLIENT_NOMIS_SECRET"
+      SPRING_SECURITY_OAUTH2_CLIENT_REGISTRATION_MANAGE-ADJUDICATIONS-API_CLIENT-ID: "SYSTEM_CLIENT_NOMIS_ID"
```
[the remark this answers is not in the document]
neilmendum: See [Jira comment](https://dsdmoj.atlassian.net/browse/NN-5363?focusedCommentId=317904) for roles which have been updated

wiremock/mappings/AdjudicationsApi_GetAdjudicationsPage.json
```diff
@@ -0,0 +1,292 @@
...
+          "agencyId": "MDI",
+          "description": "Moorland (HMP & YOI)",
+          "active": true
+        },
+        {
+          "agencyId": "WWI",
+          "description": "Wandsworth (HMP)",
+          "active": true
+        }
+      ]
```
danlivings-dxw, line 289: The `Agency` objects in this list seem to be missing the `agencyType` field as required in `AdjudicationsPage` which causes a 500 response when testing the `GET /people/{crn}/adjudications` endpoint locally. The [Swagger docs for the Adjudications API](https://manage-adjudications-api-dev.hmpps.service.justice.gov.uk/swagger-ui/index.html#/01.%20Adjudication%20Summary/getAdjudicationsByPrisonerNumber) seems to indicate that `agencyType` isn't returned, but it would be good to confirm this is the case so that we can reuse this Wiremock mapping elsewhere (like [ap-tools](https://github.com/ministryofjustice/hmpps-approved-premises-tools)).

neilmendum: Hi @danlivings-dxw thanks for digging into this! That's right. When the endpoint was created with https://github.com/ministryofjustice/hmpps-manage-adjudications-api/pull/414 (cc @Mjwillis) we didn't include `agencyType`

Seems that you, and other consumers of that endpoint don't use the field so seems fine to no longer include it. I've dropped it from `Agency` with https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/892/commits/913908c807445b9a9240002e50ee0f051dcb477e

danlivings-dxw (approved):
From my point of view I don't see any problems with this change, although it might be worth talking to @tahb about whether any infrastructure changes need to be in place to be able to deploy this (such as network/firewall rules, etc.).

helm_deploy/hmpps-approved-premises-api/values.yaml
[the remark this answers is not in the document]
tahb: From reading back on the notes from the problems on [the first release of this](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/882) it seems like [the core problem was our requests not being authorised](https://mojdt.slack.com/archives/C012FNNQ4A3/p1692093856763789?thread_ts=1692009909.588849&cid=C012FNNQ4A3).

Was that because we didn't have this variable set last time so when [we set up the previous client](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/882/files#diff-61e5abfecfb05b57d0af67e97f31c84fde2d4f9033881ac609694b0d77c78178R190) it instantiated without the correct keys. Basically, are we confident that this was the problem?

wiremock/mappings/AdjudicationsApi_GetAdjudicationsPage.json
tahb: I would be in favour of having a backing service here and in ap-tools to keep higher parity between local and live. It shouldn't be hard to do. I will check in with the CAS team.

tahb (commented):
Hey @neilmendum one question around the provision of the new secrets. Otherwise I think it's ready to take out of draft. 

If we're happy for it to go as is are you happy for us to merge it so we can be prepared to run some manual tests on the dev environment?

helm_deploy/hmpps-approved-premises-api/values.yaml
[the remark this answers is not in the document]
neilmendum: Yes - I missed that bit of config from my first PR

PR #895, opened 2023-08-17
Submit cas2 application

Harriethw:


danlivings-dxw (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/ApplicationsTransformer.kt
```diff
@@ -218,7 +218,10 @@ class ApplicationsTransformer(
     }
 
     if (entity is DomainCas2ApplicationSummary) {
-      return ApplicationStatus.inProgress
+      return when {
+        entity.getSubmittedAt() !== null -> ApplicationStatus.submitted
```
danlivings-dxw, line 222: This should be fine with `null` but for future reference `==`/`!=` and `===`/`!==` have different semantics and most of the time it's probably `==`/`!=` that you want: https://kotlinlang.org/docs/equality.html

Harriethw: ah thanks! that's good to know

PR #902, opened 2023-08-18
337 we need screens to enable cru staff to manage user accounts and permissions

garethhallberg:
Endpoints to allow staff to delete a user and to update a user's roles and permissions.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
@@ -5,7 +5,9 @@ import org.springframework.beans.factory.annotation.Value
 import org.springframework.data.domain.Sort
 import org.springframework.data.repository.findByIdOrNull
 import org.springframework.stereotype.Service
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ApprovedPremisesUserRole
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ServiceName
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.UserRolesAndQualifications
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.ClientResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.CommunityApiClient
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApplicationEntity
@@ -23,8 +25,12 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.specification.hasQua
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.community.StaffUserDetails
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.BadRequestProblem
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.results.AuthorisableActionResult
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.UserTransformer
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.transformQualifications
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.transformUserRoles
 import java.util.UUID
 import javax.servlet.http.HttpServletRequest
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.UserQualification as APIUserQualification
 
 @Service
 class UserService(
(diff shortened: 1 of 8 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
```diff
@@ -127,6 +130,50 @@ class UserService(
...
+    ApprovedPremisesUserRole.matcher -> UserRole.CAS1_MATCHER
+    ApprovedPremisesUserRole.manager -> UserRole.CAS1_MANAGER
+    ApprovedPremisesUserRole.workflowManager -> UserRole.CAS1_WORKFLOW_MANAGER
+    ApprovedPremisesUserRole.applicant -> UserRole.CAS1_APPLICANT
+    ApprovedPremisesUserRole.roleAdmin -> UserRole.CAS1_ADMIN
+    ApprovedPremisesUserRole.reportViewer -> UserRole.CAS1_REPORT_VIEWER
+    ApprovedPremisesUserRole.excludedFromAssessAllocation -> UserRole.CAS1_EXCLUDED_FROM_ASSESS_ALLOCATION
+    ApprovedPremisesUserRole.excludedFromMatchAllocation -> UserRole.CAS1_EXCLUDED_FROM_MATCH_ALLOCATION
+    ApprovedPremisesUserRole.excludedFromPlacementApplicationAllocation -> UserRole.CAS1_EXCLUDED_FROM_PLACEMENT_APPLICATION_ALLOCATION
+  }
```
pezholio, line 175: Thinking about this, I do wonder if we could put this in somewhere (maybe the `util` folder) so we can improve the test coverage and allow it to be used elsewhere.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/UserServiceTest.kt
```diff
@@ -518,4 +522,81 @@ class UserServiceTest {
...
+    private val mockHttpAuthService = mockk<HttpAuthService>()
+    private val mockOffenderService = mockk<OffenderService>()
+    private val mockCommunityApiClient = mockk<CommunityApiClient>()
+    private val mockUserRepository = mockk<UserRepository>()
+    private val mockUserRoleAssignmentRepository = mockk<UserRoleAssignmentRepository>()
+    private val mockUserQualificationAssignmentRepository = mockk<UserQualificationAssignmentRepository>()
+    private val mockProbationRegionRepository = mockk<ProbationRegionRepository>()
+    private val mockProbationAreaProbationRegionMappingRepository = mockk<ProbationAreaProbationRegionMappingRepository>()
+
+//    private val userService = spyk<UserService>(recordPrivateCalls = true)
```
pezholio, line 537: I think this was left in by mistake

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/UserServiceTest.kt
```diff
@@ -518,4 +522,81 @@ class UserServiceTest {
...
+
+    private val userService = mockk<UserService>()
+
+    private val id = UUID.fromString("21b61d19-3a96-4b88-8df9-a5e89bc6fe73")
+    private val username = "SOMEPERSON"
+    private val forename = "Jim"
+    private val surname = "Jimmerson"
+    private val staffIdentifier = 5678
```
pezholio, line 545: Do we need all this setup here, or can we rely on the defaults set by the factory?

pezholio (commented):
Looks generally good - a couple of questions / suggestions just to tighten a few things up. Also, we (dxw) tend to keep our commit history pretty tidy to make review easier and help Future Us when we're reviewing what's been going on in the codebase. There's some guidance here:

https://playbook.dxw.com/tech/how-we-do-development/#follow-our-standards

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/UserServiceTest.kt
garethhallberg: Yes we can rely on the defaults. Good shout.

richpjames:
I think there is some work that needs be done on the commit history before this can be merged.  
There's some good tips in the guidance Stu linked above. I think points [4](https://playbook.dxw.com/tech/how-we-do-development/#4-develop-the-code-and-the-tests-side-by-side), [5](https://playbook.dxw.com/tech/how-we-do-development/#5-commit-changes-in-small-steps) and [7](https://playbook.dxw.com/tech/how-we-do-development/#7-tidy-up-your-commits) need addressing here. 
It might be a good idea to build the commits around the different endpoints in the PR (PUT, `users/search` and `delius/search`) or something similar

src/main/resources/static/api.yml
```diff
@@ -3258,6 +3258,105 @@ paths:
...
+            'application/json':
+              schema:
+                $ref: '#/components/schemas/User'
+        401:
+          $ref: '#/components/responses/401Response'
+        403:
+          $ref: '#/components/responses/403Response'
+        500:
+          $ref: '#/components/responses/500Response'
+  /users/search/{name}:
```
richpjames, line 3296: I don't think we need the `/{name}` segment here. Judging from the other methods that use params it's enough to just have them specified in the `parameters:` key

garethhallberg: Good catch - I've made changes to this and the Delius search endpoint. Now using the form of /users/search?name=foo

richpjames: Thanks Gareth

pezholio (approved):

PR #911, opened 2023-08-23
Revert "Merge pull request #804 from ministryofjustice/309-replace-pe…

tahb:
…rson-with-personinfo"

This reverts commit 52c57a3118c08e235be37b7995f4836675a681b5, reversing changes made to 3fc0c6f1d7821d247c5383aae19f4383ebd9f6f0.

We are reverting this change for two reasons:

1. CAS3 tested well however the frontend needs to update its types as a result introduced by this change and is blocking the deployment pipeline until it's done. We're working on that now but want to unblock the pipeline asap.
2. CAS1 have been doing some testing and found an issue when attempting to reallocate an LAO offender. We've prompted this LAO change to the test environment to allow for more testing to take place whilst unblocking the deployment pipeline for the API


pezholio (approved):
LGTM 👍

PR #914, opened 2023-08-23
Revert "Revert "Merge pull request #804 from ministryofjustice/309-replace-pe…"

tahb:
Reverts ministryofjustice/hmpps-approved-premises-api#911

We're now reproposing [the API LAO changes originally set out here](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/804). 

We had to revert them from dev where we were conducting some manual tests:

- We needed to make CAS3 frontend changes (which also applies to CAS2 who have been looped in) which without blocked the deployment pipeline
- CAS1 found some odd behaviour during the test (which may require an additional change to this PR)

Before we go again we should check that each frontend has updated it's types and CAS1 is happy they've fixed any problems.


tahb:
Thanks @pezholio I think we'll need to now wait for Bryan to merge in the CAS3 changes. @Harriethw when do you think you'll be able to work on the CAS2 changes? We don't want to block anyones pipeline with this.

Harriethw:
@tahb this isn't going to break us massively, as far as I can tell everything should still work the same if the person is NOT an LAO? and we are nowhere near having prod or preprod environment so don't think you need to worry about us to push this - I might not get round to making the changes till Friday/Monday

PR #920, opened 2023-08-25
Return Risk To Self section of OASys report

Harriethw:
CAS2 would like to request only the Risk to Self section of an OASys.  (https://github.com/ministryofjustice/hmpps-community-accommodation-tier-2-ui/pull/176 ) 

Currently there is the `oasys/sections` endpoint which enforces returning several sections (e.g. ROSH, Risk to Self, Risk Management Plan) and optionally returns some other sections. It also does not return one of the questions we need in Risk to Self ('historical risk of self-harm')

This PR suggests an `oasys/risk-to-self` endpoint which only returns the Risk to Self section. I'd be interested in any feedback on this approach. We will shortly need to request just the ROSH section, so I thought about having a param sent through to determine what section is returned, but having it as it's own endpoint seemed easier to understand.


tahb (approved):
I think a new endpoint makes sense considering the effort involved in weaving through a new query param into the existing controller action. 

Should we find ourselves having to add more endpoints for a specific combination of oasys sections again perhaps it'd be worth iterating it then? We could add a new service class that received a list of section ids and encapsulated the orchestration logic but as we are the controller is doing this and it's not easy to extend.

For now it seems like we expose a subset of the data we were getting before with the one addition of R8.1.2 👍 

Probably worth a review from a Kotlin dev as well as me in case of any particulars.

danlivings-dxw (approved):
LGTM

PR #924, opened 2023-08-31
Seed CAS3 CI test users with the same region

tahb:
This has been updated manually on the dev and test environments but should these migrations run again then we want the regions to continue to be the same.

In the frontend we are running e2e tests that require a referrer to submit a referral, the assessor then signs in and can see that referral. The way CAS3 manages content via region (rather than nationally with CAS1) makes this important.


danlivings-dxw (approved):
LGTM 👍

PR #926, opened 2023-09-04
Provide summary data to Temporary Accommodation assessments

danlivings-dxw:
> See [ticket #1504 on the CAS3 Trello board](https://trello.com/c/Sb4KYeHX/1504-referral-data-can-be-presented-without-knowledge-of-how-to-navigate-the-question-answer-data-blob).

This PR introduces the `summaryData` field on a Temporary Accommodation assessment. This is a user-defined bag of data intended to allow the assessment summary page in the UI to present basic information from the application in a version-agnostic way. For example, whether the offender can share accommodation with others could be represented as `summaryData.canShare`, allowing the application schema to be modified independently, such as by reordering the application questions.

As the assessment is created when the application is submitted, this information is required as part of the `SubmitTemporaryAccommodationApplication` schema (i.e. when `POST`ing to `/applications/{applicationId}/submission`), which now looks like:
```jsonc
{
  "type": "CAS3",
  "translatedDocument": {
    // ...
  },
  "arrivalDate": "2000-01-01",
  "summaryData": { // This can be any object
    // ...
  }
}
```

Similarly, any endpoint that responds with an `Assessment` will include the `summaryData` field when returning a Temporary Accommodation assessment. For existing assessments, this has been set to the empty object (`{}`).

### DIFF
--- a/src/main/resources/static/api.yml
+++ b/src/main/resources/static/api.yml
@@ -5417,6 +5417,8 @@ components:
             arrivalDate:
               type: string
               format: date
+            summaryData:
+              $ref: '#/components/schemas/AnyValue'
           required:
             - arrivalDate
     ReleaseTypeOption:
@@ -5595,8 +5597,11 @@ components:
               $ref: '#/components/schemas/TemporaryAccommodationUser'
             status:
               $ref: '#/components/schemas/TemporaryAccommodationAssessmentStatus'
+            summaryData:
+              $ref: '#/components/schemas/AnyValue'
           required:
             - application
+            - summaryData
     AssessmentSummary:
       type: object
       properties:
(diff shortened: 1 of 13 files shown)


libuk:
Thanks @danlivings-dxw this makes sense to me.

src/main/resources/static/api.yml
```diff
@@ -5275,8 +5275,11 @@ components:
             arrivalDate:
               type: string
               format: date
+            summaryData:
+              $ref: '#/components/schemas/AnyValue'
           required:
             - arrivalDate
+            - summaryData
```
tahb, line 5418: Will having this as required before the frontend is ready break existing requests?

danlivings-dxw: I've modified this to be optional, and if it's not provided will default to `{}`.

tahb (approved):
LGTM 🚀

PR #927, opened 2023-09-04
(143) Add /reports/applications reports endpoint

pezholio:
This allows us to get a report about applications that have been submitted in a particular month. Not all the data we've been asked for is easy to get at the moment, so some of the fields are currently null.


pezholio:

garethhallberg (approved):

PR #928, opened 2023-09-04
Exclude cancelled bookings from being considered as overlaps in Temporary Accommodation bed search

danlivings-dxw:
> See [ticket #1509 on the CAS3 Trello board](https://trello.com/c/WFeqioTH/1509-dont-include-cancellations-in-overlapping-residents).

This PR modifies the bed search logic for Temporary Accommodation to exclude bookings with one or more cancellations. As these have been cancelled, presenting these bookings to the user has the potential to cause confusion and additional work to determine which (if any) of the bookings are actually relevant.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BookingEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BookingEntity.kt
@@ -20,8 +20,8 @@ interface BookingRepository : JpaRepository<BookingEntity, UUID> {
   @Query("SELECT b FROM BookingEntity b WHERE b.premises.id = :premisesId AND b.arrivalDate <= :endDate AND b.departureDate >= :startDate")
   fun findAllByPremisesIdAndOverlappingDate(premisesId: UUID, startDate: LocalDate, endDate: LocalDate): List<BookingEntity>
 
-  @Query("SELECT b FROM BookingEntity b WHERE b.premises.id IN :premisesIds AND b.arrivalDate <= :endDate AND b.departureDate >= :startDate")
-  fun findAllByPremisesIdsAndOverLappingDate(premisesIds: List<UUID>, startDate: LocalDate, endDate: LocalDate): List<BookingEntity>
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BookingEntity.kt
```diff
@@ -20,8 +20,8 @@ interface BookingRepository : JpaRepository<BookingEntity, UUID> {
   @Query("SELECT b FROM BookingEntity b WHERE b.premises.id = :premisesId AND b.arrivalDate <= :endDate AND b.departureDate >= :startDate")
   fun findAllByPremisesIdAndOverlappingDate(premisesId: UUID, startDate: LocalDate, endDate: LocalDate): List<BookingEntity>
 
-  @Query("SELECT b FROM BookingEntity b WHERE b.premises.id IN :premisesIds AND b.arrivalDate <= :endDate AND b.departureDate >= :startDate")
-  fun findAllByPremisesIdsAndOverLappingDate(premisesIds: List<UUID>, startDate: LocalDate, endDate: LocalDate): List<BookingEntity>
+  @Query("SELECT b FROM BookingEntity b WHERE b.premises.id IN :premisesIds AND b.arrivalDate <= :endDate AND b.departureDate >= :startDate AND SIZE(b.cancellations) = 0")
+  fun findAllNotCancelledByPremisesIdsAndOverlappingDate(premisesIds: List<UUID>, startDate: LocalDate, endDate: LocalDate): List<BookingEntity>
```
tahb, line 24: I wondered if there was a better way to structure these queries so that we don't have to try and communicate their logic in their name. If our query was complex we could get into quite a pickle. 

From a quick look it doesn't seem like Kotlin has a SQL DSL that allows query chaining(?) such that we could write something like `bookings.withoutCancelled.findOverlapsByPremises` which could make this easier to work with.

tahb (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BookingEntity.kt
danlivings-dxw: You could probably achieve something like this with an abstraction layer over [criteria queries](https://www.baeldung.com/hibernate-criteria-queries) but I don't know of anything that does this (and I don't know how well anything like that would play with Spring).

PR #931, opened 2023-09-05
Delete user flags that user with false for isActive

garethhallberg:
User delete sets user isActive to false. Unit tests and integration tests.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
@@ -244,6 +244,7 @@ class UserService(
         roles = mutableListOf(),
         qualifications = mutableListOf(),
         probationRegion = staffProbationRegion,
+        isActive = true,
       ),
     )
   }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
@@ -83,12 +83,12 @@ class UserService(
   }
 
   fun getUsersWithQualificationsAndRoles(qualifications: List<UserQualification>?, roles: List<UserRole>?) =
-    userRepository.findAll(hasQualificationsAndRoles(qualifications, roles), Sort.by(Sort.Direction.ASC, "name"))
+    userRepository.findAll(hasQualificationsAndRoles(qualifications, roles, true), Sort.by(Sort.Direction.ASC, "name"))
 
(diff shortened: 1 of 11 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
```diff
@@ -49,7 +49,9 @@ class UserService(
   private val log = LoggerFactory.getLogger(this::class.java)
 
   fun getUsersByPartialName(name: String): List<UserEntity> {
-    return userRepository.findByNameContainingIgnoreCase(name)
+    return userRepository.findByNameContainingIgnoreCase(name).filter {
+      it.isActive == true
+    }
```
pezholio, line 54: Can we update this so the filtering is done on the db side? We'll probably need to paginate this stuff eventually, so we'll need to get a consistent number of results for each page.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
```diff
@@ -81,7 +83,9 @@ class UserService(
   }
 
   fun getUsersWithQualificationsAndRoles(qualifications: List<UserQualification>?, roles: List<UserRole>?) =
-    userRepository.findAll(hasQualificationsAndRoles(qualifications, roles), Sort.by(Sort.Direction.ASC, "name"))
+    userRepository.findAll(hasQualificationsAndRoles(qualifications, roles), Sort.by(Sort.Direction.ASC, "name")).filter {
```
pezholio, line 86: Same re: filtering - I'd also expect this in a different commit (after the work that does the deletion)

pezholio (commented):
Just a couple of things, but looks good overall. It also looks like the commit message hasn't been updated either. I'd expect this work to be split into three commits to tell the story of the feature. For example:

Commit 1 would contain all the work for adding the new `isActive` field with a commit message like:

```
Add a new isActive field to Users

This adds a new field called `isActive` to users. This will allow us
to make users "inactive" without deleting them.
```

Then commit 2 would contain all the work for doing the actual deletion with a commit like:

```
Add an endpoint to soft-delete users

This adds a `DELETE /users/$id` endpoint that sets `isActive`
to false on any given user. 
```

Then the final commit would be:

```
Exclude inactive users from the user listing

Now we have an `isActive` flag, we want to hide any soft-deleted
users from the listing. We may want to use this flag elsewhere, but
for now let's just add it here to avoid confusion when someone
"deletes" a user, but still sees it in the listing.
```

Hope this makes sense, gimme a shout if not!

pezholio (approved):
LGTM - thanks for bearing with me!

PR #933, opened 2023-09-05
Allow CAS1_Manager to view assessment made by another user

garethhallberg:


pezholio (approved):
LGTM - looks like the linter is complaining though 😢

PR #935, opened 2023-09-05
Fix assessment state transition from 'ready to place' back to 'in review'

danlivings-dxw:
> See [ticket #1532 on the CAS3 Trello board](https://trello.com/c/BPzOc5fV/1532-referral-status-not-updating-properly).

When trying to mark an assessment that is 'ready to place' as being 'in review' again, the API would not correctly update the state, causing the assessment to still be shown as 'in review'.

This was caused by the `AssessmentService.reallocateTemporaryAccommodationAssessment` method not resetting the decision on the assessment to `null`. Making this change results in the expected behaviour.

The `AssessmentStateTest` class contains a test method that verifies all of the expected state transitions work as intended.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/AssessmentStateTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/AssessmentStateTest.kt
@@ -1,0 +1,169 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.integration
+
+import org.junit.jupiter.api.Test
+import org.springframework.data.repository.findByIdOrNull
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.AssessmentAcceptance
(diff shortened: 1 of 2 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/AssessmentStateTest.kt
```diff
@@ -0,0 +1,169 @@
...
+      )
+      .exchange()
+      .expectStatus()
+      .is2xxSuccessful
+
+    val persistedApplication = temporaryAccommodationApplicationRepository.findByIdOrNull(this.id)!!
+    return persistedApplication.assessments.first() as TemporaryAccommodationAssessmentEntity
+  }
+
+  private fun TemporaryAccommodationAssessmentEntity.assertApiStatus(status: TemporaryAccommodationAssessmentStatus, jwt: String) {
```
tahb, line 106: Awesome, this and the other helper functions makes that very long test much more readable!

tahb (approved):

PR #936, opened 2023-09-05
(Chore) Remove users not in Delius from Dev and Test

pezholio:
The users  “tester.testy”, “bernard.beaks” and “panesar.jaspal” do not exist in Delius. Now we call Delius to check if a user can access the offender for a assessment / placement request, we need to ensure they exist in Delius, otherwise the call will fail.


pezholio:
Gonna close this for now

PR #939, opened 2023-09-06
Ensure seed CSVs are run in order

pezholio:
We've noticed an issue in dev where sometimes, the Room seed is run before the premises seed, so the rooms do not seed correctly. This appends numbers in front of each CSV to ensure they run in the correct order (in a similar way to our repeatable migrations in Flyway). We then remove the number to ensure the correct job type is run. I’ve also updated the tests to verify that the jobs are run in the correct order. I decided to use a spy to make sure that the `seedData` method is called. This is tested in the integration tests anyway, so we don’t lose any coverage.


richpjames (approved):
Approving this with the hope it will fix our E2Es and unblock the pipelines but understand if you would rather wait until someone with more experience in this repo can review

PR #944, opened 2023-09-07
Allow Access with ROLE_PRISON and 'nomis' auth source

edavey:
This means that *across the board, for all CAS services*, access is now granted via NOMIS creds in addition to Delius creds.

NOTE: the `ValidNomisAuthorisationCodeJwt` is created using a Delius staff member so is not a realistic test. There is considerable work to do throughout the code base to handle NOMIS users rather than Delius users.


edavey:
[Create NomisUserEntity as alternative to existing (Delius) UserEntity](https://trello.com/c/pxVfFLdz/515-create-nomisuserentity-as-alternative-to-existing-delius-userentity)

edavey:
[Authenticate with `ROLE_PRISON`](https://trello.com/c/Fd5GjCvm/512-authenticate-with-roleprison)

PR #950, opened 2023-09-08
(Fix) Create custom SQL for the Application Reports

pezholio:
When making its way into production, we found that the Application reports were running out of memory. This is something we’ve seen before when listing all the applications, becuase the default query loads all the application data json into memory, which overloads the service. This creates a new repository (for tidiness, as the ApplicationEntity was getting a bit big) to generate a custom report row, containing all the things we need from the database to generate a report. This should hopefully cut down on memory usage so we can actually generate reports in prod.

I’ve had to update the unit tests to cast the applications to a `ApplicationEntityReportRow`, which is a bit long-winded, but works. I’ve also updated the integration tests to be a bit more robust.


pezholio:

garethhallberg (approved):

PR #953, opened 2023-09-11
Return RoSH section of an OASys report

patrickjfl:
Following [CAS2’s implementation of an `oasys/risk-to-self` risk to self specific endpoint ](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/920)(as an alternative to the generic OASys endpoint which returns all sections), this PR does the same for risk of serious harm (RoSH), creating an `oasys/rosh` endpoint.


pezholio (approved):
LGTM 👍

PR #955, opened 2023-09-11
Add a cron job to copy prod data to preprod

pezholio:
~~**Not to be merged until https://github.com/ministryofjustice/cloud-platform-environments/pull/16411 is in**~~

This cribs from work the interventions team have done to set a cron job to run every Sunday to copy prod data to preprod:

https://github.com/ministryofjustice/hmpps-interventions-service/blob/main/helm_deploy/hmpps-interventions-service/templates/cronjob-prod-to-preprod-refresh.yaml

This will allow us to make preprod a more realistic environment, allowing UAT and debugging easier to carry out.

### DIFF
--- a/helm_deploy/hmpps-approved-premises-api/templates/cronjob-prod-to-preprod-refresh.yaml
+++ b/helm_deploy/hmpps-approved-premises-api/templates/cronjob-prod-to-preprod-refresh.yaml
@@ -1,0 +1,131 @@
+{{- if .Values.env_details.production_env -}}
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: db-refresh-script
+data:
+  entrypoint.sh: |-
+    #!/bin/bash
+    set -e
+
+    echo "${DB_HOST}:5432:${DB_NAME}:${DB_USER}:${DB_PASS}" > ~/.pgpass
+    echo "${DB_HOST_PREPROD}:5432:${DB_NAME_PREPROD}:${DB_USER_PREPROD}:${DB_PASS_PREPROD}" >> ~/.pgpass
+    chmod 0600 ~/.pgpass
+
+    set -x
+    # Dump production data
+    pg_dump --host="$DB_HOST" \
+      --username="$DB_USER" \
+      --format=custom \
+      --no-privileges \
+      --verbose \
+      --file=/tmp/db.dump \
+      "$DB_NAME"
+    
+    # Dump existing preprod users as upserts
+    pg_dump --host="$DB_HOST_PREPROD" \
+      --username="$DB_USER_PREPROD" \
+      --on-conflict-do-nothing \
+      --column-inserts \
+      --data-only \
+      --no-privileges \
+      --verbose \
+      --table public.users \
+      --table public.user_qualification_assignments \
+      --table public.user_role_assignments \
+      --file=/tmp/users.dump \
+      "$DB_NAME_PREPROD"
+    
+    # Restore production data to preprod
+    pg_restore --host="$DB_HOST_PREPROD" \
+      --username="$DB_USER_PREPROD" \
+      --clean \
+      --no-owner \
+      --verbose \
+      --dbname="$DB_NAME_PREPROD" \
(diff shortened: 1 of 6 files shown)


pezholio:
[Copy prod data to preprod on a schedule](https://trello.com/c/3wvMdF1U/433-copy-prod-data-to-preprod-on-a-schedule)

helm_deploy/hmpps-approved-premises-api/templates/cronjob-prod-to-preprod-refresh.yaml
```diff
@@ -0,0 +1,97 @@
...
+                    secretKeyRef:
+                      name: rds-postgresql-instance-output-preprod
+                      key: rds_instance_address
+          restartPolicy: "Never"
+          volumes:
+            - name: db-refresh-script
+              configMap:
+                name: db-refresh-script
+                defaultMode: 0755
+  {{- end }}
```
tahb, line 97: I was wondering why we have this red warning, maybe a missing trailing newline? Whilst here I noticed that `{{- end }}` needs to have no indentation, I think?

helm_deploy/hmpps-approved-premises-api/templates/cronjob-prod-to-preprod-refresh.yaml
```diff
@@ -0,0 +1,97 @@
...
+    pg_restore --host="$DB_HOST_PREPROD" --username="$DB_USER_PREPROD" --clean --no-owner --verbose --dbname="$DB_NAME_PREPROD" /tmp/db.dump
+
+    rm -v /tmp/db.dump ~/.pgpass
+---
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: db-refresh-job
+spec:
+  schedule: "0 8 * * 0"
```
tahb, line 27: @danlivings-dxw fyi this could be useful if we have to do report generation up front

helm_deploy/hmpps-approved-premises-api/templates/cronjob-prod-to-preprod-refresh.yaml
```diff
@@ -0,0 +1,97 @@
...
+      template:
+        spec:
+          {{- if .Values.serviceAccountName }}
+          serviceAccountName: {{ .Values.serviceAccountName }}
+          {{- end }}
+          securityContext:
+            runAsUser: 999
+          containers:
+            - name: dbrefresh
+              image: "postgres:14"
```
tahb, line 42: Is this the version of postgres we use? I think we use postgis:postgis too so I'm not sure if there are risks to using this inconsistently?

tahb (approved):
This is a dangerous thing to get wrong isn't it.

I've left a question about postgres version and some low-level formatting for your consideration.

Otherwise it seems like this path has been taken by both interventions and offender-case-notes which is encouraging. 

As I remember our postgres database has backups taken automatically should something go wrong. We should double-check this before proceeding as well as guidance for database restores, should the worse happen. 

Then, when we're ready to merge I suggest we do it early in the day and [execute the task manually](https://github.com/ministryofjustice/offender-case-notes/tree/27ab394d717d29a3e2af80f7cec5ab7bb1835dda/database_refresh#run-an-adhoc-database-refresh) so we can check it works whilst we're around to do any database restore?

helm_deploy/hmpps-approved-premises-api/templates/cronjob-prod-to-preprod-refresh.yaml
pezholio: Also useful for us!

helm_deploy/hmpps-approved-premises-api/templates/cronjob-prod-to-preprod-refresh.yaml
pezholio: The RDS instance we use in the Cloud Platform is vanilla Postgres, version 14. As we're only doing client actions in this container, I don't think it matters.

PR #958, opened 2023-09-12
If there is associated placement request add booking to it unless the…

garethhallberg:
… request is withdrawn or the request already has a booking.

Includes tests that werify the spec.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
@@ -339,7 +339,7 @@ class BookingService(
 
       val newestSubmittedOnlineApplication = applicationService.getApplicationsForCrn(crn, ServiceName.approvedPremises)
         .filter { it.submittedAt != null }
-        .maxByOrNull { it.submittedAt!! }
+        .maxByOrNull { it.submittedAt!! } as ApprovedPremisesApplicationEntity?
       var newestOfflineApplication = applicationService.getOfflineApplicationsForCrn(crn, ServiceName.approvedPremises)
         .maxByOrNull { it.createdAt }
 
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
```diff
@@ -339,10 +339,20 @@ class BookingService(
...
 
+      var newestPlacementRequest = newestSubmittedOnlineApplication?.getLatestPlacementRequest()
+
+      if (newestPlacementRequest?.isWithdrawn == true) {
+        return@validated generalError("placementRequestIsWithdrawn")
+      }
+
+      if (newestPlacementRequest?.booking != null) {
+        return@validated newestPlacementRequest.booking!!.id hasConflictError "A Booking has already been made for this Placement Request"
+      }
```
pezholio, line 354: Thinking about this, rather than returning an error, we should just not update the placement request if either of these conditionals is true. Something like (on line 406):

```kotlin
if (newestPlacementRequest != null && newestPlacementRequest?.isWithdrawn != true && newestPlacementRequest?.booking == null) {
    newestPlacementRequest.booking = booking
    placementRequestRepository.save(newestPlacementRequest)
}
```

Then update the unit tests to ensure that `mockPlacementRequestRepository.save` does not get called, instead of checking for errors.

That way the user can still create the manual booking without any unexpected behaviour or weirdness.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/BookingServiceTest.kt
```diff
@@ -3275,6 +3275,172 @@ class BookingServiceTest {
...
+      assertThat(authorisableResult is AuthorisableActionResult.Success).isTrue
+      val validatableResult = (authorisableResult as AuthorisableActionResult.Success).entity
+      assertThat(validatableResult is ValidatableActionResult.Success).isTrue
+
+      verify(exactly = 0) {
+        mockPlacementRequestRepository.save(placementRequest)
+      }
+    }
+
+    private fun getAdHocEverys() {
```
pezholio, line 3429: You can put this at the top and add the `@BeforeEach` annotation to run it before each test in the inner class, to save calling the function

pezholio (approved):
LGTM 👍

PR #964, opened 2023-09-14
Increase parallelism on CI tests

pezholio:
This is a bit of a blunt instrument, but the CI times seem to be creeping up, so let’s throw another container at it.


edavey (approved):
👍

PR #965, opened 2023-09-14
517 Implement 'cas2' api namespace

edavey:
**NB: this should be deployed with a change to the UI so that application requests are made to the new `/cas2/applications` namespace.**

As per [Trello 517](https://trello.com/c/dtLdFRF6/517-spike-implement-cas2-api-namespace) and [draft ADRs 15 and 16 in PR 960](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/960)

Our refactoring plan to support auth via NOMIS is as follows:

- extract current CAS2 Application endpoints to a `cas2` namespace and `Cas2ApplicationsController` (in review here)
    
- pull `Cas2Application` entity out of JPA inheritance
    
- implement `NomisUser` and necessary alternate authorisation code suited to Nomis rather than Delius user
    
- allow users to authenticate with Nomis creds, associate `NomisUser` with `Cas2Application` entity and use `NomisUser` in `Cas2ApplicationsController` actions

cas2_namespace

So far so good.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ApplicationsController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ApplicationsController.kt
@@ -117,7 +117,6 @@ class ApplicationsController(
 
     val applicationResult = when (xServiceName ?: ServiceName.approvedPremises) {
       ServiceName.approvedPremises -> applicationService.createApprovedPremisesApplication(personInfo.offenderDetailSummary, user, deliusPrincipal.token.tokenValue, body.convictionId, body.deliusEventNumber, body.offenceId, createWithRisks)
-      ServiceName.cas2 -> applicationService.createCas2Application(body.crn, user, deliusPrincipal.token.tokenValue)
       ServiceName.temporaryAccommodation -> {
         when (val actionResult = applicationService.createTemporaryAccommodationApplication(body.crn, user, deliusPrincipal.token.tokenValue, body.convictionId, body.deliusEventNumber, body.offenceId, createWithRisks)) {
           is AuthorisableActionResult.NotFound -> throw NotFoundProblem(actionResult.id!!, actionResult.entityType!!)
@@ -125,6 +124,11 @@ class ApplicationsController(
           is AuthorisableActionResult.Success -> actionResult.entity
         }
       }
+
+      ServiceName.cas2 -> throw RuntimeException(
+        "CAS2 now has its own " +
+          "Cas2ApplicationsController",
+      )
     }
 
     val application = when (applicationResult) {
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ApplicationsController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ApplicationsController.kt
@@ -22,7 +22,6 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.SubmitCas2Appl
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.SubmitTemporaryAccommodationApplication
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.UpdateApplication
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.UpdateApprovedPremisesApplication
-import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.UpdateCas2Application
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.UpdateTemporaryAccommodationApplication
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApplicationEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApprovedPremisesApplicationEntity
@@ -167,7 +166,6 @@ class ApplicationsController(
         data = serializedData,
         username = user.deliusUsername,
       )
-      is UpdateCas2Application -> applicationService.updateCas2Application(applicationId = applicationId, data = serializedData, username = user.deliusUsername)
       else -> throw RuntimeException("Unsupported UpdateApplication type: ${body::class.qualifiedName}")
     }
 
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ApplicationsController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ApplicationsController.kt
@@ -18,7 +18,6 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.PlacementReque
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ServiceName
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.SubmitApplication
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.SubmitApprovedPremisesApplication
-import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.SubmitCas2Application
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.SubmitTemporaryAccommodationApplication
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.UpdateApplication
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.UpdateApprovedPremisesApplication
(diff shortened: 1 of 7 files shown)


edavey:
[SPIKE: implement cas2 API namespace](https://trello.com/c/dtLdFRF6/517-spike-implement-cas2-api-namespace)

src/main/resources/static/api.yml
```diff
@@ -5,6 +5,171 @@ info:
 servers:
 - url: /
 paths:
+  /cas2/applications:
+    post:
+      tags:
+        - Operations on CAS2 applications
+      summary: Creates a CAS2 application
+      parameters:
+        - name: createWithRisks
```
Harriethw, line 14: can we get rid of this `createWithRisks` parameter?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/Cas2ApplicationsController.kt
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    return ResponseEntity.ok(getPersonDetailAndTransform(updatedApplication, user))
+  }
+
+  override fun cas2ApplicationsApplicationIdSubmissionPost(
+    applicationId: UUID,
+    submitApplication: SubmitCas2Application,
+  ): ResponseEntity<Unit> {
+    val deliusPrincipal = httpAuthService.getDeliusPrincipalOrThrow()
```
Harriethw, line 127:
```suggestion
    httpAuthService.getDeliusPrincipalOrThrow()
```
do we need to get the username/`deliusPrincipal` as variables here?  
(I guess we don't really need to call `getDeliusPrincipalOrThrow` for CAS2, but we might want to leave it in so we can replace with something that checks we're authorising with NOMIS later on?)

Harriethw (approved):
very cool to see it coming together!
Would be good to hear from a CAS 1/3 team member too :)

src/main/resources/static/api.yml
edavey: yes, well spotted -- this has just been removed from CAS2's application entity in https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/968 -- and was not being passed to the `ApplicationService` in any case...

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/Cas2ApplicationsController.kt
edavey: Good point, we certainly aren't using that `deliusPrincipal` or its `username` but the `getDeliusPrincipalOrThrow()` is an important control in the current implementation.

I'll replace:

```kotlin
val username = deliusPrincipal.name
```
with

```kotlin
val _deliusPrincipal = httpAuthService.getDeliusPrincipalOrThrow()
```

which as you suggest might be a useful reminder...

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ApplicationsController.kt
```diff
@@ -117,14 +115,18 @@ class ApplicationsController(
...
           is AuthorisableActionResult.NotFound -> throw NotFoundProblem(actionResult.id!!, actionResult.entityType!!)
           is AuthorisableActionResult.Unauthorised -> throw ForbiddenProblem()
           is AuthorisableActionResult.Success -> actionResult.entity
         }
       }
+
+      ServiceName.cas2 -> throw RuntimeException(
+        "CAS2 now has its own " +
+          "Cas2ApplicationsController",
+      )
```
pezholio, line 129: Probably overkill, but could we redirect here?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/Cas2ApplicationsController.kt
```diff
@@ -0,0 +1,157 @@
...
+  override fun cas2ApplicationsApplicationIdPut(
+    applicationId: UUID,
+    body:
+      UpdateApplication,
+  ): ResponseEntity<Application> {
+    val user = userService.getUserForRequest()
+
+    val serializedData = objectMapper.writeValueAsString(body.data)
+
+    val applicationResult = applicationService.updateCas2Application(applicationId = applicationId, data = serializedData, username = user.deliusUsername)
```
pezholio, line 105: I wonder if now might be the time to namespace the services too? So the `applicationService` in this case would be in `services/cas2/ApplicationService.kt`?

pezholio (approved):
LGTM - I am beginning to come round to your way of thinking on this. I certainly think namespacing the controllers and services has legs, and it'll definitely cut down on some of the complexities around conditionals and versioning and make the code all round tidier.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/Cas2ApplicationsController.kt
```diff
@@ -0,0 +1,157 @@
...
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.UserService
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.ApplicationsTransformer
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.getFullInfoForPersonOrThrow
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.getInfoForPersonOrThrowInternalServerError
+import java.net.URI
+import java.util.UUID
+import javax.transaction.Transactional
+
+@Service
+class Cas2ApplicationsController(
```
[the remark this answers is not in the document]
edavey: yes, that is nice, thks!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ApplicationsController.kt
edavey: Hmm, possibly?  

At present I'm thinking that this is very much an interim pathway which exists only during the refactoring.  It's intended to handle the scenario where the UI is POSTing to `/applications` rather than to the new `cas2/applications` endpoint and give a helpful error message to developers.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/Cas2ApplicationsController.kt
```diff
@@ -0,0 +1,157 @@
...
+
+    val application = when (applicationResult) {
+      is ValidatableActionResult.GeneralValidationError -> throw BadRequestProblem(errorDetail = applicationResult.message)
+      is ValidatableActionResult.FieldValidationError -> throw BadRequestProblem(invalidParams = applicationResult.validationMessages)
+      is ValidatableActionResult.ConflictError -> throw ConflictProblem(id = applicationResult.conflictingEntityId, conflictReason = applicationResult.message)
+      is ValidatableActionResult.Success -> applicationResult.entity
+    }
+
+    return ResponseEntity
+      .created(URI.create("/applications/${application.id}"))
```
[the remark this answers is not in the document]
edavey: 👍 thanks! I failed to understand this properly -- it's setting the `location` header in the response so that the UI knows where to find the newly created resource:

https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/http/ResponseEntity.html#created(java.net.URI)

created

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/Cas2ApplicationsController.kt
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    return ResponseEntity.ok(getPersonDetailAndTransform(updatedApplication, user))
+  }
+
+  override fun cas2ApplicationsApplicationIdSubmissionPost(
+    applicationId: UUID,
+    submitApplication: SubmitCas2Application,
+  ): ResponseEntity<Unit> {
+    val _deliusPrincipal = httpAuthService.getDeliusPrincipalOrThrow()
```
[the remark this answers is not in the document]
edavey: 👍 have tweaked to remove that unused `val`.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/Cas2ApplicationsController.kt
edavey: yes good plan. I'd like to do this later refactoring step as this is already quite large?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ApplicationTest.kt
danlivings-dxw, line 1: I think splitting out the integration tests into service-specific test classes is a good idea 👍

danlivings-dxw (approved):
I'm happy for this to be merged. So far this seems like a good approach to disentangle the service-specific logic.

PR #966, opened 2023-09-14
Add MI reporting data to Temporary Accommodation application on submission

danlivings-dxw:
> See [ticket #1533 on the CAS3 Trello board](https://trello.com/c/PuhOQW6Z/1553-application-submit-accepts-mi-fields-as-first-class-fields).

This PR updates the `POST /applications/{applicationId}/submission` endpoint to add several new parameters to the request body. These parameters will be needed to enable generated booking reports to contain this information in the future, however for backwards compatibility these parameters are currently optional.

The `SubmitTemporaryAccommodationApplication` schema now looks like this:
```jsonc
{
  "type": "CAS3",
  "translatedDocument": {
    // ...
  },
  "arrivalDate": "2000-01-01",
  "isRegisteredSexOffender": false,
  "needsAccessibleProperty": false,
  "hasHistoryOfArson": false,
  "isDutyToReferSubmitted": true,
  "dutyToReferSubmissionDate": "2000-01-01",
  "isApplicationEligible": true,
  "summaryData": {
    // ...
  }
}
```

The `applicationSubmittedDate` and `riskOfSeriousHarm` properties as described in the Trello ticket have been omitted as they can be easily derived from information already stored on the application when it is created.

It is also worth noting that these fields do not have corresponding properties on the `TemporaryAccommodationApplication`/`TemporaryAccommodationApplicationSummary` schemas. This is intentional, as these fields aren't intended to be presented anywhere except in the booking report. Information to be presented directly to the user should instead be put into the `summaryData` object.


libuk:
Looks good from FE perspective 👍🏾

src/main/resources/db/migration/all/20230913145208__add_mi_reports_columns_to_temporary_accommodation_application.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE temporary_accommodation_applications ADD COLUMN is_registered_sex_offender BOOLEAN;
```
tahb, line 1: What's the behaviour for setting the default value here? Do we need DEFAULT NULL to make sure old referrals don't have boolean values of false?

tahb (approved):
One question around setting default database values for your review but otherwise LGTM 👍

PR #970, opened 2023-09-18
435 add pagination to tasks reallocatable endpoint

garethhallberg:


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/TasksTest.kt
```diff
@@ -155,6 +155,63 @@ class TasksTest : IntegrationTestBase() {
...
+          .uri("/tasks/reallocatable?type=bookingAppeal")
+          .header("Authorization", "Bearer $jwt")
+          .exchange()
+          .expectStatus()
+          .isForbidden
+      }
+    }
+
+    @Test
+    fun `Get all reallocatable tasks returns 200 with correct body when type is passed as parameter`() {
```
pezholio, line 184: Could we get a test for placement requests and placement applications too? Just for completeness.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/TasksController.kt
```diff
@@ -58,15 +58,25 @@ class TasksController(
...
+        val taskType = enumConverterFactory.getConverter(TaskType::class.java).convert(
+          type.kebabCaseToPascalCase(),
+        ) ?: throw NotFoundProblem(type, "TaskType")
+
+        when (taskType) {
+          TaskType.assessment -> return assessmentTasksResponse(user)
+          TaskType.placementRequest -> return placementRequestTasks(user)
+          TaskType.placementApplication -> return placementApplicationTasks(user)
+          else -> {
+            throw ForbiddenProblem()
```
pezholio, line 76: I feel like this should be a `BadRequestProblem` if the task type isn't a valid one.

pezholio (commented):
This is really cool! Just a couple of minor things

pezholio (approved):
LGTM - let's get it in and see how it flies in preprod!

PR #975, opened 2023-09-19
Add offence ID to Temporary Accommodation applications

danlivings-dxw:
Currently there is no way to retrieve the index offence for a referral in the Temporary Accommodation service. When creating an application, the offence ID is required and stored in the database already. By including the offence ID in the `TemporaryAccommodationApplication` schema, it becomes accessible when users of the API send a request for an application or for an assessment (which include its application), allowing the presentation of the index offence.


libuk:
Looks good, thanks for this 👍🏾

Harriethw (approved):
make sense to me!

pezholio (approved):

PR #977, opened 2023-09-20
Add a remote_db script

pezholio:
I got fed up of having to check the docs each time to remember how to do this, so I’ve added an easier method. I’ve updated the docs to detail both methods, as sometimes you’ll need access to the bash shell as well as Jaqy.

### DIFF
--- a/doc/how-to/access_database_remotely.md
+++ b/doc/how-to/access_database_remotely.md
@@ -4,13 +4,39 @@
 done as a last resort. This should ideally only be done in pairs, and
 mutating any live data is STRONGLY discouraged.
 
+Both methods detailed below use [Jaqy](https://teradata.github.io/jaqy/), which
+is a Java-native universal database client. As well as carrying out simple 
+queries, you can also do exports of data (enabling, for example, one off reporting), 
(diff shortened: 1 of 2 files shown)


doc/how-to/access_database_remotely.md
```diff
@@ -4,13 +4,39 @@
...
 
 ```bash
-script/remote_shell $ENVIRONMENT
+script/remote_db $ENVIRONMENT
 ```
 
-Where $ENVIRONMENT is the environment you want to connect to
+Where $ENVIRONMENT is the environment you want to connect to.
+
+When you're finished, just `ctrl+c` to end the session.
```
danlivings-dxw, line 28: Minor question/nitpick: what is `SIGINT` being sent to in this case? There's potentially processes/resources not being closed on the remote machine, so I'd say we should recommend using the `.quit` command to finish the session.

danlivings-dxw (approved):
I think this will be convenient, as most of the time I need to use `remote_shell` is to do what `remote_db` will do instead.

doc/how-to/access_database_remotely.md
pezholio: Oh good call. I thought it wouldn't end the session for some reason, but I've just tried `.quit` and it works!

PR #978, opened 2023-09-20
Use addWorkingDays instead of plusDays for Task's due date

richpjames:
We want to show the due date for tasks in working days, not simply days.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/TaskTransformer.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/TaskTransformer.kt
@@ -12,6 +12,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.TaskType
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.AssessmentEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.PlacementApplicationEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.PlacementRequestEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.WorkingDayCountService
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.PlacementType as ApiPlacementType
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/TaskTransformer.kt
```diff
@@ -20,13 +21,14 @@ class TaskTransformer(
   private val userTransformer: UserTransformer,
   private val risksTransformer: RisksTransformer,
   private val placementRequestTransformer: PlacementRequestTransformer,
+  private val workingDayCountService: WorkingDayCountService,
```
richpjames, line 24: @pezholio I think this is what is breaking the integration tests... I am assuming passing a service into a transformer is a no no or I'm not doing it right. Any tips? happy to do a call if easier

pezholio: I reckon you might have to add the workingDayCountService to the controller, and then pass the service to the transformer method itself. Something like:

```
@Service
class TasksController(
  private val assessmentService: AssessmentService,
  private val placementRequestService: PlacementRequestService,
  private val taskTransformer: TaskTransformer,
  private val placementApplicationService: PlacementApplicationService,
  private val enumConverterFactory: EnumConverterFactory,
  private val userTransformer: UserTransformer,
  private val taskService: TaskService,
  private val workingDayCountService: WorkingDayCountService,
) : TasksApiDelegate {

....

}
```

Then you can add the `workingDayCountService` to the transfomer method like so:

```
fun transformAssessmentToTask(assessment: AssessmentEntity, personName: String, workingDayCountService: WorkingDayCountService) 
```

And call it:

```
transformAssessmentToTask(assessment, personName, workingDayCountService)
```

in the controller.

Lemme know how you get on 👍

richpjames: thanks! I'll have a go

richpjames: yeah sure, I'm free from 1:30 tomorrow

pezholio: Cool - I've stuck an hour in the diary

PR #981, opened 2023-09-20
DRAFT: CAS2 Domain Events

Harriethw:
## Context

CAS2 would like to start emitting Domain Events:
* Initially just to log when an Application has been marked as submitted.
* We don't currently have a defined consumer of this API yet, so we will save and expose a set of basic information. 

In light of the refactoring work being done to [implement a cas2 namespace in the API](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/965) I think it makes sense to also namespace the Domain Events API. My understanding of how the CAS 1 Domain Events work currently is this:

CAS2 Domain Events and MI reporting - CAS 1 approach

When an Application is saved successfully, we create a DomainEvent which is saved to the database. Some of that data is published to the Domain Event SNS, along with a URL that any listeners can use to view the full Domain Event.
We have a Domain Events Controller with an endpoint for getting a Domain Event by ID. 

I propose we do a similar thing for CAS2, but with a namespaced endpoint, e.g. instead of `events/application-submitted/{eventId}` it would be `/events/cas2/application-submitted/{eventId}` and it will return a CAS2 specific Domain Event. 
Although there will be some shared fields on the models, if this ADR on [handling NOMIS users in CAS2 without inheritance](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/960/files) goes ahead, I think we should follow the tradition of not using inheritance/polymorphism for the same reasons - keeping it simple, less risk of touching production code. 

CAS2 Domain Events and MI reporting - CAS2 approach

**Questions**

* Controller separation:
   * it makes logical sense for the path for the CAS2 domain events to be under the `events` path, so e.g. `events/cas2/applications-submitted` etc. It also makes sense for us to have our own Cas2DomainEventsController.
   * However, this means that in the code the controller functions would be generated under the same `EventsApiDelegate`. When trying to create a `Cas2DomainEventsController` the code was complaining that it needed to have unique bean for the delegate. (I think? So I think a `@Service` annotated Controller has to have a one-to-one relationship with the `EventsApiDelegate`? )
   * So I couldn't figure out a way to extend or override the `eventsCas2ApplicationSubmittedEventIdGet` from the `DomainEventsController`, that actually worked. If anyone has any ideas would be very grateful! Atm I am just calling another method from the DomainEventsController. 

**To Do**
* create the `CAS2_ROLE_EVENTS` in dev environment


Harriethw:
closing this because of changes to API namespace and CAS3 domain event work, will open new PR

PR #984, opened 2023-09-21
(522) Extract Cas2Application from JPA inheritance structure

edavey:
In this PR we decouple the `Cas2Application` entity from the abstract `ApplicationEntity` which uses the JPA `InheritanceType.JOINED` strategy. We do this because we need our `Cas2Application` to be associated with a `User` from Nomis rather than an existing (Delius) `User`. 

We've decided (see [ADR 17: Handle Nomis users without inheritance in CAS2](https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/main/doc/architecture/decisions/0017-handle-nomis-users-without-inheritance-in-cas2.md) and [ADR 18: Introduce CAS2 API namespace](https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/main/doc/architecture/decisions/0018-introduce-cas2-api-namespace.md)) that it will be time-consuming and risky to refactor both:

- the existing `UserEntity` to support both Nomis and Delius identity sources and the 
- `ApplicationEntity` to support association with a newly polymorphic abstract `User`

Instead we will:

- extract current CAS2 Application endpoints to a `cas2` namespace and `Cas2ApplicationsController` (done in [PR 965](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/965))
- pull `Cas2Application` entity out of JPA inheritance (this PR)
- implement `NomisUser` and necessary alternate authorisation code suited to Nomis rather than Delius user
- allow users to authenticate with Nomis creds, associate `NomisUser` with `Cas2Application` entity and use `NomisUser` in `Cas2ApplicationsController` actions

We know that our needs for handling users will likely soon become more complex, e.g. CAS2 (or other CAS service) may need to support users from more than one source (e.g. Delius or NOMIS) or in CAS2's case NOMIS users and external users from NACRO. We will refactor as needed.

In this PR we introduce the following CAS2-specific classes: 

- `controller/cas2/ApplicationsController`
- `jpa/entity/Cas2ApplicationEntity`
- `service/cas2/ApplicationService`
- `service/cas2/JsonSchemaService`
- `service/cas2/UserAccessService`
- `transformer/cas2/ApplicationsTransformer`

We also a db migration to add the necessary fields, indexes and constraints to `cas_2_applications` to allow it to stand apart from the abstract `applications` table/entity.


edavey:
[Extract Cas2Application from JPA inheritance structure](https://trello.com/c/emEVeHLl/522-extract-cas2application-from-jpa-inheritance-structure)

Harriethw (approved):
thanks @edavey 🌟  it looks right to me but would appreciate some feedback from the other devs too :)

PR #985, opened 2023-09-21
Added some coroutine async to try and speed up this slow endpoint

garethhallberg:


pezholio (approved):
LGTM - let's give it a go!

PR #987, opened 2023-09-21
Don’t fail when unable to allocate a user

pezholio:
We’re getting an unusual number of allocations failing when submitting applications etc. While we attempt to work out what’s going on, we’ll log errors and make the allocated to user null. We’ll also log the failure to allocate together with the number of attempts, which will give us more of an idea why these errors are happening.

As part of this we’ve had to make the allocated to user nullable in a few places.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/TaskTransformer.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/TaskTransformer.kt
@@ -12,6 +12,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.TaskType
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.AssessmentEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.PlacementApplicationEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.PlacementRequestEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.PlacementType as ApiPlacementType
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.PlacementType as JpaPlacementType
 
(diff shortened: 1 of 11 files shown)


richpjames (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/TaskTransformer.kt
```diff
@@ -90,4 +91,12 @@ class TaskTransformer(
...
     else -> TaskStatus.notStarted
   }
+
+  private fun transformUserOrNull(userEntity: UserEntity?): ApprovedPremisesUser? {
+    return if (userEntity == null) {
+      null
+    } else {
+      userTransformer.transformJpaToApi(userEntity, ServiceName.approvedPremises) as ApprovedPremisesUser
+    }
+  }
```
danlivings-dxw, line 101: This could be expressed more idiomatically as:

```suggestion
  private fun transformUserOrNull(userEntity: UserEntity?) = userEntity?.let {
    userTransformer.transformJpaToApi(it, ServiceName.approvedPremises) as ApprovedPremisesUser
  }
```

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/AssessmentTest.kt
```diff
@@ -987,7 +987,7 @@ class AssessmentTest : IntegrationTestBase() {
 
             val persistedPlacementRequest = placementRequestRepository.findByApplication(application)!!
 
-            assertThat(persistedPlacementRequest.allocatedToUser.id).isIn(listOf(matcher1.id, matcher2.id))
+            assertThat(persistedPlacementRequest.allocatedToUser!!.id).isIn(listOf(matcher1.id, matcher2.id))
```
danlivings-dxw, line 990: It might be worth adding:
```
assertThat(persistedPlacementRequest.allocatedToUser).isNotNull
```
as an additional assertion before this one so that if it is null the test fails cleanly instead of through a `NullPointerException`.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/PlacementApplicationsTest.kt
```diff
@@ -900,7 +900,7 @@ class PlacementApplicationsTest : IntegrationTestBase() {
 
                       val createdPlacementApplication = createdPlacementRequests[0]
 
-                      assertThat(createdPlacementApplication.allocatedToUser.id).isIn(listOf(matcher1.id, matcher2.id))
+                      assertThat(createdPlacementApplication.allocatedToUser!!.id).isIn(listOf(matcher1.id, matcher2.id))
```
danlivings-dxw, line 903: It might be worth adding:
```
assertThat(createdPlacementApplication.allocatedToUser).isNotNull
```
as an additional assertion before this one so that if it is null the test fails cleanly instead of through a `NullPointerException`.

danlivings-dxw (approved):
LGTM 👍 

I've commented on a couple of minor things but I wouldn't consider them to be blockers.

PR #989, opened 2023-09-22
Improve allocation queries

pezholio:
Rather than checking if a user is excluded from allocations and running the query again, we add a subquery to exclude these users from the query altogether, which should allow us to allocate more quickly and in a less error prone fashion.


danlivings-dxw (approved):
LGTM 👍

pezholio:
Closed in favour of #992

PR #992, opened 2023-09-26
Add an AllocationsEngine class

pezholio:
The Allocations engine has been giving us a lot of grief recently. I attempted a fix in a previous PR, but I’ve since discovered that it didn’t _quite_ match up with what was expected. To try and make this a bit more configurable, and to remove some of the complex logic, I’ve added a `UserAllocationsEngine` class, which has a `getUserPool` function that uses the `CriteriaBuilder` to get a list of qualified users IDs, we then pass those IDs to get the user with the smallest workload, depending on the type of task.

It’s still a little cumbersome, but I’ve attempted to make it a little easier to grok with comments etc.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/util/UserAllocationsEngine.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/util/UserAllocationsEngine.kt
@@ -1,0 +1,136 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.util
+
+import org.springframework.data.jpa.domain.Specification
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserQualification
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserQualificationAssignmentEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserRepository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserRole
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserRoleAssignmentEntity
+import java.util.UUID
+import javax.persistence.criteria.CriteriaBuilder
+import javax.persistence.criteria.CriteriaQuery
+import javax.persistence.criteria.JoinType
+import javax.persistence.criteria.Path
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/util/UserAllocationsEngine.kt
danlivings-dxw, line 1: I think wrapping individual predicates being wrapped in an `and` when being added to the list of predicates is unnecessary as they're combined at the end anyway, so:

```kotlin
predicates.add(
  criteriaBuilder.and(
    somePredicate,
  ),
)
```
becomes:
```kotlin
predicates.add(
  somePredicate,
)
```

Additionally, it might be a good idea to separate the logic that generates each `somePredicate` from the above example into their own private functions for clarity, so that the `getUserPool` function ends up looking like:

```kotlin
val predicates = mutableListOf<Predicate>()

if (someCondition) {
  predicates.add(getSomePredicate(criteriaBuilder))
}

// ...

criteriaBuilder.and(*predicates.toTypedArray())
```

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/util/UserAllocationsEngine.kt
```diff
@@ -0,0 +1,134 @@
...
+      val predicates = mutableListOf<Predicate>()
+
+      // All active users
+      predicates.add(
+        criteriaBuilder.and(
+          criteriaBuilder.isTrue(root.get("isActive")),
+        ),
+      )
+
+      // All users with the required tole
```
danlivings-dxw, line 43: I assume this should be 'role'.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/UserServiceTest.kt
danlivings-dxw, line 1: There's a few tests with names that contain 'sends sets isLao' which I assume should be 'sets isLao'.

danlivings-dxw (approved):
I think this looks like a good approach. I've added a comment about how `UserAllocationsEngine` could be slightly refactored to improve clarity but I don't see it as a blocker.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/util/UserAllocationsEngine.kt
pezholio: Ah nice one. That's much better!

PR #993, opened 2023-09-27
(425) Use working days for task due days

pezholio:
This updates our `TaskTransformer` to use `addWorkingdays` instead of the blunt instrument `plusDays`.

It also (and potentially more controversially) changes the WorkingDayCountService to use a static file. We’re hitting the GOV.UK bank holidays API a lot to make these requests (even though they’re cached in Redis). In a seperate PR, we’ll add a Github action (similar to https://github.com/ministryofjustice/hmpps-approved-premises-ui/pull/1050) to check the file for changes, and open a PR to replace the file if it has.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/TaskTransformer.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/TaskTransformer.kt
@@ -13,6 +13,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.AssessmentEnt
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.PlacementApplicationEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.PlacementRequestEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.WorkingDayCountService
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.PlacementType as ApiPlacementType
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.PlacementType as JpaPlacementType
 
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/WorkingDayCountService.kt
```diff
@@ -1,37 +1,33 @@
...
+  private val objectMapper: ObjectMapper,
 ) {
-
   fun getWorkingDaysCount(from: LocalDate, to: LocalDate): Int {
-    val bankHolidays = when (val govUKBankHolidaysResponse = govUKBankHolidaysApiClient.getUKBankHolidays()) {
-      is ClientResult.Success -> govUKBankHolidaysResponse.body.englandAndWales.events.map { it.date }
-      is ClientResult.Failure -> govUKBankHolidaysResponse.throwException()
-    }
-    return from.getDaysUntilInclusive(to).filter { it.isWorkingDay(bankHolidays) }.size
+    return from.getDaysUntilInclusive(to).filter { it.isWorkingDay(this.getBankHolidays()) }.size
```
richpjames, line 20: So this says "Get the days from `from` until `to`. but filter out any BHs and weekends"?

richpjames (commented):
It's beyond my context + understanding of the repo but @danlivings-dxw,  now that we aren't making an API call to the the BH endpoint could `addWorkingDays` be added to the Dates.kt file and added as a method on the `LocalDate` object?

danlivings-dxw:
I'm not sure I understand the value of using a static JSON file for the list of bank holidays as it introduces a continuous manual task for developers to ensure it's up to date (even if it is just approving and merging an automatically generated PR). I don't think we need to worry about accidentally DDoSing the Bank Holidays API, as it's publicly accessible to anyone and (I believe) considered to be the canonical source of truth for bank holidays, so I imagine it's fairly robust. If we're often getting cache misses when looking up this information then it's possibly more beneficial to investigate why that's the case instead.

I think it would be worth getting a second opinion from @tahb on this approach.

pezholio:
@danlivings-dxw Yeah, I'm inclined to agree 👍 I'll remove the static stuff, and we can carry on calling (and caching) the Bank Holidays API

pezholio:
[Repeat the working days count down on other dashboards - CRU & workflow manager dashboards](https://trello.com/c/f87CF2mw/425-repeat-the-working-days-count-down-on-other-dashboards-cru-workflow-manager-dashboards)

garethhallberg (approved):

tahb:
I agree with @danlivings-dxw's points, I don't think we should remove GOV.UK Bank holidays.

I think we should strive to delegate the responsibility of bank holidays to a service that will manage this long term. If there were no devs on the team to merge pull requests for new bank holidays then the data will stagnate. A small detail but would silently affect turnaround periods and data reported on the back of it for all cas3 bookings. 

If we are making too many failed calls to the API in spite of our cache I also suggest that a better approach to this problem could be to review how we cache this bank holiday data. If we can reduce the request rate we'll hopefully reduce the failure rate.

PR #996, opened 2023-09-28
Separate caching and downstream API request logic

danlivings-dxw:
> See ticket [#1570 on the CAS3 Trello board](https://trello.com/c/KU3pJJox/1570-redis-cache-keys-are-required-for-rendering-the-assessment-page).

This PR performs a refactoring of `BaseHMPPSClient` and related classes to separate caching and outbound HTTP requests.

This was originally done as preparatory work to resolve the bug described in the Trello ticket above (ultimately from [this Sentry issue](https://ministryofjustice.sentry.io/issues/4479884804/)), however, it appears that this problem has already been resolved by some other work. An integration test verifying the fix has been added to this PR to prevent an accidental regression.

This PR refactors the specifics of interacting with the cache out of `BaseHMPPSClient.doRequest` into several methods on a new `WebClientCache` class:
- `checkPreemptiveCacheStatus`
- `tryGetCachedValue`
- `cacheSuccessfulWebClientResponse`
- `cacheFailedWebClientResponse`

Additionally, a new `ClientResult.Failure.CachedValueUnavailable` variant has been introduced to replace the original exception thrown by `getCacheEntryBody`. This improves the presentation of the error as the unhelpful exception message 'Unable to complete GET request to ' is no longer logged, and potentially allows more sophisticated handling of this particular failure mode in the future (such as by re-issuing the request to recover the data, similar to a cache timeout).

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/client/CommunityApiClient.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/client/CommunityApiClient.kt
@@ -2,11 +2,9 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.client
 
 import com.fasterxml.jackson.databind.ObjectMapper
 import org.springframework.beans.factory.annotation.Qualifier
-import org.springframework.beans.factory.annotation.Value
 import org.springframework.cache.annotation.Cacheable
 import org.springframework.core.io.buffer.DataBuffer
 import org.springframework.core.io.buffer.DataBufferUtils
-import org.springframework.data.redis.core.RedisTemplate
 import org.springframework.http.HttpMethod
 import org.springframework.http.HttpStatus
 import org.springframework.stereotype.Component
@@ -27,14 +25,18 @@ import java.time.Duration
 class CommunityApiClient(
   @Qualifier("communityApiWebClient") private val webClient: WebClient,
   objectMapper: ObjectMapper,
-  redisTemplate: RedisTemplate<String, String>,
-  @Value("\${preemptive-cache-key-prefix}") preemptiveCacheKeyPrefix: String,
-) : BaseHMPPSClient(webClient, objectMapper, redisTemplate, preemptiveCacheKeyPrefix) {
-  private val offenderDetailCacheConfig = PreemptiveCacheConfig(
+  webClientCache: WebClientCache,
+) : BaseHMPPSClient(webClient, objectMapper, webClientCache) {
+  private val offenderDetailCacheConfig = WebClientCache.PreemptiveCacheConfig(
     cacheName = "offenderDetails",
     successSoftTtlSeconds = Duration.ofHours(6).toSeconds().toInt(),
     failureSoftTtlBackoffSeconds =
-    listOf(30, Duration.ofMinutes(5).toSeconds().toInt(), Duration.ofMinutes(10).toSeconds().toInt(), Duration.ofMinutes(30).toSeconds().toInt()),
+    listOf(
+      30,
+      Duration.ofMinutes(5).toSeconds().toInt(),
+      Duration.ofMinutes(10).toSeconds().toInt(),
+      Duration.ofMinutes(30).toSeconds().toInt(),
+    ),
     hardTtlSeconds = Duration.ofHours(12).toSeconds().toInt(),
   )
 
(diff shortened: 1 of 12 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/client/CommunityApiClient.kt
```diff
@@ -27,14 +25,18 @@ import java.time.Duration
 class CommunityApiClient(
   @Qualifier("communityApiWebClient") private val webClient: WebClient,
   objectMapper: ObjectMapper,
-  redisTemplate: RedisTemplate<String, String>,
-  @Value("\${preemptive-cache-key-prefix}") preemptiveCacheKeyPrefix: String,
-) : BaseHMPPSClient(webClient, objectMapper, redisTemplate, preemptiveCacheKeyPrefix) {
-  private val offenderDetailCacheConfig = PreemptiveCacheConfig(
+  webClientCache: WebClientCache,
+) : BaseHMPPSClient(webClient, objectMapper, webClientCache) {
+  private val offenderDetailCacheConfig = WebClientCache.PreemptiveCacheConfig(
```
tahb, line 30: If we wanted to add an environment variable to feature flag caching so we could quickly turn it off, would the best way be to set this variable in each API client to null?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/PreemptiveCacheTest.kt
```diff
@@ -214,6 +215,49 @@ class PreemptiveCacheTest : IntegrationTestBase() {
...
+    assertThat((firstResult as ClientResult.Success).body).isEqualTo(offenderDetailsResponse)
+    wiremockServer.verify(exactly(1), getRequestedFor(urlEqualTo("/secure/offenders/crn/${offenderDetailsResponse.otherIds.crn}")))
+
+    // Subsequent calls up to successSoftTtlSeconds should return the cached value without making an upstream request
+    every { Instant.now() } returns fourSecondsLaterInstant
+
+    val secondResult = preemptivelyCachedClient.getOffenderDetailSummaryWithCall(crn)
+    assertThat(secondResult is ClientResult.Success).isTrue
+    assertThat((secondResult as ClientResult.Success).body).isEqualTo(offenderDetailsResponse)
+    wiremockServer.verify(exactly(1), getRequestedFor(urlEqualTo("/secure/offenders/crn/${offenderDetailsResponse.otherIds.crn}")))
```
tahb, line 249: I was expecting to read that we'd expect no external call to be made, since we expect the cache to be the one to respond. Should this be exactly(0)?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/AssessmentTest.kt
```diff
@@ -133,6 +134,70 @@ class AssessmentTest : IntegrationTestBase() {
...
+
+        // Simulate https://ministryofjustice.sentry.io/issues/4479884804 by deleting the data key from the cache while
+        // preserving the metadata key.
+        val cacheKeys = CacheKeySet(preemptiveCacheKeyPrefix, "inmateDetails", inmateDetails.offenderNo)
+        redisTemplate.delete(cacheKeys.dataKey)
+
+        webTestClient.get()
+          .uri("/assessments")
+          .header("Authorization", "Bearer $jwt")
+          .header("X-Service-Name", ServiceName.approvedPremises.value)
```
tahb, line 182: Curious as to how this got fixed. Presumably this test doesn't start failing if we use temporaryAccommodation here.

tahb (approved):
Great work in identifying and abstracting the caching responsibilities, it's easier to read now. 

I wonder if now, at the point of more knowledge we should try to wrap the caching of each integration with feature flags so we can easily turn it off in future. @pezholio what do you think? 

Given the fact this code wraps every request it feels like we should do a little testing on the dev environment to exercise each of the caches.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/PreemptiveCacheTest.kt
danlivings-dxw: It's `exactly(1)` because of the first call, which does make the upstream request - Wiremock keeps track of the total number of times that the endpoint was requested.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/AssessmentTest.kt
pezholio: Maybe we can try both with a `ParameterizedTest`?

pezholio (approved):
LGTM - I think adding flags by way of env vars as @tahb suggests would be a good move though

PR #997, opened 2023-09-29
Fix user region not refreshing to the latest Delius value

tahb:
https://trello.com/c/CyPr93lv

We had a support request for a user whom had been unable to access the service because they were appearing as "West Midlands" rather than "Kent, Surrey & Sussex".

After confirming the regional data is now set correctly after a presumed update, our users table still had the WM region associated with it.

We extend the existing update mechanism to now also update the user's region. As this is part of a scheduled job my hope is that this change will automatically keep up to date with ndelius as the canonical store for regional ids.

- add a helper method for finding the users region so we can reuse it in create and update logic

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
@@ -315,6 +315,6 @@ class UserService(
   }
 
   private fun userHasChanged(user: UserEntity, deliusUser: StaffUserDetails): Boolean {
-    return (deliusUser.email !== user.email) || (deliusUser.telephoneNumber !== user.telephoneNumber) || (deliusUser.staff.fullName != user.name) || (deliusUser.staffCode != user.deliusStaffCode)
+    return (deliusUser.email !== user.email) || (deliusUser.telephoneNumber !== user.telephoneNumber) || (deliusUser.staff.fullName != user.name) || (deliusUser.staffCode != user.deliusStaffCode) || (deliusUser.probationArea.code != user.probationRegion.deliusCode)
   }
 }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
@@ -194,6 +194,12 @@ class UserService(
       user.telephoneNumber = deliusUser.telephoneNumber
       user.deliusStaffCode = deliusUser.staffCode
 
+      deliusUser.probationArea?.let { probationArea ->
+        findProbationRegionFromArea(probationArea)?.let { probationRegion ->
+          user.probationRegion = probationRegion
+        }
+      }
+
       user = userRepository.save(user)
     }
 
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
@@ -15,6 +15,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.ClientResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.CommunityApiClient
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApplicationEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ProbationAreaProbationRegionMappingRepository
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
```diff
@@ -191,7 +193,9 @@ class UserService(
       user.email = deliusUser.email.toString()
       user.telephoneNumber = deliusUser.telephoneNumber
       user.deliusStaffCode = deliusUser.staffCode
-
+      user.probationRegion = findProbationRegionFromArea(deliusUser.probationArea)?.apply {
```
tahb, line 196: There is probably a much nicer way to do this. It didn't feel right to copy the create logic and default it to "North West" so if it can't be found, don't update it?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/UserServiceTest.kt
```diff
@@ -564,11 +583,7 @@ class UserServiceTest {
 
       var entity = result.entity
 
-      assertThat(entity.id).isEqualTo(user.id)
-      assertThat(entity.name).isEqualTo("$forename $surname")
-      assertThat(entity.deliusUsername).isEqualTo(user.deliusUsername)
       assertThat(entity.email).isEqualTo("null")
-      assertThat(entity.telephoneNumber).isEqualTo(deliusUser.telephoneNumber)
```
tahb, line 571: As this test describes itself as testing the email and all the other fields are tested in `it returns the user's details from the Community API and saves the email address, telephone, staff code and probation region` this seemed redundant. 

It also seems like the checks have diverged with the other test checking for deliusStaffCode where this one isn't. Let's keep this test focused around email?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/UserServiceTest.kt
```diff
@@ -538,24 +548,33 @@ class UserServiceTest {
...
         .withoutEmail()
+        .withProbationAreaCode(user.probationRegion.deliusCode)
         .produce()
 
       every { mockUserRepository.findByIdOrNull(id) } returns user
       every { mockCommunityApiClient.getStaffUserDetails(username) } returns ClientResult.Success(
         HttpStatus.OK,
         deliusUser,
       )
+      every { mockProbationAreaProbationRegionMappingRepository.findByProbationAreaDeliusCode(user.probationRegion.deliusCode) } returns ProbationAreaProbationRegionMappingEntityFactory()
```
tahb, line 574: I would prefer not to have to add a stub in here because the checking for regions shouldn't impact this behaviour. It seems though the global mocking requires it. Should we move this out of the test and up into the global (or whichever level) context?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
pezholio: Yeah, this feels like a bit of a blunt object. Maybe if the probation region can't be found, we don't update it?

pezholio: Also, this will only run if `userHasChanged` returns true, and we don't currently compare the UserEntity and Delius User's probation code, see https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/b829a1d3d219ec43cf4ec9f74d02bc24fe6090bd/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt#L311

tahb:
@pezholio Heya thanks for looking! This is hopefully ready for another review now

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/UserServiceTest.kt
pezholio: Yeah, we could hoist it to the top in a `@BeforeEach` function. Something like:

```kotlin
@BeforeEach
fun setup() {
  every { mockProbationAreaProbationRegionMappingRepository.findByProbationAreaDeliusCode(user.probationRegion.deliusCode) } returns ProbationAreaProbationRegionMappingEntityFactory()
}
```

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/UserServiceTest.kt
pezholio: Sounds sensible 👍

pezholio (approved):
LGTM 👍

PR #998, opened 2023-09-29
454 performance enhancements on premises premises id bookings

garethhallberg:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PremisesEntity.kt
```diff
@@ -235,11 +235,10 @@ data class TemporaryAccommodationPremisesSummary(
   val bedCount: Int,
 )
 
-
 interface BookingSummary {
   fun getID(): UUID
-  fun getArrivalDate(): Timestamp
-  fun getDepartureDate(): Timestamp
+  fun getArrivalDate(): LocalDate
+  fun getDepartureDate(): LocalDate
```
pezholio: This should probs be squashed into the previous commit.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PremisesEntity.kt
```diff
@@ -65,18 +65,18 @@ interface PremisesRepository : JpaRepository<PremisesEntity, UUID> {
   @Query(
     """
   SELECT
-  booking.id as id,
+  cast(booking.id as TEXT) as id,
```
pezholio, line 68: Again, I think this should be squashed into that first commit

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/util/PremisesUtils.kt
```diff
@@ -0,0 +1,28 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.util
+
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.DateCapacity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.PremisesEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.PremisesService
+import java.time.LocalDate
+
+fun getDateCapacities(premises: PremisesEntity, premisesService: PremisesService): List<DateCapacity> {
```
pezholio: Does this belong in the previous commit?

pezholio: Ah, gotcha. This gets moved back into the service in the next commit doesn't it? Probably worth deleting it in that case 👍

pezholio (commented):
Looking good, just a bit of tidying up, but I'm looking forward to trying this out!

pezholio (approved):
LGTM 👍

PR #1000, opened 2023-10-02
Use domain event to generate application reports

pezholio:
This updates the ApplicationEntityReportRow to use the domain events, rather than the actual entities to generate the reports. This simplifies the report generator considerably, and also allows us to get the information as a point in time report, rather than us having to get data (especially data from external services) at the point the application is generated.

Rather than creating the domain events manually in test, I’ve updated the ApplicationReportsTest to be more like an end to end test, so we submit applications, assess them and go through the bookings process like an actual user, so the domain events get generated in the backend. This way, if the implementation of the domain events changes in any way, these tests will fail and alert us to the fact that something has changed.

I’ve removed the ApplicationReportGenerator unit test too now, as the integration test already covers this pretty well, and the implementation on the generator is vastly simplified now.


pezholio:
[Application report tweaks](https://trello.com/c/COAbUuWI/460-application-report-tweaks)

garethhallberg (approved):

PR #1003, opened 2023-10-03
(476) Fix allocations engine

pezholio:
The new allocations engine is not behaving as expected, because the SQL `IN` query was including qualifications and roles that were included in other rows. This now updates the criteria builder to do selective joins on each qualification/role and include / exclude them as necessary.

I’ve also updated the tests to be more robust and:

a) create a full corpus of users (not just ones with the roles we need) 
b) test that `getUserPool` only returns users with the required roles and qualifications.


garethhallberg (approved):

pezholio:
[Fix allocations](https://trello.com/c/raYbkcR7/476-fix-allocations)

PR #1004, opened 2023-10-03
Recording an arrival to a CAS3 premises emits a domain event

danlivings-dxw:
> See ticket [#1526 on the CAS3 Trello board](https://trello.com/c/aytIIXiB/1526-send-a-domain-event-when-a-person-arrives-at-a-property).

This PR is the first of several intended to provide domain events to Delius upon certain actions occurring in the CAS3 service. It introduces the *person arrived* event, which occurs when a request is made to the `POST /premises/{premisesId}/bookings/{bookingId}/arrivals` endpoint for the CAS3 service.

The structure of the CAS3 *person arrived* event differs from the CAS1 *person arrived* event, and looks like this:
```jsonc
{
 "id": "364145f9-0af8-488e-9901-b4c46cd9ba37",
  "timestamp": "2022-11-30T14:53:44",
  "eventType": "accommodation.cas3.person.arrived",
  "eventDetails": {
    "personReference": {
      "crn": "C123456",
      "noms": "A1234ZX", // if we have it
    },
    "deliusEventNumber": "7",
    "applicationId": "484b8b5e-6c3b-4400-b200-425bbe410713", // optional
    "bookingId": "14c80733-4b6d-4f35-b724-66955aac320c",
    "premises": {
      "addressLine1": "",
      "addressLine2": "", // if we have it
      "postcode": "",
      "town": "", // if we have it,
      "region": ""
    },
    "arrivedAt": "2022-11-30T12:00:00", // default time to midday
    "expectedDepartureOn": "2023-02-28",
    "notes": "The actual time of arrival is unknown so has been defaulted, this should be updated manually."
  }
}
```

An event can be retrieved by sending a request to the `GET /events/cas3/person-arrived/{eventId}` endpoint. This and other CAS3 event endpoints have been namespaced in accordance with the approach proposed by CAS2 in [ADR 18](doc/architecture/decisions/0018-introduce-cas2-api-namespace.md).


helm_deploy/values-dev.yaml
```diff
@@ -26,7 +26,8 @@ generic-service:
     ASSIGN-DEFAULT-REGION-TO-USERS-WITH-UNKNOWN-REGION: true
     SENTRY_ENVIRONMENT: dev
     CACHES_STAFFMEMBERS_EXPIRY-SECONDS: 20
-    DOMAIN-EVENTS_EMIT-ENABLED: true
+    DOMAIN-EVENTS_CAS1_EMIT-ENABLED: true
```
tahb, line 29: This [commit](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1004/commits/bfc8a27ed28439a408ef4d5e7fd457bbac3246f5) where we move cas1 events into a namespace is probably one for you to be aware of @pezholio

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas3/DomainEventService.kt
```diff
@@ -0,0 +1,135 @@
...
+    val domainEventEntity = domainEventRepository.findByIdOrNull(id) ?: return null
+
+    val data = when {
+      enumTypeFromDataType(T::class) == domainEventEntity.type ->
+        objectMapper.readValue(domainEventEntity.data, T::class.java)
+      else -> throw RuntimeException("Unsupported DomainEventData type ${T::class.qualifiedName}/${domainEventEntity.type.name}")
+    }
+    return DomainEvent(
+      id = domainEventEntity.id,
+      applicationId = domainEventEntity.applicationId,
```
tahb, line 57: Is applicationId our application/referral uuid? Just checking that we're still prepared for this to be null.

Should we send the bookingId here which is something we will always have?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
```diff
@@ -893,6 +901,45 @@ class BookingService(
     booking.departureDate = expectedDepartureDate
     updateBooking(booking)
 
+    if (booking.premises is TemporaryAccommodationPremisesEntity) {
+      val domainEventId = UUID.randomUUID()
+
+      val application = booking.application as? TemporaryAccommodationApplicationEntity
+
+      cas3DomainEventService.savePersonArrivedEvent(
```
tahb, line 909: Now we have everything laid out on how to build this domain event, I wonder if we know enough to abstract this knowledge and responsibility out of the booking service and push it down a layer. This new bit of indirection feels worth it and it'd make/keep the booking class simpler.

Something like `cas3DomainEventService.savePersonArrivedEvent(booking)` ?

tahb (commented):
Thanks for this 👍  

Looks good to me aside from perhaps the booking service doing too much here. Most of our domain events hang off things happening to a booking so perhaps worth improving on early on?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas3/DomainEventService.kt
danlivings-dxw: The application ID is required by `SnsEventAdditionalInformation` when the event is published, and as this is part of the data that's directly exposed to an external service that I know very little about I've opted not to make this nullable. If we don't have an application ID available we could either choose not to emit an event or to use a zero UUID as a stand-in, which is the approach I've taken here.

tahb: Sounds like we need to figure this one out. I'm not sure what a zero UUID means but sending a stand in to a receiver who is expecting something else could cause problems.

[It seems like CAS1 added the domain event record and included applicationId on the entity](https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/main/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/DomainEvent.kt#L8) and they then presumably expose it in every(?) domain event, and have as a side affect made that tightly coupled?

CAS3 will be sending several domain events based on a booking and one based on an application(referral.submitted).

Assuming someone is consuming this data (and we should check with the integrations team) is what we need here a change to the domain events to make this polymorphic? 

```kotlin
data class DomainEvent<T> (
  val id: UUID,
  val recordId: UUID,
  val recordType: Booking | Application | Any?,
  val crn: String,
  val occurredAt: Instant,
  val data: T,
)
```

This would allow us to always send an identifier for the primary record of interest with a booking id or an application id together with the detail of type of record it belongs to.

Alternatively we make it nullable though we should also check if the receiver is ok with that as presumably they had a need for it originally.

danlivings-dxw: I'd expect the worst case scenario for using a zero UUID (`00000000-0000-0000-0000-000000000000`) would be that the service at the other end of the message queue would use it to do a request to the `GET /applications/{applicationId}` endpoint, which would then give them a 404 response.

This suggestion to overhaul what a domain event looks like is a larger breaking change of the contract between the API and the consumer of the queue. While it's probably a more accurate model, I wouldn't feel comfortable doing any further work on domain events until it's been agreed with all relevant teams, which would put us in significant danger of missing our deadline on this feature.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
danlivings-dxw: I've introduced a `DomainEventBuilder` class that can be used to obtain an instance of an event.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas3/DomainEventService.kt
tahb: [For reference the conversation is continuing in the Slack channel with the probation integrations team.](https://mojdt.slack.com/archives/C03KVQUFSER/p1696846088073159)

tahb (approved):
LGTM 👍

PR #1008, opened 2023-10-04
AP has email field. When booking made emails go to the user that book…

garethhallberg:
…ed it and the AP itself. Amends to all the tests and seeds.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/EmailNotificationService.kt
```diff
@@ -16,21 +13,26 @@ class EmailNotificationService(
   private val notifyConfig: NotifyConfig,
   @Qualifier("normalNotificationClient") private val normalNotificationClient: NotificationClient?,
   @Qualifier("guestListNotificationClient") private val guestListNotificationClient: NotificationClient?,
-  private val guestListUserRepository: NotifyGuestListUserRepository,
 ) {
   private val log = LoggerFactory.getLogger(this::class.java)
 
-  fun sendEmail(user: UserEntity, templateId: String, personalisation: Map<String, *>) {
+  fun sendEmail(userEmail: String?, templateId: String, personalisation: Map<String, *>, premisesEmailAddress: String? = null) {
```
pezholio, line 19: Probs can just call the `userEmail` variable `email`

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/EmailNotificationService.kt
```diff
@@ -16,21 +13,26 @@ class EmailNotificationService(
...
-        log.info("Email sending is disabled - would have sent template $templateId to user ${user.id}")
+        log.info("Email sending is disabled - would have sent template $templateId to user $userEmail")
         return
       }
 
-      if (notifyConfig.mode == NotifyMode.TEST_AND_GUEST_LIST && guestListUserRepository.findByIdOrNull(user.id) != null) {
-        guestListNotificationClient!!.sendEmail(templateId, user.email, personalisation, null)
+      if (notifyConfig.mode == NotifyMode.TEST_AND_GUEST_LIST) {
+        guestListNotificationClient!!.sendEmail(templateId, userEmail, personalisation, null)
+        if (premisesEmailAddress != null) {
```
pezholio, line 28: Do we need `premisesEmailAddress` now?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
```diff
@@ -238,7 +238,7 @@ class BookingService(
       val lengthOfStayWeeksWholeNumber = (lengthOfStayDays.toDouble() % 7) == 0.0
 
       emailNotificationService.sendEmail(
```
pezholio, line 240: I think we should have two calls to `emailNotificationService.sendEmail` here, as it makes the function more generic, and gives us the flexibility to use another template for the premises if we need it

pezholio (commented):
Coupla comments on the service method - I think we can make it more generic

pezholio (commented):
LGTM - just need to make sure the email address is present before sending the email to the premises. I think an empty email address will cause the upstream API to fail

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
```diff
@@ -237,22 +237,44 @@ class BookingService(
...
+            "crn" to placementRequest.application.crn,
+            "startDate" to arrivalDate.toString(),
+            "endDate" to departureDate.toString(),
+            "lengthStay" to if (lengthOfStayWeeksWholeNumber) lengthOfStayWeeks else lengthOfStayDays,
+            "lengthStayUnit" to if (lengthOfStayWeeksWholeNumber) "weeks" else "days",
+          ),
+        )
+      }
+      if (premises != null) {
+        if (premises!!.emailAddress != null) {
```
pezholio, line 259: I think you can just do `if (premises?.emailAddress != null) {` here rather than the two conditionals

PR #1009, opened 2023-10-04
542: Move CAS2 specific api endpoints into separate OpenAPI spec file

edavey:
See [Trello 542](https://trello.com/c/CRkvuGW9/542-move-cas2-specific-api-endpoints-into-separate-openapi-spec-file)

In CAS2 we now need to add a further `/cas2/` endpoint to
the API. This will be in order to search for people using
a new `Cas2OffenderService`, which will not rely on the
logged in user being a Delius user. However, when
implementing:

`/cas2/people`

we discovered that there's no way to define a true `cas2`
url namespace in our existing single `api.yml` OpenAPI spec.

### CODE GENERATION

The way the generator is working is that it takes the first
part of the url:

```
/cas2/people/search
 ^^^^
```

and assumes that this is the name of your REST "resource"
and that you wish to have your models and controllers named
accordingly: i.e. `Cas2` and `Cas2ApiController`

It creates a Kotlin class named for this, i.e. `Cas2Api`
which your own controller e.g. `cas.PeopleController`
delegates to `Cas2ApiDelegate`.

It's not possible to delegate to the same
generated controller with `Cas2ApiDelegate` from two
overriding classes e.g. `cas2.ApplicationsController` and
`cas2.PeopleController`.

The solution is to split our OpenAPI specs so we have:

- `cas2-api.yml` for the cas2 namespace
- `api.yml` for the other endpoints

and put the common `components` (`responses`, `headers`
and `schemas`) into a shared file which is referenced using
the [OpenAPI $ref property].

### PROBLEM

However, there is a [bug (or missing feature)] in the Kotlin
OpenAPI generator which causes code generation to fail when
a 'remote' `ref` to a external resource is used.

### SOLUTION

In this PR we introduce a build stage for code generation using an 
`openApiPreCompilation` gradle task to:

1. create a build artefact named `./codegen/built-{specName}-spec.yml`

2. populate it with the given spec's paths plus all of the
   shared components

3. define a new `openapitools` gradle task `openApiGenerateCas2Namespace`
   to generate CAS2 namespace code -- NB: with these
   important configuration settings:

    i) `apiPackage.set("uk.gov.justice.digital.hmpps.approvedpremisesapi.api.cas2")`
       to use a `cas2` package within the API controller
       namespace

    ii) `put("apiSuffix", "Cas2")` to give our generated
        CAS2 controller a 'Bean' identity which won't clash
        with the controller defined by `api.yml` (i.e.
        `ApplicationsCas2Delegate` rather than `ApplicationsApiDelegate`)

  See the docs for the [Kotlin-Spring OpenAPI code generator]
  for details of these config options.

### DEPENDENCY: GENERATING TYPES FOR UI PROJECTS

We need these build artefacts (`codegen/built-*-spec.yml`) in version
control in order to generate Typescript types in the UI projects.

See the [`generate_types` script] in the CAS2 UI project for
example. This will be tweaked to use the generated `built-cas2-api-spec.yml`
file as this process is also unable to follow `$ref` links.

[bug (or missing feature)]: https://github.com/OpenAPITools/openapi-generator/issues/11251 

[OpenAPI $ref property]: https://swagger.io/docs/specification/using-ref/

[Kotlin-Spring OpenAPI code generator]: https://openapi-generator.tech/docs/generators/kotlin-spring/

[`generate_types` script]: https://github.com/ministryofjustice/hmpps-community-accommodation-tier-2-ui/blob/main/script/generate-types

cas2_swagger


src/main/resources/static/codegen/built-api-spec.yml
```diff
@@ -0,0 +1,7411 @@
...
+        - name: X-Service-Name
+          in: header
+          required: true
+          description: Only lost beds for this service will be returned
+          schema:
+            $ref: '#/components/schemas/ServiceName'
+        - name: probationRegionId
+          in: query
+          required: false
+          description: If provided, only lost beds for this region will be returned (CAS2)
```
Harriethw, line 3692: minor: It looks like CAS2 is mentioned in this description when it probably doesn't need to be?

Harriethw (approved):
Thanks @edavey 👏 
Would be good to get the others eyes on this too @tahb @pezholio @danlivings-dxw 

and for other projects to be aware it requires frontend changes to the file referenced in `generate-types`

pezholio (approved):
This looks really good - thanks 👍

src/main/resources/static/codegen/built-api-spec.yml
edavey: Well spotted! In this PR I've tried to limit changes to the structural to make the PR as clear as possible. i.e. this PR is "just" concerned with: 

- splitting the api.yml spec into two
- developing the Gradle tasks to generate Kotlin from 2 build artefacts

But yes, we certainly need to remove references like this to CAS2 in places where it no longer makes sense or is necessary. But I'd prefer to do that tidying in later PRs...

PR #1021, opened 2023-10-06
(480) Get managing team for CRN

pezholio:
As part of our reporting needs, we need to get the reporting team for a CRN (e.g. the Case Manager). Before, we were just fetching the first team of the logged in user, but it turns out this isn't what we need.

As part of the ambition to move away from Community API, the integrations team have built us a shiny new endpoint to give us all the information we need about a case, including the managing team.

This PR adds the client integration for this endpoint, and uses it when submitting an application and storing the domain event info.

Going forward, we'll want to replace all our calls to getting offender details with this new endpoint (as well as some bulk fetch endpoints, which have also been built), but let's start with this first.


pezholio:
[Get managing team for CRN](https://trello.com/c/Kkv6sA7E/480-get-managing-team-for-crn)

garethhallberg (approved):

PR #1025, opened 2023-10-10
Update type generation workflows

pezholio:
We now need to watch the build YML files for changes, not the source ones. I’ve also added a seperate CAS2 workflow to watch the CAS2 YAML file.


edavey (approved):
Thanks Stu for fixing this up.

PR #1026, opened 2023-10-10
Add support for CAS3 domain events

danlivings-dxw:
> See ticket [#1623 on the CAS3 Trello board](https://trello.com/c/JAHewQrO/1623-we-can-send-domain-events-without-requiring-an-application-id).

Following [this discussion](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1004#discussion_r1348479356) on #1004 where a need for a meaningful record identifier when the application ID is unavailable was raised, and [this Slack discussion with the probation integrations team](https://mojdt.slack.com/archives/C03KVQUFSER/p1696846088073159), we've reached an agreement to modify the structure of the domain events that we send, such that the `additionalInformation` section of the event can accept an optional `applicationId` and/or an optional `bookingId`.

While this change on its own is rather small, the above PR (#1004) at the time of writing already represents quite a large changeset, so we've decided to extract all of the setup work from that PR, apply the agreed changes to the domain event structure, and bundle it into a separate PR (this one) to be merged first, thus allowing the other PR to be solely focused on the implementation the 'person arrived' CAS3 domain event specifically.

### DIFF
--- a/src/main/resources/db/migration/all/20231009162205__add_support_for_cas3_to_domain_events_table.sql
+++ b/src/main/resources/db/migration/all/20231009162205__add_support_for_cas3_to_domain_events_table.sql
@@ -1,0 +1,9 @@
+ALTER TABLE domain_events ALTER COLUMN application_id DROP NOT NULL;
+
+ALTER TABLE domain_events ADD COLUMN booking_id UUID;
+ALTER TABLE domain_events ADD COLUMN service TEXT;
+
+UPDATE domain_events
+SET service = 'CAS1';
+
+ALTER TABLE domain_events ALTER COLUMN service SET NOT NULL;
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -189,6 +189,25 @@ tasks.register<org.openapitools.generator.gradle.plugin.tasks.GenerateTask>("ope
   importMappings.put("Instant", "java.time.Instant")
 }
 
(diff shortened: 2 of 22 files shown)


src/main/resources/db/migration/all/20231009162205__add_support_for_cas3_to_domain_events_table.sql
```diff
@@ -0,0 +1,9 @@
+ALTER TABLE domain_events ALTER COLUMN application_id DROP NOT NULL;
```
tahb, line 1: @pezholio just an FYI of this change in case you might want to put additional requirement protections in the layer above?

src/main/resources/db/migration/all/20231009162205__add_support_for_cas3_to_domain_events_table.sql
```diff
@@ -0,0 +1,9 @@
...
+
+ALTER TABLE domain_events ADD COLUMN booking_id UUID;
+ALTER TABLE domain_events ADD COLUMN service TEXT;
+
+UPDATE domain_events
+SET service = 'CAS1';
```
tahb, line 7: [It doesn't appear like CAS2 have merged their domain event work](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/981) in yet so this should be safe for all environments. @Harriethw FYI this would be a change to consider ontop of that draft.

tahb (approved):
👍

PR #1029, opened 2023-10-11
(143) Add daily metrics report

pezholio:
This adds a new report endpoint to fetch daily metrics for applications started, applications submitted, assessments made and bookings made as outlined in the Google sheet [here](https://docs.google.com/spreadsheets/d/12gF4sYmEk8p-7MMNLdEOmgFhPQ4Jhflt8fGGSeVUGWs/edit#gid=1592196053). Because the report has a dimension of days, rather than a row for each entity, the implementation is a bit more janky than I'd like, but it's better to do it this way than impelment a whole new report generator I think.


pezholio:

garethhallberg (approved):

PR #1044, opened 2023-10-16
Add application URL and booking URL to CAS3 domain events

danlivings-dxw:
> See [ticket #1639 on the CAS3 Trello board](https://trello.com/c/sPimu5Mx/1639-add-urls-to-the-person-arrived-and-departed-domain-event-endpoints).

Following a discussion with the integrations team, we've come to the agreement to provide URLs to the relevant booking and application (if available) associated with the event. This PR introduces `applicationUrl` and `bookingUrl` as properties on the two existing CAS3 domain events, as well as the infrastructure to generate appropriate URLs for future event types.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas3/DomainEventBuilder.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas3/DomainEventBuilder.kt
@@ -1,5 +1,6 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas3
 
+import org.springframework.beans.factory.annotation.Value
 import org.springframework.stereotype.Component
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.cas3.model.CAS3PersonArrivedEvent
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.cas3.model.CAS3PersonArrivedEventDetails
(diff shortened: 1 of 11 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas3/DomainEventBuilder.kt
```diff
@@ -98,9 +106,16 @@ class DomainEventBuilder {
...
           reasonDetail = null,
         ),
       ),
     )
   }
+
+  private fun TemporaryAccommodationApplicationEntity?.toUrl(): URI? =
+    this?.let { URI(applicationUrlTemplate.replace("#applicationId", it.id.toString())) }
+
+  private fun BookingEntity.toUrl(): URI =
```
tahb, line 119: When we do the referral.submitted event we won't have a booking but will have a referral so this may also need to be optional. Happy to leave that change until later if you'd prefer.

tahb (approved):
LGTM 🚀

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas3/DomainEventBuilder.kt
danlivings-dxw: I think in this situation we could just not have a `bookingId` field on the event and not call the method.

PR #1046, opened 2023-10-17
Show application timeline with working tests

garethhallberg:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ApplicationService.kt
```diff
@@ -778,4 +781,26 @@ class ApplicationService(
...
+
+  fun getAllInformationRequestEventsForApplication(applicationId: UUID): List<DomainEventSummary> {
+    val assessments = applicationRepository.findAllAssessmentsById(applicationId)
+
+    var allClarifications = mutableListOf<AssessmentClarificationNoteEntity>()
+    assessments.map {
+      allClarifications.addAll(it.clarificationNotes)
+    }
+    return allClarifications.map {
+      DomainEventSummary(
```
pezholio, line 799: This doesn't feel _quite_ right to me - this isn't really a Domain event as such, so I don't think it's right to marshal it into a `DomainEventSummary`. I reckon it might be better here to use a transformer in the service method itself and have two functions - one to convert a Domain event to the API object, and one to convert the clarification note into the API object.

src/main/resources/static/api.yml
```diff
@@ -1866,6 +1866,46 @@ paths:
...
+            $ref: '_shared.yml#/components/schemas/ServiceName'
+      responses:
+        200:
+          description: successful operation
+          content:
+            'application/json':
+              schema:
+                type: array
+                items:
+                  $ref: '_shared.yml#/components/schemas/DomainEventSummary'
```
pezholio, line 1896: I'd feel a bit happier if we had a new object for this - something like `TimelineEvent`, that way if we want to change any implementation of the `DomainEventSummary`, then we can keep the `TimelineEvent` the same (or vice versa)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/ApplicationsTransformer.kt
```diff
@@ -190,4 +191,15 @@ class ApplicationsTransformer(
     AssessmentDecision.REJECTED -> uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.AssessmentDecision.rejected
     null -> null
   }
+
+  fun transformTimelineToApi(events: List<JpaDomainEventSummary>): List<APIDomainEventSummary> {
+    return events.map {
+      it
+      APIDomainEventSummary(
+        type = it.type.toString(),
```
pezholio, line 199: Similarly, I think we might want to have a new TimelineEvent type here - again, they don't 
_quite_ map 1:1 with the domain events we have, and I think it gives us more flexbility to seperate the domain events from the timelines

pezholio (commented):
Looking good! Just some slightly nitpicky things about naming - happy to chat through it if you need 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/DomainEventEntity.kt
```diff
@@ -40,4 +44,11 @@ enum class DomainEventType {
   APPROVED_PREMISES_BOOKING_CANCELLED,
   APPROVED_PREMISES_BOOKING_CHANGED,
   APPROVED_PREMISES_APPLICATION_WITHDRAWN,
+  APPROVED_PREMISES_INFORMATION_REQUEST,
 }
+
+data class TimelineEvent(
```
pezholio, line 50: This might fit better in the models directory, as it should sit seperately from Domain Events

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/DomainEventEntity.kt
```diff
@@ -40,4 +44,11 @@ enum class DomainEventType {
   APPROVED_PREMISES_BOOKING_CANCELLED,
   APPROVED_PREMISES_BOOKING_CHANGED,
   APPROVED_PREMISES_APPLICATION_WITHDRAWN,
+  APPROVED_PREMISES_INFORMATION_REQUEST,
```
pezholio, line 47: I think we need a new enum for timeline event types, as we want them to sit apart from domain events, then we can have a mapping from one to the other.

pezholio (approved):
LGTM 👍

PR #1043, opened 2023-10-16
Making a provisional booking on a CAS3 premises emits a domain event

danlivings-dxw:
> See ticket [#1548 on the CAS3 Trello board](https://trello.com/c/t5XNjx7y/1548-send-a-domain-event-when-a-booking-is-made-provisionally).

This PR is the third of several intended to provide domain events to Delius upon certain actions occurring in the CAS3 service. It introduces the *booking provisionally made* event, which occurs when a request is made to the `POST /premises/{premisesId}/bookings` endpoint for the CAS3 service.

The structure of the CAS3 *booking provisionally made* event looks like this:
```jsonc
{
   "id": "364145f9-0af8-488e-9901-b4c46cd9ba37",
   "timestamp": "2022-11-30T14:53:44",
   "eventType": "accommodation.cas3.booking.provisionally-made",
   "eventDetails": {
     "personReference": {
       "crn": "C123456",
       "noms": "A1234ZX"
     },
     "applicationId": "484b8b5e-6c3b-4400-b200-425bbe410713", // optional
     "applicationUrl": "", // optional
     "bookingId": "14c80733-4b6d-4f35-b724-66955aac320c",
     "bookingUrl": "",
     "expectedArrivedAt": "2022-11-30T12:00:00", // We don't have a time so we default to midday
     "notes": "The actual time of arrival is unknown so has been defaulted to midday."
   }
 }
```

An event can be retrieved by sending a request to the `GET /events/cas3/booking-provisionally-made/{eventId}` endpoint. This and other CAS3 event endpoints have been namespaced in accordance with the approach proposed by CAS2 in [ADR 18](doc/architecture/decisions/0018-introduce-cas2-api-namespace.md).

PR #1047, opened 2023-10-17
Enable CAS2 domain events when application submitted

Harriethw:
## Context

CAS2 would like to start emitting Domain Events:
* Initially just to log when an Application has been marked as submitted.
* We don't currently have a defined consumer of this API yet, so we will save and expose a set of basic information. 

CAS2 Domain Events and MI reporting - CAS2 approach

This follows the implementation from cas3 - [introduced here](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1026) and [used here](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1043/files#diff-5bb0d41ddffa51dc89d73ebc6a4067d6eb5d0060569a9a2251db70a50b3fbfab) - and the work done to [implement a cas2 namespace in the API](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/965).

There are two parts to the PR:

* allowing CAS2 Domain Events to be viewed on an endpoint
  * this involves adding a new pattern to the Oauth configuration. I've added a test to ensure this does not affect access permissions for CAS1 events
* saving and emitting the Domain Event
  * This is triggered by an application being submitted
  
 I could split up if we think this is too complex, but I wanted to get feedback on the complete picture of the end-to-end process, and the approach has been mostly copied from CAS3 so hopefully not too surprising.

## questions
* I've noticed the CAS3 events endpoint for `OAuth2ResourceServerSecurityConfiguration.kt` doesn't have a specific role/auth assigned to it. Is that TBC?
* The timeline work that was recently introduced requires an exhaustive `when` in the `transformDomainEventTypeToTimelineEventType` function - I've set a default to throw an error is domain event type is not recognised as we aren't yet handling timelines in CAS2, not sure if this is the right approach or should just fail silently.
* More of a style question - Elsewhere when we've used CAS2 in a camelcase context it's been capitalised like so `ApplicationsCas2Delegate`, `Cas2ApplicationsTest` . In the CAS3 implementation of domain events the classes use all caps `CAS3PersonArrivedEvent`. I'm going to continue to use camel case for ours to stick with our current patterns but open to feedback.


src/main/resources/static/cas2-domain-events-api.yml
```diff
@@ -0,0 +1,156 @@
...
+        staffIdentifier:
+          type: integer
+          format: int64
+          example: 1501234567
+        name:
+          type: string
+          example: John Smith
+        username:
+          type: string
+          example: JohnSmithNPS
```
edavey, line 112: `JohnSmithNPS` is a probation style username (the suffix is for "National Probation Service" I think).

`SMITHJ_GEN` would be a more accurate example of a NOMIS/DPS username

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/DomainEventService.kt
```diff
@@ -0,0 +1,128 @@
...
+
+    val data = when {
+      enumTypeFromDataType(T::class) == domainEventEntity.type ->
+        objectMapper.readValue(domainEventEntity.data, T::class.java)
+      else -> throw RuntimeException("Unsupported DomainEventData type ${T::class.qualifiedName}/${domainEventEntity.type.name}")
+    }
+    return DomainEvent(
+      id = domainEventEntity.id,
+      applicationId = domainEventEntity.applicationId,
+      bookingId = domainEventEntity.bookingId,
```
edavey, line 58: hmm, this `bookingId`... As we don't have bookings would it be clearer to explicitly set this to `null`?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2DomainEventTest.kt
```diff
@@ -0,0 +1,84 @@
...
+      .uri("/events/cas2/application-submitted/e4b004f8-bdb2-4bf6-9958-db602be71ed3")
+      .exchange()
+      .expectStatus()
+      .isUnauthorized
+  }
+
+  @Test
+  fun `Get CAS2 Application Submitted Event without ROLE_CAS2_EVENTS returns 403`() {
+    val jwt = jwtAuthHelper.createClientCredentialsJwt(
+      username = "username",
```
edavey, line 26: do you think it would be clearer if you set the `roles` explicitly, even better to a real role which is not permitted access? e.g.

```kt
    val jwt = jwtAuthHelper.createClientCredentialsJwt(
      username = "username",
      roles = listOf("ROLE_APPROVED_PREMISES_EVENTS"),
    )
```

edavey (approved):
This is a thing of beauty! 😎 

Great work on a favourite topic of mine! I've made a couple of tiny suggestions.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/DomainEventService.kt
Harriethw: ah yeah, thanks this one slipped through 🐟

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2DomainEventTest.kt
Harriethw: great idea!

pezholio (approved):
This looks great from my POV 👍

danlivings-dxw (approved):
LGTM 👍 

Regarding question 1: we've decided to reuse the events role for Approved Premises in CAS3 as it's the same integration service on the other end getting the event information.
Regarding question 3: there's no particular reason to prefer `CAS3WhateverEvent` over `Cas3WhateverEvent`. In my head the general unofficial rule is that if it's defined in the YAML file it's capitalised as `CAS3` but if it's defined as a Kotlin class it's capitalised as `Cas3`, so that I can tell at a glance whether I'm dealing with an API model or some internal structure.

PR #1049, opened 2023-10-18
(518) External NACRO assessor logs in to view submitted CAS2 applications

edavey:
External users from NACRO will be logging into CAS to perform specific operations on *submitted* CAS2 applications only.

The users are managed on HMPPS auth and will have the role `CAS2_ASSESSOR` (the JWT 'authority' `ROLE_CAS2_ASSESSOR`). The `auth_source` of their claim will be `auth`. 

See AP-Tools [PR 48](https://github.com/ministryofjustice/hmpps-approved-premises-tools/pull/48) for a CAS2_ASSESSOR user for use in local development.

In this PR we:

- allow an external user to authenticate and create a representation for each one in a new `external_users` table
- allow holders of the `CAS2_ASSESSOR` role to `GET` a list of submitted applications or an individual submitted application.

#### A note on ApplicationStatus versus 'has been submitted'
  
Note that for "submitted" we test for the presence of a  
value in the `Cas2Application.submittedAt` timestamp. An  
application can be submitted exactly once and can never  
become "unsubmitted".  
  
i.e. "submitted" is an event rather than a status which  
changes from time to time. This is contrary to what is  
currently implied by ApplicationStatus:  
  
```  
ApplicationStatus:  
  type: string  
  enum:  
    - inProgress  
    - submitted  
    - requestedFurtherInformation  
    - pending  
    - rejected  
    - awaitingPlacement  
    - placed  
    - inapplicable  
    - withdrawn  
```  
  
In CAS2 we will be requiring external users from NACRO  
(logging in with the `CAS2_ASSESSOR` role) to provide updates on  
their progress in taking submitted applications through the  
assessment and placement process which they manage in their  
own NACRO / OpenHousing systems.  
  
So it will be clearer for CAS2 to model something  
like a `CAS2ExternalWorkflowStatus` which can go though  
states such as:  
  
- `pendingAssessment`  
- `furtherInfoRequested`  
- `applicationWithdrawn`  
- `accepted / rejected`  
- `placed`  
- `bookingCancelled`  
  
We probably need to implement a lightweight state-machine to  
ensure that only valid transitions are made. e.g. that an  
application isn't moved from `pendingAssessment` to  
`bookingCancelled` without having been first `accepted`.  
  
As these workflow updates will be made by NACRO outside of  
their actual workflow perhaps it will be necessary for us to  
ask for the date on which the status change occurred. e.g.  
"When was this application accepted?" as this is not  
necessarily when the event is being reported in our  
service...

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/ApplicationsController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/ApplicationsController.kt
@@ -11,7 +11,6 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.NewApplication
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.SubmitCas2Application
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.UpdateApplication
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas2ApplicationEntity
-import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.NomisUserEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.BadRequestProblem
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.ConflictProblem
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.ForbiddenProblem
@@ -47,7 +46,7 @@ class ApplicationsController(
 
     val applications = applicationService.getAllApplicationsForUser(user)
 
-    return ResponseEntity.ok(applications.map { getPersonDetailAndTransformToSummary(it, user) })
+    return ResponseEntity.ok(applications.map { getPersonDetailAndTransformToSummary(it) })
   }
 
   override fun applicationsApplicationIdGet(applicationId: UUID):
@@ -64,7 +63,7 @@ class ApplicationsController(
     }
 
     if (application != null) {
-      return ResponseEntity.ok(getPersonDetailAndTransform(application, user))
+      return ResponseEntity.ok(getPersonDetailAndTransform(application))
     }
     throw NotFoundProblem(applicationId, "Application")
   }
@@ -75,7 +74,7 @@ class ApplicationsController(
     val nomisPrincipal = httpAuthService.getNomisPrincipalOrThrow()
     val user = userService.getUserForRequest()
 
-    val personInfo = offenderService.getFullInfoForPersonOrThrow(body.crn, user)
+    val personInfo = offenderService.getFullInfoForPersonOrThrow(body.crn)
 
     val applicationResult = applicationService.createApplication(
       body.crn,
@@ -125,7 +124,7 @@ class ApplicationsController(
       is ValidatableActionResult.Success -> validationResult.entity
     }
 
-    return ResponseEntity.ok(getPersonDetailAndTransform(updatedApplication, user))
+    return ResponseEntity.ok(getPersonDetailAndTransform(updatedApplication))
   }
 
   override fun applicationsApplicationIdSubmissionPost(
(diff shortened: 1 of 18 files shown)


edavey:
[NACRO assessor can login](https://trello.com/c/hJnd84x0/518-nacro-assessor-can-login)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/ApplicationsController.kt
```diff
@@ -40,31 +41,41 @@ class ApplicationsController(
...
+  private val externalUserService: ExternalUserService,
 ) : ApplicationsCas2Delegate {
 
   override fun applicationsGet(): ResponseEntity<List<ApplicationSummary>> {
-    val user = userService.getUserForRequest()
-
-    val applications = applicationService.getAllApplicationsForUser(user)
-
-    return ResponseEntity.ok(applications.map { getPersonDetailAndTransformToSummary(it, user) })
+    val authenticatedPrincipal = httpAuthService.getCas2AuthenticatedPrincipalOrThrow()
```
Harriethw, line 48: do I understand it correctly then that there are broadly two places where we impose permissions:
- the `OAuth2ResourceServerSecurityConfiguration.kt` file where we impose _role-based authorization_ on paths. So, e.g. only roles `CAS2_ASSESSOR` and `PRISON` will be able to reach the cas2/applications endpoints.
- the `getCas2AuthenticatedPrincipalOrThrow` which imposes _auth_source authorization_, so e.g. we look at the auth source and if it isn't `nomis` or `auth` they are logged out?

Harriethw (approved):
thanks @edavey 🙌

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/ApplicationsController.kt
edavey: yes, that's how it works at present. I think that we should develop this towards:

#### 1. `OAuth2ResourceServerSecurityConfiguration`
Continue to enforce role-based access rules working on both permitted HTTP method and path. But as previously discussed use custom matchers in the Spring Security filter chain, something like:
```kotlin
val cas2ApplicantAccess: RequestMatcher = { (request) ->
  request.getServletPath().matches("cas2/applications/**")
  // ...other conditions
}

val cas2AssessorAccess: RequestMatcher = { (request) ->
  request.getServletPath().matches("cas2/submissions/**")
  // ...other conditions
}

val cas2AdminAccess: RequestMatcher = { (request) ->
  request.getServletPath().matches("cas2/applications/**")
  // ...other conditions
}

http {
  authorizeHttpRequests {
    authorize(cas1Access, hasRole("PROBATION"))
    authorize(cas2ApplicantAccess), hasAnyRole("POM_USER", "CAS2_ADMIN"))
    authorize(cas2AssessorAccess), hasRole("CAS2_ASSESSOR"))
    authorize(cas2AdminAccess, hasRole("CAS2_ADMIN"))
    authorize(cas2MangementInfoAccess, hasRole("CAS2_MI"))
    authorize(cas3Access, hasRole("PROBATION"))
  }
}
```

These custom matchers should be unit-testable, allowing us to remove slow access tests at the integration test level.

### 2. `HttpAuthService`

`HttpAuthService.getCas2AuthenticatedPrincipalOrThrow()` tests the logged in user's JWT's claim `auth_source` and now ejects anyone who hasn't been authenticated by NOMIS or externally. We use it in our controller to i) perform that control and ii) to distinguish between external NACRO and internal NOMIS/DPS users. I think it would be good to develop this into a method such as `getCurrentuser()` which is run on every controller action to provide a persisted `Cas2User` record which is built from either a `NomisUser` or an `ExternalUser` and offers a standard interface e.g.

```
type (nomis/nacro)
userIdentifier
name
email
phone
roles (POM_USER, CAS2_ADMIN)
isEnabled
```

This user could then be passed to services by the controller where more fine-grained controls are needed, or where auditing or domain events need user information.

PR #1054, opened 2023-10-18
Feature/1644 expect a eligibility reason for cas3 referrals

tahb:
The `eligiblityReason` is a new field we have to put into our MI reports. This sits alongside the `isEligible` which we leave as it is since it's already been released.

[I used this earlier PR as a reference.](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/966)

[We will need to migrate the data out of the blob and into this field as a separate operation.](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1059)

Frontend changes: https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/693


danlivings-dxw (approved):
LGTM 👍

PR #1057, opened 2023-10-19
(485) Retire “Error in Booking” reason

pezholio:
This is a duplicate, so we should retire it.


pezholio:
[There seem to be two `Error in booking` cancellation reasons - we should remove one](https://trello.com/c/IRhKlvMZ/485-there-seem-to-be-two-error-in-booking-cancellation-reasons-we-should-remove-one)

garethhallberg (approved):

PR #1059, opened 2023-10-19
Backfill cas3 referrals so they have eligibility reasons

tahb:
We’ll need this historic data for referrals created so far in production.


danlivings-dxw (approved):
LGTM 👍 

Interesting to see that Postgres gives convenient operators for getting information out of JSON data.

PR #1060, opened 2023-10-19
Fix/revert confirmed and cancelled booking domain event

tahb:
[It appears booking cancellation domain events is broken in prod and a handled error is bubbling to the frontend](https://ministryofjustice.sentry.io/issues/4558221352/?referrer=slack&notification_uuid=e873c170-82c5-4587-9524-41fc7dba2bec&alert_rule_id=12408391&alert_type=issue). It also failed e2e tests but was pushed through with the **confirmed** domain event which may be ok in of itself.

Given these two features have been rebased on top of each other I revert both features (neither used in prod) to enable a faster revert.


pezholio (approved):
LGTM - Reverts make me a bit squishy, but given the urgency, let's get it in

PR #1061, opened 2023-10-19
Add an `isRestricted` argument to a `FullPerson`

pezholio:
This will allow us to tell a user that the Person they have accessed is a Limited Access Offender, meaning even though they may be able to acces this offender’s details, other users may not be able to.


pezholio:
[LAO next steps](https://trello.com/c/dqA5q4tb/405-lao-next-steps)

garethhallberg (approved):

PR #1065, opened 2023-10-20
Fix   increase memory size

garethhallberg:


pezholio (approved):

tahb:
I'm looking back at this change to try and understand what's going on with the memory problems. What was the rationale behind this change and the selection of 40?

pezholio:
Going through a bunch of other repos, it seems like 40MB was about the average other projects had chosen. This was a red herring anyway, so can probably be reverted

PR #1069, opened 2023-10-23
(413) Add sentence type to Application Report

pezholio:
This is needed for performance reporting. The original plan was to have a breakddown by release type and sentence type, but we’ve agreed that adding this to the application report was sufficient.


pezholio:

garethhallberg (approved):

PR #1070, opened 2023-10-23
Bump Guava version

pezholio:
Addresses https://nvd.nist.gov/vuln/detail/CVE-2023-2976

This is used by kotlinx/dataframe, but is not included directly.


edavey (approved):
Thanks!

PR #1072, opened 2023-10-23
463 withdraw options for submitted applications

garethhallberg:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AssessmentService.kt
```diff
@@ -783,6 +793,11 @@ class AssessmentService(
     return AuthorisableActionResult.Success(referralHistoryNoteEntity)
   }
 
+  fun updateAssessmentWithdrawn(assessmentId: UUID) {
```
pezholio, line 796: I'm not sure this will save the assessment here, so you'll need to call `assessmentRepository.save` after you set `isWithdrawn`.

pezholio (commented):
Just one thing here - I'm not sure `updateAssessmentWithdrawn` is saving the assessment.

pezholio (approved):

PR #1075, opened 2023-10-24
(APS-14) Hide withdrawn assessments

pezholio:
Now assessments can be withdrawn, we need to hide them on the assessments and application listing pages.


pezholio:
We're going to do this filtering on the frontend, so closing this

garethhallberg (approved):

PR #1076, opened 2023-10-25
Implement submissions endpoint for assessors

edavey:
## Add a new `/cas2/submissions` endpoint to represent submitted applications

In this PR I make a case for introducing a new REST resource to represent applications 
which have been submitted.

The other CAS services introduce a new DB entity (`Assessment`), but I think this is
overkill for our needs. We can achieve the "separation of concerns" which we need
by introducing a new API model (REST representation) to present the existing application
record in the post-submission world.

### In detail

When the UI asks that API for a list of applications, what is the appropriate
range of filtering which should be implicitly applied by the API?

`GET /applications`
  - APPLICANT: all applications which I've created (filter on associated user)
  - ASSESSOR: all application which have been submitted (filter on submittedAt
    presence)
  - ADMIN: all applications (no filter)

`GET /application/{applicationId}`
  - APPLICANT: 403 if I didn't create the application
  - ASSESSOR: 404 if the application hasn't yet been submitted

I think it's reasonable to implicitly filter a list of resources according to
some access rules, which could become slightly complex, such as:

- APPLICANT can see applications they've created
- APPLICANT can see applications made to their region
- APPLICANT can see applications where an assessor has named them in a request
  for further information

It becomes more opaque if we combine this with rules which are based on the
**state of the application** as well as on the **role of the user**, e.g.

- ASSESSOR can see only applications which have been submitted

### Beyond "apply"

The submission of an application is an important milestone (consider the
`application.submitted` domain event). The mutable form `data` is stored in a
permanent read-only `document` as a full and reliable record of the risks and
needs at the time of application.

The lifecycle of the accommodation workflow moves to a new stage on submission
of the application. It's not a  stage out which the application can transition,
it's an event which cannot be undone.

If we agree that the API should not handle filtering submitted applications
implicitly, how should the UI request them explicitly? Two options:

1. url params, e.g. `GET /cas2/applications?filters=[submitted]`. In this
   approach the complexity at the API is retained: the controller actions still
   needs to handle the logic of whether to return all applications or just those
   which have been submitted.
2. dedicated url e.g. `GET /cas2/submissions`. In this approach a new endpoint
   and controller (and API representation) can be used.

### Re-presenting Application as Submission / SubmittedApplication

I believe that by giving this different entity its own name e.g. **submission**
we can make our code simpler.

The new `cas2/submissions` endpoint can return a new representation of an
application which has been through the submission process. This
SubmittedApplication or SubmittedApplicationSummary can do a much better (more
specialised) job of meeting our needs as we leave the "apply" journey behind.
e.g.

- `data` is no longer relevant or helpful. Only exposing `document` is clearer
  and less error prone in displaying a submitted application.
- we can add in a list of `StatusUpdates` to our new `SubmittedApplication`
  representations

submissions_endpoints

cas2_submitted_application

cas2_submitted_application_summary

### Not included

This PR does not include moving `POST /cas2/applications/submission` to the new controller as 
`POST /cas2/submissions` -- but this would now be the correct way to "submit an application":
i.e. "create a submission".

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/OAuth2ResourceServerSecurityConfiguration.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/OAuth2ResourceServerSecurityConfiguration.kt
@@ -60,6 +60,7 @@ class OAuth2ResourceServerSecurityConfiguration {
         authorize(HttpMethod.GET, "/events/cas2/application-submitted/*", hasAuthority("ROLE_CAS2_EVENTS"))
         authorize(HttpMethod.GET, "/events/**", hasAuthority("ROLE_APPROVED_PREMISES_EVENTS"))
         authorize(HttpMethod.GET, "/cas2/applications/**", hasAnyRole("CAS2_ASSESSOR", "PRISON"))
+        authorize(HttpMethod.GET, "/cas2/submissions/**", hasRole("CAS2_ASSESSOR"))
         authorize("/cas2/**", hasAuthority("ROLE_PRISON"))
         authorize(anyRequest, hasAuthority("ROLE_PROBATION"))
       }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/OAuth2ResourceServerSecurityConfiguration.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/OAuth2ResourceServerSecurityConfiguration.kt
@@ -59,7 +59,6 @@ class OAuth2ResourceServerSecurityConfiguration {
         authorize(HttpMethod.DELETE, "/internal/room/*", permitAll)
         authorize(HttpMethod.GET, "/events/cas2/application-submitted/*", hasAuthority("ROLE_CAS2_EVENTS"))
         authorize(HttpMethod.GET, "/events/**", hasAuthority("ROLE_APPROVED_PREMISES_EVENTS"))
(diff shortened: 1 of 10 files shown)


edavey:
The UI will now have "submitted" versions of the Application types returned from these endpoints.

```
server/@types/shared/models/Cas2SubmittedApplication.ts

-> 

/* generated using openapi-typescript-codegen -- do no edit */
/* istanbul ignore file */
/* tslint:disable */
/* eslint-disable */

import type { AnyValue } from './AnyValue';
import type { ApplicationStatus } from './ApplicationStatus';
import type { Person } from './Person';

export type Cas2SubmittedApplication = {
    id: string;
    person: Person;
    createdAt: string;
    createdByUserId: string;
    schemaVersion: string;
    outdatedSchema: boolean;
    document?: AnyValue;
    status: ApplicationStatus;
    submittedAt?: string;
};


server/@types/shared/models/Cas2SubmittedApplicationSummary.ts

-> 

/* generated using openapi-typescript-codegen -- do no edit */
/* istanbul ignore file */
/* tslint:disable */
/* eslint-disable */

import type { ApplicationStatus } from './ApplicationStatus';
import type { Person } from './Person';

export type Cas2SubmittedApplicationSummary = {
    id: string;
    createdByUserId: string;
    status: ApplicationStatus;
    person: Person;
    createdAt: string;
    submittedAt?: string;
};
```

edavey:
[Spike to demonstrate /submissions endpoint to represent SubmittedApplications](https://trello.com/c/Pmb740n0/582-spike-to-demonstrate-submissions-endpoint-to-represent-submittedapplications)

Harriethw (approved):
Thanks Ed, seeing it like this can see it solves quite a few problems. I'm happy to go ahead with this way

PR #1078, opened 2023-10-25
Include referral columns in booking report

danlivings-dxw:
> See [ticket #1516 on the CAS3 Trello board](https://trello.com/c/fixhLvcQ/1516-include-new-referral-columns-in-the-mi-report-download).

There are several columns that are now required for MI reporting for bookings that have applications. For bookings not associated with an application these must be empty, but otherwise should be included in the booking report.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/reporting/generator/BookingsReportGenerator.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/reporting/generator/BookingsReportGenerator.kt
@@ -1,6 +1,7 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.reporting.generator
 
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.BookingEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.TemporaryAccommodationApplicationEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.reporting.model.BookingsReportRow
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.reporting.properties.BookingsReportProperties
 import java.time.LocalDate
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/reporting/model/BookingsReportRow.kt
```diff
@@ -3,6 +3,15 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.reporting.model
...
 data class BookingsReportRow(
+  val referralId: String?,
+  val referralDate: LocalDate?,
+  val riskOfSeriousHarm: String?,
+  val sexOffender: Boolean?,
+  val needForAccessibleProperty: Boolean?,
+  val historyOfArsonOffence: Boolean?,
+  val dutyToReferMade: Boolean?,
+  val dateDutyToReferMade: LocalDate?,
+  val isReferralEligibleForCas3: Boolean?,
```
tahb, line 14: `referralEligibility` is now ready to add so we could add it here or as another new PR?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/reporting/generator/BookingsReportGeneratorTest.kt
```diff
@@ -634,4 +639,93 @@ class BookingsReportGeneratorTest {
     assertThat(actual.count()).isEqualTo(1)
     assertThat(actual[0][BookingsReportRow::accommodationOutcome]).isNull()
   }
+
+  @Test
+  fun `The referral columns are empty if there is no application for the booking`() {
```
tahb, line 644: 👍

tahb (commented):
I think we should avoid hashing/obfuscating our application id, am I missing something?

We might also want to add the new `eligibilityReason` column in now for convenience but I'll leave that one up to you.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/reporting/model/BookingsReportRow.kt
danlivings-dxw: I've added this column as `referralEligibilityReason`.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/reporting/generator/BookingsReportGenerator.kt
```diff
@@ -1,16 +1,29 @@
...
 import java.time.temporal.ChronoUnit
 
 class BookingsReportGenerator : ReportGenerator<BookingEntity, BookingsReportRow, BookingsReportProperties>(BookingsReportRow::class) {
 
   override val convert: BookingEntity.(properties: BookingsReportProperties) -> List<BookingsReportRow> = {
+    val application = this.application as? TemporaryAccommodationApplicationEntity
+
     listOf(
       BookingsReportRow(
+        referralId = application?.id?.toShortBase58(),
```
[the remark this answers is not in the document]
danlivings-dxw: I'd assumed that we just wanted to present an arbitrary unique reference here similar to what we do in the bed usage/utilisation reports that can be used to quickly refer to a particular application by the users beyond the boundaries of the service (such as email or Slack messages), rather than this being a potential source of support requests. I think UUIDs aren't especially human readable as they're pretty long and if you're not familiar with them have characters that can be easily confused with invalid ones (**0** and **O** for example). Regardless, I've changed this to a UUID.

tahb (approved):
👍

PR #1085, opened 2023-10-30
Move CAS2 application submission endpoint to /cas2/submissions

edavey:
**This should be merged with the corresponding change at the UI https://github.com/ministryofjustice/hmpps-community-accommodation-tier-2-ui/pull/273**

CAS2 Applications are now submitted using a POST to the submissions endpoint:

`POST /cas2/submissions` with
  `{ translatedDocument: JSON, applicationId: UUID}`

rather than

`POST /cas2/applications/{applicationId}/submission` with
  `{ translatedDocument: JSON, type: 'CAS2'}`

Representing a submitted application as a "submission" resource with a type of `CAS2SubmittedApplication`
allows us to avoid adding to complexity to the `ApplicationsController`.

Note also that:

- the `applicationId` is now received as the `applicationId`
  form data param rather than being in the path.

- the `type` param is no longer part of the interface, as this was not being used.

post_submissions


Harriethw (approved):
Thanks Ed, really easy to follow and good to see some things being tidied up.

PR #1091, opened 2023-10-31
Withdrawing placement application also withdraws placement requests a…

garethhallberg:
…nd emails user. Integration and unit tests to prove it works.


pezholio (approved):
LGTM 👍

PR #1092, opened 2023-10-31
Install detekt as a static code analyser

tahb:
As part of the investigation into memory errors it was decided that we would review options for static code analysers. The hope was that this would be a quicker way to get more insight into performance shaped problems, improving the chances of us catching them before being released.

## Is this going to catch all performance problems?

No. [Changes to sql queries](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1069/files#diff-9c2bd1878eadcb62699d706a836c2185895e52ccee5b05983525b613f6bac386R32) are unlikely to be caught. [A list of performance problems that will be targetted can be read here](https://detekt.dev/docs/rules/performance/).

We probably need to keep going and introduce Hibernate Profiler to check our queries using the ORM.

As for checking raw sql I'm not sure we can put checks in place for that?

## Changes

- [We asked around and detekt was recommended.](https://mojdt.slack.com/archives/C69NWE339/p1698668913072199)
- We start with the default detekt rules for everything including style, naming, complexity and performance. We include instructions on how to configure our choices going forwards
- We use the tool to create a baseline, ignoring all past rule violations allowing us to move quickly and catch any new violations whilst tolerating the old until we have time to address them
- We add it as a check to CI

## Notes

- For the first release I've made it pass regardless so I can check the CI output without blocking the pipeline for everyone. If I check it and there's no feedback I know the baseline works and I can quickly enable it with a follow up change.
- We have ktlint installed and provided by the central moj library. ktlint appears to be focused on format whilst detekt focuses on code analysis for things like complexity. [Apparently both can be installed in harmony](https://hofstede-matheus.medium.com/improve-code-quality-with-ktlint-detekt-and-git-hooks-d173722594e4).


script/test
```diff
@@ -14,4 +14,5 @@ echo "==> Running ktlint..."
 echo "==> Running tests..."
 ./gradlew test
 
-
+echo "==> Running static code analyser..."
+./gradlew detekt
```
pezholio, line 18: I feel like we should run this first, so we fail fast on any detekt issues before running the whole test suite

.circleci/config.yml
```diff
@@ -145,6 +145,19 @@ jobs:
...
+      _JAVA_OPTIONS: "-Xmx3g"
+      GRADLE_OPTS: "-Dorg.gradle.daemon=false -Dorg.gradle.workers.max=2"
+    docker:
+      - image: cimg/openjdk:17.0
+    steps:
+      - checkout
+      - attach_workspace:
+          at: .
+      - run:
+          name: Run ktlint
```
pezholio, line 159: This should be `Run detekt` I think

pezholio (approved):
LGTM - thanks for doing this. Just a couple of teeny things - I take it we'll revert https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1092/commits/d7be70670a02bcc03069d0bcac583ac9bb41bbee before merging?

script/test
tahb: I considered this too. Running it latter has the advantage of getting test feedback as you go without getting interrupted by style or naming nitpicking, especially as we code we might be aware of these problems but are immediately interested in test feedback.

How do other api devs run the tests? Given they take so long I suppose they are running the unit tests outside of script/test for faster feedback? In which case yes adding it to the top here feels right.

tahb:
I was thinking to revert https://github.com/ministryofjustice/hmpps-approved-premises-api/commit/d7be70670a02bcc03069d0bcac583ac9bb41bbee after merging and seeing it run through CI, as to not block the pipeline due to missing something.

script/test
pezholio: To be honest, I tend to run the tests in IntelliJ, and only run the full suite in CI, so very rarely use this script

tahb: From a quick poll it seems that this isn't used regularly so the ordering arguable doesn't matter much. Let's keep it grouped with linting that happens up front as you suggest 👍

PR #1100, opened 2023-11-03
Include personal data for bookings report

danlivings-dxw:
> See [ticket #1535 on the CAS3 Trello board](https://trello.com/c/b8NYSvNw/1535-report-mi-spreadsheet-includes-personal-data).

This PR introduces columns containing personal data to the bookings report. These columns are:
- `personName`
- `pncNumber`
- `gender`
- `ethnicity`
- `dateOfBirth`

Of these columns, four are populated as expected - `pncNumber` is currently not available and is always presented as a blank cell. This is because we do not store personal information within our service and have to obtain it elsewhere. Although we cache results from the Community API, the number of results expected in a typical spreadsheet is between 200 and 400 rows, and performing a request to the cache for each CRN incurs an unacceptable performance degradation (see below). The alternative solution is to use the AP-Delius Integration API, which provides an endpoint for accessing personal data in bulk, but this endpoint only provides a subset of the personal data and does not currently include the PNC number.

### Performance testing
We are beginning to run into performance limits in various aspects of the service. Including another external integration has the potential to introduce significant performance impacts, so a Gatling test has been introduced to quantify the impact of this integration on creating booking reports.

The performance test is structured in the following way:
- 1200 bookings are created for the same CRN, distributed randomly over a period spanning from 60 days before to 60 days after the current date, using the following steps:
  - creating a premises
  - creating a room in that premises
  - creating an application
  - updating that application
  - submitting that application
  - creating a booking for that room
- 100 reports are generated for random months over this period.

Each run was performed on an M1 Macbook Pro under similar background load conditions. Prior to each test, the database was cleared and all caches purged to ensure each run was fully independent.

The control test was measured using the current `main` branch:
Control test results measured using the current `main` branch

The single CRN endpoint test:
Test results measured using the single CRN endpoint

The bulk data endpoint test:
Test results measured using the bulk data endpoint

| Control | 100.0% | 8.658 | 49.901 | 29.526 | 31.704 | 11.488 |
| Single CRN[2] |  9.8% | 47.878 | 57.193 | 53.311 | 54.885 | 3.429 |
| Bulk CRNs | 100.0% | 8.609 | 48.642 | 26.669 | 26.485 | 12.527 |

[2]: Calculated from a reduced sample of 61 requests due to how Gatling schedules users in its simulations.

### DIFF
--- a/src/gatling/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/gatling/simulations/ApplyJourneyStressSimulation.kt
+++ b/src/gatling/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/gatling/simulations/ApplyJourneyStressSimulation.kt
@@ -12,11 +12,13 @@ import io.gatling.javaapi.http.HttpDsl.status
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.NewApplication
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ServiceName
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.SubmitTemporaryAccommodationApplication
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.UpdateApplicationType
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.UpdateTemporaryAccommodationApplication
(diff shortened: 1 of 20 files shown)


sudhakar-plr:
It looks ok to me.

sudhakar-plr (approved):

PR #1101, opened 2023-11-03
Aps23 users must be able to add manual notes into the timeline view

garethhallberg:
1. New endpoint /applications/{applicationId}/notes. 
2. Database migration.
3. New ApplicationTimelineNoteEntity and ApplicationTimelineNoteRepository and ApplicationTimelineNoteService.
4. Tests for creating notes.
5. Tweaked Timeline endpoint return to include these notes as a new TimlineEventType.
6. Tests that show the timeline endpoint can return these notes in payload.


pezholio (approved):
LGTM - would be good to spend some time doing some further refactoring.

Not for now (becuase that would be extremely annoying!) but it'd be good in future to seperate out the refactors from the features to make it easier to review. I know that's hard though, because we don't often see the Detekt errors until we run the tests!

PR #1103, opened 2023-11-06
Exclude Spring autowiring from `complexity.LongParameterList` detekt rule

danlivings-dxw:


detekt.yml
```diff
@@ -22,6 +25,11 @@ complexity:
   LargeClass:
     threshold: 1000
   LongParameterList:
-    threshold: 12
```
tahb, line 25: Nice. Does this really account for all previous instances of this violation?

tahb (approved):

PR #1109, opened 2023-11-07
Improve handling of situations where person information is unavailable

danlivings-dxw:
An [outage in the Community API this morning](https://mojdt.slack.com/archives/C03KVQUFSER/p1699359142045559) resulted in various errors being returned by our service, including [this `ClassCastException`](https://ministryofjustice.sentry.io/issues/4608905206/?alert_rule_id=14565024&alert_type=issue&notification_uuid=02bbb030-e496-4646-8a9a-9da39dd81bc5&project=4503931792392192&referrer=slack) when trying to cast a `PersonInfoResult.Unknown` to a `PersonInfoResult.Success.Full`.

This PR addresses the `ClassCastException` and several related exceptional responses where a successful one is possible. This has been achieved through two steps:

1. Widening JPA-to-API transformers that restrict the range of allowed `PersonInfoResult`s as parameters to accept all variants. This is safe as in every instance the transformer only delegates this to the `PersonTransformer` to be transformed, which can handle all variants.
2. Replacing calls to throwing methods of getting `PersonInfoResult`s found in `PersonUtils.kt` with the non-throwing\* `OffenderService.getInfoForPerson`.

The `PersonUtils.kt` file has also been removed as no functions within it are in use following this refactoring.

---

\*: `OffenderService.getInfoForPerson` *does* still throw exceptions in certain weird situations at the interface with the Community API, but unpicking those situations to handle them gracefully seems fairly complex and is beyond the scope of this PR.


tahb (approved):
Am I following correctly here.

- The error handling within `offenderService.getInfoForPerson` already returns `Success`, `NotFound`, `Unknown` and `Failure` for person information 
-  When an `Unknown` or `Failure` event occurs we **can** return a limited person info object which is compatible with the frontend
- We delete `getFullInfoForPersonOrThrow` etc in favour of the above so we throw less often
- `ApplicationPost` does need to do some custom error handling on the write because restricted should still result in a throw

I think this makes sense to me. @pezholio this affects a bunch of things so be good to have your eyeballs on this too.

tahb:
Oh and thanks once again for taking the initiative on this one 🙌

edavey (approved):
Nice work, thanks

tahb:
@edavey I was speaking with Harriet. We wondered if to apply this improvement to CAS2 (due to the namespaces) more work may be required eg. https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/fc98157bb57c005ea2ff7002a3107f2e009d7fde/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/ApplicationsController.kt#L76C11-L76C11 and the redundancy of https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/fc98157bb57c005ea2ff7002a3107f2e009d7fde/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/util/cas2/PersonUtils.kt#L17. Will leave that with you both, just cautious of this not helping you too.

pezholio (approved):
LGTM 👍

PR #1110, opened 2023-11-07
Chore/seed data for south west in local

tahb:
In order to test booking search pagination I wanted to view more than bookings in an individual state.

In order to set this up we first seed local data for the jimsnowldap user. This sits in addition to premises data set up for e2e tests run by the frontend, in case that's run locally for the EofE region.

### DIFF
--- a/src/main/resources/db/migration/local/R__4_create_bookings_for_cas3_south_west.sql
+++ b/src/main/resources/db/migration/local/R__4_create_bookings_for_cas3_south_west.sql
@@ -50,6 +50,486 @@ VALUES
   )
 ON CONFLICT(id) DO NOTHING;
 
+--- Add a Booking arriving soon ---
+
+INSERT INTO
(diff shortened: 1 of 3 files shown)


src/main/resources/db/migration/local/R__4_create_bookings_for_cas3_south_west.sql
```diff
@@ -50,6 +50,486 @@ VALUES
...
+    "premises_id",
+    "bed_id",
+    "service",
+    "created_at",
+    "noms_number"
+  )
+VALUES
+  (
+    'c628b111-d1c4-49b5-9b7a-2f4aad9c2755',
+    CURRENT_DATE + (4 * 2) + 1,
```
tahb, line 72: We mostly copy these these statements but we push the arrival and departure dates forward on each one so they don't overlap.

PR #1111, opened 2023-11-07
Add CAS2 application status reference data

edavey:
## Add /cas2/reference-data/application-status endpoint

This endpoint will allow the UI to build a form for an assessor to provide a status update.

The status options are currently defined inline in `model/reference/Cas2ApplicationStatusSeeding.kt`
but we'll introduce database seeding to manage them.

cas2_reference_data

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/reference/Cas2ApplicationStatusSeeding.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/reference/Cas2ApplicationStatusSeeding.kt
@@ -1,0 +1,60 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.model.reference
+
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Cas2ApplicationStatus
+import java.util.UUID
+
+object Cas2ApplicationStatusSeeding {
+
+  fun statusList(): List<Cas2ApplicationStatus> {
+    return listOf(
+      Cas2ApplicationStatus(
+        id = UUID.fromString("c989f05f-c574-49a2-8381-11d332d98d40"),
+        name = "received",
+        label = "Received",
+        description = "The application has been received and is yet to be processed",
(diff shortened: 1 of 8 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/reference/Cas2ApplicationStatusSeeding.kt
```diff
@@ -0,0 +1,60 @@
...
+      Cas2ApplicationStatus(
+        id = UUID.fromString("176bbda0-0766-4d77-8d56-18ed8f9a4ef2"),
+        name = "placeOffered",
+        label = "Place offered",
+        description = "A place has been offered which has been sent by email.",
+      ),
+      Cas2ApplicationStatus(
+        id = UUID.fromString("a919097d-b324-471c-9834-756f255e87ea"),
+        name = "onWaitingList",
+        label = "On waiting list",
```
Harriethw, line 37: I think in the prototype this is now `On waitlist` and it looks like we're waiting on a description from Chanel... might be worth checking if she has anything in mind, otherwise going ahead and we make copy changes later.

edavey: These are just to get us started and it's understood that the copy will change and quite likely the list of possible states also.

I took the liberty of putting the "wait list" Americanism into British English!

Harriethw (approved):
Nice work!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/reference/Cas2ApplicationStatusSeeding.kt
Chanel-Diep: Ah great spot, thanks @edavey! I was so focused on making the labels short I didn't realise I'd slipped into an Americanism 😅

PR #1112, opened 2023-11-08
Fix: correctly return user's email in NomisUser representation

edavey:


edavey:
[BUG: NomisUser.email returns username](https://trello.com/c/yaINpbnN/630-bug-nomisuseremail-returns-username)

Harriethw (approved):

PR #1113, opened 2023-11-08
(629) New endpoint for CAS2 status updates on submitted applications

edavey:
## Assessor can now post status updates on submitted applications

See individual commit messages / OpenAPI specs for details.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/SubmissionsController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/SubmissionsController.kt
@@ -4,6 +4,7 @@ import org.springframework.http.HttpStatus
 import org.springframework.http.ResponseEntity
 import org.springframework.stereotype.Service
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.cas2.SubmissionsCas2Delegate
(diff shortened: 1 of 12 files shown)


edavey:
[create new POST /cas2/submissions/{applicationId}/status-updates endpoint](https://trello.com/c/ydLd4kYZ/629-create-new-post-cas2-submissions-applicationid-status-updates-endpoint)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/SubmissionsController.kt
```diff
@@ -79,6 +82,22 @@ class SubmissionsController(
     return ResponseEntity(HttpStatus.OK)
   }
 
+  override fun submissionsApplicationIdStatusUpdatesPost(
+    applicationId: UUID,
+    statusUpdate: Cas2ApplicationStatusUpdate,
+  ): ResponseEntity<Unit> {
+    if (!statusUpdateService.isValidStatus(statusUpdate)) {
+      return ResponseEntity(HttpStatus.NOT_FOUND)
```
Harriethw, line 90: Are we doing the check twice if it's a valid status? here in the controller and then again in the service? Is that so we can return a 404 here?

Harriethw (approved):
Thanks Ed, this is so clear and it all makes sense to me.

PR #1114, opened 2023-11-08
Reduce number of queries needed for placement requests and applications

tahb:
## Context
[Slack thread where we debugged this initially](https://mojdt.slack.com/archives/C044ZUGJRC7/p1699440674447369).

- We think that at 8:29 on Nov 8th 2023 user 17073fa0-f6c8-46a3-a6be-fcda599d2d22 requests `/placement-requests` from CAS1
- We think [that request failed to load all the results from the database and ended up in an out of memory error which bubbled up into a result set exception ](https://ministryofjustice.sentry.io/issues/4612537965/?project=4503931792392192&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=24h&stream_index=10)
- We checked preprod (which is synced from prod) and found that user had nearly 400 placement requests assigned
- The team expect that to be normal for CRU's
- We think the long term solution is to paginate these results so we don't ever have to load ~400 records into memory. CAS1 have been advised to prioritise this in their backlog
- As a short term mitigation we remove columns we don't care about from the result set to make it more efficient. This includes the small json blob stored within  the `data` column. The way we do this is to replace the derived Hibernate query and replace it with our explicit query which should honour the same interface

## Changes

As part of this work we identify N+1 queries. By moving away from the derived Hibernate queries and (sadly) writing our own SQL, we use some tactical JOINs to grab the data we need for transformation, upfront. [We hope that if we load this Hibernate will be smart enough not to execute additional queries.](https://stackoverflow.com/questions/60321266/jpa-hibernate-eager-loading-performing-subsequent-queries-to-fetch-all-data)

We first tackle `PlacementRequests` for the `/placement-requests` endpoint and then `PlacementApplications`for the`/tasks` endpoint as I could see a similar pattern of problem there.

I would really like to avoid writing out SQL like this but I'm unable to find a better way to do this with Hibernate. In rails we can do `Model.eager_load(association_1, association_2, …)` which removes the need to write low level SQL ourselves?

This has a side benefit that we are then able to remove `placement_application.data` from when looking for placement_requests where they aren't needed. These blobs aren't big but when a user needs to see 400 this could make a difference to both memory usage and sql performance. 

## How to view these sql changes?

Set this in your API's application.yml:

``` 
jpa:
    hibernate:
      ddl-auto: none
    properties:
      hibernate:
        dialect: org.hibernate.spatial.dialect.postgis.PostgisDialect
        show_sql: true
        format_sql: true
```

Restart the server (brought up originally by AP Tools).

I'd like to have figured out how to seed CAS1 so there were placement requests and applications. Then I could demonstrate the saving in queries. A production example of all the sql queries which actually times out the latest request to this endpoint can be viewed in [App Insights here.](https://portal.azure.com/#view/AppInsightsExtension/DetailsV2Blade/ComponentId~/%7B%22SubscriptionId%22%3A%22a5ddf257-3b21-4ba9-a28c-ab30f751b383%22%2C%22ResourceGroup%22%3A%22nomisapi-prod-rg%22%2C%22Name%22%3A%22nomisapi-prod%22%2C%22LinkedApplicationType%22%3A0%2C%22ResourceId%22%3A%22%252Fsubscriptions%252Fa5ddf257-3b21-4ba9-a28c-ab30f751b383%252FresourceGroups%252Fnomisapi-prod-rg%252Fproviders%252FMicrosoft.Insights%252Fcomponents%252Fnomisapi-prod%22%2C%22ResourceType%22%3A%22microsoft.insights%252Fcomponents%22%2C%22IsAzureFirst%22%3Afalse%7D/DataModel~/%7B%22eventId%22%3A%221f58084e-29f6-425e-8ebb-a04bc48426c3%22%2C%22timestamp%22%3A%222023-11-10T00%3A18%3A03.895Z%22%2C%22cacheId%22%3A%22f7d0d315-003c-47c6-8901-b4dee7bbccd6%22%2C%22eventTable%22%3A%22requests%22%2C%22timeContext%22%3A%7B%22durationMs%22%3A86400000%2C%22endTime%22%3A%222023-11-10T17%3A38%3A06.339Z%22%7D%7D)


tahb:
It became apparent these queries weren't calling application.data and that something else must be so the value of doing this is diminished.

pezholio (approved):
LGTM - the fact that we're having to do all this manual SQL makes me sad, but hopefully this will help the issues we're having

tahb:
Now I've got test data working locally I'm not convinced Hibernate does respect this way querying so closing this for now.

PR #1115, opened 2023-11-08
Enable CAS3 domain events to be enabled by event type

danlivings-dxw:
> See [ticket #1647 on the CAS3 Trello board](https://trello.com/c/tvJ7hfcR/1647-facility-to-turn-on-just-referralsubmitted-domain-event).

This PR introduces the ability for CAS3 to configure whether events are emitted to the SNS queue based on the event type.

This is achieved by setting the value of the `DOMAIN-EVENTS_CAS3_EMIT-ENABLED` environment variable with a comma-separated list of the names of the event types.

The values `false` and `true` are no longer supported, and have been replaced with the empty string and the full list of event types respectively.


sudhakar-plr (approved):
Looks ok to me.

tahb (approved):
LGTM. 

The unit test coverage is good. I wonder if we could extract any common set up to make them more readable?

PR #1117, opened 2023-11-09
Expose representations of CAS2 status updates within SubmittedApplication

edavey:
## Expose status updates within submitted application

Our representation of a `SubmittedApplication` now includes a list of `StatusUpdate`s

status_updates_in_submitted_application

Note that:

- `updatedAt` is the creation timestamp of the status update
- `updatedBy` is a new representation of an `ExternalUser`
- status updates are ordered by date descending i.e. most recent first

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas2/StatusUpdateTransformer.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas2/StatusUpdateTransformer.kt
@@ -1,0 +1,26 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.cas2
+
+import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Cas2StatusUpdate
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas2StatusUpdateEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.ExternalUserTransformer
+
+@Component("Cas2StatusUpdateTransformer")
+class StatusUpdateTransformer(
+  private val externalUserTransformer: ExternalUserTransformer,
+) {
+
+  fun transformJpaToApi(
+    jpa: Cas2StatusUpdateEntity,
+  ):
+    Cas2StatusUpdate {
+    return Cas2StatusUpdate(
+      id = jpa.id,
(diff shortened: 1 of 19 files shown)


edavey:
[Expose status updates with SubmittedApplication](https://trello.com/c/0ygNzRsV/581-expose-status-updates-with-submittedapplication)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas2/ApplicationsTransformer.kt
```diff
@@ -36,26 +32,6 @@ class ApplicationsTransformer(
     )
   }
 
-  fun transformJpaToSubmittedApplication(
```
Harriethw, line 39: nice

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas2/StatusUpdateTransformer.kt
```diff
@@ -0,0 +1,26 @@
...
+  private val externalUserTransformer: ExternalUserTransformer,
+) {
+
+  fun transformJpaToApi(
+    jpa: Cas2StatusUpdateEntity,
+  ):
+    Cas2StatusUpdate {
+    return Cas2StatusUpdate(
+      id = jpa.id,
+      name = jpa.status().name,
```
Harriethw, line 19: Is the reason `name` isn't a regular field, because while the `label` and `description` might change, the `name` will always stay the same, so we have to get that from `Cas2ApplicationStatusSeeding` directly?

Harriethw (approved):
Thanks Ed, makes sense to me!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas2/StatusUpdateTransformer.kt
edavey: yes that's right -- the 'name' of the status comes the associated `Cas2ApplicationStatus`. At present these are just simple "model" or API objects hard-coded in `Cas2ApplicationStatusSeeding` but I guess we'll move them into the db and have the association managed by JPA with the `@ManyToOne` annotation, something like:

```kt
@ManyToOne
@JoinColumn(name = "status_id")
val status: Cas2ApplicationStatus,
```

 At that point the line in the transformer would become:

```kt
name = jpa.status.name,
```

I think...

PR #1120, opened 2023-11-09
Increase memory limit for responses from Delius Integration API

danlivings-dxw:
In certain situations, downloading a booking report can result in a `WebClientResponseException`:

```
200 OK from GET https://approved-premises-and-delius.hmpps.service.justice.gov.uk/probation-cases/summaries; nested exception is org.springframework.core.io.buffer.DataBufferLimitException: Exceeded limit on max bytes to buffer : 262144
```

The memory limit for a `WebClient` is configurable and has already been increased to 750000 bytes for the Prison API. This uses the same configurable limit value on a provisional basis to alleviate the problem, although in future being able to configure this limit on a per-API basis may be preferable.


pezholio (approved):
LGTM 👍

PR #1125, opened 2023-11-09
Buffer `WebClient` responses to avoid `DataBufferLimitException`

danlivings-dxw:
The memory limit for responses from the Delius Integration API has recently been increased to 750,000 bytes (see #1120). However, this was still not sufficient when generating booking reports, as personal information is requested for several hundred CRNs concurrently. Continuing to increase this memory limit can only achieve so much as the API is limited by how much memory it has access to. This introduces a buffering approach for the responses which circumvents the limit and also avoids constructing potentially large strings as an intermediate step.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/client/BaseHMPPSClient.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/client/BaseHMPPSClient.kt
@@ -3,11 +3,16 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.client
 import com.fasterxml.jackson.core.type.TypeReference
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
+import com.fasterxml.jackson.module.kotlin.readValue
+import org.springframework.core.io.buffer.DataBuffer
+import org.springframework.core.io.buffer.DataBufferUtils
 import org.springframework.http.HttpHeaders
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/client/BaseHMPPSClient.kt
```diff
@@ -86,6 +81,37 @@ abstract class BaseHMPPSClient(
...
+
+    val pipedOutputStream = PipedOutputStream()
+    val pipedInputStream = PipedInputStream(pipedOutputStream)
+
+    val body = result
+      .body
+      ?.doOnError { pipedInputStream.close() }
+      ?.doFinally { pipedOutputStream.close() }
+
+    DataBufferUtils
```
tahb, line 106: This should help us on the large query for offender summaries though I worry what the effect might be for all/most other requests. Is the overhead of setting this up each time for every small request worth the trade off? Are there other trade offs?

Could/should we introduce a parameter to stream only specific requests?

danlivings-dxw: I did consider doing this, but I'm inclined to think this extra complexity wouldn't be worth it from a performance perspective, as I suspect that we'd only see a very small overhead if any, and there's a good chance that constructing the `String` instance as an intermediate step and passing that to the `ObjectMapper` would negate the benefits from skipping that overhead. Also, I suspect that internally something like this happens anyway, but just for a single buffer (<250kB/750kB depending on configuration).

PR #1134, opened 2023-11-13
Use native query to retrieve data for bookings report

danlivings-dxw:
This PR replaces the Hibernate-managed JPQL query for `BookingRepository.findAllByOverlappingDate` with a handwritten native SQL query called from `BookingsReportRepository.findAllByOverlappingDate`.

This native SQL query returns just the set of data that is needed to populate the booking report and avoids the N+1 query problem caused by lazily loading dependent entities on the booking.

### Performance testing
The performance test used to assess the performance of this change was performed using the same code and methodology as for #1100.

The original control test from `main` as of #1100, using JPQL and allowing Hibernate to generate multiple queries:
Control test results measured using the `main` branch

The results of the test as a result of using a handwritten native query:
image

| Control (JPQL query) | 100.0% | 8.658 | 49.901 | 29.526 | 31.704 | 11.488 |
| Native query |  100.0% | 0.057 | 0.576 | 0.210 | 0.218 | 0.061 |


tahb (approved):
Awesome 🙌

PR #1136, opened 2023-11-14
Fix CAS2 application status copy

patrickjfl:


edavey:
good spot! But I think this is now redundant as that initial "pseudo-status" should not be listed in our reference data... See https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1150

patrickjfl:
> good spot! But I think this is now redundant as that initial "pseudo-status" should not be listed in our reference data... See #1150

Aha, yes! Will close now.

PR #1143, opened 2023-11-14
(APS 45) Update migration to only update status

pezholio:
Currently, when running `save` against the application, we save the entire object, which may cause issues in production when real users are using the system. This detaches the entity to remove it from the [JPA entity lifecycle](https://dzone.com/articles/jpa-entity-lifecycle) and then runs a custom update to only update the status field.


tahb (approved):
LGTM!

PR #1145, opened 2023-11-14
PI-1624 - Move from Community API to AP and Delius Service for offender details

anthony-britton-moj:
Move from Community API to AP and Delius Service for offender details


docker-compose.test.yml
```diff
@@ -1,7 +1,7 @@
 version: "3"
 services:
   integration_test_database:
-    image: "postgis/postgis"
+    image: "ghcr.io/baosystems/postgis:15-3.3"
```
pezholio, line 4: What's the context behind this change?

tahb:
I believe we're now well underway with a series of changes that achieves the same goal (for which we have been using this as a reference) but with extras we need of removing the dependency on our caching and giving us a way to toggle back if something goes wrong.

https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1200
https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1220

There is still more to do before we're done but I think we're now safe to close this. Thanks for giving us the foundation for this work!

PR #1148, opened 2023-11-15
Remove `status` on CAS2 SubmittedApplications

Harriethw:
We do not need this field, as it relates to the `ApplicationStatus` types which are not revelant to our submissions, and do not want them confused with our `StatusUpdates` which are added by assessors to a submitted application.


edavey (approved):
Nice tidy-up!

PR #1150, opened 2023-11-15
update copy for CAS2 Status updates

Harriethw:
This should reflect the copy provided https://trello.com/c/dw06tLpm/635-nacro-status-update-copy

We have removed 'retrieved' because it is not technically a status that a user can add, but the default state it's in before a status has been added.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/reference/Cas2ApplicationStatusSeeding.kt
```diff
@@ -12,53 +12,47 @@ object Cas2ApplicationStatusSeeding {
...
-        name = "received",
-        label = "Received",
-        description = "The application has been received and is yet to be processed",
-      ),
       Cas2ApplicationStatus(
         id = UUID.fromString("f5cd423b-08eb-4efb-96ff-5cc6bb073905"),
         name = "moreInfoRequested",
         label = "More information requested",
-        description = "More information about the application has been requested from the POM (Prison Offender Manager).",
+              description = "The prison offender manager (POM) must provide information requested for the application to progress.",
```
edavey, line 19: I think something has gone haywire with the indentation here, see before `description` and `label` properties....

edavey (approved):
yes, this "pseudo-state" should not have been included.

BTW I think some wonky indentation has crept in here...

PR #1152, opened 2023-11-15
Search for person by NOMIS ID for CAS2

Harriethw:
This depends on the AP Tools PR being merged: https://github.com/ministryofjustice/hmpps-approved-premises-tools/pull/51 
If you'd like to call the added controller function from the UI there is this WIP PR: https://github.com/ministryofjustice/hmpps-community-accommodation-tier-2-ui/pull/310

We need to allow a CAS2 user to search for an offender by prison number (nomsNumber), from an API that returns a CRN in order to make subsequent requests for OASyS data. This PR makes a request to [the `probation-offender-search` API](https://github.com/ministryofjustice/probation-offender-search). The shape of the [data returned should match what's returned from the probation API](http://localhost:8084/swagger-ui/index.html#/offender-search/search_1), which is a list of OffenderDetails.

See their controller tests:
https://github.com/ministryofjustice/probation-offender-search/blob/7093fdf716f969de3b80827675f8389cef01f95c/src/test/kotlin/uk/gov/justice/hmpps/probationsearch/controllers/OffenderSearchControllerTest.kt
And Swagger: http://localhost:8084/swagger-ui/index.html#/offender-search/search_1 

I have some outstanding questions on the implementation:

~~1. If an offender is an LAO, we don't want to allow the user to continue. If I understand correctly if `currentExclusion` or `currentRestriction` are `true` then the offender is an LAO? In which case we can just return a Forbidden response? I'm not entirely sure what those fields mean on this probation offender search API.~~ the implementation for LAO should mean a forbidden response is returned if they are LAO
2. The current `FullPerson` requires a date of birth, which according to the Probation API model might potentially be `null`. How should we handle that? Do we want to continue on CAS2 if there is no DoB, as it could potentially be the wrong person? Or, should there be the concept of a CAS2 Person that might not have a DoB? Orrrr, do we make another call to another Delius API to get that DoB.
~~3. We don't get an `inOutStatus` from the Probation API. Can we assume the person is in custody, because the POM is making an application for CAS2? Would this have implications if it was wrong?~~ we are making call to Prison API to get extra prison info
4. Are we making the right call to the `probation-offender-search` API, and have I understood the shape of the data we should expect back?

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/client/ProbationOffenderSearchApiClient.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/client/ProbationOffenderSearchApiClient.kt
@@ -1,0 +1,21 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.client
+
+import com.fasterxml.jackson.databind.ObjectMapper
+import org.springframework.beans.factory.annotation.Qualifier
+import org.springframework.stereotype.Component
+import org.springframework.web.reactive.function.client.WebClient
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.probationoffendersearchapi.ProbationOffenderDetail
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.probationoffendersearchapi.ProbationOffenderSearchNomsRequest
+
+@Component
+class ProbationOffenderSearchApiClient(
+  @Qualifier("probationOffenderSearchApiWebClient") webClient: WebClient,
+  objectMapper: ObjectMapper,
+  webClientCache: WebClientCache,
+) : BaseHMPPSClient(webClient, objectMapper, webClientCache) {
+
+  fun searchOffenderByNomsNumber(nomsNumber: String) = postRequest<List<ProbationOffenderDetail>> {
+    path = "/search"
+    body = ProbationOffenderSearchNomsRequest(nomsNumber = nomsNumber)
+  }
+}
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/OffenderService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/OffenderService.kt
@@ -7,23 +7,77 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.ApOASysContextApi
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.ClientResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.CommunityApiClient
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.PrisonsApiClient
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.ProbationOffenderSearchApiClient
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.PersonInfoResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.PersonRisks
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.ProbationOffenderSearchResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.RiskStatus
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.RiskWithStatus
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.RoshRisks
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.community.OffenderDetailSummary
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.prisonsapi.InmateDetail
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.probationoffendersearchapi.ProbationOffenderDetail
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.results.AuthorisableActionResult
 
 @Service("CAS2OffenderService")
+@Suppress(
+  "ReturnCount",
+)
 class OffenderService(
   private val communityApiClient: CommunityApiClient,
   private val prisonsApiClient: PrisonsApiClient,
+  private val probationOffenderSearchApiClient: ProbationOffenderSearchApiClient,
   private val apOASysContextApiClient: ApOASysContextApiClient,
 ) {
 
   private val log = LoggerFactory.getLogger(this::class.java)
+
+  fun getPersonByNomsNumber(nomsNumber: String): ProbationOffenderSearchResult {
+    fun logFailedResponse(probationResponse: ClientResult.Failure<List<ProbationOffenderDetail>>) =
+      log.warn("Could not get inmate details for $nomsNumber", probationResponse.toException())
+
+    val probationResponse = probationOffenderSearchApiClient.searchOffenderByNomsNumber(nomsNumber)
+
+    val probationOffenderDetailList = when (probationResponse) {
+      is ClientResult.Success -> probationResponse.body
+      is ClientResult.Failure.StatusCode -> when (probationResponse.status) {
+        HttpStatus.NOT_FOUND -> return ProbationOffenderSearchResult.NotFound(nomsNumber)
+        HttpStatus.FORBIDDEN -> return ProbationOffenderSearchResult.Forbidden(nomsNumber, probationResponse.toException())
+        else -> {
+          logFailedResponse(probationResponse)
+          return ProbationOffenderSearchResult.Unknown(nomsNumber, probationResponse.toException())
+        }
+      }
+
+      is ClientResult.Failure -> {
+        logFailedResponse(probationResponse)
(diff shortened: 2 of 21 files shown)


src/main/resources/static/codegen/built-cas2-api-spec.yml
```diff
@@ -241,11 +241,11 @@ paths:
     get:
       tags:
         - People operations
-      summary: Searches for a Person by their CRN
+      summary: Searches for a Person by their Prison Number (NOMIS ID)
```
edavey, line 244: Would it be clearer if this were part of the same commit where `cas2-api.yml` is modified?

edavey (commented):
this looks great, thks! Just some little comments and suggestions. It does look like we need to make the call to the prison side to find the name of the prison where the person is located.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/client/ProbationOffenderSearchApiClient.kt
```diff
@@ -0,0 +1,21 @@
...
+
+@Component
+class ProbationOffenderSearchApiClient(
+  @Qualifier("probationOffenderSearchApiWebClient") webClient: WebClient,
+  objectMapper: ObjectMapper,
+  webClientCache: WebClientCache,
+) : BaseHMPPSClient(webClient, objectMapper, webClientCache) {
+
+  fun searchOffenderByNomsNumber(nomsNumber: String) = postRequest<List<ProbationOffenderDetail>> {
+    path = "/search"
```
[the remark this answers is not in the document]
Harriethw: hmm, maybe it's one to chat about as I've just seen there is also a `nomsNumber` endpoint! https://probation-offender-search-dev.hmpps.service.justice.gov.uk/swagger-ui/index.html#/offender-search/findByNomsNumbers 

I think I'd prefer it to be more explicit, especially having worked with opensearch APIs before I've had trouble getting exact matches (that might have been an issue with that API though!)

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas2/OffenderServiceTest.kt
```diff
@@ -44,9 +49,10 @@ class OffenderServiceTest {
   }
 
   private val offenderService = OffenderService(
-    mockCommunityApiClient,
-    mockPrisonsApiClient,
-    mockApOASysContextApiClient,
+          mockCommunityApiClient,
+          mockPrisonsApiClient,
+          mockProbationOffenderSearchClient,
+          mockApOASysContextApiClient,
```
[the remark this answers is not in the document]
Harriethw: no, I think something has changed about my Kotlin linting in the IDE...

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/PersonTransformer.kt
```diff
@@ -68,6 +69,21 @@ class PersonTransformer {
...
+  fun transformProbationOffenderToPersonApi(probationOffenderResult: ProbationOffenderSearchResult.Success.Full, nomsNumber: String) =
+          FullPerson(
+                  type = PersonType.fullPerson,
+                  crn = probationOffenderResult.probationOffenderDetail.otherIds.crn,
+                  name = "${probationOffenderResult.probationOffenderDetail.firstName} ${probationOffenderResult.probationOffenderDetail.surname}",
+                  // TODO: we are not handling if DoB is not there, what should we do?
+                  dateOfBirth = probationOffenderResult.probationOffenderDetail.dateOfBirth!!,
+                  sex = probationOffenderResult.probationOffenderDetail.gender ?: "Not Found",
+                  // TODO: do we need to make call to prison api to determine status?
+                  status = FullPerson.Status.inCustody,
```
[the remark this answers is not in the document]
Harriethw: brill, thanks - have added that call now.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/PersonTransformer.kt
```diff
@@ -68,6 +69,21 @@ class PersonTransformer {
     )
   }
 
+  fun transformProbationOffenderToPersonApi(probationOffenderResult: ProbationOffenderSearchResult.Success.Full, nomsNumber: String) =
+          FullPerson(
+                  type = PersonType.fullPerson,
+                  crn = probationOffenderResult.probationOffenderDetail.otherIds.crn,
+                  name = "${probationOffenderResult.probationOffenderDetail.firstName} ${probationOffenderResult.probationOffenderDetail.surname}",
+                  // TODO: we are not handling if DoB is not there, what should we do?
+                  dateOfBirth = probationOffenderResult.probationOffenderDetail.dateOfBirth!!,
```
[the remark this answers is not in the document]
Harriethw: hmm `dateOfBirth` is a required field on `FullPerson`... I'm wondering if we let it blow up for now then, or change the Person class? who would we ask to check who's missing dob?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/OffenderService.kt
```diff
@@ -7,24 +7,71 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.ApOASysContextApi
...
 ) {
 
   private val log = LoggerFactory.getLogger(this::class.java)
 
+  fun getPersonByNomsNumber(nomsNumber: String): ProbationOffenderSearchResult {
+    fun logFailedResponse(probationResponse: ClientResult.Failure<List<ProbationOffenderDetail>>) =
+            log.warn("Could not get inmate details for $nomsNumber", probationResponse.toException())
+
+    // TODO: do we need to use cache, as below?
+    val probationResponse = probationOffenderSearchApiClient.searchOffenderByNomsNumber(nomsNumber)
```
Harriethw, line 37: @edavey do you have any thoughts on using the cache?

edavey: no I don't think we need to cache calls to Probation Offender Search -- as it is in itself a cache of sorts, on top of Delius/Community API...

pezholio: I think we're leaning towards moving away from so much caching, often the response is faster than Redis anyway!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/PersonTransformer.kt
```diff
@@ -68,6 +69,24 @@ class PersonTransformer {
     )
   }
 
+  fun transformProbationOffenderToPersonApi(probationOffenderResult: ProbationOffenderSearchResult.Success.Full, nomsNumber: String) =
+          FullPerson(
+            type = PersonType.fullPerson,
+            crn = probationOffenderResult.probationOffenderDetail.otherIds.crn,
+            name = "${probationOffenderResult.probationOffenderDetail.firstName} ${probationOffenderResult.probationOffenderDetail.surname}",
+            // TODO: we are not handling if DoB is not there, what should we do?
+            dateOfBirth = probationOffenderResult.probationOffenderDetail.dateOfBirth!!,
```
pezholio, line 78: Having that `!!` there makes me nervous. Could you extend the CAS2 `FullPerson` to make `dateOfBirth` optional?

pezholio (approved):
LGTM from where I'm standing. Just a couple of things to think about

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/PersonTransformer.kt
Harriethw: I'm using the shared `FullPerson` model at the moment, I'm wondering if I should create a CAS2 version or if there's a way to extend the `FullPerson` with just the change to that one field

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/client/ProbationOffenderSearchApiClient.kt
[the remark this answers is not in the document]
Harriethw: we decided will use the current implementation 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/PersonTransformer.kt
Harriethw: after chatting with @edavey we decided to take the risk of having the !! for now as it's very unlikely we won't have DoB (we're assuming!), and we don't want to create yet another model for this one difference to `FullPerson`,  and I'm also thinking it should be required as a piece of data that identifies the right person.

PR #1154, opened 2023-11-15
Include personal data for bookings report

danlivings-dxw:
> See [ticket #1535 on the CAS3 Trello board](https://trello.com/c/b8NYSvNw/1535-report-mi-spreadsheet-includes-personal-data).

> See PR #1100, which was the original version of this PR that was reverted. 

This PR introduces columns containing personal data to the bookings report. These columns are:
- `personName`
- `pncNumber`
- `gender`
- `ethnicity`
- `dateOfBirth`

Of these columns, four are populated as expected - `pncNumber` is currently not available and is always presented as a blank cell. This is because we do not store personal information within our service and have to obtain it elsewhere. Although we cache results from the Community API, the number of results expected in a typical spreadsheet is between 200 and 400 rows, and performing a request to the cache for each CRN incurs an unacceptable performance degradation (see below). The alternative solution is to use the AP-Delius Integration API, which provides an endpoint for accessing personal data in bulk, but this endpoint only provides a subset of the personal data and does not currently include the PNC number.

### Performance testing
The performance test used to assess the performance of this change was performed using the same code and methodology as for #1100.

The current control test from `main` as of #1134, using the native SQL query:
image

Introducing personal data using the bulk CRN data endpoint:
image

| Control |  100.0% | 0.057 | 0.576 | 0.210 | 0.218 | 0.061 |
| Including personal data | 100.0% | 0.079 | 0.945 | 0.229 | 0.218 | 0.091 |


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/reporting/generator/BookingsReportGenerator.kt
```diff
@@ -19,10 +19,13 @@ class BookingsReportGenerator : ReportGenerator<BookingsReportDataAndPersonInfo,
...
-        personName = personInfo.tryGetDetails { "${it.firstName} ${it.surname}".trim() },
-        pncNumber = personInfo.tryGetDetails { it.otherIds.pncNumber },
+        personName = personInfo.tryGetDetails {
+          val nameParts = listOf(it.name.forename) + it.name.middleNames + it.name.surname
+          nameParts.joinToString(" ")
+        },
+        pncNumber = null,
         gender = personInfo.tryGetDetails { it.gender },
-        ethnicity = personInfo.tryGetDetails { it.offenderProfile.ethnicity },
+        ethnicity = personInfo.tryGetDetails { it.profile?.ethnicity },
```
tahb, line 28: Q; Looks like we've changed from `PersonInfoResult` to `PersonSummaryInfoResult` which then changes how we get ethnicity. [I don't think we did this before on attempt 1](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1100/files#diff-f07bbcf9104e9ae593b098a2ae7dd2fad95e30e3ad8fd9e96175991e78ae1c21R29). Is PersonSummaryInfoResult a newer and perhaps efficient way of handling this too?

tahb (approved):
Looks good! One minor non blocking question out of curiosity.

tahb (approved):
Looks good! One minor non blocking question out of curiosity

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/reporting/generator/BookingsReportGenerator.kt
danlivings-dxw: This was originally done [in this commit](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1100/commits/98f5b53b2ac11b86323d515892ff39f900ba88c5). The two types are unrelated beyond the similarities in their name, as `PersonInfoResult`s come from the Community API and `PersonSummaryInfoResult`s come from the Delius Integration API, and they return data of different shapes. The value of `PersonSummaryInfoResult` is that we can get many of them per request instead of just one but it has less information than `PersonInfoResult`, so I don't think it would be appropriate to replace `PersonInfoResult` with `PersonSummaryInfoResult` in every situation.

PR #1155, opened 2023-11-15
(636) Use logged-in Nomis User's JWT to auth with Nomis-User-Roles API

edavey:
The Nomis-User-Roles `/users/{username}` endpoint is
authorised to use roles:

- `ROLE_MANAGE_NOMIS_USER_ACCOUNT`
- `ROLE_MAINTAIN_ACCESS_ROLES_ADMIN`
- `ROLE_MAINTAIN_ACCESS_ROLES`

See [UserResource.kt][UserResource]

These are roles which are granted to users (["authorisation code" flow][AuthorisationCode]), rather than to systems (["client credentials" flow][ClientCredentials]).

To avoid using excessive privileges on a system-to-system connection, we change our use of this API so that when retrieving details of the logged in Nomis user we:

- query the Nomis-User-Roles' `/me` endpoint rather than `/users/{username}`
- authenticate using the logged in users JWT through the `Authorization` header, rather than using the  system-to-system "nomis-user-roles-api" OAuth client

[UserResource]:
https://github.com/ministryofjustice/nomis-user-roles-api/blob/1500d7799df6e79b7439c87e70a8f5f743b9747a/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisuserrolesapi/resource/UserResource.kt

[AuthorisationCode]:
https://github.com/ministryofjustice/hmpps-auth#using-authorisation-code

[ClientCredentials]:
https://github.com/ministryofjustice/hmpps-auth#using-client-credentials


edavey:
[Get pre-prod credentials for API connection to Nomis-User-Roles API](https://trello.com/c/RTMTs8qk/636-get-pre-prod-credentials-for-api-connection-to-nomis-user-roles-api)

Harriethw (approved):
thanks Ed

PR #1156, opened 2023-11-15
(649) Fix: ensure new assessor is persisted on GET to individual SubmittedApplication

edavey:
At present we rely on an explicit `ensureExternalUserPersisted()` function within controller actions to ensure that the logged in assessor is recorded in our `external_users` table.

It would be good to develop this, perhaps:

- making it the responsibility of `HttpAuthService.getCas2AuthenticatedPrincipalOrThrow()`, or
- do something more elaborate using Spring's `UserContext` see [NomisUserRoles](https://github.com/ministryofjustice/nomis-user-roles-api/blob/main/src/main/kotlin/uk/gov/justice/digital/hmpps/nomisuserrolesapi/utils/UserContext.kt#L6) and [Baeldung: Get User in Spring Security](https://www.baeldung.com/get-user-in-spring-security)


edavey:
[Bug: ensure that any previously unseen Assessor has an ExternalUser record persisted when viewing an individual SubmittedApplication](https://trello.com/c/yUkOQCWG/649-bug-ensure-that-any-previously-unseen-assessor-has-an-externaluser-record-persisted-when-viewing-an-individual-submittedapplican)

patrickjfl (approved):
LGTM

PR #1167, opened 2023-11-20
633 Add name and email address to `ExternalUser`

edavey:
See [Trello 633](https://trello.com/c/ojaf1IBa/633-add-name-and-email-address-to-external-users)

Note that to run locally [AP-Tools now provides HMPPS Manage-Users API on port 9560](https://github.com/ministryofjustice/hmpps-approved-premises-tools/pull/52)

### Add name and email address to `ExternalUser`

The representation of an `ExternalUser` which is used in `Cas2StatusUpdate.updatedBy` needs to have more information than we were initially extracting from the logged in user's JWT.

Our `ExternalUserService` now connects to the HMPPS-Manage-Users' `/externalusers/{username}` endpoint and we add `name` and `email` properties to:

- the record which we persist in the `external_users table` and also to 
- the JSON representation which we provide to the UI.

extended_external_user


Harriethw (approved):
LGTM, thanks

PR #1170, opened 2023-11-21
APS50 - New endpoint so we return tasks based on types.

garethhallberg:
Tests that show it works. Some minor reformatting to ease the detekt complaints.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/TasksController.kt
```diff
@@ -89,9 +89,53 @@ class TasksController(
...
+    val tasks = mutableListOf<Task>()
+    val type = enumConverterFactory.getConverter(TaskType::class.java).convert(
+      taskType.kebabCaseToPascalCase(),
+    ) ?: throw NotFoundProblem(taskType, "TaskType")
+
+    if (user.hasRole(UserRole.CAS1_MATCHER)) {
+      when (type) {
+        TaskType.placementApplication -> {
+          async {
+            tasks += getPlacementApplicationTasks(
```
pezholio, line 121: We probs don't need to push into a list here, given that the list will only contain one type of task

pezholio (commented):
LGTM, just one minor thing

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/TasksController.kt
```diff
@@ -89,9 +89,53 @@ class TasksController(
...
+        )
+      }
+    }
+
+    return@runBlocking ResponseEntity.ok(tasks)
+  }
+
+  override fun tasksTaskTypeGet(taskType: String): ResponseEntity<List<Task>> = runBlocking {
+    val user = userService.getUserForRequest()
+    var tasks = listOf<Task>()
```
pezholio, line 112: I think this will just return an empty list now?

Probs better to do:

```suggestion
var tasks = if (user.hasRole(UserRole.CAS1_MATCHER)) {
      when (type) {
        TaskType.placementApplication -> {
          async {
            tasks = getPlacementApplicationTasks(
              placementApplicationService.getVisiblePlacementApplicationsForUser(user),
              user,
            )
          }
        }
        TaskType.placementRequest -> {
          async {
            tasks = getPlacementRequestTasks(
              placementRequestService.getVisiblePlacementRequestsForUser(user),
              user,
            )
          }
        }
        else -> {
          throw BadRequestProblem()
        }
      }
```

pezholio (approved):
LGTM - sorry for the misdirection!

PR #1172, opened 2023-11-21
Only show placement applications without decision

pezholio:
Listing all placement applications for a user has limited use and also means we can be loading a ton of applications into memory. We’re going to be adding pagination soon, but in the meantime, let’s make sure we’re only loading the applications we care about.


garethhallberg (approved):
Nicely done

PR #1175, opened 2023-11-22
[Bugfix-1668]Calculate the number of night stays by considering midnight time in reports

sudhakar-plr:
Refer 1668 for more information.

This PR is to fix a bug in reports where the nights are calculated as days instead of only the stayed night. By using the midnight time, we can calculate the number of nights stays.


tahb (approved):
LGTM 🚀

PR #1176, opened 2023-11-22
Remove the ability to hard delete a booking as team utility

tahb:
We previously offered this function to help early users recover from human error.

The last valid use case we have for this is to allow a very small number of CAS3 users who enter an incorrect arrival date to recover. We’d hard delete the booking and allow them to redo the whole thing and set the new date. We are now adding this as a user journey so users can do this themselves.

Hard deleting a booking should then be an unexpected scenario.

The reason we need to take the nuclear step of removing this code is that using it now has unintended consequences. Enabling Domain events means we will be sharing references to bookings with downstream services like ndelius. They’ll be listening for updates as a booking moves through the different states. If we were to start emitting and then hard delete the booking, that would orphan the booking in those systems and they’d never recieve further updates. Attempts by them to access the details for that event would result in an error, likely a 404. In order to keep the integrity of data between systems we can no longer hard delete bookings as the solution.

To revert this I mostly revert the relveant parts of the pull request that introduces it https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/437/files


tahb:
@pezholio I presume you haven't been using this utility and it's just one that affects us?

pezholio:
Nah, you're good to get rid 👍

pezholio (approved):

PR #1179, opened 2023-11-23
Aps53 return workload data in tasks endpoint

garethhallberg:


pezholio (commented):
Looks good, just a few small suggestions

pezholio (approved):

PR #1187, opened 2023-11-28
Filter by allocated or unallocated reallocatable tasks

garethhallberg:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/PlacementRequestService.kt
```diff
@@ -73,10 +76,22 @@ class PlacementRequestService(
   fun getAllReallocatable(
     page: Int?,
     sortDirection: SortDirection?,
+    allocatedFilter: AllocatedFilter?,
   ): Pair<List<PlacementRequestEntity>, PaginationMetadata?> {
     val sortField = "created_at"
     val pageable = getPageable(sortField, sortDirection, page)
-    val allReallocatable = placementRequestRepository.findAllReallocatable(pageable)
+    val allReallocatable: Page<PlacementRequestEntity>?
```
pezholio, line 83: Again, `when` would be nicer here

pezholio (commented):
Looks good - just a few comments. It's a shame the other autocorrected stuff snuck into the main commit, but not worth losing sleep over!

pezholio (approved):

PR #1189, opened 2023-11-28
Add PNC number to FullPerson

Harriethw:
We need to display a person's PNC number to our CAS2 users.
We can get this when we search for an offender when we call `searchOffenderByNomsNumber` on the ProbationOffenderSearch. https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/61284aa89c71b09a6d095ebc6fe8b491c26dc3b3/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/OffenderService.kt#L39 

This PR adds the number as an optional field to the `FullPerson` model and the `PersonTransformer`.


pezholio (approved):
LGTM 👍

PR #1190, opened 2023-11-28
[1710] - Return restricted person response when LAO call fails with runtime exception

sudhakar-plr:
# Changes in this PR
Please refer [1710](https://trello.com/c/jvMr1obF/1710-listing-assessments-should-not-500-due-to-limited-access-to-any-one-of-the-referrals) for more information.

This PR to handle bug in offender service where runtime exception thrown when LAO API call failed with 403 without response body. 

Expected Result:
We should return Restricted Person response instead of throwing runtime exception. Additionally we will send the error to Sentry to report  the issue for further investigation.

Background:
We get forbidden exception when HPT shouldn’t have access to the personal details within the referral (AKA a limited access offender or an LAO), But for some reason LAO calls return 403 status with empty body which caused offenderService to throw runtime exception due response marshalling failure. 

Decision:
For CAS3, anyone within a regional HPT team will be loading all referrals assigned to that region - regardless which they intend to work on. So the referral can be a restricted access to HPT and this is an expected LAO scenario. So we expect it technically handle it safely. Hence we are fixing by returning Restricted person response instead of throwing runtimeException

Note: This fix is applied only in this instance of LAO calls, There may be similar calls in different places, which we should monitor and fix as needed in the future.

# Release checklist

confirm:

## Pre merge checklist

    advance?

## Post merge checklist

considering it done.

fix it as soon as possible.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/OffenderService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/OffenderService.kt
@@ -1,5 +1,6 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.service
 
+import io.sentry.Sentry
 import org.slf4j.LoggerFactory
 import org.springframework.http.HttpStatus
 import org.springframework.stereotype.Service
@@ -493,7 +494,8 @@ class OffenderService(
                   accessResponse.deserializeTo<UserOffenderAccess>()
                   return PersonInfoResult.Success.Restricted(crn, offender.otherIds.nomsNumber)
                 } catch (exception: Exception) {
-                  accessResponse.throwException()
+                  Sentry.captureException(RuntimeException("LAO calls returns but failed to marshal the response", exception))
+                  return PersonInfoResult.Success.Restricted(crn, offender.otherIds.nomsNumber)
                 }
               }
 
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/OffenderService.kt
```diff
@@ -489,7 +490,8 @@ class OffenderService(
                   accessResponse.deserializeTo<UserOffenderAccess>()
                   return PersonInfoResult.Success.Restricted(crn, offender.otherIds.nomsNumber)
                 } catch (exception: Exception) {
-                  accessResponse.throwException()
+                  Sentry.captureException(RuntimeException("LAO calls returns but failed to marshal the response", exception))
+                  return PersonInfoResult.Success.Restricted(crn, offender.otherIds.nomsNumber)
```
tahb, line 494: @pezholio this is worth making you aware of as even though CAS1 have manual assignments it could theoretically affect CAS1 users too. 

I think failing safely and returning a restricted person should there be _any_ problems getting the offender details on an LAO check is better than failing badly and breaking wider functionality.

The likelihood? We have only one recorded case so far (maybe we missed earlier cases) where the response isn't something we can deserialise. This means it could happen again and is worth closing off in my view.

The risk? We may return offender info presented as an LAO when in fact they aren't LAO but there was a problem accessing their info - this could be confusing to staff.

We could go further and change the type of response to something like `PersonInfoResult.Failed` and handle that explicitly into the frontend. We could say "Limited offender details" instead of "limited access offender". Whether we do that now or wait to see if this happens via Sentry alerts is a question.

pezholio: Yep, I'm happy with that. Seems like a safer option anyway

pezholio (approved):

tahb (approved):
Thanks Sudhakar ⭐

PR #1193, opened 2023-11-29
(APS 72) Add a job to update the sentence type/situation

pezholio:
This adds a new data migration job to backfill the sentence type and situation. To keep things contained I’ve created a new repository and put the helper classes in the job itself.

As we’re only fetching a small subset of data, we _should_ be able to get away without having to paginate through the results, but the proof of the pudding will be in the eating when we run it in preprod!


garethhallberg (approved):

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/MigrationJobScaffoldingTest.kt
```diff
@@ -13,7 +13,7 @@ class MigrationJobScaffoldingTest : SeedTestBase() {
       .uri("/migration-job")
       .bodyValue(
         MigrationJobRequest(
-          jobType = MigrationJobType.updateAllUsersFromCommunityApi,
+          jobType = MigrationJobType.allUsersFromCommunityApi,
```
pezholio, line 16: I don't know why every time I make changes to the migration jobs, the generator changes the name

PR #1194, opened 2023-11-29
Paginate get all reallocatable tasks with allocated filter, unallocat…

garethhallberg:
…ed filter and no filter

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/TasksController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/TasksController.kt
@@ -44,11 +44,12 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.UserService
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.TaskTransformer
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.UserTransformer
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.extractEntityFromAuthorisableActionResult
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.getIndices
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.getMetadata
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.getNameFromOffenderDetailSummaryResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.kebabCaseToPascalCase
 import java.util.UUID
 import javax.transaction.Transactional
-
 @Service
 class TasksController(
   private val userService: UserService,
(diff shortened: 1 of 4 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/TasksTest.kt
```diff
@@ -744,6 +744,264 @@ class TasksTest : IntegrationTestBase() {
...
+            webTestClient.get()
+              .uri("/tasks/reallocatable?page=2")
+              .header("Authorization", "Bearer $jwt")
+              .exchange()
+              .expectStatus()
+              .isOk
+              .expectHeader().valueEquals("X-Pagination-CurrentPage", 2)
+              .expectHeader().valueEquals("X-Pagination-TotalPages", 3)
+              .expectHeader().valueEquals("X-Pagination-TotalResults", 27)
+              .expectHeader().valueEquals("X-Pagination-PageSize", 10)
```
pezholio, line 827: Would be good for this to be a sum of  numAllocatedPlacementApplications,  numUnallocatedPlacementApplications etc etc to save this being a magic number

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/util/PaginationUtils.kt
```diff
@@ -40,3 +40,29 @@ fun <T>getMetadata(response: Page<T>, page: Int?): PaginationMetadata? {
     null
   }
 }
+
+fun getIndices(page: Int, taskCount: Int): Pair<Int, Int> {
```
pezholio, line 44: Might be a good idea to get some unit testing in here, just to be sure it copes with things like uneven counts and things like that. It'll also document the behaviour for future us so the behaviour's not buried in the integration tests.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/TasksController.kt
```diff
@@ -440,9 +442,26 @@ class TasksController(
       )
 
     val tasks: MutableList<Task> = ArrayList()
-    async { tasks.addAll(assessmentTasks) }
-    async { tasks.addAll(placementRequestTasks) }
-    async { tasks.addAll(placementApplicationTasks) }
+    tasks.addAll(assessmentTasks)
+    tasks.addAll(placementRequestTasks)
+    tasks.addAll(placementApplicationTasks)
```
pezholio, line 447: Don't quote me on this, but you might be able to retain the behaviour you had before with:

```
val jobs = listOf(
  async { tasks.addAll(assessmentTasks) }
  async { tasks.addAll(placementRequestTasks) }
  async { tasks.addAll(placementApplicationTasks) }
)
jobs.awaitAll()
```

That way we can build all the tasks up asynchronously and not lose the performance benefits we had before.

pezholio (commented):
Just a coupla comments - looks good in the main though 👍

pezholio (approved):
LGTM 👍

PR #1200, opened 2023-11-30
[APS-67] Introduce facade for offender details

danlivings-dxw:

This PR introduces a facade for retrieving offender details, allowing us to abstract to some degree over the source of this information, simplifying the planned transition away from the Community API in favour of the AP Delius Context API.

This facade is implemented such that it can be configured through the Spring property `data-sources.offender-details` or the environment variable `DATA-SOURCES_OFFENDER-DETAILS`, which accepts values of either `community_api` or `ap_delius_context_api`. This allows for control over when and in what environments the transition to the AP Delius Context API occurs, and allows us to rollback the transition without reverting significant code changes should problems occur.

Currently, this facade only supports the Community API for individual offenders, and it replicates the existing behaviour. Future work will expand the facade to support listings, as well as providing an implementation backed by the AP Delius Context API.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/datasource/OffenderDetailsDataSource.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/datasource/OffenderDetailsDataSource.kt
@@ -1,0 +1,62 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.datasource
+
+import org.slf4j.LoggerFactory
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.context.annotation.Primary
+import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.ClientResult
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.community.OffenderDetailSummary
+
+interface OffenderDetailsDataSource {
+  val name: OffenderDetailsDataSourceName
+
+  fun getOffenderDetailSummary(crn: String): ClientResult<OffenderDetailSummary>
+}
+
+enum class OffenderDetailsDataSourceName {
+  COMMUNITY_API,
+  AP_DELIUS_CONTEXT_API,
+}
+
+@Component
+@Primary
+class ConfiguredOffenderDetailsDataSource(
+  dataSources: List<OffenderDetailsDataSource>,
+  @Value("\${data-sources.offender-details}")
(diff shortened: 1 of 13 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/datasource/OffenderDetailsDataSource.kt
```diff
@@ -42,12 +43,20 @@ class ConfiguredOffenderDetailsDataSource(
...
+  private val communityApiClient: CommunityApiClient,
+) : OffenderDetailsDataSource {
   override val name: OffenderDetailsDataSourceName
     get() = OffenderDetailsDataSourceName.COMMUNITY_API
 
   override fun getOffenderDetailSummary(crn: String): ClientResult<OffenderDetailSummary> {
-    throw NotImplementedError("Getting details for individual offenders from the Community API is not currently supported")
+    var offenderResponse = communityApiClient.getOffenderDetailSummaryWithWait(crn)
+
+    if (offenderResponse is ClientResult.Failure.PreemptiveCacheTimeout) {
```
pezholio, line 58: I'm hoping we'll be able to make the preemptive caching go away soon, but it makes sense to keep it in place for the Community API calls in the meantime

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/datasource/CommunityApiOffenderDetailsDataSourceTest.kt
```diff
@@ -46,41 +47,63 @@ class CommunityApiOffenderDetailsDataSourceTest {
...
+      val successBody = UserOffenderAccess(
+        userRestricted = false,
+        userExcluded = false,
+        restrictionMessage = null,
+      )
+
+      return allClientResults(successBody).intoArgumentStream()
+    }
+
+    private fun <T> allClientResults(successBody: T): List<ClientResult<T>> = listOf(
```
pezholio, line 89: This is nice!

pezholio (approved):
This is great, thanks! Nice neat reversible solution - look forward to seeing how it progresses 👍

edavey (approved):
Thanks Dan, this is really easy to follow and looks like a good way to switch these endpoints. 

I do find the `@ParameterizedTest`s with `@MethodSource`s quite hard to parse but I guess my eyes will become accustomed!

tahb (approved):
Nice abstraction and well crafted!

PR #1201, opened 2023-12-01
Turn on Sentry Performance Monitoring

pezholio:
This sets the Sample Rate to send all traces to Sentry. If this has a negative effect on performance or billing, we can reduce it or disable it completely.


tahb (approved):
With this type of change I wonder if setting the trace to 1.0 for all environments is going to use up Sentry credits/funding. Do we want to turn on in dev to 1.0, check it works, scale that down and then only turn this on in prod?

PR #1196, opened 2023-11-30
593 Seeding for local development environment

edavey:
This PR introduces "auto-seeding" for the local development environment.

**See also https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1207 which introduces "AutoScript" seeding for the dev/test environments where CSV seed files are not required / appropriate.**

There are two seeding mechanisms currently available:

### 1) auto-seeding 
This runs seeding jobs based on the CSV files found in `resources/db/seed/local+dev+test`. Auto-seeding is orchestrated by `SeedService.autoSeed()`. 

As the name of the seed directory suggests, this is run in three environments: local, dev and test. 

This is suitable for seed data which is not sensitive or private and which will be committed to our public git repo.

### 2) manual seeding
This uses the same seed jobs, but the seed CSV files are manually uploaded to a Kubernetes pod and the seeding job is triggered manually, as per [doc/how-to/run_seed_job_remotely.md](https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/72b5f3e32d484645f5893c7e9bff70fe98cc51fa/doc/how-to/run_seed_job_remotely.md). This is the right mechanism for adding sensitive or personally-identifiable information such as names, usernames, real CRNs or NOMS numbers.

In this PR we add auto-seeding for CAS2:

- **Nomis users**: prison officers who create applications
- **External users**: assessors from Nacro who add status updates to submitted applications
- Applications which are in 3 forms:
  - `IN_PROGRESS` - a full set of form `data` but no translated read-only `document`. These have a creation date of **Oct 2023**
  - `SUBMITTED` - with both `data` and `document`. These have a creation date of **Nov 2023**
  - `IN_REVIEW` - with status updates. These have a submitted date of **Dec 2023**

As we only have one person in prison set up in our local AP-Tools development environment (Aadland Bertrand CRN X320741 / NOMS A1234AI) I've used these dates (Oct / Nov / Dec) to differentiate between the applications which are seeded in different states.

### Dev and Test environments

This auto-seeding will not currently be very useful in dev/test environments:

- `POM_USER` does not exist in that environment so won't be able to log in as the applicant and see the dashboard of in progress/submitted applications
- submitted applications won't be shown correctly to a logged in assessor as the API uses the `PersonTransformer` (`OffenderDetail` and `InmateDetail`) and the subject of our application Aadland Bertrand (NOMS A1234AI) doesn't exist in dev/test

So we need to do some **more work to provide seeding in the dev/test environment**:

[...]

PR #1207, opened 2023-12-04
593 (pt 2) "auto script" seeding for dev and test environments

edavey:
In this PR we propose a new form of automated seeding: Auto-scripting.

**This PR builds on https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1196 which is waiting for review/to merge.**

AutoScripting is a form of seeding which does not use CSV
seeding files. i.e. the seeding is entirely defined in a
script. In CAS2 we will use this for seeding using real
NomisUsers and ExternalUser (in dev and test environments)
whose usernames we don't want expose in a seed file.

If AutoScript seeding is enabled (it's a sub-set of Auto-seeding),
then the `Cas2AutoScript` is run. 

That script creates three `Cas2Application`s for each
NomisUser found in the test/dev database:

- IN_PROGRESS - with a full set of in-flight form `data`, but
  not yet submitted
- SUBMITTED - recently submitted, so with full `document`
  data
- IN_REVIEW - submitted, with a few status updates

Our applications are for a test/dev user called Bertrand
Aadland (Noms G9542VP / CRN X320741) who exists in both
Community API and the Prison API. See:

https://github.com/ministryofjustice/community-api/tree/main/src/main/resources/db/data

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas2/Cas2AutoScript.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas2/Cas2AutoScript.kt
@@ -1,13 +1,138 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.seed.cas2
 
+import com.microsoft.applicationinsights.boot.dependencies.apachecommons.io.FileUtils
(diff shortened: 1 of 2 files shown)


edavey:

pezholio (approved):
This is really useful! 👍

PR #1209, opened 2023-12-05
[1724] - Creating more CAS3 booking for E2E testing

sudhakar-plr:


tahb (approved):
Looks good we now have 12 departed bookings 👍

PR #1213, opened 2023-12-06
671 Create CAS2 application.status-updated domain event 

edavey:
We now create and store a CAS2 `application.status-updated` domain event when it's received from an assessor. If the system is configured to 'emit' events (`emitDomainEventsEnabled `) then a message is sent to the domain event broadcast channel.

See https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/main/doc/architecture/decisions/0004-generate-domain-events-when-ap-data-is-changed.md for information on how we use domain events.

status-updated


edavey:
[Record <application.status-updated> domain event when assessor creates a status update](https://trello.com/c/qn54tH5T/671-record-applicationstatus-updated-domain-event-when-assessor-creates-a-status-update)

Harriethw (approved):
Thanks Ed, this is nice and clear. Great refactoring of the tests as well :)

PR #1220, opened 2023-12-07
[APS-67] Introduce facade for offender risks

danlivings-dxw:

This PR introduces a facade for retrieving offender risks, allowing us to abstract to some degree over the source of this information, simplifying the planned transition away from the Community API in favour of the AP Delius Context API.

This facade is implemented such that it can be configured through the Spring property `data-sources.offender-risks` or the environment variable `DATA-SOURCES_OFFENDER-RISKS`, which accepts values of either `community_api` or `ap_delius_context_api`. This allows for control over when and in what environments the transition to the AP Delius Context API occurs, and allows us to rollback the transition without reverting significant code changes should problems occur.

Currently, this facade only supports the Community API and it replicates the existing behaviour. Future work will provide an implementation backed by the AP Delius Context API.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/datasource/CommunityApiOffenderRisksDataSourceTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/datasource/CommunityApiOffenderRisksDataSourceTest.kt
@@ -1,0 +1,413 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.unit.datasource
+
+import io.mockk.every
+import io.mockk.mockk
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import org.springframework.http.HttpMethod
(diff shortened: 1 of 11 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/datasource/CommunityApiOffenderRisksDataSourceTest.kt
```diff
@@ -0,0 +1,212 @@
...
+    mockHMPPSTierApiClient,
+  )
+
+  @Test
+  fun `getPersonRisks returns NotFound envelopes for RoSH, Tier, Mappa & flags when respective Clients return 404`() {
+    val crn = "a-crn"
+
+    mock404RoSH(crn)
+    mock404Tier(crn)
+    mock404Registrations(crn)
```
pezholio, line 44: I like this approach very much - makes the tests so much more readable!

pezholio (approved):
LGTM - looking forward to seeing how this develops 👍

PR #1221, opened 2023-12-08
Feature/APS-68 add new columns to raw data reports

davidatkinsuk:
APS-68 Adding new columns to Application Reports, as detailed on the ticket.

Whilst the ticket details 5 columns, 2 were already being provided in the report. An example report generated locally via teh UI - [applications-2023-12 (3).xlsx](https://github.com/ministryofjustice/hmpps-approved-premises-api/files/13613686/applications-2023-12.3.xlsx)
is attached.

The following has now been resolved by populated the domain_event.booking_id column, indexing the column and backfilling the existing entries via a migration script

> The updated SQL includes a JOIN on a value taken from a 'json' column to help us find if a placement_application exists for the given Booking ID:
> 
> `left join placement_requests on cast(placement_requests.booking_id as text) = booking_made_event.data -> 'eventDetails' ->> 'bookingId'
> left join placement_applications on placement_applications.id = placement_requests.placement_application_id`
> 
> According to [postgres json type documentation](https://www.postgresql.org/docs/current/datatype-json.html), values in the json type cannot be indexed, and as such the join will most likely have performance issues with large sets of data. At a minimum this should be sanity checked against a realistic data set in pre-prod once deployed to the environment.
> 
> We could look at converting the column type to jsonb, which then allows indexing of specific values, but there are some concerns:
> 
> - switching to JSONB will lose insignificant formatting and order is not preserved. It's not clear if this will be an issue
> - switching to JSONB will drop duplicate keys - this _shouldnt_ be an issue if JSON is well defined
> - we'd have to add the index onto the domain_event.data, making the definition of the column specific to a given domain event type
> 
> Alternatively, we could look at sourcing the relationship straight from a postgres column. There is a column 'booking_id' in domain_events (added for CAS3) which isn't populated for CAS1. If this was populated for the APPROVED_PREMISES_BOOKING_MADE event, we could use this instead. This data is readily available in the BookingService that creates the event. For existing data we could use a  migration query to populate this for existing domain event entries
>

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApplicationEntityReportRow.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApplicationEntityReportRow.kt
@@ -41,7 +41,7 @@ interface ApplicationEntityReportRowRepository : JpaRepository<ApplicationEntity
       submission_event.data -> 'eventDetails' ->> 'targetLocation' as targetLocation,
       cast(withdrawl_event.data -> 'eventDetails' ->> 'withdrawnAt' as date) as applicationWithdrawalDate,
       withdrawl_event.data -> 'eventDetails' ->> 'withdrawalReason' as applicationWithdrawalReason,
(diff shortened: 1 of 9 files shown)


pezholio (approved):
LGTM - thanks! 👍

PR #1225, opened 2023-12-11
593: CAS2 seeding for dev and local envs

edavey:
This PR is a reprise of https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1196 which had to be reverted due to a problem in the way that seed file fixtures were being loaded.

See final addition refactoring:

Refactor seed fixture fixture loading: as 'resource'

Loading these fixture files with `FileUtils.readFileToString()`
does not work in our deployed environments on the Cloud
Platform where artefacts are bundled into a JAR artefact.

In order to load a file from the 'resources' bundle we need
to use some special functions: we use a special "resource
loader" to generate a stream which we then pass to
`FileCopyUtils.copyToString()`

See https://www.baeldung.com/spring-load-resource-as-string

We wrap the resource loading code in a `try` block
so that the seeding migration process will continue in event
of problems. In this way this "auto-script" seeding should
no longer cause a deployment failure.


Harriethw (approved):
amazing work Ed, thanks for figuring this out.
If possible would be good to get someone more experienced with seeding than myself to look at it as well, but it looks good as far as I'm concerned.

edavey:

PR #1228, opened 2023-12-11
Change cas3 e2e test booking to use non-lao crn

sudhakar-plr:
This PR is to change the CAS3 e2e test booking to use Non-LAO CRN to make the booking pagination to work in E2E.


tahb (approved):
Looks good. We can drop all those comments that attempt to describe what's unique about that CRN "-- Multiple offences" just incase that misdirects someone in future

sudhakar-plr:
> Looks good. We can drop all those comments that attempt to describe what's unique about that CRN "-- Multiple offences" just incase that misdirects someone in future

Done

PR #1231, opened 2023-12-12
593 pt 2 CAS2 "auto-script" seeding for dev test

edavey:
This is a reprise of https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1207 which was reverted due to a problem loading fixture files from the "resources" bundle.

In this version of the PR the loading of files is:

- done using a "resource loader"
- wrapped in a "try" block to avoid causing a error which blocks deployment

See https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/bde98f81a698c61a979a498f4831f569759bf5d1/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas2/Cas2AutoScript.kt#L133-L144


edavey:

Harriethw (approved):

PR #1235, opened 2023-12-12
Booking search query optimisations

tahb:
- Exists should stop counting when it matches the condition. Using ‘Select 1’ instead of ‘count 1’ should return the first record in a simpler way
- Add a database index for each subquery, it didn't seem like we had these indexes:

Screenshot 2023-12-12 at 16 34 04

This isn't going to get us out of the N+5 problem we have but it might buy us some time.


pezholio (approved):
I think this would be a big improvement. There's still work to do on the query, but I think adding the indexes in particular will really help

PR #1237, opened 2023-12-13
Run performance tests with Gatling

pezholio:
Temporarily trying to run these on push. These should probably be run as part of the deploy pipeline, or on a schedule.


nsmj63-dxw:
I've picked this up to attempt to get tests working correctly on the dev environment.

tahb:
I ended up doing [a dive into getting CAS2 Gatling tests](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1797) working and came to some conclusions that suggests this is worth closing. 

The problem was that CAS3 tests worked locally and locally when pointed at dev. When making it run on CI we couldn't prevent it returning 403.

After digging into why it was broken for CAS2 I don't think Gatling tests can be run on CI due to how it attempts to imitate the authentication exchange with HMPPS Auth.

The Gatling tests try to imitate the frontend by crafting requests to start a basic authentication exchange which results in a limited token. Given we are running from within the API context there is no way to request an expanded access token since there is no callback endpoint for the HMPPS Auth response that we can intercept, this returns the important expanded access token with extra data. In the CAS2 context that token and additional data is required to authenticate with the nomis-roles-api. Adding an API auth callback just for Gatling to intercept a token is convoluted.

I suggest that load tests are probably better designed so that they are replayed through the whole stack - including the frontend - which first logs a machine user in before beginning the tests. Something like https://k6.io/.

PR #1238, opened 2023-12-13
aps-48 stop using static total_beds column and calculate on-the-fly

davidatkinsuk:
Before this commit, the bed count shown when listing premises did not match the number of beds configured in the system. This was because the count was being populated from premises.total_beds, which was not neccessarily being kept in sync with the number of beds registered in the system (e.g. when a new bed was added via the API, the count was not being modified).

This commit removes all code relying on the premises.total_bed column and instead always calculates the bed count on-the-fly.

When performing find operations in the PremisesRepository, a sub-query is used to determine the bed count. Whilst this does not create a particularly performant execution plan (a subplan us executed for each row), this was deemed acceptable given that the size of the premises table is small, and if performance was an issue we could add paging to the find operations.

I attempted to find a way to make JPA/Hibernate return the total bed count alongside the PremisesEntity when performing get operations in the PremisesRepository, but this was outside the capabilities of JPQL without resorting to listing out all entity fields in the SELECT, which would be cumbersome and hard to maintain. Therefore, we make a separate SQL call to count the number of beds once we have retrieved the PremisesEntity. This is preferable to using lazy loads over the premises->rooms->beds relationship which would trigger multiple SQL calls.

This commit does not remove the column as this would complicate rollback/rollforward fixing if there’s any issues. Instead the column will be removed once we’re happy that this change is stable.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PremisesEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PremisesEntity.kt
@@ -26,7 +26,20 @@ import javax.persistence.Table
 
 @Repository
 interface PremisesRepository : JpaRepository<PremisesEntity, UUID> {
-  fun findAllByProbationRegion_Id(probationRegionId: UUID): List<PremisesEntity>
+
+  @Query("SELECT p as premises, (SELECT CAST(COUNT(b) as int) FROM p.rooms r JOIN r.beds b WHERE r.premises = p GROUP BY p) as roomCount FROM PremisesEntity p")
+  fun findAllWithRoomCount(): List<PremisesWithRoomCount>
+
+  @Query(
+    """
+        SELECT 
+          p as premises, 
+          (SELECT CAST(COUNT(b) as int) FROM p.rooms r JOIN r.beds b WHERE r.premises = p GROUP BY p) as roomCount 
+        FROM PremisesEntity p
+        WHERE p.probationRegion.id = :probationRegionId
+    """,
+  )
+  fun findAllByProbationRegion(probationRegionId: UUID): List<PremisesWithRoomCount>
 
   @Query(
     """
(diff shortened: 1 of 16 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PremisesEntity.kt
```diff
@@ -26,7 +26,20 @@ import javax.persistence.Table
...
 interface PremisesRepository : JpaRepository<PremisesEntity, UUID> {
-  fun findAllByProbationRegion_Id(probationRegionId: UUID): List<PremisesEntity>
+
+  @Query("SELECT p as premises, (SELECT CAST(COUNT(b) as int) FROM p.rooms r JOIN r.beds b WHERE r.premises = p GROUP BY p) as roomCount FROM PremisesEntity p")
+  fun findAllWithRoomCount(): List<PremisesWithRoomCount>
+
+  @Query(
+    """
+        SELECT 
+          p as premises,
```
pezholio, line 36: Ah, this is very useful!

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/PremisesTest.kt
```diff
@@ -1179,23 +1168,21 @@ class PremisesTest : IntegrationTestBase() {
           probationRegionEntityFactory.produceAndPersist { withYieldedApArea { apAreaEntityFactory.produceAndPersist() } }
         }
         withService("CAS1")
-        withTotalBeds(20)
-      }
+      }.onEach { addRoomsAndBeds(it, 5, 4) }
```
pezholio, line 1171: Nice little helper here 👍

pezholio: Might be good to have the room and bed counts as named parameters so they're not so "magic number"-y

pezholio (approved):
LGTM - a lot of fiddly tests to unpick, so I can see why the commit's so big!

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/PremisesTest.kt
davidatkinsuk: Good idea, i've done this 👍

PR #1239, opened 2023-12-13
Exclude entities from LongParameterList rule

davidatkinsuk:
Entities typically have large constructors, reflecting the columns in the database tables. Trying to reduce the number of arguments is non-trivial and may result in worse performance and an unnecessarily fragmented database design.

This commit globally excludes Entities from that rule, and removes any detekt-baseline.xml entries that are now covered by this exclude.


detekt.yml
```diff
@@ -33,5 +33,9 @@ complexity:
       # autowiring works. Spring beans will therefore require a larger-scale refactoring to reduce the number of
       # dependencies required by each one that would violate this rule.
       - org.springframework.stereotype.*
+      # Entities typically have large constructors, reflecting the columns in the database tables. Trying to reduce
+      # the number of arguments is non-trivial and may result in worse performance and an unnecessarily fragmented
+      # database design.
```
danlivings-dxw, line 38: This is a good, succinct explanation of the configured rule.

danlivings-dxw (approved):
LGTM 👍

pezholio (approved):
Yeah, happy with this - we've got a couple of `@Suppress` annotations with `LongParameterList` included, so would be good to get rid of these in this branch too if we can 👍

davidatkinsuk:
> Yeah, happy with this - we've got a couple of `@Suppress` annotations with `LongParameterList` included, so would be good to get rid of these in this branch too if we can 👍

Good thinking, I didn't consider checking for existing @Suppress annotations. I've just checked and all our existing LongParameterList annotations are against types other than entities, so i'll leave them as-is

PR #1242, opened 2023-12-13
Return full User on CAS2 Application

Harriethw:
We need to display the user's email address and name on Applications, so we are now returning the full user rather than just the id.


pezholio (approved):
LGTM 👍

PR #1243, opened 2023-12-13
Add performance test for booking search API

davidatkinsuk:
Added an additional journey into the BookingsReportGenerationTimeSimulation to measure performance of the booking search API

### DIFF
--- a/doc/how-to/run_performance_tests.md
+++ b/doc/how-to/run_performance_tests.md
@@ -11,6 +11,12 @@ Once the API has started locally, just run:
 ./gradlew gatlingRun
 ```
 
+To run a specific test define the FQN of the simulation:
+
+```shell
+./gradlew gatlingRun-uk.gov.justice.digital.hmpps.approvedpremisesapi.gatling.simulations.BookingsTimeSimulation```
+```
+
 ## Running against a remote environment
 This requires some configuration, which can be done through environment variables:
 ```shell
--- a/src/gatling/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/gatling/simulations/ApplyJourneyStressSimulation.kt
+++ b/src/gatling/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/gatling/simulations/ApplyJourneyStressSimulation.kt
@@ -1,5 +1,6 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.gatling.simulations
 
+import io.gatling.javaapi.core.CoreDsl
 import io.gatling.javaapi.core.CoreDsl.constantUsersPerSec
 import io.gatling.javaapi.core.CoreDsl.repeat
 import io.gatling.javaapi.core.CoreDsl.scenario
@@ -13,10 +14,12 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.gatling.util.authorizeUs
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.gatling.util.getUUID
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.gatling.util.withAuthorizedUserHttpProtocol
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.randomInt
+import kotlin.time.Duration.Companion.milliseconds
 import kotlin.time.Duration.Companion.minutes
 import kotlin.time.Duration.Companion.seconds
 import kotlin.time.toJavaDuration
 
+@Suppress("MagicNumber")
 class ApplyJourneyStressSimulation : Simulation() {
   private val applicationIdKey = "application_id"
   private val getApplicationId = { session: Session -> session.getUUID(applicationIdKey) }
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -81,6 +81,20 @@ Then either:
  - Run or Debug the `verification`, `test` Task from the "Gradle" panel
  - Open an individual test class and click the green play icon in the gutter at the class level or on a specific test method
 
+## Running Performance Tests
+
+To run the performance tests use gatlingRun gradle task
```
tahb, line 86: Having something in the readme is a good idea however I think it could be a pointer to the [existing how-to doc](https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/main/doc/how-to/run_performance_tests.md) instead to avoid the duplication and keep the readme concise. 

We should add those instructions for running a specific too, that's very helpful.

src/gatling/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/gatling/simulations/BookingsReportGenerationTimeSimulation.kt
```diff
@@ -127,6 +127,19 @@ class BookingsReportGenerationTimeSimulation : Simulation() {
   )
     .pause(1.seconds.toJavaDuration())
 
+  private val bookingSearch = exec(
```
tahb, line 130: What are your thoughts around the existing name of this simulation? We've got BookingsReportGenerationTimeSimulation which implies it's only used for reports, which isn't the case with the need to test booking search. 

I suppose this file is already doing a lot of booking set up that we'd like to reuse. I think we could either rename this file to make it more generic, BookingsTimeSimulation? Or make a new one alongside it that extracts and reuses the same booking creation parts?

src/gatling/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/gatling/simulations/BookingsReportGenerationTimeSimulation.kt
```diff
@@ -140,6 +153,11 @@ class BookingsReportGenerationTimeSimulation : Simulation() {
       setupBooking.injectOpen(
         constantUsersPerSec(20.0).during(60.seconds.toJavaDuration()),
       )
+        .andThen(
+          bookingsSearchJourney.injectOpen(
```
tahb, line 158: We're currently trying to figure out how we write assertions for the other tests so that these fail if any request takes longer than say 5 seconds.

We had a quick test and if we follow this https://gatling.io/docs/gatling/reference/current/core/assertions/ we get global() undefined or similar. Probably not importing something correctly I suspect. I appreciate this probably wasn't in the original ask but is something we need to figure out before the test will notify anyone of a problem. [Thread here if you have any ideas or can make it work](https://mojdt.slack.com/archives/C03K0HB0LBE/p1702573213708589?thread_ts=1702474801.256889&cid=C03K0HB0LBE)

tahb (commented):
Thanks for this. I've left a couple of suggestions. I do think we might need to think where these new tests live.

src/gatling/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/gatling/simulations/BookingsReportGenerationTimeSimulation.kt
davidatkinsuk: I agree the name isn't really reflective and undersells how much the test is doing. I've renamed it to the proposed BookingsTimeSimulation as that's far more reflective of what it does.

Regarding splitting the test. I think there's a trade of here between cohesive tests and having tests that won't take forever to run in circle. My instinct is to go for fewer large tests, or possibly even one simulation that hits on all areas of interest concurrently. Maybe we break our simulations in to re-usable steps, and then have a CI-specific simulation that makes use of all the parts (if that's at all manageable!)

src/gatling/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/gatling/simulations/BookingsReportGenerationTimeSimulation.kt
davidatkinsuk: I'll take a look once i've finished my current ticket  👍

README.md
davidatkinsuk: cool, i wasn't aware of the how-to docs. i've removed this and updated the how-to

src/gatling/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/gatling/simulations/BookingsReportGenerationTimeSimulation.kt
davidatkinsuk: I've added some global() assertions in for both simulations and i can see the build failing if they're sufficient low. They're very high at the moment to get tests running in dev after a few runs. I guess we at least we have a high water mark we can start pulling down

pezholio (approved):
LGTM 👍

PR #1244, opened 2023-12-13
Turn off all domain events in dev

tahb:
We are trying to add load tests as part of CI via Gatling. As these will send many, many thousands of requests we should prevent this level of traffic being consumed by integrations and all that data going into ndelius etc.

This level of data may affect their performance and/or produce a huge amount of data that we and other services are then unable to load.

Domain events should still be enabled in all other environments and can be tested by the integrations and ndelius team in pre-prod.

[It doesn't look like the Integrations team use dev for domain event based testing](https://github.com/ministryofjustice/hmpps-probation-integration-e2e-tests/tree/bd26450ed5fdd57bc2fb38e53f725686d947930f/steps/approved-premises). If they were to start, they probably ought to use the test environment.


tahb:
@Harriethw @pezholio Can we turn off your domain events in dev?

Harriethw (approved):
I don't see that this would affect us on CAS2 as we are not yet doing anything with the domain events. 
Tagging @edavey for when he's back next week though as he's been starting to create domain events for our MI reports.

pezholio (changes requested):
This isn't a goer I don't think, as the Integration Team's integration tests rely on domain events being on to work

tahb:
We'll have to find another way, perhaps a dedicated load test environment?

PR #1247, opened 2023-12-14
Add Person fields to CAS2 Application Entity

Harriethw:
This is another try at introducing 
* https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1226

> Currently, we store the CRN of a person on an application, and then use that CRN to make an API call to retrieve the rest of the person's data whenever we return an application or application summary.
> 
> Instead, we would like to store a "snapshot" of the information about the Person on an CAS2 application, at the moment when the application is created.

This involves adding new columns to the cas2_applications table, some of which are `NOT NULL`. It previously failed on the dev environment, I think it might have been because I hadn't included the new columns in the `cas2_applications.csv`. 

I will have to exec into preprod to delete any existing cas2 applications in that environment, but I'm working on the assumption that dev/test databases get wiped before these migrations are run.


tahb:
> but I'm working on the assumption that dev/test databases get wiped before these migrations are run.

[This is what should be wiping applications and assessments on dev+test environments](https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/main/src/main/resources/db/migration/dev%2Btest/R__3_clear_applications.sql).

tahb:
As a higher level change this is interesting to see because I think it's a divergence from how we've been treating this category of data so far. We avoid being a store for personal details and take ourselves out of scope of being a handler of PII and instead suffer the consequences of loading any information on demand.

Have we considered the DPIA implications of this, are we confident that we can take this position, pass a higher level of scrutiny and follow-up with any requests to do things like scrub it after a certain period or on request? Just trying to think of how this might bite us in the future. 

If we are confident then I can imagine the other CAS services might want to pivot to this approach and gain the performance improvements.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas2ApplicationEntity.kt
```diff
@@ -25,6 +28,14 @@ interface Cas2ApplicationRepository : JpaRepository<Cas2ApplicationEntity, UUID>
 SELECT
     CAST(a.id AS TEXT) as id,
     a.crn,
+    a.noms_number as nomsNumber,
```
tahb, line 31: I wonder why we have to list these all out instead of *? As we're using a summary object I presume we don't have to list them out in order to avoid returning the `data` blob.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/transformer/cas2/ApplicationsTransformerTest.kt
```diff
@@ -85,35 +90,64 @@ class ApplicationsTransformerTest {
...
+        override fun getPncNumber() = "PNC321"
+        override fun getName() = "Robert Smith"
+        override fun getDateOfBirth() = LocalDate.now().minusYears(60)
+        override fun getNationality() = "British"
+        override fun getSex() = "Male"
+        override fun getPrisonName() = "HMP Bristol"
+        override fun getPersonStatus() = Cas2NewApplication.PersonStatus.inCustody.name
       }
 
+      val person = FullPerson(
```
tahb, line 107: We could have a test that tests that a person with none of the optional fields handles correctly, presumably returning a `FullPerson` with null for some attributes, which sounds a bit confusing because you might expect an object called 'full' to have everything.

src/main/resources/static/codegen/built-api-spec.yml
```diff
@@ -6066,6 +6066,38 @@ components:
...
+          type: string
+        personStatus:
+          type: string
+          enum:
+            - InCustody
+            - InCommunity
+            - Unknown
+        prisonName:
+          type: string
+        pncNumber:
```
tahb, line 6093: This might be different for your users but PNC was certainly a very important number in cas3 reports.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas2/Cas2ApplicationsSeedJob.kt
```diff
@@ -17,9 +17,11 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.seed.SeedJob
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas2.JsonSchemaService
 import java.io.IOException
 import java.io.InputStreamReader
+import java.time.LocalDate
 import java.time.OffsetDateTime
 import java.util.UUID
 
+@Suppress("MagicNumber")
```
tahb, line 24: What's this rule, do we think it's reasonable to suppress it or is this a signal of something we ought to change?

Harriethw:
>Have we considered the DPIA implications of this, are we confident that we can take this position, pass a higher level of scrutiny and follow-up with any requests to do things like scrub it after a certain period or on request? Just trying to think of how this might bite us in the future.

This might be one for @edavey and Ben to be looped in - I think the thinking is for CAS2 we're sending a snapshot of the data for an application at a certain time to Nacro, and we need to have a concrete record of what the referrer is agreeing to send over, and we need to surface that data on in-progress applications and submitted applications. 

I can check but I think our DPIA has been done with the assumption that we're storing and sharing this PII with Nacro. 

I don't think we have a plan for how to scrub data except that "it can be done manually if/when requested", which would still be possible?

tahb (commented):
Left some thoughts and though I've read through it all  a more experienced kotlin eye is probably a good idea.

My main concern is in a comment around our approach to personal data and if the API starts here, that has implications for handling and for CAS1 and 3.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas2/Cas2ApplicationsSeedJob.kt
Harriethw: good spot - accidentally left this in, can take this out now

src/main/resources/static/codegen/built-api-spec.yml
Harriethw: It's come up as something Nacro ask for but not sure how required it is for a referral - it's an optional field on offenders returned from the probation-offender-search so we have to account for some cases where it might be missing.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas2ApplicationEntity.kt
Harriethw: I don't think I understand what exactly is happening well enough to answer.
It's taking `FROM cas_2_applications` so unless we did want all the columns from a full application (including `data` we need to list out the columns we do want, that pertain to an `ApplicationSummary`? Or is the Java side able to take a whole application returned from the sql query and only return a summary?

Harriethw:
Going to pause this then while we have conversation  https://mojdt.slack.com/archives/C04T6SC31HA/p1702563783184729?thread_ts=1702563252.090339&cid=C04T6SC31HA

Harriethw:
closing because we don't know if we will do this yet.

PR #1249, opened 2023-12-14
add pncNumber to Person returned on application

Harriethw:
We need to show the PNC number of a person on an application. This adds it to the Person Transformer method that is used in the Application Transformer `transformJpaToApi` so that when we get an Application this will be returned.


pezholio (approved):
LGTM 👍

PR #1250, opened 2023-12-14
Gatling can be configured with a CRN

tahb:
Depending on the user being used, they will have access to different CRNs in different environments.

Jimsnowldap can see X320741 locally.

However we want to use a dedicated load test on a real environment. This means our user won’t have access to that CRN so we need to make this configurable to avoid 403s during the application create process.

We will need GATLING_CRN=X371199 set for these gatling tests to work on a real environment.


tahb:
I've added it to CircleCI.

pezholio (approved):
Nice find!

PR #1252, opened 2023-12-14
[1742] - Add new booking status column and migrate CAS3 booking status

sudhakar-plr:
# Changes in this PR
Refer [1742](https://trello.com/c/6XMIEPmD/1742-improve-booking-search-by-using-a-status-column) for more information.

This PR contains changes to introduce new `status` column in `bookings` table.  This PR contains 2 changes

1. The `status` column will be populated based on the below booking API :
``` 
/placement-requests/{id}/booking  --> confirmed
/premises/{premisesId}/bookings.  --> confirmed
/premises/{premisesId}/bookings(CAS3)  --> provisional
/premises/{premisesId}/bookings/{bookingId}/arrivals --> arrived
/premises/{premisesId}/bookings/{bookingId}/cancellations --> cancelled
/premises/{premisesId}/bookings/{bookingId}/confirmations --> confirmed
/premises/{premisesId}/bookings/{bookingId}/departures --> departed

```

2. Migrate existing CAS3 booking status.
The migration of existing CAS3 booking status is done using migrationJob, which pick existing CAS3 booking with status=null and determine the migration status using below logic:

```
If 'cancellation' record exists for the booking then 'cancelled'
If 'departure' record exists for the booking then 'departure'
If 'arrivals' record exists for the booking then 'arrived'
If 'confirmation' record exists for the booking then 'confirmed'
anything else 'provisional'
```

# Release checklist

confirm:

## Pre merge checklist

    advance?

## Post merge checklist

considering it done.

fix it as soon as possible.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BookingEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BookingEntity.kt
@@ -2,7 +2,9 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity
 
 import org.springframework.data.domain.Page
 import org.springframework.data.domain.Pageable
+import org.springframework.data.domain.Slice
 import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.data.jpa.repository.Modifying
 import org.springframework.data.jpa.repository.Query
 import org.springframework.stereotype.Repository
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.BookingStatus
@@ -152,6 +154,16 @@ interface BookingRepository : JpaRepository<BookingEntity, UUID> {
     nativeQuery = true,
   )
   fun findBookings(serviceName: String, status: BookingStatus?, probationRegionId: UUID?, pageable: Pageable?): Page<BookingSearchResult>
+
+  @Modifying
+  @Query("UPDATE BookingEntity b set b.status = :status where b.id = :bookingId")
+  fun updateBookingStatus(bookingId: UUID, status: BookingStatus)
+
+  @Query(
+    "SELECT * FROM bookings WHERE status IS NULL AND service='temporary-accommodation' ",
+    nativeQuery = true,
+  )
+  fun findAllCas3bookingsWithNullStatus(pageable: Pageable?): Slice<BookingEntity>
 }
 
 @EntityListeners(BookingListener::class)
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BookingEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BookingEntity.kt
@@ -15,6 +15,8 @@ import java.util.Objects
 import java.util.UUID
 import javax.persistence.Entity
 import javax.persistence.EntityListeners
+import javax.persistence.EnumType
+import javax.persistence.Enumerated
 import javax.persistence.Id
 import javax.persistence.JoinColumn
 import javax.persistence.ManyToOne
@@ -197,6 +199,8 @@ data class BookingEntity(
   var nomsNumber: String?,
   @OneToOne(mappedBy = "booking")
   var placementRequest: PlacementRequestEntity?,
+  @Enumerated(value = EnumType.STRING)
+  var status: BookingStatus?,
 ) {
   val departure: DepartureEntity?
     get() = departures.maxByOrNull { it.createdAt }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/migration/BookingStatusMigrationJob.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/migration/BookingStatusMigrationJob.kt
@@ -1,0 +1,52 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.migration
+
+import org.slf4j.LoggerFactory
+import org.springframework.data.domain.PageRequest
+import org.springframework.data.domain.Slice
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.BookingStatus
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.BookingEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.BookingRepository
+import javax.persistence.EntityManager
+
+class BookingStatusMigrationJob(
+  private val bookingRepository: BookingRepository,
+  private val entityManager: EntityManager,
+  private val pageSize: Int,
+) : MigrationJob() {
+  private val log = LoggerFactory.getLogger(this::class.java)
+  override val shouldRunInTransaction = true
+
+  override fun process() {
+    log.info("Starting Booking Migration process...")
+
+    var page = 1
+    var hasNext = true
+    var slice: Slice<BookingEntity>
+
+    while (hasNext) {
+      log.info("Getting page $page")
+      slice = bookingRepository.findAllCas3bookingsWithNullStatus(PageRequest.of(0, pageSize))
+      slice.content.forEach {
+        setStatus(it)
+      }
+      entityManager.clear()
+      hasNext = slice.hasNext()
+      page += 1
+    }
+    log.info("Booking Status Migration process complete!")
+  }
+
+  private fun setStatus(booking: BookingEntity) {
+    booking.status = when {
+      booking.cancellations.isNotEmpty() -> BookingStatus.cancelled
+      booking.departures.isNotEmpty() -> BookingStatus.departed
+      booking.arrivals.isNotEmpty() -> BookingStatus.arrived
+      booking.confirmation != null -> BookingStatus.confirmed
+      else -> BookingStatus.provisional
+    }
+
+    log.info("Updating booking status ${booking.id} to ${booking.status}")
+    entityManager.detach(booking)
+    bookingRepository.updateBookingStatus(booking.id, booking.status!!)
+  }
+}
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
@@ -25,6 +25,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.PersonN
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.PersonReference
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.Premises
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.StaffMember
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.BookingStatus
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.PlacementDates
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ServiceName
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.ClientResult
@@ -225,7 +226,7 @@ class BookingService(
           turnarounds = mutableListOf(),
           nomsNumber = placementRequest.application.nomsNumber,
           placementRequest = null,
-          status = null,
+          status = BookingStatus.confirmed,
         ),
       )
 
@@ -379,7 +380,7 @@ class BookingService(
           dateChanges = mutableListOf(),
           nomsNumber = nomsNumber,
           placementRequest = null,
-          status = null,
+          status = BookingStatus.confirmed,
         ),
       )
 
@@ -713,7 +714,7 @@ class BookingService(
           offlineApplication = null,
           turnarounds = mutableListOf(),
           placementRequest = null,
-          status = null,
+          status = BookingStatus.provisional,
         ),
       )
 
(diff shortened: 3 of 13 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
```diff
@@ -825,6 +826,7 @@ class BookingService(
 
     booking.arrivalDate = arrivalDate
     booking.departureDate = expectedDepartureDate
+    booking.status = BookingStatus.arrived
```
tahb, line 829: Setting this after the ArrivalEntity is created makes sense because if that fails, we don't want to set the status. The way this is order dependent feels a touch fragile and probably applies to the other state changes too.

I'm wondering if an after create hook (if that's even possible) could be set up on the arrival entity so that if this line gets moved around the two states can't be out of sync. Whenever an arrival entity is created that entity goes and updates the parent booking, hopefully this would provide a better guarantee. That would mean even more UPDATEs though which isn't great either.

What are everyone's thoughts?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BookingEntity.kt
```diff
@@ -152,6 +154,16 @@ interface BookingRepository : JpaRepository<BookingEntity, UUID> {
...
+
+  @Modifying
+  @Query("UPDATE BookingEntity b set b.status = :status where b.id = :bookingId")
+  fun updateBookingStatus(bookingId: UUID, status: BookingStatus)
+
+  @Query(
+    "SELECT * FROM bookings WHERE status IS NULL AND service='temporary-accommodation' ",
+    nativeQuery = true,
+  )
+  fun findAllBookingWithNullStatus(pageable: Pageable?): Slice<BookingEntity>
```
tahb, line 166: Maybe we can name this as `findAllCas3bookingsWithNullStatus` since it's no longer derived?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BookingEntity.kt
```diff
@@ -152,6 +154,16 @@ interface BookingRepository : JpaRepository<BookingEntity, UUID> {
     nativeQuery = true,
   )
   fun findBookings(serviceName: String, status: BookingStatus?, probationRegionId: UUID?, pageable: Pageable?): Page<BookingSearchResult>
+
+  @Modifying
+  @Query("UPDATE BookingEntity b set b.status = :status where b.id = :bookingId")
+  fun updateBookingStatus(bookingId: UUID, status: BookingStatus)
+
+  @Query(
+    "SELECT * FROM bookings WHERE status IS NULL AND service='temporary-accommodation' ",
```
tahb, line 163: Is there any way we can eager load the associations up front in such a way that Hibernate doesn't have to make all the subqueries? I think i've tried this in the past but couldn't make it do it, interested to maybe learn if this just isn't possible on a query-by-query basis.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/MigrateBookingStatusTest.kt
```diff
@@ -0,0 +1,321 @@
...
+
+        migrationJobService.runMigrationJob(MigrationJobType.bookingStatus, 1)
+
+        assertBookingStatusIsNull(booking, approvedPremises)
+      }
+    }
+  }
+
+  @Test
+  fun `Should not migrate CAS3 booking with existing status 'arrived' with existing departure entity and returns 202 response`() {
```
tahb, line 36: If a booking with the arrived status has a departure record (I'm not sure we expect this situation to happen), should it not be updated to departed after all?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/MigrateBookingStatusTest.kt
```diff
@@ -0,0 +1,321 @@
...
+        assertBookingStatusIsNull(departedBooking2, ServiceName.temporaryAccommodation)
+        assertBookingStatusIsNull(arrivedBooking, ServiceName.temporaryAccommodation)
+        assertBookingStatusIsNull(confirmedBooking, ServiceName.temporaryAccommodation)
+        assertBookingStatusIsNull(provisionalBooking, ServiceName.temporaryAccommodation)
+
+        migrationJobService.runMigrationJob(MigrationJobType.bookingStatus, 1)
+
+        assertBookingStatus(departedBooking1, BookingStatus.departed, ServiceName.temporaryAccommodation)
+        assertBookingStatus(departedBooking2, BookingStatus.departed, ServiceName.temporaryAccommodation)
+        assertBookingStatus(arrivedBooking, BookingStatus.arrived, ServiceName.temporaryAccommodation)
```
tahb, line 180: This test describes itself as

>  fun `Successfully migrate CAS3 only 'departed' bookings when request is for 'departed' bookings and multiple booking exists in the DB`() {

But I think it's actually updating the booking statuses for arrived, confirmed and provisional too. Perhaps the description is wrong or we need to make sure that these bookings still have a null status?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/MigrateBookingStatusTest.kt
```diff
@@ -0,0 +1,321 @@
...
+        migrationJobService.runMigrationJob(MigrationJobType.bookingStatus, 1)
+
+        assertBookingStatus(booking1, BookingStatus.confirmed, ServiceName.temporaryAccommodation)
+        assertBookingStatus(booking2, BookingStatus.confirmed, ServiceName.temporaryAccommodation)
+      }
+    }
+  }
+
+  @Test
+  fun `Successfully migrate CAS3 only 'departed' bookings when request is for 'departed' bookings and multiple booking exists in the DB`() {
```
tahb, line 161: I was hoping to see a test that tried a combination of bookings in various states, woop! 🙌

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/MigrateBookingStatusTest.kt
```diff
@@ -0,0 +1,321 @@
...
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.BookingStatus
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.MigrationJobType
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ServiceName
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.integration.givens.`Given a User`
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.integration.givens.`Given an Offender`
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.BookingEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.BookingRepository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserEntity
+
+class MigrateBookingStatusTest : MigrationJobTestBase() {
```
tahb, line 15: Being able to write these tests is exactly what we hoped for! Nice find and use of these jobs 👍

tahb (commented):
This looks great, very easy to follow, thank you.

I've left a couple of questions and one main suggestion about whether after commit hooks might be a more reliable way to keep our status field in sync with the records we create.

Otherwise I have questions for CAS1 @pezholio:

- are you happy that we default your statuses to confirmed?
- do you need anything from us to backfill the legacy CAS1 bookings with this status? 

I approve though I'll wait for CAS1 to review too and hit the green button if all looks good.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/MigrateBookingStatusTest.kt
sudhakar-plr: Its my fault, i forget to change the description. Initially i developed an API to migrate 'status' by status and then realised we already got migration tool, so i changed the approach but forget to change the description.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
pezholio: We do a similar thing for applications using Listeners (see https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1119)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/migration/BookingStatusMigrationJob.kt
```diff
@@ -0,0 +1,52 @@
...
+        setStatus(it)
+      }
+      entityManager.clear()
+      hasNext = slice.hasNext()
+      page += 1
+    }
+    log.info("Booking Status Migration process complete!")
+  }
+
+  private fun setStatus(booking: BookingEntity) {
```
pezholio, line 39: It'd be good if we could update this to handle CAS1 Bookings too - I think we'd be fine to have bookings default to `provisional` for now as we'll be introducing the concept of confirmations soon

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BookingEntity.kt
sudhakar-plr: I hve tried it but getting 'org.hibernate.loader.MultipleBagFetchException: cannot simultaneously fetch multiple bags' . Looks like its not straight forward changes for this SQL.
I even dont want to change the entity to annotate @Eager Fetch.

pezholio (commented):
I'm happy for this to go in as is - but it'd be good to migrate the CAS1 statuses as well (even if we won't be using them), just to avoid confusion. That said, I don't want to be a blocker, so equally, it's fine to leave alone 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
sudhakar-plr: I would prefer the service owns the responsibility of updating the booking status , so that the service has control over when and how a booking status get updated.
i'm not sure even we can add EventListener for this as such we are updating different entities.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BookingEntity.kt
sudhakar-plr: done

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/MigrateBookingStatusTest.kt
sudhakar-plr: If booking already has status, then we dont want migrate the booking. We only pick up booking for migration when the status=null and service=temporary-accommodation.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/migration/BookingStatusMigrationJob.kt
sudhakar-plr: I left the migration of CAS1 and CAS2 booking intentionally, because i dont have the requirement and I dont want to assume and create wrong status. 
I would request you to add story in respective team to migrate the status as per the requirement.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BookingEntity.kt
pezholio: That `MultipleBagFetchException` error it the bane of my life!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/migration/BookingStatusMigrationJob.kt
pezholio: Yeah, that's totally fair 👍

pezholio (approved):
Looks good from my POV

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
tahb: I've just spotted that these methods are marked with `@Transactional` so if the arrival fails to create then we should never have the booking.status=arrived. 

So long as we have this on all the service methods I think we are protected from someone in the future unknowingly changing the ordering of these calls. 👍 

On the subject of listeners I think it comes down to appetite to risk. We should be able to trust our services to take this responsibility however there is always a risk that in future one of the many services are changed unknowingly, a new service method takes over or even records start being created outside of services(!). Using listeners does, I think, provide a better guarantee of the status always being updated. 

Perhaps we go with this and if it becomes flakey we pivot to listeners?

I hope we could [write a post-create action that fetches the parent booking and updates it](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1119/files#diff-3554486b057c692da1ac1f3b75addc76f71670c8efbdc4594e42eda5b8edfed8R11).

tahb (approved):
LGTM. I think we should release this before lunch so that we are still around this PM to roll it back, if we can't then I think it's best we wait until our return. @pezholio if it goes go in today we are both off next week so any problems please revert this.

PR #1254, opened 2023-12-15
(APS-66) Add new referrals report

pezholio:
This returns a list of all unique applications made. If an application has an associated booking (i.e. it’s been made through the standard Apply - Assess - Placement flow), then we include information about the associated booking too. This is in addition to the main Applications report, which we’re keeping around for now as it contains more information. Once the Placement Requests report is done, we’ll remove the old-style Applications report.


davidatkinsuk (approved):
LGTM

PR #1256, opened 2023-12-15
Allow a CAS2 Admin to view submitted applications

Harriethw:
We have the need on CAS2 for an admin account who can view any submitted application This is likely to be an internal user, so I am assuming they will be using a NOMIS account to login.


edavey (approved):
This looks spot-on, thks.

PR #1257, opened 2023-12-15
[APS-67] Add aggregate methods to offender details facade

danlivings-dxw:

This PR introduces aggregate methods to the `OffenderDetailsDataSource`. These aggregate methods present an interface equivalent to calling the singular methods within a `map` function, e.g.:

```kotlin
// The following are equivalent
val summaries1 = crns.map { dataSource.getOffenderDetailSummary(it) }
val summaries2 = dataSource.getOffenderDetailSummaries(crns)

// The following are equivalent
val userAccess1 = crns.map {
  dataSource.getUserAccessForOffenderCrn(username, it)
}
val userAccess2 = dataSource.getUserAccessForOffenderCrns(username, crns)
```

However, these aggregate methods will allow for optimisation in the future by coalescing multiple CRNs into a single request, as the AP Delius Context API provides bulk data endpoints.

This PR also provides an alternate implementation for `OffenderService.getOffenderSummariesByCrns` that delegates to these facade methods instead of directly interacting with the AP Delius Context API client. Additionally, it refactors several classes where `OffenderService.getOffenderByCrn` is called within a loop to use the above method. These are:
- [ ] `TasksController` (**NB 2023-12-15**: This has been refactored to make the switch to the aggregate method easier but the above method is not yet being used.)
- [ ] `BookingSearchService` (**NB 2023-12-15**: Not yet started but relevant method call is very close to the loop so should be straightforward to replace.)
- [ ] `CalendarService` (**NB 2023-12-15**: Not yet started but relevant method call is very close to the loop so should be straightforward to replace.)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingSearchService.kt
```diff
@@ -44,28 +44,40 @@ class BookingSearchService(
...
+      forceApDeliusContextApi = false,
+    )
+
+    return bookingSearchResultDtos
+      .map { result -> result to offenderSummaries.first { it.crn == result.personCrn } }
+      .filter { (_, offenderSummary) -> offenderSummary !is PersonSummaryInfoResult.Success.Restricted }
+      .map { (result, offenderSummary) ->
+        result
+        result.personName = when (offenderSummary) {
+          is PersonSummaryInfoResult.Success.Full -> "${offenderSummary.summary.name.forename} ${offenderSummary.summary.name.surname}"
```
davidatkinsuk, line 75: we _could_ use the new OffenderDetailsUtil#getNameFromPersonSummaryInfoResult here which would be cleaner, but It would slightly change the behaviour:

* we'd include middle names in the result (presumably this is a good thing)
* for unknown we'd return "Unknown" (instead of null)
* for not found we'd return "Not Found" (instead of null)

I tried to look in the corresponding UI code and couldn't figure out if nulls in the later two values are relied upon to control logic in the UI (e.g. hide a row?).

Any thoughts from a reviewer?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/CalendarService.kt
```diff
@@ -19,16 +19,18 @@ class CalendarService(
...
-            is AuthorisableActionResult.Success -> "${offenderResult.entity.firstName} ${offenderResult.entity.surname}"
-            is AuthorisableActionResult.Unauthorised -> "LAO Offender"
-          }
+    calendarInfo.values.forEach { bedList ->
+      bedList.filterIsInstance<CalendarBookingInfo>().forEach { bed ->
+        val offenderSummary = offenderSummaries.first { it.crn == bed.crn }
+        bed.personName = when (offenderSummary) {
+          is PersonSummaryInfoResult.Success.Full -> "${offenderSummary.summary.name.forename} ${offenderSummary.summary.name.surname}"
+          is PersonSummaryInfoResult.Success.Restricted -> "LAO Offender"
+          else -> "Unknown"
```
davidatkinsuk, line 33: we _could_ use the new OffenderDetailsUtil#getNameFromPersonSummaryInfoResult here which would be cleaner, but It would slightly change the behaviour:

- we'd include middle names in the result (presumably this is a good thing)
- for not found we'd return "Not Found" (instead of 'Unknown')

Any thoughts from a reviewer?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/OffenderService.kt
```diff
@@ -81,6 +83,41 @@ class OffenderService(
     )
   }
 
+  fun getOffenderSummariesByCrns(
+    crns: List<String>,
```
davidatkinsuk, line 87: Is it worth using a Set instead to filter out potential duplicates?

pezholio: Good plan! 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/CalendarService.kt
pezholio: I think we should probably update `getNameFromPersonSummaryInfoResult` to return `Unknown`, just for consistency. I agree on middle names too 👍

davidatkinsuk: Done

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/OffenderService.kt
davidatkinsuk: Done

PR #1269, opened 2023-12-19
Don’t show reallocated placement applications

pezholio:
When viewing placement applications for an application, we should only show ones that haven’t been reallocated. I’ve updated the query, as well as updated the test to confirm that reallocated placement applications don’t show.


davidatkinsuk (approved):
LGTM

PR #1270, opened 2023-12-19
Show type as referral when no arrival date

pezholio:
Previously, we were returning `referral` for a type when a placement application is not present. There are some circumstances when an application can be made, but have subsequent placement applications made, so we need to change the rule to check for the presence of a start date.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApplicationEntityReportRow.kt
```diff
@@ -57,8 +57,8 @@ interface ApplicationEntityReportRowRepository : JpaRepository<ApplicationEntity
       cast(placement_applications.data -> 'request-a-placement' -> 'decision-to-release' ->> 'decisionToReleaseDate' as date) as paroleDecisionDate,
       (
         CASE
-          WHEN placement_applications.id IS NULL THEN 'referral'
-          ELSE 'placement request'
+          WHEN apa.arrival_date IS NULL THEN 'placement request'
```
davidatkinsuk, line 60: presumably this field isn't populated when the application is only used for placement requests? should we consider introducing a better marker for this on an application in a subsequent change and use this as a fall back?

davidatkinsuk (approved):
I think we're good to merge but should consider my comment for a future change?

pezholio:
Gonna park this as I think I got the wrong end of the stick

PR #1274, opened 2023-12-20
Fix test env var for async save enabled

davidatkinsuk:
This commit fixes the names of the env vars in test, matching the changes made on 317760c25d72008b9a300f31dca47a5cc4553858 for other envs


pezholio (approved):
Nice catch!

PR #1281, opened 2023-12-21
(APS 153) Add sorting to the reallocatable tasks endpoint

pezholio:
This adds sorting and native SQL pagination to the `/tasks/reallocatable` endpoint. This is a bit tricky as we have to fetch three different entity types. Before we've been a bit manual with our approach, but this makes pagination and sorting tricky because we have to load all the entities into memory to do the sort. This will only get harder once the service gets opened to more regions, and the Tasks page is already incredibly slow.

I've got around this by having a `UNION` query that fetches the bare minimum of fields from all the taskable tables, then fetching the entities by their ID in three separate queries. We then have to re-sort the fields after fetching to ensure the sort order is preserved.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/PlacementRequestService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/PlacementRequestService.kt
@@ -15,6 +15,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.PlacementDates
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.PlacementRequestSortField
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.PlacementRequestStatus
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.SortDirection
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.TaskSortField
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.ClientResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.CommunityApiClient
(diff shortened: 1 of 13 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/PlacementRequestService.kt
```diff
@@ -75,11 +76,18 @@ class PlacementRequestService(
...
     sortDirection: SortDirection?,
     allocatedFilter: AllocatedFilter?,
   ): Pair<List<PlacementRequestEntity>, PaginationMetadata?> {
-    val sortField = "created_at"
-    val pageable = getPageable(sortField, sortDirection, page)
+    val pageable = getPageable(
+      // Convert to snake_case, because the findAllReallocatable* methods are native SQL queries
+      when (sortField) {
+        TaskSortField.createdAt -> "created_at"
+      },
```
danlivings-dxw, line 87: Optional, but this could be replaced with `sortField.value` if the `TaskSortField` enum is defined in the spec to use snake case for its acceptable values.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/TaskService.kt
```diff
@@ -23,7 +35,55 @@ class TaskService(
...
+    allocatedFilter: AllocatedFilter?,
+    page: Int?,
+    sortField: TaskSortField,
+    sortDirection: SortDirection,
+  ): Pair<List<TypedTask>, PaginationMetadata?> {
+    val pageable = getPageable(
+      // Convert to snake_case, because getAllReallocatable is a native SQL query
+      when (sortField) {
+        TaskSortField.createdAt -> "created_at"
+      },
```
danlivings-dxw, line 53: Optional, but this could be replaced with `sortField.value` if the `TaskSortField` enum is defined in the spec to use snake case for its acceptable values.

danlivings-dxw (approved):
Looks good to me 👍

PR #1285, opened 2024-01-03
(Fix) Update tasks endpoint

pezholio:
`TaskWrapper` now returns a `UserWithWorkload` array, but the spec wasn’t updated. For some reason we weren’t getting compilation errors!


richpjames (approved):
I wonder if it was compiling because UserWithWorkload meets and exceeds the contract

PR #1287, opened 2024-01-03
Amended the preUpdate, appended to the enum, verified with tests

garethhallberg:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/listeners/AssessmentListener.kt
```diff
@@ -13,20 +13,34 @@ class AssessmentListener {
...
     } else if (assessment.decision == AssessmentDecision.ACCEPTED) {
-      (assessment.application as ApprovedPremisesApplicationEntity).status = ApprovedPremisesApplicationStatus.AWAITING_PLACEMENT
+      val application = (assessment.application as ApprovedPremisesApplicationEntity)
+      application.status = applicationGetStatusFromArrivalDate(application)
     } else if (assessment.decision == AssessmentDecision.REJECTED) {
       (assessment.application as ApprovedPremisesApplicationEntity).status = ApprovedPremisesApplicationStatus.REJECTED
     }
   }
+
+  fun applicationGetStatusFromArrivalDate(
```
pezholio, line 37: As we're only using the arrival date here, we could just pass the arrival date in

pezholio (approved):
LGTM - just one minor thing

pezholio (changes requested):
Whoops - I've just remembered we need to translate the `ApprovedPremisesApplicationStatus` to the `ApiApprovedPremisesApplicationStatus`

pezholio:
We need to update `ApprovedPremisesApplicationStatus` in the API spec to add a pending placement request entry:

```
    ApprovedPremisesApplicationStatus:
      type: string
      enum:
        - started
        - submitted
        - rejected
        - awaitingAssesment
        - unallocatedAssesment
        - assesmentInProgress
        - awaitingPlacement
        - placementAllocated
        - inapplicable
        - withdrawn
        - requestedFurtherInformation
        - pendingPlacementRequest
```

And then update the transformer lookup in https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/36b4239096d7a17dfdabb4ad231fd77a50d97a31/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/ApplicationsTransformer.kt#L39

pezholio (approved):
LGTM 👍

PR #1293, opened 2024-01-04
Hide withdrawn placement requests for `notMatched`

pezholio:
When looking at not matched placement requests, it's showing withdrawn placement requests too, which is not what we want at all. When `notMatched` is provided we need to filter out placement requests that are withdrawn.


danlivings-dxw (approved):
Looks good to me 👍

PR #1296, opened 2024-01-05
Fix/user role seed duplication

tahb:
[We noticed today that seeded dev roles had many duplicate roles](https://mojdt.slack.com/archives/C03K0HB0LBE/p1704448977587099):

image (2)

This looks to be due to role inserts not conflicting properly on insert and so were reinserting themselves on every server start.

We had this protection in create_dev_users:

```
ON CONFLICT (id)
DO
  NOTHING;
```

I don't think this was doing anything because in this context the `id` was a randomly generated uuid and so would never conflict.

This dev seeding issue could be hacked together by removing each set of roles before insertion each time however I think this actually highlights a more general problem: the database can allow users to have duplicate roles. I suspect this is never a situation we expect to get into and so I'm proposing we add a uniqueness constraint, which will then raise a conflict correctly and prevent future duplicates.

I have some recordings on Slack of me testing these that I'll put in [the original thread](https://mojdt.slack.com/archives/C03K0HB0LBE/p1704448977587099) since they're too big to upload to GitHub.

### DIFF
--- a/src/main/resources/db/migration/all/20240105113951__user_role_assignments_unique_by_role_and_user_id.sql
+++ b/src/main/resources/db/migration/all/20240105113951__user_role_assignments_unique_by_role_and_user_id.sql
@@ -1,0 +1,3 @@
+ALTER TABLE user_role_assignments
+ADD CONSTRAINT unique_role_user_constraint
+UNIQUE (role, user_id);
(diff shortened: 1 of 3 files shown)


src/main/resources/db/migration/all/20240105113951__user_role_assignments_unique_by_role_and_user_id.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE user_role_assignments
```
tahb, line 1: This whole approach relies on this constraint being reasonable and succeeding on dev.

Whilst I remove the seeded dev users, if users can really have multiple of the same roles this will not deploy and will need reverting.

PR #1301, opened 2024-01-08
Feature/aps 39 display offender status at time of submission

davidatkinsuk:
This PR adds a new column to Approved Premises Applications to capture the offender's 'in out' status at the point the application was submitted. This is subsequently exposed to the UI via the API. A migration job is also provided to backfill this column for already submitted tasks

A few things to note:

* I've refactored the API to move the FullPerson.status into a standalone named type. This didn't break the CAS1 UI because the model returned is unchanged, please let me know if you think it's something worth checking with CAS2 (i guess the CAS3 E2E tests will capture any issues there)
* I've tested the backfill job as best i can locally, but test data is limited. Discussed with Stu and agreed we can test this in pre-prod
* The backfill logic is based upon a conversation with the Prison API team. Outcomes of that conversation are detailed on https://dsdmoj.atlassian.net/browse/APS-39
* A UI PR will follow once we've got the backfill ran successfully


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/migration/InmateStatusOnSubmissionMigrationJobTest.kt
```diff
@@ -0,0 +1,310 @@
...
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.prisonsapi.SignificantMovements
+import java.time.LocalDateTime
+import java.time.OffsetDateTime
+import java.time.ZoneOffset
+
+class InmateStatusOnSubmissionMigrationJobTest {
+
+  @Test
+  fun `Determine In Out Status returns OUT if no Prison Periods provided`() {
+    val result = determineInOutStatus(
```
pezholio, line 18: Big fan of unit testing all this stuff here, rather than testing all the possible permutations in the integration tests

pezholio (approved):
LGTM 👍 - big fan of the comprehensive unit tests too

PR #1302, opened 2024-01-08
(Chore) Get all task counts

pezholio:
Now we're surfacing workload data, it occurred to me that we're only surfacing assessments, not the other types of task. This PR refactors the workload fetching service method to fetch all workloads in one query and fetches all other task types, summing them at the end.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/UserEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/UserEntity.kt
@@ -154,7 +154,71 @@ interface UserRepository : JpaRepository<UserEntity, UUID>, JpaSpecificationExec
           a.allocated_to_user_id = u.id
           and a.reallocated_at is null
           and a.submitted_at > current_date - interval '30' day
-      ) as completedAssessmentsInTheLastThirtyDays
+      ) as completedAssessmentsInTheLastThirtyDays,
+      (
+        SELECT
+          count(*)
+        from
+          placement_applications placement_application
+        where
+          placement_application.allocated_to_user_id = u.id
+          and placement_application.reallocated_at is null
+          and placement_application.submitted_at is null
+      ) as pendingPlacementApplications,
+      (
+        SELECT
+          count(*)
+        from
+          placement_applications placement_application
+        where
+          placement_application.allocated_to_user_id = u.id
+          and placement_application.reallocated_at is null
+          and placement_application.submitted_at > current_date - interval '7' day
+      ) as completedPlacementApplicationsInTheLastSevenDays,
+      (
+        SELECT
+          count(*)
+        from
+          placement_applications placement_application
+        where
+          placement_application.allocated_to_user_id = u.id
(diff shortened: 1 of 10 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/UserEntity.kt
```diff
@@ -154,7 +154,70 @@ interface UserRepository : JpaRepository<UserEntity, UUID>, JpaSpecificationExec
...
+      (
+        SELECT
+          count(*)
+        from
+          placement_requests placement_request
+          left join bookings booking on booking.id = placement_request.booking_id
+        where
+          placement_request.allocated_to_user_id = u.id
+          and placement_request.reallocated_at is null
+          and booking.created_at > current_date - interval '30' day
```
davidatkinsuk, line 219: the 7 day version of this query had 'placement_request.booking_id is not null', but this one doesn't. Is that deliberate?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/TasksTest.kt
```diff
@@ -1467,44 +1468,61 @@ class TasksTest : IntegrationTestBase() {
...
+                repeat(numPlacementRequestsCompleted7Days) {
+                  val days = kotlin.random.Random.nextInt(1, 7).toLong()
+                  createTask(TaskType.placementRequest, OffsetDateTime.now().minusDays(days), allocatableUser, user, offenderDetails.otherIds.crn)
                 }
 
-                val numAssesmentsCompleted30Days = 9
-                repeat(numAssesmentsCompleted30Days) {
+                val numAssessmentsCompleted30Days = 4
+                repeat(numAssessmentsCompleted30Days) {
                   val days = kotlin.random.Random.nextInt(8, 30).toLong()
```
davidatkinsuk, line 1506: i think this should be nextInt(1,30) as we still want to include those in the first 7 days? (this is repeated a few times in this test)

davidatkinsuk (changes requested):
Just a few comments from me 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/UserEntity.kt
pezholio: Good catch!

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/TasksTest.kt
pezholio: If we had that, then we'd potentially have a flickering test scenario, because if they'd been completed between 1 and 7 days ago, they'd show up as being completed in the last 7 days. I'll update some naming to make that clearer, as `numAssessmentsCompleted30Days` suggests they cover ALL that 30 day period

davidatkinsuk (approved):
LGTM 👍

PR #1305, opened 2024-01-08
(APS-174) Fix timeline transformer

pezholio:
Before, changed bookings were being translated to `approvedPremisesBookingCancelled`. This isn’t what we wanted, so this PR changes the transformer to return the correct type. To improve the coverage, I’ve also added a test to test the transformations from domain event to timeline event are correct.


davidatkinsuk (approved):
LGTM, ParameterizedTest FTW 👍

PR #1306, opened 2024-01-09
(APS-167) Add pagination to assessment listing

pezholio:
This adds support for pagination on the `/assessments` endpoint. This is currently only done by splitting the resulting list and returning the resulting page. We should do this at the database level, but we're doing it this way so we can get on with the frontend implementation. I've also refactored a bunch of tests to DRY them up a bit.

I've deferred from the usual implementation and added a `perPage` argument (which we'll only use in tests) so we don't have to build a ton of objects when doing the integration tests.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/AssessmentController.kt
```diff
@@ -75,17 +80,30 @@ class AssessmentController(
...
+      .filterByStatuses(statuses)
+
+    // TODO: We should carry out the pagination at the database level
+    if (page !== null) {
+      val pageSize = perPage ?: defaultPageSize
+      val chunkedSummaries = filteredSummaries.chunked(pageSize)
+      val pageIndex = page - 1
+
+      metadata = PaginationMetadata(page, chunkedSummaries.size, filteredSummaries.size.toLong(), pageSize)
+      filteredSummaries = filteredSummaries.chunked(pageSize)[pageIndex]
```
davidatkinsuk, line 101: can do chunkedSummaries[pageIndex] instead here?

davidatkinsuk (approved):
one minor comment but as this is a temporary solution, upto you if you want to fix

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/AssessmentController.kt
pezholio: Oh good catch 👍

PR #1311, opened 2024-01-09
Amended the entity query, test to show it works, comments explaining …

garethhallberg:
…the Hibernate bug


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
```diff
@@ -1661,6 +1702,7 @@ class BookingService(
     thisEntityId: UUID?,
     bedId: UUID,
   ): BookingEntity? {
+    log.info("PleaseStop")
```
pezholio, line 1705: Think this was left in by mistake 😄

pezholio (commented):
LGTM - just some debug stuff that's slipped in

pezholio (approved):
LGTM 👍

PR #1314, opened 2024-01-10
Feature/aps 120 allow bookings without bed ids

davidatkinsuk:
To achieve this the POST /premises/{premisesId}/bookings endpoint was updated to make bed id optional.

Because the temporary accommodation booking function still requires a Bed ID, it has been updated to explicitly check this itself as part of validation.

This pushed the cyclomatic complexity above the limit for the adhoc booking function. A partially successful effort was made to reduce this below the limit, but I think it would need a non-trivial amount of work so i've left the warning as suppressed it for now


pezholio (approved):
LGTM - thanks for the quick turnaround!

PR #1315, opened 2024-01-10
(765) Stream new CAS2 'application status updates' report

edavey:
[Trello 765](https://trello.com/c/GV3HWCzy/765-download-application-status-updates-report)

This management information report lists application status updates, based on the CAS2 `application.status-updated` domain event.

The spreadsheet which is streamed in response to a request to:

`/cas2/reports/application-status-updates`

has the following format:

```
eventId: "364145f9-0af8-488e-9901-b4c46cd9ba37"
applicationId "3fa85f64-5717-4562-b3fc-2c963f66afa6"
personCrn: "X320741"
personNoms: "A1234AI"
updatedBy: "CAS2_ASSESSOR_USER"
updatedAt: "2024-01-09T09:17:55"
newStatus: "moreInfoRequested"
```

NOTES:

- the `updatedAt` timestamp is formatted in the ISO8601 "Calendar date and local time (extended)" style, as in the above example

- we add a detekt exception to allow a "LongParameterList" in the new `ApplicationStatusUpdatesReportRow` data class


tahb (approved):
This looks good to merge 👍 

My main worry is how well this will do at scale when there are hundreds/thousands of domain events and we're digging through json for each one. The query or the connection will presumably time out at some point and it'd be good to understand when this is likely to be. CAS3 are for instance considering pre-generating these over night so they don't need to be calculated on demand.

edavey:
> My main worry is how well this will do at scale when there are hundreds/thousands of domain events and we're digging through json for each one. The query or the connection will presumably time out at some point and it'd be good to understand when this is likely to be. CAS3 are for instance considering pre-generating these over night so they don't need to be calculated on demand.

I think this PR https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1331 should allow us to sleep easy regarding performance. The volume of CAS2 activity is low and these reports are now constrained by `domain_events.occurred_at` or `cas_2_applications.created_at` -- i.e. without digging into the JSON blobs.

PR #1316, opened 2024-01-10
Filter out inactive users when showing a task

pezholio:
When showing users, we're allowing inactive users to be allocated, which is causing issues. This updates the endpoint so we ignore inactive users.


davidatkinsuk (approved):
LGTM 👍

garethhallberg (approved):

danlivings-dxw (approved):
LGTM 👍

PR #1317, opened 2024-01-10
[APS-67] Implement AP Delius Context API offender risks data source

danlivings-dxw:

This PR provides the ability to use the AP Delius Context API as a data source for offender risks. This is achieved by implementing the `ApDeliusContextApiOffenderRisksDataSource` class.

This PR does not change the data source for offender risks as this is down to the individual configuration for each environment.

**NB**: The behaviour for retrieving the MAPPA details has changed - the current behaviour in the Community API implementation is derived from be95ffa, which ensures a registration exists that corresponds to the MAPPA details and returns an error if not. However, the structure of registrations from the AP Delius Context API is different, and does not include the necessary information to perform this check, and so it has been omitted.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/datasource/OffenderRisksDataSource.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/datasource/OffenderRisksDataSource.kt
@@ -5,6 +5,7 @@ import org.springframework.beans.factory.annotation.Value
 import org.springframework.context.annotation.Primary
 import org.springframework.http.HttpStatus
 import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.ApDeliusContextApiClient
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.ApOASysContextApiClient
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.ClientResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.CommunityApiClient
@@ -17,6 +18,8 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.RiskWithStatus
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.RoshRisks
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.community.Registration
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.community.Registrations
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.deliuscontext.CaseDetail
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.deliuscontext.MappaDetail
 
 interface OffenderRisksDataSource {
   val name: OffenderRisksDataSourceName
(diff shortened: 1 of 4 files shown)


danlivings-dxw:
@tahb I'd like to get your input on whether the changes to retrieving MAPPA details will cause problems for CAS3 as I believe the change to implement the current behaviour was made as part of CAS3

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/datasource/OffenderRisksDataSource.kt
```diff
@@ -195,11 +197,104 @@ class CommunityApiOffenderRisksDataSource(
...
+
+        when (summary.anyRisksAreNull()) {
+          true -> RiskWithStatus(status = RiskStatus.NotFound)
+          false -> RiskWithStatus(
+            value = RoshRisks(
+              overallRisk = summary.determineOverallRiskLevel().text,
+              riskToChildren = summary.riskChildrenCommunity!!.text,
+              riskToPublic = summary.riskPublicCommunity!!.text,
+              riskToKnownAdult = summary.riskKnownAdultCommunity!!.text,
+              riskToStaff = summary.riskStaffCommunity!!.text,
```
pezholio, line 235: These bangs make me nervous - can we guarantee these fields will be available?

danlivings-dxw: The `summary.anyRisksAreNull()` method would return `true` if any of these fields were null, so this branch would never be executed.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/datasource/OffenderRisksDataSource.kt
```diff
@@ -195,11 +197,104 @@ class CommunityApiOffenderRisksDataSource(
...
+      HttpStatus.NOT_FOUND -> RiskWithStatus(status = RiskStatus.NotFound)
+      else -> RiskWithStatus(status = RiskStatus.Error)
+    }
+    is ClientResult.Failure -> RiskWithStatus(status = RiskStatus.Error)
+  }
+
+  fun CaseDetail.toMappa(): RiskWithStatus<Mappa> = this.mappaDetail?.let {
+    RiskWithStatus(
+      value = Mappa(
+        "CAT ${it.categoryDescription}/LEVEL ${it.levelDescription}",
```
tahb, line 277: 👋 The reason CAS3 [recently changed this to add a guard](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1215/files#diff-4a8c3767d589a6283081bfd38ef72be5ebc2dd5ec9d401127ded27adb5ec015cR567) was because [we found bad data that blocked a new referral from being created](https://trello.com/c/mSSzt3eH). When we tried to load in the risk info into a brand new application, we got a 500 when unpacking the data because the required `category` was missing.

As we now are we aren't using `!!` I guess we wouldn't 500 but the value would be stripped back to a pretty useless "CAT /LEVEL"? In those cases I think it's best we throw an error instead and allow the frontend to display the "We couldn't find data for this right now" etc in the risk widget.

Though the names of these values have changed I should think we ought to continue guarding against this and other values being nil?

tahb: It could well be worth creating a CAS1 and CAS3 referral locally with this change and use the CRNs that have missing data to test how this works.

danlivings-dxw: I couldn't do a local manual test with the list of CRNs that you sent me, but I've implemented a check to make sure that `categoryDescription` and `levelDescription` are both not null. If either are, it'll return the `NotFound` variant of the `RiskWithStatus`.

pezholio (approved):
LGTM - thanks for persisting with this!

tahb (approved):
LGTM 👍 

Before we put this into prod I think we should make sure to do testing on dev/test with [the test crns](https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/4467949648/CRNs+for+Testing+in+Development+Environment). Making sure that when the crn has missing combinations of risk data the user sees the expected data on their referral. You've mentioned that this isn't going to be used until a configuration change is made so it sounds like we're going to need to remember to facilitate that separately from this.

PR #1320, opened 2024-01-11
Seed CAS2 domain events

edavey:
This PR extends the existing "auto-script" seeding of CAS2 applications to include CAS2 domain events:

- `application.submitted`
- `application.status-updated`

This will make it easier to test management information reporting which relies on these domain events.


Harriethw (approved):
Thanks for putting this together so clearly 👍

PR #1321, opened 2024-01-11
Adding tests for all possible assessment statuses

davidatkinsuk:
As part of the work to update the assessment endpoint to use ‘in database’ paging, we need to move the logic that filters on status into the database queries.

As a pre-requisite to this work this commit ensures there is an integration test for each possible CAS1/CAS3 status, and tests to ensure multiple statuses can be used for filtering

This commit also adds a script to ensure docker is running locally before starting starting the test database

### DIFF
--- a/script/test_database
+++ b/script/test_database
@@ -6,6 +6,8 @@ set -e
 
 cd "$(dirname "$0")/.."
 
+script/utils/launch-docker.sh
+
 echo "==> Starting integration test database..."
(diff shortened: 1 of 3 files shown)


tahb:
@sudhakar-plr I know we were working on making CAS3 application statuses explicit with the use of a status field rather than being inferred so this may be useful for you to be involved with 👍

pezholio (approved):
LGTM 👍

PR #1325, opened 2024-01-12
Restrict CAS2 referrer users to ROLE_POM

Harriethw:
Previously on CAS2 we were granting access to all users with ROLE_PRISON, a role
granted automatically to all nomis authorised users. This is too broad a
restriction, because it means any users logging in via nomis will be
able to search for prisoners and make applications.

As our referrer users for private beta will be POMs, we should restrict
to the POM role for now. This can be assigned to devs to test in
different environments. 

There is some question around whether our users 
in production will definitely have this role, but we've been told that this should
be the role applied to all POMs. We can mitigate this risk by testing with a trusted
user shortly before going live.

Contract managers will be given the role CAS2_ADMIN so they can see
submitted applications, but they shouldn't need ability to create
applications.


edavey (approved):
This change from ROLE_PRISON to ROLE_POM looks correct technically.

PR #1326, opened 2024-01-12
Move CAS1 Assessments filtering into SQL

davidatkinsuk:
This change is a pre-requisite to enable paging/sorting in SQL instead of programmatically in the controller code

Not only is the status filtered in SQL, but the SQL now also returns a Status String, meaning we no longer have to determine it programatically when listing assessments.

Unfortunately, we still determine the status programmatically when retrieving an individual assessment (in AssessmentTransformer), so we have a duplication of this logic in both Java and SQL. The presumed way to address this issue is to explicitly store the status in the assessment table itself, which then allow us simplify the SQL added by this commit.

Note that we no longer return dateOfInfoRequest and isStarted from the assessment search queries, because this was only ever used to determine the CAS1 status programatically for search results


pezholio (approved):
LGTM 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/AssessmentController.kt
```diff
@@ -64,10 +62,15 @@ class AssessmentController(
...
+    val manuallyFilter = xServiceName == ServiceName.temporaryAccommodation
 
     val summaries = when {
       xServiceName == ServiceName.temporaryAccommodation && crn != null -> assessmentService.getAssessmentSummariesByCrnForUser(user, crn, xServiceName)
-      else -> assessmentService.getVisibleAssessmentSummariesForUser(user, xServiceName)
+      xServiceName == ServiceName.temporaryAccommodation -> assessmentService.getVisibleAssessmentSummariesForUserCAS3(user)
+      else -> {
+        val domainSummaryStatuses = statuses?.map { assessmentTransformer.transformApiStatusToDomainSummaryState(it) } ?: emptyList()
+        assessmentService.getVisibleAssessmentSummariesForUserCAS1(user, domainSummaryStatuses)
+      }
```
danlivings-dxw, line 73: Having this change being to the `else` branch might change the behaviour for CAS2, so we should make sure that this won't cause them problems.

@edavey, @Harriethw: does CAS2 use this assessment logic?

davidatkinsuk: I think the previous behaviour here was to return an emptyList() for CAS2 (this was returned by the assessmentSrvice function i've changed), now it will return approved premises assessments, which was not an intentional change on my part, so good spot! i'll add an explicit CAS2 branch in here to retain existing behaviour

Harriethw: We don't have assessments in CAS2 so we aren't ever calling this atm

davidatkinsuk: Thanks @Harriethw, in that case i'll update the code to throw an unsupported operation exception for CAS2 to make it clear going forward 👍

PR #1328, opened 2024-01-12
Add telephone number as "first class" field to CAS2 Applications

Harriethw:
We would like to save the referrer's telephone number as a "first class"
field when a CAS2 Application is submitted, and surface it when a submitted application is viewed. 

This will allow us to surface it on submitted applications to Assessors, without having to dig
into the document JSON blob.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas2/Cas2AutoScript.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas2/Cas2AutoScript.kt
@@ -98,6 +98,7 @@ class Cas2AutoScript(
         preferredAreas = "Luton | Hertford"
         hdcEligibilityDate = LocalDate.parse("2024-02-28")
         conditionalReleaseDate = LocalDate.parse("2024-02-22")
+        telephoneNumber = "0800 123 456"
       },
     )
   }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas2ApplicationEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas2ApplicationEntity.kt
@@ -103,6 +103,7 @@ data class Cas2ApplicationEntity(
   var preferredAreas: String? = null,
   var hdcEligibilityDate: LocalDate? = null,
   var conditionalReleaseDate: LocalDate? = null,
+  var telephoneNumber: String? = null,
 ) {
   override fun toString() = "Cas2ApplicationEntity: $id"
 }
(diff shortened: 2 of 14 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas2/Cas2AutoScript.kt
```diff
@@ -71,6 +71,7 @@ class Cas2AutoScript(
         submittedAt = submittedAt,
         schemaVersion = jsonSchemaService.getNewestSchema(Cas2ApplicationJsonSchemaEntity::class.java),
         schemaUpToDate = true,
+        telephoneNumber = "123 456 789",
```
edavey, line 74: hmm, I see that this is necessary and works fine. 

But I wonder if we should make an effort to keep the seeded data realistic: as is, this change would add the `telephoneNumber` property to unsubmitted applications as well as to submitted ones. 

If you see https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/9aa9062b07ee4f76b780f0e53891132dd0a16191/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas2/Cas2AutoScript.kt#L81 from my open PR https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1320, we could introduce something like `setFirstClassProperties()` to just affect the applications which are in either the `SUBMITTED` or `IN_REVIEW` states?

edavey (approved):
This v clear, thks.

A couple of notes:

- I think we need to merge the UI part of the work (submit `telephoneNumber` on submission) first as otherwise the POST to `submissions` will fail due to the lack of the new 'required' property?
- I've suggested we keep the Auto-Script seeding more realistic. (I think we should build on https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1320 and then also add the other first-class properties which are promoted from `application.data` on submission...)

Harriethw:
> I think we need to merge the UI part of the work (submit telephoneNumber on submission) first as otherwise the POST to submissions will fail due to the lack of the new 'required' property?

@edavey that's a really good point! I'll move this back to draft while I get the UI PR read

>I've suggested we keep the Auto-Script seeding more realistic. (I think we should build on https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1320 and then also add the other first-class properties which are promoted from application.data on submission...)

This sounds like a better solution, thanks, will take a look

edavey:
@Harriethw I've added a `applyFirstClassProperties()` function to the auto-script process in https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1340/commits/f3918a8459fa31269678239e909eb3f53ae2802b Hopefully I can get this merged ahead of this PR, and you could add `telephoneNumber` to the seeding here?

Harriethw:
@edavey please could you take another look at this? I've added the field in the script here https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1328/commits/b512ce2312cce0d730dac65200a51ba0151799e8

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas2/Cas2AutoScript.kt
```diff
@@ -98,6 +98,7 @@ class Cas2AutoScript(
         preferredAreas = "Luton | Hertford"
         hdcEligibilityDate = LocalDate.parse("2024-02-28")
         conditionalReleaseDate = LocalDate.parse("2024-02-22")
+        telephoneNumber = "0800 123 456"
```
edavey, line 101: nice! thks for using a realistic number!

edavey (approved):
Really clear, thks.

PR #1329, opened 2024-01-12
Use `equals` to match `HttpStatus` in OffenderService

pezholio:
In #1324 we introduced `equals` when matching `HttpStatus`. However, we missed this in the `OffenderService` class, which is the one we're currently experiencing bugs. I had changed this to match on the value, which should work, but we're still experiencing problems. I've not got much confidence this will fix the issue, but this will at least give us consistency.


davidatkinsuk (approved):
LGTM

PR #1330, opened 2024-01-15
Remove ResultSetMapping for Assessments

davidatkinsuk:
Before this change @SqlResultSetMapping was used to map results from several native queries in the AssessmentRepository into instances of DomainAssessmentSummary, via a @NamedNativeQuery definition

Whilst use of ResultSetMapping helps overcome issues mapping some SQL types into their equivalent kotlin-native types (1), the convention in all other repositories is to use Spring Data’s interface-based projection and work around the aforementioned mapping issues by mapping UUIDs to text in the SQL, or mapping timestamptz to Timestamp instead of OffsetDateTime.

Furthermore, Spring Data’s automated paging functionality does not fully work when using @NamedNativeQuery. On startup you see warning logs as follows:

“Finder [insert method name] is backed by a NamedQuery but contains a Pageable parameter! Sorting delivered via this Pageable will not be applied!”

Therefore, to enable database-based paging and improve repository consistency, all usages of @SqlResultSetMapping and @NamedNativeQuery in the Assessment Repository have been changed to instead use interface-based projection on a self-defined @Query annotation

1. See https://github.com/spring-projects/spring-data-jpa/issues/1796

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/util/TimeExtensions.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/util/TimeExtensions.kt
@@ -1,0 +1,20 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.util
+
+import java.sql.Timestamp
+import java.time.OffsetDateTime
+
+const val NANO_MAX = 1E6
+
+fun OffsetDateTime?.toTimestampOrNull(): Timestamp? {
+  return this?.toTimestamp()
+}
+
+fun OffsetDateTime.toTimestamp(): Timestamp {
+  return Timestamp.from(this.toInstant())
+}
+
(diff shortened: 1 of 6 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/util/TimeExtensions.kt
```diff
@@ -0,0 +1,20 @@
...
+
+fun OffsetDateTime.toTimestamp(): Timestamp {
+  return Timestamp.from(this.toInstant())
+}
+
+fun OffsetDateTime.roundNanosToMillisToAccountForLossOfPrecisionInPostgres(): OffsetDateTime {
+  val millis = Math.round(this.nano / NANO_MAX)
+  val roundedNanos = (millis * NANO_MAX).toInt()
+  return this.withNano(roundedNanos)
+}
```
pezholio, line 20: These are super useful!

pezholio (approved):
LGTM - much more consistent too!

PR #1334, opened 2024-01-16
Feature/aps 179 db paging sorting for assessments

davidatkinsuk:
This is the final PR for APS-179

This PR moves sorting and paging into the SQL for CAS1 only, increasing the number of tests for sorting on specific columns.

As part of this change a default sort field is now assumed if one is not provided by the API call. This simplifies handling of the paging criteria and ensures we have consistent results from paging.

Unused methods have been removed from paging utils, and the default page size is now taken from config instead of hard-coded


pezholio (approved):
Looking really good, thanks for this - some proper smart quality of life improvements too 🙏

PR #1335, opened 2024-01-16
Emit CAS2 domain events in prod

tahb:
We don’t believe these are being consumed yet and that this work is in progress after specs have been exchanged[1].

We leave the test environment as having these domain events off as I think the dev and test environments probably send events to the same queue and may cause odd behaviour (such as duplicate contacts) in downstream systems like nDelius?

[1] https://mojdt.slack.com/archives/C05GEPLBRN3/p1705400557152239


edavey (approved):
👍

PR #1338, opened 2024-01-16
Persist booking domain events when emit disabled

davidatkinsuk:
The ‘arrived-departed-domain-events-disabled’ configuration is used to determine if domain events should be emitted for arrival, non-arrivel and departure booking events.

Even if we don’t want to emit domain events to SNS, it’s still useful to persist them to the domain_events table as this information can be used for reporting purposes.

This change ensures that when arrived-departed-domain-events-disabled’ is true, we still persist the events


pezholio (approved):
LGTM 👍

PR #1340, opened 2024-01-17
778 CAS2: improve seeding to include first class fields on submitted applications

edavey:
These properties are set on submission of the application in
`ApplicationService.submitApplication()`:

"promoted" up out of the application.data JSON blob:

- `preferredAreas`
- `hdcEligibilityDate`
- `conditionalReleaseDate`

and also `referringPrisonCode` which is looked up from the
Prison API.

Another property (`telephoneNumber`) will be added soon.

These first-class properties are used in the 'metadata'
which is shown in the "application summary" which
accompanies a submitted application. They are also included
in the submitted.application domain event and through this,
the "Submitted applications" management information report.


Harriethw (approved):
fab!

PR #1344, opened 2024-01-17
(APS-208) Ignore cancelled lost beds in lost beds report

pezholio:
In addition to this, I’ve had to change the `LostBedReportRow#id` type to a String to allow the integration tests to pass. The won’t make any difference as the typing information is discarded when converting to Excel.


davidatkinsuk (approved):
LGTM 👍

PR #1346, opened 2024-01-17
CircleCI runs CAS2 e2e tests

tahb:
We should circle back and make these mandatory for the release pipeline once CAS2 get this green.


edavey (approved):
👍 thanks for picking this up

PR #1348, opened 2024-01-17
[APS-57] Introduce rule-based user allocator

danlivings-dxw:
> See [APS-57 on the Approved Premises Jira board](https://dsdmoj.atlassian.net/browse/APS-57).

This PR introduces a new approach to allocating users to assessments, placement requests, and placement applications. In the future, most allocations will be performed manually, with only tasks meeting specific conditions being assigned automatically to a user. To achieve this, the `UserAllocator` class has been implemented, which processes a configured collection of rules that implement the `UserAllocatorRule` interface in order to determine to which (if any) user the task should be allocated.

The `UserAllocatorRule` interface provides the `evaluateAssessment`, `evaluatePlacementRequest`, and `evaluatePlacementApplication` methods, which can be overridden to implement the rule logic. Rules can be ordered by defining their `priority`; a rule with a lower value will be evaluated earlier.

When a rule is evaluated, it will return a `UserAllocatorRuleOutcome` representing what action the allocator should take. Currently the available outcomes are:
- `Skip`: The default rule outcome. This indicates that the allocator should proceed to the next rule.
- `DoNotAllocate`: This indicates that the allocator should abort and not allocate a user.
- `AllocateToUser`: This indicates that the allocator should finish evaluating rules and allocate to the given user. If this allocation fails, however, a warning is logged, and the rule is skipped.

The existing user allocation methods on `UserService` have been removed and replaced with the `LegacyAllocatorRule`. This rule serves as a facade to bring the legacy allocation logic under the new system. This rule can be enabled/disabled or reprioritised using the `user-allocations.rules.legacy-allocator.enabled` and `user-allocations.rules.legacy-allocator.priority` configuration values. To allow the new user allocator to be introduced without any external change in behaviour, this rule is enabled and assigned a higher than default priority.


pezholio (approved):
LGTM - This looks really clear and well-architected. Thanks! 👍

PR #1349, opened 2024-01-17
Fix frontend CAS2 `applicationUrl` in application.status-updated domain event

edavey:
The CAS2 `StatusUpdateService` was using the `url-templates.frontend.application` (Approved Premises/CAS1) setting rather than CAS2's `url-templates.frontend.cas2.application`.

In this PR we:

- set the test config to a clear literal value (`http://cas2.frontend/applications/#id`)
- extend our integration tests to verify that the expected value has been used in the creation of both `application.submitted` and `application.status-updated` events

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/StatusUpdateService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/StatusUpdateService.kt
@@ -27,7 +27,7 @@ class StatusUpdateService(
   private val applicationRepository: Cas2ApplicationRepository,
   private val statusUpdateRepository: Cas2StatusUpdateRepository,
   private val domainEventService: DomainEventService,
-  @Value("\${url-templates.frontend.application}") private val applicationUrlTemplate: String,
+  @Value("\${url-templates.frontend.cas2.application}") private val applicationUrlTemplate: String,
 ) {
 
   fun isValidStatus(statusUpdate: Cas2ApplicationStatusUpdate): Boolean {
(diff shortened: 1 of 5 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2SubmissionTest.kt
```diff
@@ -646,6 +651,15 @@ class Cas2SubmissionTest : IntegrationTestBase() {
...
           }
+
+          // verify that generated 'application.submitted' domain event links to the CAS2 domain
+          val persistedDomainEvent = domainEventRepository.findFirstByOrderByCreatedAtDesc()
+          val domainEventFromJson = objectMapper.readValue(
+            persistedDomainEvent!!.data,
+            Cas2ApplicationSubmittedEvent::class.java,
+          )
+          Assertions.assertThat(domainEventFromJson.eventDetails.applicationUrl)
+            .matches("http://cas2.frontend/applications/.+")
```
pezholio, line 662: This makes me a tad nervous because it's linked to the configuration. It might be better to set a variable in the class to get the value of `url-templates.frontend.cas2.application` and check the value is correct.  Something like:

```
@Value("\${url-templates.frontend.cas2.application}") 
private val applicationUrlTemplate: String
```

In the body of the test class, and then:

```kotlin
val expectedUrl = applicationUrlTemplate.replace('#applicationId', application.id)
...
Assertions.assertThat(domainEventFromJson.eventDetails.applicationUrl).isEqualTo(expectedUrl)
```

I have no idea if that would work, but just a thought

pezholio (approved):
LGTM - just a minor suggestion on the tests, but may be a tad overengineered!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/StatusUpdateService.kt
```diff
@@ -95,7 +95,7 @@ class StatusUpdateService(
           eventType = EventType.applicationStatusUpdated,
           eventDetails = Cas2ApplicationStatusUpdatedEventDetails(
             applicationId = application.id,
-            applicationUrl = applicationUrlTemplate.replace("id", application.id.toString()),
+            applicationUrl = applicationUrlTemplate.replace("#id", application.id.toString()),
```
edavey, line 98: @pezholio thanks Stu, your suggestion worked well and nudged me towards verifying that the whole URL is correct -- which allowed me to pick up a further bug! see above line 98

PR #1351, opened 2024-01-18
APS-193 add urls to timeline reponses

davidatkinsuk:
To support this change we need to add an assessment_id column to domain_events, and backfill it for existing entries

### DIFF
--- a/src/main/resources/db/migration/all/20240117111637__add_assessment_id_to_domain_event.sql
+++ b/src/main/resources/db/migration/all/20240117111637__add_assessment_id_to_domain_event.sql
@@ -1,0 +1,1 @@
+ALTER TABLE domain_events ADD COLUMN assessment_id UUID NULL;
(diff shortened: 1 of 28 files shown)


src/main/resources/db/migration/all/20240117111637__add_assessment_id_to_domain_event.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE domain_events ADD COLUMN assessment_id TEXT;
```
pezholio, line 1: Should this be a UUID?

pezholio (approved):
LGTM - just one query about the assessment IDs

src/main/resources/db/migration/all/20240117111637__add_assessment_id_to_domain_event.sql
davidatkinsuk: Good spot! it absolutely should be. I've updated it to UUID and also explicitly stated its nullable (which is default anyway). Code was already using UUID thankfully :D

PR #1352, opened 2024-01-18
(APS-206) Remove middle names from tasks endpoint

pezholio:
This is confusing to users, so we need to be consistent.


davidatkinsuk (approved):
LGTM 👍

PR #1356, opened 2024-01-18
Add missing application indexes

pezholio:
This adds some missing indexes that relate to an application, which should hopefully speed the application up. There's still a few more in other places, but these seem to cover a good chunk of the whole application.

### DIFF
--- a/src/main/resources/db/migration/all/20240118152118__add_missing_application_indexes.sql
+++ b/src/main/resources/db/migration/all/20240118152118__add_missing_application_indexes.sql
@@ -1,0 +1,11 @@
+CREATE INDEX ON application_timeline_notes(application_id);
+
+CREATE INDEX ON applications(created_by_user_id);
+
+CREATE INDEX ON approved_premises_applications(probation_region_id);
+
+CREATE INDEX ON placement_requests(application_id);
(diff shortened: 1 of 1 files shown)


src/main/resources/db/migration/all/20240118152118__add_missing_application_indexes.sql
```diff
@@ -0,0 +1,12 @@
+CREATE INDEX ON application_timeline_notes(application_id);
+
+CREATE INDEX ON applications(created_by_user_id);
+CREATE INDEX ON applications(schema_version);
```
davidatkinsuk, line 4: i'm not sure this one is required as I dont _think_ we'll ever be searching for applications with a particular schema (or traversing a hibernate relationship from JsonSchemaEntity to Applications?

davidatkinsuk (approved):
One comment from me, approved regardless 👍

src/main/resources/db/migration/all/20240118152118__add_missing_application_indexes.sql
pezholio: Ah, good point, I'll get rid 👍

PR #1357, opened 2024-01-18
(Fix) Open timeline endpoint up to all roles

pezholio:
We originally locked this down to only admins and workflow managers, but there is no need to do this, and users are getting kicked out. I’ve also changed the endpoint to return a 501 if the service is not CAS1, which seems more appropriate.


davidatkinsuk (approved):
LGTM 👍

PR #1359, opened 2024-01-18
Fix: CAS2: add magic flyway timestamp to make migration 'repeatable'

edavey:
This "repeatable" migration was not running as expected on each deployment to dev and to test.

This is because it was missing the magic "timestamp" placeholder which is used to make migrations repeat.

See https://flywaydb.org/blog/flyway-timestampsandrepeatables

This came to our attention because the domain events table on 'dev' contained old `cas2.application.submitted` events which were missing certain attributes which are required by our CAS 'Submitted applications' MI report.


edavey:
[Error generating "Submitted applications" MI report](https://trello.com/c/KDGyqCAk/782-error-generating-submitted-applications-mi-report)

tahb (approved):

PR #1363, opened 2024-01-19
Enhance Placement Request Dashboard Filtering and Sorting

davidatkinsuk:
Relates to https://dsdmoj.atlassian.net/browse/APS-198

Add requestType filtering
Add sorting for risks, name and request type

This PR also introduces an enum type for a Placement Request ‘request type’ which is included in the model returned to API callers


pezholio (approved):
Looks great - thanks!

PR #1367, opened 2024-01-19
API prod deploys require all frontend e2e to pass

tahb:
Previously this was an informational step to alert the person releasing that their might be an issue.

We are now in the position where CAS1&3 are in public beta and CAS2 is about to go into private beta. The risk of deploying a breaking API change has increased to a point where we should now make this a required step when deploying to preprod and prod.

Instead we now run the risk of flakey e2e tests from any of the projects from failing with false positives and blocking an API deploy.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -373,11 +373,9 @@ workflows:
       - request-preprod-approval:
           type: approval
           requires:
-#            TODO: Uncomment this when we have confirmed the e2es work correctly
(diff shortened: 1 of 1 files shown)


pezholio (approved):
LGTM 👍

edavey (approved):
yes I agree that this we ought to require all 3 services' E2E tests to be green before deploying. If this proves difficult in practice, we can revisit!

PR #1370, opened 2024-01-22
Return entire user object when returning timeline notes

pezholio:
This updates the timeline notes and timeline events to return an entire user object, rather than the user ID. There's also a couple of refactors/tweaks I discovered on the way.


davidatkinsuk (approved):
LGTM 👍

PR #1371, opened 2024-01-22
APS-196 add optional AP Area ID filtering to tasks endpoints

davidatkinsuk:
Relates to https://dsdmoj.atlassian.net/browse/APS-196

Add optional AP Area ID filtering to /tasks and /tasks/{taskType}

/tasks/reallocatable will be updated in a subsequent PR


pezholio (approved):
LGTM 👍

PR #1374, opened 2024-01-22
(APS-215) return AP Areas for users and filter by AP Area

pezholio:


src/main/resources/static/_shared.yml
```diff
@@ -2634,6 +2634,22 @@ components:
...
+        - type: object
+          properties:
+            qualifications:
+              type: array
+              items:
+                $ref: '#/components/schemas/UserQualification'
+            roles:
+              type: array
+              items:
+                $ref: '#/components/schemas/ApprovedPremisesUserRole'
```
davidatkinsuk, line 2649: Do we want to change this to not be ap speciifc?

src/main/resources/static/api.yml
```diff
@@ -3649,6 +3649,12 @@ paths:
           schema:
             type: string
             format: uuid
+        - name: apArea
```
davidatkinsuk, line 3652: I think maybe this should be called 'apAreaId'?

davidatkinsuk (commented):
LGTM from a CAS1 POV, just a few minor comments on potential improvements

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/UserTransformer.kt
```diff
@@ -38,7 +40,19 @@ class UserTransformer(
     )
   }
   fun transformJpaToApi(jpa: UserEntity, serviceName: ServiceName) = when (serviceName) {
-    ServiceName.approvedPremises, ServiceName.cas2 -> ApprovedPremisesUser(
+    ServiceName.cas2 -> Cas2User(
```
edavey, line 43: I think it might be simpler (as well as more meaningful) if you were to throw an exception if the `ServiceName` is "cas2" -- because a `Cas2User` of this form doesn't exist. CAS2 users are from the NOMIS/DPS side and so:

- don't have a Delius username
- don't have a probation region
- don't have qualifications

edavey (commented):
I think it will be clearer if we don't have this Cas2User...

src/main/resources/static/api.yml
pezholio: Good call 👍

davidatkinsuk (approved):
All comments resolved. LGTM 👍

PR #1377, opened 2024-01-23
Simplified Reallocated Task Logic

davidatkinsuk:
This commits simplifies the reallocation code In preparation for adding filtering on AP Area ID.

The 3 ‘allocation queries’ for placement requests, placement applications and assessments have been flattened into a single query per entity type.

This commit has no functional change, just refactoring


pezholio (approved):
Very nice! 👍

PR #1378, opened 2024-01-23
(APS-214) Change Probation Region ID to AP Area ID

pezholio:
This updates the `ApprovedPremisesApplicationEntity` to take an `apArea`, rather than a `probationRegion`. This has resulted in some changes to the queries when filtering tasks, as well as changes to the submission endpoint.

I've kept the legacy `probation_region_id` in the database for now (there's a migration to backfill any applications that already have probation regions), but once all applications have AP Areas by way of a backfill, we can remove this column.

### DIFF
--- a/src/main/resources/db/migration/all/20240123104559__migrate_application_probation_ids_to_ap_area_ids.sql
+++ b/src/main/resources/db/migration/all/20240123104559__migrate_application_probation_ids_to_ap_area_ids.sql
@@ -1,0 +1,13 @@
+UPDATE
+    approved_premises_applications
+set
+    ap_area_id = (
+        SELECT
+            ap_area_id
+        from
+            probation_regions
+        where
+                id = approved_premises_applications.probation_region_id
+    )
+where
+    approved_premises_applications.probation_region_id is not null;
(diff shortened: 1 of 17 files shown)


src/main/resources/db/migration/all/20240123104559__migrate_application_probation_ids_to_ap_area_ids.sql
```diff
@@ -0,0 +1,13 @@
...
+set
+    ap_area_id = (
+        SELECT
+            ap_area_id
+        from
+            probation_regions
+        where
+                id = approved_premises_applications.probation_region_id
+    )
+where
```
davidatkinsuk, line 12: i guess we want to drop the probation_region_id column at somepoint but understand why not now until happy wiht backfill ... shall we create a ticket to do this?

davidatkinsuk (approved):
approved with one comment that doesn't really affect the PR itself 👍

src/main/resources/db/migration/all/20240123104559__migrate_application_probation_ids_to_ap_area_ids.sql
pezholio: I'll do it now, I'm going to move onto that next 👍

PR #1353, opened 2024-01-18
APS-197-API placement request dashboard filter by region.

garethhallberg:
Entity changed, query changed, tests added and or amended

PR #1379, opened 2024-01-23
/placement-requests/dashboard now allows filtering on ap area id

davidatkinsuk:
This supersedes Gareth's PR here -> https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1353 as we're now using AP Area ID instead of Probation Region ID, and it was easier to cherry pick changes into a new branch


pezholio (approved):
LGTM 👍

PR #1380, opened 2024-01-23
APS-150 point adjudications call back at nomis

timoj58:
code base will need updating to use the new adjudications endpoints and is incompatible at present, and adjudications does not want to host this services call to nomis after data migration


davidatkinsuk:
Closing this as it's quite old now and the switch appears to have been done via a different PR

PR #1381, opened 2024-01-23
(APS-219) Add migration job to add AP Areas to applications

pezholio:
This adds a migration job to add an AP area to all Approved Premises applications. In most circumstances, these will be the AP Area of the submitting user, but there are a small subset of applications, where the user and area do not match up. After a review with Sararh, we’ve got a list of these handful of applications, and have the IDs of all the applications and associated AP Area.

Once these are run, we can then update the remaining applications with the AP area of the submitting user.


davidatkinsuk (approved):
LGTM 👍

PR #1382, opened 2024-01-23
(APS-165) Add new licence release type options

pezholio:
This is required to set the correct referral type values in Delius. Once this is in, we can update the frontend to allow the relevant types to be set.


davidatkinsuk (approved):
LGTM 👍

PR #1384, opened 2024-01-23
Delete two production CAS2 applications used for testing

tahb:
We had planned to use realistic referral data and leave these referrals in the service. In practice there were a couple of problems with this which mean we want to remove these:

1. Authorship was from the team
2. These referrals might not receive update from Nacro
3. The referral might be double counted somewhere downstream in reporting

We could delete this manually on the prod command line but want to get this more sensitive change in as code where it can be reviewed.

The application IDs are taken from prod:
Screenshot 2024-01-23 at 16 09 30

Are there any associations I've missed?

[Testing recording in Slack](https://mojdt.slack.com/archives/C05GEPLBRN3/p1706026107319099?thread_ts=1706026069.051109&cid=C05GEPLBRN3).


Harriethw (approved):

edavey (approved):
This looks correct and safe.

PR #1386, opened 2024-01-23
[CAS-183]-Add ignore unknown property for jackson object mapper

sudhakar-plr:


davidatkinsuk (approved):
Given that we trust the service we're calling to not rename existing properties (i.e. changes will always be backwards compatible), this LGTM and probably a good idea as a default for all marshalling 👍

PR #1393, opened 2024-01-25
Placement Request Notifications

davidatkinsuk:
APS-159 - Notify PP when placement request created

1) An application with placement dates included is assessed, leading to an entry in placement_requests
2) PP creates a placement_application

APS-160 - Notify PP when placement_application is allocated

1) placement_application is created and automatically allocated
2) placement_application not allocated to anyone is manually reallocated to someone

APS-161 - Notify PP when placement_applciation has been assessed and accepted or rejected (different notifications for each)

Note that whilst there are ‘withdrawn’ options on the decision screen currently, because they’re being removed in the near future we’re using the ‘rejected’ email for this


pezholio (approved):
LGTM 👍

davidatkinsuk:
Going to hold off merging this because i want to add feature flags in so we can test in pre-prod without the risk of it inadvertently live before testing is complete

davidatkinsuk:
@pezholio i've updated this to add a feature flag so the changes don't go live before tested. Would you mind double checking i've taken the correct approach? I've created a subtask to remove the flag once changes are live -> https://dsdmoj.atlassian.net/browse/APS-256

pezholio (approved):
LGTM - nice to get that feature flag in too for safety's sake 👍

PR #1395, opened 2024-01-25
(142) Change CAS2 application update statuses

edavey:
In this PR we make some changes to the list of possible statuses which can be applied to a `Cas2StatusUpdate` when an assessor from NACRO provides an update to a submitted application.

[Jira 142](https://dsdmoj.atlassian.net/browse/CAS2-142)

### DISABLING

Rather than deleting statuses, which could lead to orphan `cas_2_status_updates` records we mark unwanted statues as inactive:

Set 2 statuses to 'inactive' state

- Could not be placed
- Incomplete

### ADDING

Four new statuses are added:

- Offer accepted
- Offer declined or withdrawn
- Referral cancelled
- Awaiting arrival

### EDITING

One status has a changed label:

- 'Withdrawn' becomes 'Referral withdrawn'

## Notes

The bulk of the work here is in handling the active/inactive distinction. We:

- introduce a new internal model class (`Cas2PersistedApplicationStatus`) to accommodate the additional `isActive` attribute which we don't wish to expose via the `/reference-data/application-status` endpoint and a "transformer" to translate this internal representation into the JSON API representation for the reference data endpoint
- introduce a "finder" class (`Cas2PersistedApplicationStatusFinder`) to offer both `all()` and `active()` lists of statuses

updated_statuses


Harriethw (approved):
Thanks Ed!

PR #1397, opened 2024-01-26
Re-order CAS2 Status updates

Harriethw:
These should be in a certain order for our assessors.

```
More information requested

Awaiting decision

On waiting list

Place offered

Offer accepted

Offer declined or withdrawn

Referral withdrawn

Referral cancelled

Awaiting arrival
```

https://dsdmoj.atlassian.net/browse/CAS2-142

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/reference/Cas2ApplicationStatusSeeding.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/reference/Cas2ApplicationStatusSeeding.kt
@@ -19,16 +19,22 @@ object Cas2ApplicationStatusSeeding {
         description = "The CAS-2 team has the information they need and will make a decision.",
       ),
       Cas2PersistedApplicationStatus(
+        id = UUID.fromString("a919097d-b324-471c-9834-756f255e87ea"),
+        name = "onWaitingList",
+        label = "On waiting list",
+        description = "The applicant has been added to the waiting list for Short-Term Accommodation (CAS-2).",
+      ),
+      Cas2PersistedApplicationStatus(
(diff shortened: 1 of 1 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/reference/Cas2ApplicationStatusSeeding.kt
```diff
@@ -18,17 +18,23 @@ object Cas2ApplicationStatusSeeding {
...
-        id = UUID.fromString("a919097d-b324-471c-9834-756f255e87ea"),
-        name = "onWaitingList",
-        label = "On waiting list",
-        description = "The applicant has been added to the waiting list for Short-Term Accommodation (CAS-2).",
+        id = UUID.fromString("fe254d88-ce1d-4cd8-8bd6-88de88f39019"),
+        name = "offerAccepted",
+        label = "Offer accepted",
+        description = "The accommodation offered has been accepted.",
       ),
       Cas2PersistedApplicationStatus(
```
davidatkinsuk, line 39: could not be placed and incomplete aren't mentioned in the ticket. presumably these are ok to be there?

davidatkinsuk (approved):
LGTM, with one sanity check comment as i'm not familiar with the requirements 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/reference/Cas2ApplicationStatusSeeding.kt
Harriethw: good spot thanks, but yep those ones are meant to be in there - they have `isActive` currently set to `false` so they won't be returned to UI.

PR #1405, opened 2024-01-29
Add withdrawal reasons for placement requests and placement applications

davidatkinsuk:
https://dsdmoj.atlassian.net/browse/APS-246

Also includes a logic change, allowing applicant to withdraw their own placement requests (previously this could only be done by a manager)

### DIFF
--- a/src/main/resources/static/_shared.yml
+++ b/src/main/resources/static/_shared.yml
@@ -505,6 +505,16 @@ components:
       required:
         - arrivalDate
         - departureDate
+    WithdrawPlacementRequest:
+      type: object
+      properties:
+        reason:
+          type: string
+          enum:
+            - DuplicatePlacementRequest
+            - AlternativeProvisionIdentified
+      required:
+        - reason
     Booking:
       allOf:
         - $ref: '#/components/schemas/BookingBody'
--- a/src/main/resources/static/_shared.yml
+++ b/src/main/resources/static/_shared.yml
@@ -3392,6 +3392,16 @@ components:
         - rotl
         - release_following_decision
         - additional_placement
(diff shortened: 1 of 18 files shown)


src/main/resources/static/_shared.yml
```diff
@@ -505,6 +505,16 @@ components:
...
         - arrivalDate
         - departureDate
+    WithdrawPlacementRequest:
+      type: object
+      properties:
+        reason:
+          type: string
+          enum:
+            - DuplicatePlacementRequest
+            - AlternativeProvisionIdentified
```
pezholio, line 515: Could we make this enum a named object? That way it'll be easier to use in the frontend

src/main/resources/static/_shared.yml
```diff
@@ -3392,6 +3392,16 @@ components:
...
         - release_following_decision
         - additional_placement
+    WithdrawPlacementApplication:
+      type: object
+      properties:
+        reason:
+          type: string
+          enum:
+            - DuplicatePlacementRequest
+            - AlternativeProvisionIdentified
```
pezholio, line 3402: Same here vis a vis the named object for the enum. Could potentially share it with the `WithdrawPlacementRequest` object too

pezholio (approved):
Looks good. Minor suggestion about the API objects

src/main/resources/static/_shared.yml
davidatkinsuk: Yep, i'll remember to do that for new enums going forward

src/main/resources/static/_shared.yml
davidatkinsuk: Good idea! have done this

PR #1406, opened 2024-01-29
(APS-253) Ensure filtering by role returns distinct users

pezholio:
When filtering by role, the search sometimes returns multiple users as roles/qualifications can sometimes be applied multiple times. This adds a `distinct` option to the `UserSpecification` to ensure that users are only returned once.


davidatkinsuk (approved):
LGTM 👍

PR #1408, opened 2024-01-29
[APS-186] Implement female assessments allocation rule

danlivings-dxw:
> See [APS-186 on the Approved Premises Jira board](https://dsdmoj.atlassian.net/browse/APS-155).

This PR implements a user allocator rule for women's assessments. The rule is defined in the [Regional Allocations Miro board](https://miro.com/app/board/uXjVNWmfgeM=/?moveToWidget=3458764575277007894&cot=14).

This rule has three configuration options:
- `user-allocations.rules.female-assessments.enabled`: whether the rule should be used.
- `user-allocations.rules.female-assessments.priority`: the priority of the rule (a lower value will evaluate the rule sooner).
- ~~`user-allocations.rules.female-assessments.allocate-to-user`: the username of the user to whom assessments should be allocated if the rule's requirements are met. Currently all women's assessments should be assigned to the same user.~~ This configuration option is no longer used as the rule now dictates that a user with the `WOMENS` qualification should be allocated.

This rule has been enabled and assigned a higher priority than the legacy allocator rule for local development and testing. All non-local environments have this rule disabled for now.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/allocations/rules/FemaleAssessmentRule.kt
```diff
@@ -0,0 +1,33 @@
...
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApprovedPremisesApplicationEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.AssessmentEntity
+
+@Component
+@ConditionalOnProperty(name = ["user-allocations.rules.female-assessments.enabled"])
+class FemaleAssessmentRule(
+  @Value("\${user-allocations.rules.female-assessments.priority:0}")
+  override val priority: Int,
+  @Value("\${user-allocations.rules.female-assessments.allocate-to-user}")
+  private val allocateToUser: String,
```
pezholio, line 17: I think as with the ESAP rule, we're more than likely going to have multiple people qualified to carry out Female assessments, so I'd feel happier if we used `AllocateByQualification` here

danlivings-dxw: I've updated this to use `AllocateByQualification(UserQualification.WOMENS)` instead of a specific user.

pezholio (approved):
LGTM 👍

PR #1409, opened 2024-01-30
Move get booking logic from controller into booking service

davidatkinsuk:
Previously the logic to get a booking for a given premises was fully defined in the premises controller. Whilst some of this logic is specific to premises (e.g. checking if the booking belongs to the given premises), the majority of the logic is generic when retreiving booking.

Because we’re going to add an endpoint to retrieve a booking by its ID only (i.e. no premise ID) to support new withdrawal logic, the common ‘get booking’ logic needs moving into the BookingService

This commit also adds a small premises summary to the Booking API response. This allows the controller to validate that the premises id is as expected, and the premises name will be of use to the UI in subsequent changes

There are no functional changes on this PR, just refactoring

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
@@ -71,6 +71,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserQualification
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserRole
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.DomainEvent
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.PersonInfoResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.validated
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.InternalServerErrorProblem
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.results.AuthorisableActionResult
@@ -122,6 +123,8 @@ class BookingService(
   @Value("\${url-templates.frontend.application}") private val applicationUrlTemplate: String,
   @Value("\${url-templates.frontend.booking}") private val bookingUrlTemplate: String,
   @Value("\${arrived-departed-domain-events-disabled}") private val arrivedAndDepartedDomainEventsDisabled: Boolean,
+  private val userService: UserService,
+  private val userAccessService: UserAccessService,
 ) {
   val approvedPremisesBookingAppealedCancellationReasonId: UUID =
     UUID.fromString("acba3547-ab22-442d-acec-2652e49895f2")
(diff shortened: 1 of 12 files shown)


pezholio (approved):
LGTM - just a couple of Detekt errors that need addressing

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
```diff
@@ -122,14 +124,46 @@ class BookingService(
...
+  private val bookingTransformer: BookingTransformer,
 ) {
   val approvedPremisesBookingAppealedCancellationReasonId: UUID =
     UUID.fromString("acba3547-ab22-442d-acec-2652e49895f2")
 
   private val log = LoggerFactory.getLogger(this::class.java)
 
   fun updateBooking(bookingEntity: BookingEntity): BookingEntity = bookingRepository.save(bookingEntity)
-  fun getBooking(id: UUID) = bookingRepository.findByIdOrNull(id)
+
```
sudhakar-plr, line 137: Just a question, `getBooking()` method is used by `getBookingForPremises` which is then used in multiple places in the premisesController. So not sure the additional function you added here are applicable to other places, i expect the test should fail to reflect if not then just a question.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
```diff
@@ -122,14 +124,46 @@ class BookingService(
...
+      val staffMemberResult = staffMemberService.getStaffMemberByCode(keyWorkerStaffCode, premises.qCode)
+
+      if (staffMemberResult !is AuthorisableActionResult.Success) {
+        throw InternalServerErrorProblem("Unable to get Key Worker via Staff Code: $keyWorkerStaffCode / Q Code: ${premises.qCode}")
+      }
+
+      staffMemberResult.entity
+    }
+
+    return AuthorisableActionResult.Success(bookingTransformer.transformJpaToApi(booking, personInfo, staffMember))
```
sudhakar-plr, line 165: I believe this transformation is to build the API response, so am not sure this should happen in the service. Service should return agnostic response and controller should transform based on the respective API structure.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
davidatkinsuk: Hi - good spot. I addressed this in my first commit - https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1409/commits/330d6deaa0359db97351c4a8904e87d6b1cfdda8

The few services were using this are now using the repository directly to ensure i didn't break them

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
davidatkinsuk: Yes, really good point. I did quickly check this because i couldn't recall where transformers sit and I got unlucky in that i seemingly found the only service (ApplicationService) that uses transformers ... I'll make a note to move those too in the future! I've updated the code to move the transformer into the controller.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
sudhakar-plr: oh got it, how did i missed it?!

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/BookingServiceTest.kt
```diff
@@ -160,82 +163,53 @@ class BookingServiceTest {
...
-    bedMoveRepository = mockBedMoveRepository,
-    premisesRepository = mockPremisesRepository,
-    assessmentRepository = mockAssessmentRepository,
-    notifyConfig = NotifyConfig(),
-    applicationUrlTemplate = "http://frontend/applications/#id",
-    bookingUrlTemplate = "http://frontend/premises/#premisesId/bookings/#bookingId",
-    arrivedAndDepartedDomainEventsDisabled = true,
-  )
+  private val bookingService = createBookingService(arrivedAndDepartedDomainEventsDisabled = false)
+  private val bookingServiceWithArrivedAndDepartedDomainEventsDisabled = createBookingService(arrivedAndDepartedDomainEventsDisabled = true)
```
sudhakar-plr, line 213: would be nice if we can add unit test for the new changes in getBooking() method. Previously it was just a DB call, but now you added couple of service class within the method , so would be helpful if we test the negative flow. I believe you have added integration test to test positive flow.

davidatkinsuk: Hi @sudhakar-plr , the logic in the new version of getBooking() is existing code, moved from the PremisesController#premisesPremisesIdBookingsBookingIdGet function (which now calls this new function instead). The only changes I made to the existing code was to make it return AuthroisationActionResult instead of directly throwing Problems. Therefore, the existing integration tests in BookingTest give us good coverage, as shown below:

Screenshot 2024-01-30 at 11 44 56

Therefore, whilst i agree it would be good to have service level unit tests, I think the coverage we have the moment is sufficient to at least get this change merged as it's blocking some work we need for UI

sudhakar-plr (approved):

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/BookingServiceTest.kt
davidatkinsuk: Yes, i agree on establishing strategy for unit vs integration tests ... i'll make a note for the next all-cas API call 👍

PR #1413, opened 2024-01-30
Paginate CAS2 submissions end point

Harriethw:
UI PR is here https://github.com/ministryofjustice/hmpps-community-accommodation-tier-2-ui/pull/498

This adds pagination to `/submissions`, which lists all submitted CAS2
applications.

We are using the existing utility functions and `Page` response already
in the API, and utilised by CAS1/3's `getAllApprovedPremisesApplications` method. 

* The sorting is hardcoded as ascending "submitted_at" because we do not yet have UI/need
to set the sort order from the frontend. 
* We're using the default page size for now, but may need to customise that in the near future.
* We have set the page`param` to be optional in order not to break the existing frontend - if
no page param is given, then all the results are returned, which means
we will not break the existing UX on the frontend.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/ApplicationService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/ApplicationService.kt
@@ -19,6 +19,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas2Applicati
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.NomisUserEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.NomisUserRepository
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.DomainEvent
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.PaginationMetadata
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.ValidationErrors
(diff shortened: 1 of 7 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/ApplicationService.kt
```diff
@@ -51,8 +55,16 @@ class ApplicationService(
     return applicationRepository.findAllCas2ApplicationSummariesCreatedByUser(user.id)
   }
 
-  fun getAllSubmittedApplicationsForAssessor(): List<Cas2ApplicationSummary> {
-    return applicationRepository.findAllSubmittedCas2ApplicationSummaries()
+  fun getAllSubmittedApplicationsForAssessor(page: Int?): Pair<List<Cas2ApplicationSummary>, PaginationMetadata?> {
```
pezholio, line 58: Probs not necessary for now as you're not doing anything in the way of sorting, but we have a nice `PageCriteria` data class, which makes passing around pagination and sorting data more easy

pezholio (commented):
LGTM 👍

edavey (approved):
Nice work, v clear!

If you have time to look into it now, it might be worth seeing if adopting the `PageCriteria` class improves this? https://github.com/ministryofjustice/approved-premises-api/blob/main/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/util/PaginationUtils.kt#L25-L25

Harriethw:
@edavey @pezholio thanks both, yep that `PageCriteria` class looks good, I'll take a look at using it before merging. (and don't think this work is super urgent to get merged based on the rate of applications being submitted)

PR #1414, opened 2024-01-30
Add endpoint to get booking by id

davidatkinsuk:


edavey (approved):
Very clear, thks, though perhaps it would be worth explaining in the commit message why it now makes sense to rename the controller -- now that it's providing more than just "search"?

davidatkinsuk:
> Very clear, thks, though perhaps it would be worth explaining in the commit message why it now makes sense to rename the controller -- now that it's providing more than just "search"?

Thanks Ed. Good idea on explaining the name change - I've pushed a new commit message to clarify this

PR #1416, opened 2024-01-30
Refactor ‘withdrawable’ logic from SQL into Kotlin Code

davidatkinsuk:
Initially the logic to discover which placement applications, placement requests and bookings where withdrawable was written as a single SQL query. Whilst this may be the most efficient way to find these entities, it was duplicating logic that was, in some cases, already represented in kotlin code. Duplication of logic across multiple languages (SQL/Kotlin) is not ideal. Kotlin typically provides us with a more fluent, type safe, cohesive and reusable way to express such logic.

Whilst it’s recognised that in some cases we have no choice but to encode logic in SQL due to the size of the data set we’re searching (e.g. if we’re scanning the whole table with sortingm paging etc.), for this scenario we have a very limited set of entities to deal with (those linked to the application), and the withdrawable endpoint is going to be one hit far less frequently than others.

For this reason we’ve converted the SQL query into java code, distributed between the placement applciation service, placement request service and booking service, defining the logic in a far more cohesive manner.


pezholio (approved):
Nice work - this is much more succinct and readable 👍

PR #1417, opened 2024-01-30
(APS-259) Allow assessors with LAO to see full person

pezholio:
If someone with permissions to see LAO offenders gets assigned an LAO offender, they can’t proceed, as they can’t click through to the assessment. As they’re able to view the assessment (and all the associated personal details) it follows that they should be able to see the details

I’ve only made the change for CAS1 at this point to avoid any unexpected issues on CAS3


davidatkinsuk (approved):
LGTM 👍

PR #1421, opened 2024-01-31
(APS-250) Remove withdrawal reasons

pezholio:
These are no longer needed. We'll need to make sure the corresponding UI changes are deployed before / as soon as these changes are live, as this will be a breaking change


davidatkinsuk (approved):
LGTM 👍

PR #1422, opened 2024-01-31
Fix mapping bug when using delius offender api

davidatkinsuk:
Relates to https://dsdmoj.atlassian.net/browse/APS-263

The code previously implicitly assumed that the results from the delius offender details API were being returned in the same order as the list of the CRNs requested. We’ve changed the code to remove this assumption and link results by matching on CRNs.

This PR also fixes an inconsistency between delius API and community API behaviour in the event of error, ensuring a ClientResult.Failure is returned for each CRN requested.


pezholio (approved):
LGTM - nice work on building on the feature but keeping the behaviour predictable at each step - makes my life as a reviewer much easier! 👍

PR #1423, opened 2024-01-31
Return region with `ApprovedPremisesApplication`

pezholio:
In some cases it's going to be useful to show an application's region when listing (for example) tasks.


davidatkinsuk (approved):
LGTM 👍

PR #1424, opened 2024-01-31
(APS-265) Loosen permissions when viewing assessments

pezholio:
All users will need to view an assessment (except when the offender is a limited access offender and they don’t have permissions to view), so we’ll open up permissions to all logged in users.


davidatkinsuk (approved):
LGTM 👍

PR #1426, opened 2024-01-31
[APS-236] Introduce appeals for CAS1 applications

danlivings-dxw:
> See [APS-236 on the Approved Premises Jira board](https://dsdmoj.atlassian.net/browse/APS-236).

This PR introduces appeals for when an application is rejected. An appeal can be created for an application using the `POST /applications/{applicationId}/appeals` endpoint. The body of a request to this endpoint looks like:

```json
{
    "appealDate": "2024-01-01",
    "appealDetail": "Some information about the appeal",
    "reviewer": "A Name",
    "decision": "accepted",
    "decisionDetail": "Some information about the decision"
}
```

An appeal can only be created by a user if they have the `CAS1_APPEALS_MANAGER` role (see #1411) and can access the application.

This PR does not include an endpoint for viewing an appeal, as this is part of APS-238.

### DIFF
--- a/src/main/resources/static/_shared.yml
+++ b/src/main/resources/static/_shared.yml
@@ -4124,3 +4124,67 @@ components:
         - A1
         - A2
         - A3
+    NewAppeal:
+      type: object
+      properties:
+        appealDate:
+          type: string
+          format: date
+        appealDetail:
+          type: string
+        reviewer:
+          type: string
+        decision:
+          $ref: '#/components/schemas/AppealDecision'
+        decisionDetail:
+          type: string
+      required:
+        - appealDate
+        - appealDetail
+        - reviewer
+        - decision
+        - decisionDetail
+    Appeal:
+      type: object
+      properties:
+        id:
+          type: string
+          format: uuid
+        appealDate:
+          type: string
+          format: date
+        appealDetail:
+          type: string
+        reviewer:
+          type: string
+        decision:
+          $ref: '#/components/schemas/AppealDecision'
+        decisionDetail:
+          type: string
+        createdAt:
+          type: string
+          format: date-time
+        applicationId:
+          type: string
+          format: uuid
+        assessmentId:
+          type: string
+          format: uuid
+        createdByUserId:
+          type: string
+          format: uuid
+      required:
+        - id
+        - appealDate
+        - appealDetail
+        - reviewer
+        - decision
+        - decisionDetail
+        - createdAt
+        - applicationId
+        - createdByUserId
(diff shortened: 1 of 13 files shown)


src/main/resources/static/_shared.yml
```diff
@@ -4114,3 +4114,67 @@ components:
...
+          type: string
+          format: date
+        appealDetail:
+          type: string
+        reviewer:
+          type: string
+      required:
+        - appealDate
+        - appealDetail
+        - reviewer
```
pezholio, line 4130: As discussed, I think we want to create appeals endpoint to be a one-shot endpoint (at least for now). This makes the UI work easier to implement

pezholio (changes requested):

src/main/resources/static/_shared.yml
danlivings-dxw: This has now been changed to not require a separate endpoint for adding the decision.

pezholio (approved):
LGTM - really clear 👍

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/AppealServiceTest.kt
```diff
@@ -0,0 +1,194 @@
...
+          it.application == application &&
+          it.createdBy == createdByUser
+      }
+      verify(exactly = 1) {
+        appealRepository.save(
+          match {
+            it.appealDate == now &&
+              it.appealDetail == "Some information about why the appeal is being made" &&
+              it.reviewer == "ReviewBot 9000" &&
+              it.application == application &&
```
edavey, line 187: do we also want to verify that the expected decision is being recorded by the repository? e.g. something like:

```
it.decision == AppealDecision.accepted
```

edavey (approved):
This looks ideal. Thanks for structuring the PR so clearly. 

I've just made a tiny comment on the service unit test's description of the collaboration with the repository -- to suggest including the decision itself in the description of the message passed. Apologies if I've misunderstood how that test is working..

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/AppealServiceTest.kt
danlivings-dxw: Good catch, I've updated the test to include the `decision` and the `decisionDetail` fields.

PR #1431, opened 2024-02-01
Update CAS1 Booking Cancellation Reasons

davidatkinsuk:
Relates to https://dsdmoj.atlassian.net/browse/APS-268


pezholio (approved):

sudhakar-plr (approved):
Ok to me.
I hope changing the reason name via app is already done and its migrating existing one.

davidatkinsuk:
> Ok to me. I hope changing the reason name via app is already done and its migrating existing one.

Good call - Rich has confirmed that the App just displays what's pulled back from the database, so no change required on UI

PR #1435, opened 2024-02-02
Refactor GetReallocatedTasks Functions to reduce duplication and provide generic functionality

davidatkinsuk:
Relates to https://dsdmoj.atlassian.net/browse/APS-223

This PR is in preparation for sharing the functionality with the /tasks endpoint, reducing duplication. At which point we can then add additional functionality required by APS-223. There is no functional change

The changes significantly reduce duplication when retreiving tasks for reallocation. Previously this request could be redirected to one of four different services (one for each specific type and one for all types). Now all requests go to a single service and a single SQL query.

This may be slightly less performant than the previous approach when requesting a specific task type. If we find there are performance problems it's an easy fix because all requests now go through a single function. In this case we'd add a specific repository function on the TaskRepository for task types, but do in such a way that we use the same SQL through constants.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/TasksController.kt
```diff
@@ -514,7 +514,7 @@ class TasksController(
     )
   }
 
-  private fun placementApplicationTasks(
+  private fun reallocatbleResponseForPlacementApplications(
```
pezholio: Typo here

pezholio (approved):
This is a VAST improvement, thanks! 👍 One comment about a typo, but given the code is deleted as part of this PR, then I don't think it matters

PR #1436, opened 2024-02-02
(APS-267) Ensure all things related to an application are withdrawn when an application is withdrawn

pezholio:
This adds a new `WithdrawablesService` that contains the source of truth for all things that are "withdrawable", we then can add a `withdrawAllForApplication` function that withdraws everything that is "withdrawable" for an application in one shot. I've had to add a new internal reason for placement application / requests that shows they were withdrawn as a side effect of the whole application being withdrawn.


davidatkinsuk (approved):
LGTM, nice to see some of that logic being moved out of the controller too. 👍

PR #1439, opened 2024-02-02
(APS-273) Hide withdrawable bookings from non CRU users

pezholio:
This adds a configurable `showAllBookings` option to the `allWithdrawables` service function. If true, we use the `UserAccessService` to see if the user can manage the booking. If they can’t we filter them out.

In reality, `currentUserCanManagePremisesBookings` for CAS1 only looks at a user’s role, so this will only return true for CRU or AP managers.

When using the withdrawAllForApplication, we set `showAllBookings` to true, because it doesn’t make sense to withdraw an application and not withdraw the associated bookings. This may change as the service evolves, but I think this makes sense as things currently stand.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/WithdrawableService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/WithdrawableService.kt
@@ -16,24 +16,25 @@ class WithdrawableService(
   @Lazy private val placementRequestService: PlacementRequestService,
   @Lazy private val bookingService: BookingService,
   @Lazy private val placementApplicationService: PlacementApplicationService,
+  @Lazy private val userAccessService: UserAccessService,
 ) {
   val approvedPremisesWithdrawnByPPBookingWithdrawnReasonId: UUID =
     UUID.fromString("d39572ea-9e42-460c-ae88-b6b30fca0b09")
 
-  fun allWithdrawables(application: ApprovedPremisesApplicationEntity): Withdrawables {
+  fun allWithdrawables(application: ApprovedPremisesApplicationEntity, onlyUserManageableBookings: Boolean = true): Withdrawables {
     val placementRequests = placementRequestService.getWithdrawablePlacementRequests(application)
     val bookings = bookingService.getCancelleableBookings(application)
     val placementApplications = placementApplicationService.getWithdrawablePlacementApplications(application)
 
     return Withdrawables(
       placementRequests = placementRequests,
-      bookings = bookings,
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/WithdrawableService.kt
```diff
@@ -16,24 +16,25 @@ class WithdrawableService(
   @Lazy private val placementRequestService: PlacementRequestService,
   @Lazy private val bookingService: BookingService,
   @Lazy private val placementApplicationService: PlacementApplicationService,
+  @Lazy private val userAccessService: UserAccessService,
 ) {
   val approvedPremisesWithdrawnByPPBookingWithdrawnReasonId: UUID =
     UUID.fromString("d39572ea-9e42-460c-ae88-b6b30fca0b09")
 
-  fun allWithdrawables(application: ApprovedPremisesApplicationEntity): Withdrawables {
+  fun allWithdrawables(application: ApprovedPremisesApplicationEntity, showAllBookings: Boolean = false): Withdrawables {
```
davidatkinsuk, line 24: could we maybe flip this boolean and call it something like 'onlyShowUserManageableBookings'? (or something a bit more concise :) )

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/WithdrawableService.kt
```diff
@@ -16,24 +16,25 @@ class WithdrawableService(
...
       placementRequests = placementRequests,
-      bookings = bookings,
+      bookings = if (showAllBookings) { bookings } else { bookings.filter { userAccessService.currentUserCanManagePremisesBookings(it.premises) } },
       placementApplications = placementApplications,
     )
   }
 
   fun withdrawAllForApplication(application: ApprovedPremisesApplicationEntity, user: UserEntity) {
-    val withdrawables = allWithdrawables(application)
+    val withdrawables = allWithdrawables(application, true)
```
davidatkinsuk, line 37: what's the implication of this if the user doesn't have permissions to withdraw a booking that exists? Or will that never happen?

davidatkinsuk (changes requested):
A question and suggestion from me

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/WithdrawableService.kt
pezholio: This won't happen because the permissions are handled in the controller (we should be much more consistent about where this kind of stuff happens!)

davidatkinsuk (approved):
LGTM, i's great how you can see this being nicely built up 👍

PR #1440, opened 2024-02-02
Add an optional list of `StatusDetails` to CAS2 Application `StatusUpdates` 

Harriethw:
* I'm not a Kotlin/SQL expert, would appreciate any feedback about better way of doing this.
* This is quite a big PR because we are in a rush to have work un-blocked for Frontend. 
* DOES NOT INCLUDE: 
  * updating Domain Events or MI reports - this can come in near future.
  * the Details for Further Information - this not confirmed yet

We [previously introduced ](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1113)a set of CAS2 specific Application Statuses, that are applied by Assessors.

Screenshot 2024-02-02 at 17 01 25

We now need to introduce an optional sub-level of category on those statuses, which we are calling 'Status Details'.
A Status (e.g. 'Offer declined') could have none or several Details on it (e.g. 'Change of circumstances', 'No Reply/No Response'). 

This is proposing

1. Display the new details to the UI
Screenshot 2024-02-02 at 17 07 00

* We add the concept of a `Cas2PersistedApplicationStatusDetail` as a field on the `Cas2PersistedApplicationStatus` models seeded in `Cas2PersistedApplicationStatusSeeding.kt` file. So that the Details are returned when the UI calls `/reference-data/application-status`.

2. Add Details to the saved Status Updates on an Application

Screenshot 2024-02-02 at 17 14 19

* The endpoint `/submissions/{applicationId}/status-updates` takes an optional param `newStatusDetails`.
* We create a new table `cas_2_status_update_details` , which has a ManyToOne bi-directional relationship with the existing `cas2_status_updates`
* When a Status Update is posted with a list of Details, the `StatusUpdateService` iterates through the list of Details and saves them in the db.

3. Return the Details on Status Updates on an Application

Screenshot 2024-02-02 at 17 18 20

* So that we can render these Details on the UI, we update the StatusUpdateTransformer to return Details, which are then returned on a SubmittedApplication when making a GET request to `/submissions/{applicationId}`

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/StatusUpdateDetailEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/StatusUpdateDetailEntity.kt
@@ -1,0 +1,36 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity
+
+import org.hibernate.annotations.CreationTimestamp
+import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.stereotype.Repository
+import java.time.OffsetDateTime
+import java.util.UUID
+import javax.persistence.Entity
+import javax.persistence.Id
+import javax.persistence.JoinColumn
+import javax.persistence.ManyToOne
+import javax.persistence.Table
+
+@Repository
+interface Cas2StatusUpdateDetailRepository : JpaRepository<Cas2StatusUpdateDetailEntity, UUID> {
+  fun findFirstByStatusUpdateIdOrderByCreatedAtDesc(statusUpdateId: UUID): Cas2StatusUpdateDetailEntity?
+}
+
+@Entity
+@Table(name = "cas_2_status_update_details")
+data class Cas2StatusUpdateDetailEntity(
+  @Id
+  val id: UUID,
+  val statusDetailId: UUID,
+
+  val label: String,
+
+  @ManyToOne
+  @JoinColumn(name = "status_update_id")
+  val statusUpdate: Cas2StatusUpdateEntity,
+
+  @CreationTimestamp
+  var createdAt: OffsetDateTime? = null,
+) {
+  override fun toString() = "Cas2StatusDetailEntity: $id"
+}
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/StatusUpdateDetailEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/StatusUpdateDetailEntity.kt
@@ -3,6 +3,8 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity
 import org.hibernate.annotations.CreationTimestamp
 import org.springframework.data.jpa.repository.JpaRepository
 import org.springframework.stereotype.Repository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.reference.Cas2PersistedApplicationStatusDetail
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.reference.Cas2PersistedApplicationStatusFinder
 import java.time.OffsetDateTime
 import java.util.UUID
 import javax.persistence.Entity
@@ -33,4 +35,9 @@ data class Cas2StatusUpdateDetailEntity(
   var createdAt: OffsetDateTime? = null,
 ) {
   override fun toString() = "Cas2StatusDetailEntity: $id"
+
+  fun statusDetail(statusId: UUID, detailId: UUID): Cas2PersistedApplicationStatusDetail {
+    return Cas2PersistedApplicationStatusFinder().getById(statusId).statusDetails?.find { detail -> detail.id == detailId }
+      ?: error("Status detail with id $detailId not found")
+  }
 }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/reference/Cas2ApplicationStatusSeeding.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/reference/Cas2ApplicationStatusSeeding.kt
@@ -55,18 +55,99 @@ object Cas2ApplicationStatusSeeding {
         name = "offerDeclined",
         label = "Offer declined or withdrawn",
         description = "The accommodation offered has been declined or withdrawn.",
+        statusDetails = listOf(
+          Cas2PersistedApplicationStatusDetail(
+            id = UUID.fromString("62645779-242d-4601-a8f8-d2cbf1d41dfa"),
+            name = "areaUnsuitable",
+            label = "Area unsuitable",
+          ),
+          Cas2PersistedApplicationStatusDetail(
+            id = UUID.fromString("fc38f750-e9d2-4270-b542-d38286b9855c"),
+            name = "changeOfCircumstances",
+            label = "Change of circumstances",
+          ),
+          Cas2PersistedApplicationStatusDetail(
+            id = UUID.fromString("31122b89-e087-4b5f-b59a-f7ffa0dd3e0c"),
+            name = "noResponse",
+            label = "No response",
+          ),
+          Cas2PersistedApplicationStatusDetail(
+            id = UUID.fromString("32e62af3-6ea5-4496-a82c-b7bad67080a5"),
+            name = "offerWithdrawnByNacro",
+            label = "Offer withdrawn by Nacro",
+          ),
+          Cas2PersistedApplicationStatusDetail(
+            id = UUID.fromString("b5bfbff4-aaa6-4fb0-ba36-5bca58927dc5"),
+            name = "propertyUnsuitable",
+            label = "Property unsuitable for needs",
+          ),
+          Cas2PersistedApplicationStatusDetail(
+            id = UUID.fromString("f58267b8-f91b-4a4f-9aa2-80089ba111e4"),
+            name = "withdrawnByReferrer",
+            label = "Withdrawn by referrer",
+          ),
+          Cas2PersistedApplicationStatusDetail(
+            id = UUID.fromString("ed5d529c-d7d1-4f29-a0c0-89fd104cc320"),
+            name = "rehousedByAnotherLandlord",
+            label = "Rehoused by another landlord",
+          ),
+        ),
       ),
       Cas2PersistedApplicationStatus(
         id = UUID.fromString("004e2419-9614-4c1e-a207-a8418009f23d"),
         name = "withdrawn",
         label = "Referral withdrawn",
         description = "The prison offender manager (POM) withdrew the application.",
+        statusDetails = listOf(
+          Cas2PersistedApplicationStatusDetail(
+            id = UUID.fromString("e4a2391e-e847-427a-a913-51e0b0ad9f52"),
+            name = "hdcNoLongerEligible",
+            label = "HDC - No longer eligible",
+          ),
+          Cas2PersistedApplicationStatusDetail(
+            id = UUID.fromString("c5dce0d2-fc05-4a07-8157-25b8821cdb06"),
+            name = "governorDecidedUnsuitable",
+            label = "Governor - Decided unsuitable",
+          ),
+          Cas2PersistedApplicationStatusDetail(
+            id = UUID.fromString("3fb37f85-be88-4eee-812d-af122e268eef"),
+            name = "governorChosenAlternative",
+            label = "Governor - Chosen alternative accommodation",
+          ),
+          Cas2PersistedApplicationStatusDetail(
+            id = UUID.fromString("8a857a7d-94f9-43ec-963c-2a2528e88a6e"),
+            name = "governorOther",
+            label = "Governor - Other",
+          ),
+          Cas2PersistedApplicationStatusDetail(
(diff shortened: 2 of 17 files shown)


detekt-baseline-main.xml
```diff
@@ -930,6 +930,10 @@
     <ID>ReturnCount:ExceptionHandling.kt$ExceptionHandling$override fun toProblem(throwable: Throwable, status: StatusType): ThrowableProblem?</ID>
     <ID>ReturnCount:OffenderService.kt$OffenderService$fun getInfoForPerson(crn: String, deliusUsername: String, ignoreLao: Boolean): PersonInfoResult</ID>
     <ID>ReturnCount:TaskService.kt$TaskService$fun reallocateTask(requestUser: UserEntity, taskType: TaskType, userToAllocateToId: UUID, id: UUID): AuthorisableActionResult&lt;ValidatableActionResult&lt;Reallocation&gt;&gt;</ID>
+    <ID>ReturnCount:StatusUpdateService.kt$StatusUpdateService$fun create( applicationId: UUID,
+      statusUpdate: Cas2ApplicationStatusUpdate,
```
Harriethw, line 934: I don't know if this is best practice, but couldn't see a way of having less returns.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/StatusUpdateDetailEntity.kt
```diff
@@ -0,0 +1,44 @@
...
+}
+
+@Entity
+@Table(name = "cas_2_status_update_details")
+data class Cas2StatusUpdateDetailEntity(
+  @Id
+  val id: UUID,
+  val statusDetailId: UUID,
+
+  val description: String,
```
Harriethw, line 28: we might not need a description, depends on content requirements

detekt-baseline-main.xml
Harriethw: also not sure if the ignore should be in `detekt-baseline-main` or `detekt-baseline`?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/StatusUpdateService.kt
```diff
@@ -48,6 +52,16 @@ class StatusUpdateService(
         ValidatableActionResult.GeneralValidationError("The status ${statusUpdate.newStatus} is not valid"),
       )
 
+    var statusDetails: List<Cas2PersistedApplicationStatusDetail>? = null
```
Harriethw, line 55: this adds to many 'returns' to the `create` method according to detekt, but not sure how to refactor it - from what I understand Kotlin doesn't like functions that could return more than one type, so how could I extract some of these returns to a separate method? 🤔

Harriethw: added a suppression flag, following convo here https://mojdt.slack.com/archives/C03K0HB0LBE/p1707221283595289

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/reference/Cas2ApplicationStatusSeeding.kt
```diff
@@ -11,6 +11,29 @@ object Cas2ApplicationStatusSeeding {
         name = "moreInfoRequested",
         label = "More information requested",
         description = "The prison offender manager (POM) must provide information requested for the application to progress.",
+        statusDetails = listOf(
+          Cas2PersistedApplicationStatusDetail(
+            id = UUID.fromString("9019c886-dcba-4277-b667-423a2ab847c3"),
+            statusId = UUID.fromString("f5cd423b-08eb-4efb-96ff-5cc6bb073905"),
+            name = "aboutTheApplicant",
+            description = "About the applicant",
```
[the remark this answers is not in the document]
Harriethw: we didn't need description so took it out

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/reference/Cas2PersistedApplicationStatus.kt
```diff
@@ -7,5 +7,6 @@ data class Cas2PersistedApplicationStatus(
   val name: String,
   val label: String,
   val description: String,
+  val statusDetails: List<Cas2PersistedApplicationStatusDetail>? = null,
```
davidatkinsuk, line 10: would it be possible to make this non-nullable and initialise to emptyList()? Doesn't matter too much in kotlin as we have nice nullable control but can make calling code a bit simpler maybe?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/reference/Cas2ApplicationStatusSeeding.kt
```diff
@@ -55,18 +55,114 @@ object Cas2ApplicationStatusSeeding {
...
       ),
       Cas2PersistedApplicationStatus(
         id = UUID.fromString("f13bbdd6-44f1-4362-b9d3-e6f1298b1bf9"),
         name = "cancelled",
         label = "Referral cancelled",
         description = "The application has been cancelled.",
+        statusDetails = listOf(
+          Cas2PersistedApplicationStatusDetail(
+            id = UUID.fromString("f13bbdd6-44f1-4362-b9d3-e6f1298b1bf9"),
+            statusId = UUID.fromString("fc0b7e04-7eb1-427b-b2b7-85a109a067b0"),
```
davidatkinsuk, line 149: are these statusIds meant to match the 'parent' status id? if yes, they don't appear to here? do we actually need this statusId field (I had a quick look and couldn't find any usage, but i could be wrong!)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/StatusUpdateService.kt
```diff
@@ -35,6 +39,8 @@ class StatusUpdateService(
     return findActiveStatusByName(statusUpdate.newStatus) != null
   }
 
+  // too many returns, but unsure how to refactor to reduce
```
davidatkinsuk, line 42: i'm ok with not having the comment here as the SuppressWarnings makes it clear it was considered

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/StatusUpdateService.kt
```diff
@@ -48,6 +54,16 @@ class StatusUpdateService(
...
 
+    var statusDetails: List<Cas2PersistedApplicationStatusDetail>? = null
+    if (!statusUpdate.newStatusDetails.isNullOrEmpty()) {
+      statusDetails = statusUpdate.newStatusDetails.map { detail ->
+        findStatusDetailOnStatus(status, detail)
+          ?: return AuthorisableActionResult.Success(
+            ValidatableActionResult.GeneralValidationError("The status detail $detail is not valid"),
+          )
+      }
+    }
```
davidatkinsuk, line 65: instead of having a mutable list 'var statusDetails' how about doing something like the following code. Then statusDetails will be non nullable removing need for ? when accessing it for the forEach

```
val statusDetails = if (statusUpdate.newStatusDetails.isNullOrEmpty()) {
  emptyList()
} else {
  statusUpdate.newStatusDetails.map { detail ->
    findStatusDetailOnStatus(status, detail)
      ?: return AuthorisableActionResult.Success(
        ValidatableActionResult.GeneralValidationError("The status detail $detail is not valid"),
      )
  }
}
```

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/StatusUpdateService.kt
```diff
@@ -65,13 +81,27 @@ class StatusUpdateService(
...
+
     createStatusUpdatedDomainEvent(createdStatusUpdate)
 
     return AuthorisableActionResult.Success(
       ValidatableActionResult.Success(createdStatusUpdate),
     )
   }
 
+  private fun findStatusDetailOnStatus(status: Cas2PersistedApplicationStatus, detailName: String) =
```
davidatkinsuk, line 102: could this be a function on the Cas2PersistedApplicationStatus class?

davidatkinsuk (changes requested):
Hi - looks good, I can see this follows the same patterns established by the existing Cas2PersistedApplicationStatus work.

I've made a few comments - i think they're mostly stylistic so please feel free to ignore, The only one I think needs attention is where statusId is defined in the Cas2ApplicationStatusSeeding.kt, hence review being 'Request changes'

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/reference/Cas2ApplicationStatusSeeding.kt
```diff
@@ -55,18 +55,114 @@ object Cas2ApplicationStatusSeeding {
...
+            id = UUID.fromString("31122b89-e087-4b5f-b59a-f7ffa0dd3e0c"),
+            statusId = UUID.fromString("9a381bc6-22d3-41d6-804d-4e49f428c1de"),
+            name = "noResponse",
+            label = "No Response",
+          ),
+          Cas2PersistedApplicationStatusDetail(
+            id = UUID.fromString("32e62af3-6ea5-4496-a82c-b7bad67080a5"),
+            statusId = UUID.fromString("9a381bc6-22d3-41d6-804d-4e49f428c1de"),
+            name = "offerWithdrawnByNacro",
+            label = "Other Withdrawn by Nacro",
```
tahb, line 81: Should this be "Offer"?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/reference/Cas2ApplicationStatusSeeding.kt
Harriethw: ah yeah that's a good point, I don't think we need the `statusId` at all here...

src/main/resources/static/codegen/built-cas2-api-spec.yml
```diff
@@ -2268,11 +2268,35 @@ components:
         description:
           type: string
           example: 'More information about the application has been requested from the POM (Prison Offender Manager).'
+        statusDetails:
```
tahb, line 2271: Just above here in the OpenAPI spec should we include the created_at timestamp of the Cas2ApplicationStatus so we can give that to the frontend to present?

src/main/resources/db/migration/all/20240206092803__add_cas2_status_update_detail_table.sql
```diff
@@ -0,0 +1,18 @@
...
+
+    PRIMARY KEY (id),
+
+    CONSTRAINT fk_status_update_id
+        FOREIGN KEY(status_update_id)
+    	    REFERENCES cas_2_status_updates(id)
+
+);
+-- index for status updates to cas_2_status_updates table
+CREATE INDEX ON cas_2_status_update_details(status_update_id);
```
tahb, line 17: Well remembered! 👍

tahb (commented):
Couple of small things from me but generally looks good to me. Given David's request for more info after a thoughtful review I'll leave the approval to them.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/StatusUpdateService.kt
Harriethw: that's some nice Kotlin-y syntax, will use!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/reference/Cas2ApplicationStatusSeeding.kt
Harriethw: oof yeah good spot!

src/main/resources/static/codegen/built-cas2-api-spec.yml
Harriethw: that would be good, but it would also mean adding to this big PR 🙈 should be done as part of this ticket? https://dsdmoj.atlassian.net/jira/software/c/projects/CAS2/boards/1347?selectedIssue=CAS2-213

Harriethw: oh actually, this model doesn't have a `created_at` because it's the seeded data, it's the `Cas2StatusUpdate` which has the timestamp, and that is already being returned and rendered on the timeline as it stands

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/StatusUpdateService.kt
Harriethw: nice one, done 👍

davidatkinsuk (approved):
Thanks for the quick turn around on updates. I've also checked Tom's comments are resolved. LGTM 👍

PR #1441, opened 2024-02-02
[APS-216] Update emergency rules and enable new allocations engine

pezholio:
This updates the emergency rules to be managed by config (basically cribbing from the work @danlivings-dxw did and I asked him to change - sorry Dan!), as well as disabling the legacy allocation engine in all environments. 

This will need to be merged on Monday as soon as we have the Delius usernames of the people we want to allocate to.


edavey:
Is there any more background information on the allocations system? I'm not clear on the purpose and status of "legacy allocations". It looks like this PR disables the use of  the `UserAllocationsEngine` entirely? Is that right? Are we planning to remove that code?

edavey (approved):
This looks clear and thorough. 

However, I don't have the context to say whether it is "correct" in scope!

pezholio:
Yeah, the plan is to remove the legacy allocations engine entirely eventually. This just gives us the flexibility to put it back if there are any issues

PR #1442, opened 2024-02-02
Use /tasks/reallocatable code to service /tasks requests

davidatkinsuk:
This PR makes several enhancements to the /tasks/reallocatable endpoint such that it can support requests sent to the /tasks endpoint. These are

1. Allow requests from users with the role CAS1_MATCHER
2. Add the option to filter on the allocated user

Subsequently, requests for /tasks endpoint is now forwarded to the reallocatable logic

The end goal is to have a single one-size-fits-all tasks endpoint


pezholio (approved):
LGTM - this should hopefully go a long way to simplifying the Tasks controller!

PR #1447, opened 2024-02-05
Redefine /tasks endpoint to provide generic search 

davidatkinsuk:
Relates to https://dsdmoj.atlassian.net/browse/APS-223

Before this PR, the /tasks endpoint would return tasks assigned to the calling user, with some additional limited filtering options. This version of the endpoint is not currently used by the UI and as such we can make breaking changes safely

ThisPR redefines the /tasks endpoint to return tasks for any user, with an optional ‘allocatedUserToId’ to replicate the previous /tasks endpoint behaviour. The end point also provides paging and various other filters.

This allows us to deprecate /tasks/reallocatable and /tasks/{taskType} as the same functionality is now provided by the generic /tasks endpoint

https://dsdmoj.atlassian.net/browse/APS-283 and https://dsdmoj.atlassian.net/browse/APS-284 have been created to subsequently remove usages of the deprecated endpoints on the UI, and then remove the deprecated endpoints themselves


pezholio (approved):
This is some great work untangling this controller - thanks! 👍

PR #1448, opened 2024-02-05
More verbose error logging for notify

pezholio:
This adds some more verbose error logging for Notify, so we can confirm email sending will work as expected in production.


davidatkinsuk (approved):
LGTM, will definitely make support and testing easier!

PR #1449, opened 2024-02-05
(APS-282) Add APArea to `UserWithWorkload`

pezholio:
This is needed to filter by AP Area. Ordinarily, we’d extend the `ApprovedPremisesUser` here, but the code generator doesn’t like extensions of an already extended type.


davidatkinsuk (approved):
LGTM 👍

PR #1450, opened 2024-02-05
Update emergency region config to have dashes

pezholio:
Any underscores will be stripped out of the config env vars, so `NORTH_WEST` will be normalised to `northwest`. Replacing the dahses with underscores should fix this


pezholio:
@danlivings-dxw Good call!

PR #1452, opened 2024-02-05
Send notifications to assessor on app withdrawal

davidatkinsuk:
If there is an ‘active’ assessment when an application is withdrawn, a notification will be sent to the person currently allocated to the assessment.

A temporary feature flag has been added to allow us to control when this feature goes live in production. A JIRA ticket https://dsdmoj.atlassian.net/browse/APS-285 has been created to track its removal.


pezholio (approved):
LGTM 👍

PR #1455, opened 2024-02-06
Return Status Updates on CAS2 Submitted Applications

Harriethw:
We'd like referrers to see status updates on an
application. These are only applied to
applications after they've been submitted.

This adds the `statusUpdates` via the ApplicationsTransformer.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/transformer/cas2/ApplicationsTransformerTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/transformer/cas2/ApplicationsTransformerTest.kt
@@ -11,13 +11,16 @@ import org.junit.jupiter.api.BeforeEach
 import org.junit.jupiter.api.Nested
 import org.junit.jupiter.api.Test
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ApplicationStatus
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Cas2StatusUpdate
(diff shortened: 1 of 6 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/transformer/cas2/ApplicationsTransformerTest.kt
```diff
@@ -85,6 +91,33 @@ class ApplicationsTransformerTest {
...
+      val application = submittedCas2ApplicationFactory.withStatusUpdates(
+        mutableListOf(
+          statusUpdateEntity,
+        ),
+      )
+        .produce()
+
+      val result = applicationsTransformer.transformJpaToApi(application, mockk())
+
+      assertThat(result.id).isEqualTo(application.id)
```
tahb, line 116: A minor idea that we could ignore for now: if this test is all about status updates I don't think we get value by repeating assertions covered elsewhere. There's a cost (that is currently small) where should we need to change those assertions later we'd need to involve ourselves in potentially changing many other tests.

tahb (approved):
LGTM 👍

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/transformer/cas2/ApplicationsTransformerTest.kt
Harriethw: makes sense. as I had to rebase anyway I went in and removed check for `status` and `telephoneNumber`

PR #1456, opened 2024-02-06
Fix placement request withdrawable logic bug

davidatkinsuk:
If a placement request has a booking associated with it, but that booking has been cancelled, as per existing behaviour, it will have a status of ‘not matched’. Therefore, a placement request with this state should also be withdrawable.

As part of this commit some logic defined in the PlacementRequestTransformer has been moved into a common entity function to reduce duplication

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PlacementApplicationEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PlacementApplicationEntity.kt
@@ -1,8 +1,6 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity
 
 import org.hibernate.annotations.Type
-import org.springframework.data.domain.Page
-import org.springframework.data.domain.Pageable
 import org.springframework.data.jpa.repository.JpaRepository
 import org.springframework.data.jpa.repository.Query
 import org.springframework.stereotype.Repository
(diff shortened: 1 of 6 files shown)


pezholio (approved):
LGTM - I'm all for getting one source of truth for this stuff 👍

edavey:
Nice fix.

Could we be a little more descriptive with the commit title?

e.g. 

Title: "Prevent placement requests with active bookings from being 'withdrawable'" ?

In the commit message you mention:

>   some logic defined in the PlacementRequestTransformer has been moved into a common entity function to reduce duplication

I'm not seeing any change like that in the `PlacementRequestTransformer`. Probably I'm misunderstanding? or maybe that part of the commit message is no longer intended?

davidatkinsuk:
> Nice fix.
> 
> Could we be a little more descriptive with the commit title?
> 
> e.g.
> 
> Title: "Prevent placement requests with active bookings from being 'withdrawable'" ?
> 
> In the commit message you mention:
> 
> > some logic defined in the PlacementRequestTransformer has been moved into a common entity function to reduce duplication
> 
> I'm not seeing any change like that in the `PlacementRequestTransformer`. Probably I'm misunderstanding? or maybe that part of the commit message is no longer intended?

Hi Ed - sorry this was on auto merge so it merged before I saw this. I'll agree, commit title should of been clearer, i'll look to improve on that going forward.

The common logic is creating the hasActiveBooking() on the entity and using that in the transformer instead of some duplicated predicates. it was a pretty small de-duplication tbf!

PR #1458, opened 2024-02-07
Added missing release types to PR transformer

davidatkinsuk:
New application release types were added to the open api in commit 54c6fdb, but the placement request transformer was not updated to include these new types. These leads to RuntimeExceptions when attempting to view a placement request task for an application for one of these new release types.

This commit adds those missing mappings and improves regression testing to ensure all release types are mapped going forward

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/transformer/PlacementRequestTransformerTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/transformer/PlacementRequestTransformerTest.kt
@@ -5,6 +5,8 @@ import io.mockk.mockk
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.jupiter.api.BeforeEach
 import org.junit.jupiter.api.Test
+import org.junit.jupiter.params.ParameterizedTest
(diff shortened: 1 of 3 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/transformer/PlacementRequestTransformerTest.kt
```diff
@@ -322,4 +324,41 @@ class PlacementRequestTransformerTest {
       ),
     )
   }
+
+  @ParameterizedTest
+  @EnumSource(value = ReleaseTypeOption::class)
```
edavey, line 329: so test scenario loops for every release type, right? Nice.

edavey (approved):
thks 👍

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/transformer/PlacementRequestTransformerTest.kt
davidatkinsuk: correct. i'm not sure why we're not enumerating these values but suspect that could open a can of worms so this at least gives us safety going forward.

PR #1461, opened 2024-02-07
Add ‘application’ type to /withdrawables response

davidatkinsuk:
This relates to https://dsdmoj.atlassian.net/browse/APS-290

An additional type has been added to the list of types that can be returned by the /withdrawables response. This is used to indicate if the application as a whole can be withdrawn by the calling user.

This commit includes a new extension function for WebTestClient, adding a method that will automatically marshal an object in json before asserting it against the response. This will help reduce duplicate code.

A subsequent PR will be created to modify the logic used to determine if a user can withdraw an application, allowing users with the role 'A CRU Manager can withdraw an application once it has been submitted' to withdraw applications.


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ApplicationTest.kt
```diff
@@ -3408,61 +3439,64 @@ class ApplicationTest : IntegrationTestBase() {
...
-                WithdrawableType.placementRequest,
-                listOf(datePeriodForDuration(placementRequest.expectedArrival, placementRequest.duration)),
-              ),
-            )
+          val expected = listOf(
+            Withdrawable(
+              application.id,
+              WithdrawableType.application,
+              emptyList(),
+            ),
```
edavey, line 3478: Do you agree that this PR could be easier to review (i.e. easier for the reviewer to spot issues / misunderstandings) if the change to behaviour were isolated in a single commit, separate from the various helpful refactorings? e.g.

- preliminary commit to fixup existing formatting
- commit to reformat `integration/ApplicationTest.kt`, whitespace, renaming block variables (`user` becomes `allocatedTo`)
- commit to introduce `jsonForObject` helper
- commit to change behaviour of `/withdrawables` endpoint to include new `application` type

edavey (approved):
This looks good to me, though would be clearer if the change to behaviour were contained in a commit apart from other refactorings.

I think it would also be good if the commit message contained a little more context on why this `application` type is being added to the list of things which can be withdrawn. 

At present, a whole application can already be withdrawn, so perhaps this change is designed to reduce the logic on the front-end and model concepts more consistently? Perhaps it's also a preliminary to improving the process of 'cascading' withdrawals? Whereby associated entities are also withdrawn? Just guessing!

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ApplicationTest.kt
davidatkinsuk: i've moved the jsonForObject change to its own commit as I agree it shouldn't be mixed in with the new functionality

As for formatting, alot of the whitespace change was introduced because i had to add an additional 'Given a User' block into the test for the change. If we moved away from using lots of embedded code blocks that would massively reduce noise for changes like this, but most of the code is done this way unfortunately

As for variable name change from 'user' to 'allocatedTo', i agree it probably should be on its own commit but i feel like small variable names should be ok in 'unrelated' commits as the change is pretty clear? I'll try to do such changes on a separate commit going forward.

davidatkinsuk:
> I think it would also be good if the commit message contained a little more context on why this `application` type is being added to the list of things which can be withdrawn.

Good idea, i've updated the commit to explain the purpose behind the change

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ApplicationTest.kt
edavey: 👍 thks David, I think I failed to appreciate that some of the whitespace/indentation changes were related to that additional Given block! apologies 🙇

PR #1464, opened 2024-02-07
Allow CAS1 Workflow Manager to withdraw apps

davidatkinsuk:
Relates to https://dsdmoj.atlassian.net/browse/APS-290

If an application is submitted, a workflow manager should be able to withdraw it. This commit updates the application withdrawal logic which will be automatically used when returning a list of withdrawables to the UI

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApplicationEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApplicationEntity.kt
@@ -350,6 +350,7 @@ class ApprovedPremisesApplicationEntity(
 
   fun getLatestPlacementRequest(): PlacementRequestEntity? = this.placementRequests.maxByOrNull { it.createdAt }
   fun getLatestBooking(): BookingEntity? = getLatestPlacementRequest()?.booking
+  fun isSubmitted() = submittedAt != null
 }
 
 @Repository
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ApplicationTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ApplicationTest.kt
@@ -3360,7 +3360,12 @@ class ApplicationTest : IntegrationTestBase() {
               withBooking(bookingWithArrival)
             }
 
-            val expected = listOf(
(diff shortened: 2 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApplicationEntity.kt
```diff
@@ -350,6 +350,7 @@ class ApprovedPremisesApplicationEntity(
 
   fun getLatestPlacementRequest(): PlacementRequestEntity? = this.placementRequests.maxByOrNull { it.createdAt }
   fun getLatestBooking(): BookingEntity? = getLatestPlacementRequest()?.booking
+  fun isSubmitted() = submittedAt != null
```
edavey, line 353: interesting to see how this is done, very neat.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ApplicationTest.kt
```diff
@@ -3360,7 +3360,12 @@ class ApplicationTest : IntegrationTestBase() {
               withBooking(bookingWithArrival)
             }
 
-            val expected = listOf(
+            val expected = listOfNotNull(
```
edavey, line 3363: interesting, hadn't come across `listOfNotNull()`

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/UserAccessServiceTest.kt
```diff
@@ -1495,7 +1497,7 @@ class UserAccessServiceTest {
   }
 
   @Nested
-  inner class UserCanWithdrawApplication {
+  inner class UserCanWithdrawAdpplication {
```
edavey, line 1500: I think you've introduced a tiny typo here

edavey (approved):
very clear. Moving the auth control to the `UserAccessService` looks like an excellent refactoring.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/UserAccessServiceTest.kt
davidatkinsuk: oops! good spot. Fixed.

PR #1465, opened 2024-02-07
Placement Requests withdrawals should cascade to Bookings

davidatkinsuk:
This relates to the second bullet point on https://dsdmoj.atlassian.net/browse/APS-292

Most of the work on this PR is changing the placement request and booking withdrawal/cancellation endpoints to be idempotent to ensure that when withdrawals are cascaded (particularly via the Withdrawal service), we don't end up withdrawing the same entity multiple times (which would lead to multiple notifications)

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/PlacementRequestService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/PlacementRequestService.kt
@@ -287,6 +287,10 @@ class PlacementRequestService(
     val placementRequest = placementRequestRepository.findByIdOrNull(placementRequestId)
       ?: return AuthorisableActionResult.NotFound("PlacementRequest", placementRequestId.toString())
 
+    if(placementRequest.isWithdrawn) {
+      return AuthorisableActionResult.Success(Unit)
+    }
+
     if (!user.hasRole(UserRole.CAS1_WORKFLOW_MANAGER) && placementRequest.application.createdByUser != user) {
       return AuthorisableActionResult.Unauthorised()
     }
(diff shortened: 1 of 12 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/PlacementRequestService.kt
```diff
@@ -296,6 +308,25 @@ class PlacementRequestService(
 
     placementRequestRepository.save(placementRequest)
 
+    placementRequest.booking?.let { booking ->
+      val bookingCancellationResult = bookingService.createCancellation(
+        user,
+        booking,
+        LocalDate.now(),
+        CAS1_WITHDRAWN_BY_PP_ID,
+        "Automatically withdrawn as placement request was withdrawn",
```
edavey, line 317: do you think it would be useful from an auditing and debugging perspective to include the ID of the "initiating" entity here eg.

> Automatically withdrawn as placement request abc123 was withdrawn

edavey (approved):
👍 really clear, and the approach of making the withdrawals idempotent feels neat and appropriate.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/PlacementRequestService.kt
davidatkinsuk: I agree it would be good to track but capturing here wouldn't fully solve the issue because we have no free text fields on Placement Application and Placement Requests where it would also be good to capture similar information. I think that maybe is a nice to have feature (along with withdrawal date as you suggested on the other PR, which could be used to 'group' withdrawals i guess if we used the exact same snapshot)

PR #1466, opened 2024-02-07
[CAS-50]-Automatically close CAS3 assessments after booking is made

sudhakar-plr:
# Changes in this PR
Refer [CAS-50](https://dsdmoj.atlassian.net/jira/software/c/projects/CAS/boards/1331?selectedIssue=CAS-50) for more information.

This PR is to automatically close the CAS3 assessments after the booking is made successfully.

Business wants to close CAS3 assessment automatically after booking is done, So that a referral is no longer appears in the `Ready to Place` queue to avoid confusion or duplicate booking. We have separate story(cas-176) to reinstate assessment when the booking is cancelled.

The change is in `BookingService` to call an existing `AssessmentService.closeAssessment()` method to close the assessment automatically as part of booking creation.

Additionally, our requirement is not to fail the booking if the closure of assessment is failed for some reason, instead we complete the existing booking process and raise an sentry alert with the reason of failure for further investigation by support team. Alternatively, the user continue to use existing process of closing the assessment from UI.

# Release checklist

confirm:

## Pre merge checklist

    advance?

## Post merge checklist

considering it done.

fix it as soon as possible.

[CAS-50]: https://dsdmoj.atlassian.net/browse/CAS-50?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
@@ -1,6 +1,7 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.service
 
 import arrow.core.Either
+import io.sentry.Sentry
 import org.slf4j.LoggerFactory
 import org.springframework.beans.factory.annotation.Value
 import org.springframework.data.repository.findByIdOrNull
@@ -74,7 +75,11 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserRole
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.DomainEvent
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.PersonInfoResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.validated
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.BadRequestProblem
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.ConflictProblem
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.ForbiddenProblem
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.InternalServerErrorProblem
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.NotFoundProblem
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.results.AuthorisableActionResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.results.ValidatableActionResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.getDaysUntilInclusive
@@ -126,6 +131,7 @@ class BookingService(
   @Value("\${arrived-departed-domain-events-disabled}") private val arrivedAndDepartedDomainEventsDisabled: Boolean,
   private val userService: UserService,
   private val userAccessService: UserAccessService,
+  private val assessmentService: AssessmentService,
 ) {
   val approvedPremisesBookingAppealedCancellationReasonId: UUID =
     UUID.fromString("acba3547-ab22-442d-acec-2652e49895f2")
@@ -801,6 +807,10 @@ class BookingService(
       booking.turnarounds += turnaround
 
       cas3DomainEventService.saveBookingProvisionallyMadeEvent(booking)
+
+      if (assessmentId != null) {
+        closeTransitionalAccommodationAssessment(assessmentId, user, booking)
+      }
 
       success(booking)
     }
@@ -1805,6 +1815,31 @@ class BookingService(
     val submittedAt = application?.submittedAt ?: offlineApplication?.createdAt as OffsetDateTime
 
     return Triple(applicationId, eventNumber, submittedAt)
+  }
+
+  @SuppressWarnings("ThrowsCount", "TooGenericExceptionCaught")
+  private fun closeTransitionalAccommodationAssessment(
+    assessmentId: UUID,
+    user: UserEntity,
+    booking: BookingEntity,
+  ) {
+    try {
+      val assessmentValidationResult = when (val assessmentAuthResult = assessmentService.closeAssessment(user, assessmentId)) {
+        is AuthorisableActionResult.Success -> assessmentAuthResult.entity
+        is AuthorisableActionResult.NotFound -> throw NotFoundProblem(assessmentId, "Assessment")
+        is AuthorisableActionResult.Unauthorised -> throw ForbiddenProblem()
+      }
+
+      when (assessmentValidationResult) {
+        is ValidatableActionResult.GeneralValidationError -> throw BadRequestProblem(errorDetail = assessmentValidationResult.message)
+        is ValidatableActionResult.FieldValidationError -> throw BadRequestProblem(invalidParams = assessmentValidationResult.validationMessages)
+        is ValidatableActionResult.ConflictError -> throw ConflictProblem(id = assessmentValidationResult.conflictingEntityId, conflictReason = assessmentValidationResult.message)
+        is ValidatableActionResult.Success -> Unit
+      }
+    } catch (exception: Exception) {
+      log.error("Unable to close CAS3 assessment $assessmentId for booking ${booking.id} ", exception)
+      Sentry.captureException(RuntimeException("Unable to close CAS3 assessment $assessmentId for booking ${booking.id} ", exception))
+    }
   }
 }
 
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
```diff
@@ -1794,6 +1804,31 @@ class BookingService(
...
+        is AuthorisableActionResult.Unauthorised -> throw ForbiddenProblem()
+      }
+
+      when (assessmentValidationResult) {
+        is ValidatableActionResult.GeneralValidationError -> throw BadRequestProblem(errorDetail = assessmentValidationResult.message)
+        is ValidatableActionResult.FieldValidationError -> throw BadRequestProblem(invalidParams = assessmentValidationResult.validationMessages)
+        is ValidatableActionResult.ConflictError -> throw ConflictProblem(id = assessmentValidationResult.conflictingEntityId, conflictReason = assessmentValidationResult.message)
+        is ValidatableActionResult.Success -> Unit
+      }
+    } catch (exception: Exception) {
```
davidatkinsuk, line 1827: would changing this to catch ThrowableProblem work? (trying to get rid of the TooGenericExceptionCaught warning)

I don't think we should be capturing RuntimeExceptions (e.g. database issues), IMO these should really propagate back to the caller because if they retry it _should_ work (instead of leaving the assessment in an unexpected state)

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/BookingTest.kt
```diff
@@ -914,6 +916,97 @@ class BookingTest : IntegrationTestBase() {
           .jsonPath("$.bed.id").isEqualTo(bed.id.toString())
           .jsonPath("$.bed.name").isEqualTo("test-bed")
           .jsonPath("$.assessmentId").isEqualTo("${assessment.id}")
+
+        // Assert the assessment is closed
```
davidatkinsuk, line 920: i don't think this comment is required? if it's not clear without the comment we could chuck in a private function like.   isClosed(assessment): Boolean?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/BookingTest.kt
```diff
@@ -1013,6 +1107,10 @@ class BookingTest : IntegrationTestBase() {
...
+        val temporaryAccommodationAssessmentEntity = temporaryAccommodationAssessmentRepository.findByIdOrNull(assessment.id)
```
davidatkinsuk, line 1112: pls. see other comment regarding //

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/BookingTest.kt
```diff
@@ -914,6 +916,97 @@ class BookingTest : IntegrationTestBase() {
...
+        }
+
+        val application = temporaryAccommodationApplicationEntityFactory.produceAndPersist {
+          withCreatedByUser(userEntity)
+          withCrn(offenderDetails.otherIds.crn)
+          withProbationRegion(userEntity.probationRegion)
+          withApplicationSchema(applicationSchema)
+        }
+
+        val assessmentSchema = temporaryAccommodationAssessmentJsonSchemaEntityFactory.produceAndPersist {
```
davidatkinsuk, line 959: for my understanding - i guess this schema is invalid which leads to closure failing. If that's correct could the variable please be called 'invalidAssessmentSchema'?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/BookingServiceTest.kt
```diff
@@ -4302,6 +4328,486 @@ class BookingServiceTest {
...
+      verify(exactly = 1) {
+        mockAssessmentService.closeAssessment(user, application.id)
+      }
+      verify(exactly = 1) {
+        Sentry.captureException(any())
+      }
+    }
+
+    @Test
+    fun `createTemporaryAccommodationBooking saves Booking and creates domain event and log error when closing assessment failed with runtime exception`() {
```
davidatkinsuk, line 4545: as mentioned in other comment, i don't think we should stop runtime exceptions

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/BookingServiceTest.kt
```diff
@@ -4302,6 +4328,486 @@ class BookingServiceTest {
...
+        entry("$.assessmentId", "doesNotExist"),
+      )
+
+      verify(exactly = 0) {
+        mockAssessmentService.closeAssessment(user, assessmentId)
+      }
+    }
+
+    @Test
+    fun `createTemporaryAccommodationBooking fail to save booking and not closing assessment when saving booking triggered Database exception`() {
```
davidatkinsuk, line 4713: is there anyway to reduce duplication in these test methods (e.g. move the app/assessment creation into a common function?)

davidatkinsuk (changes requested):
Hi. A few comments from me, always fun to see how other CAS' use common entities :)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
sudhakar-plr: As i mentioned in the PR comments, we dont want to fail the booking when if we cant close the assessment and that the reason we capture all exception and log the response.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/BookingServiceTest.kt
sudhakar-plr: Its a requirement to not block the booking flow while cancelling the assessment.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/BookingTest.kt
sudhakar-plr: Its a valid schema only. But in the service layer we check the assessment.schema_id is equal to the as latest schema definition. The Date is used to find the latest schema id, if latest schema is not same as the assessment.schema_id then we get validation error. 

So inorder to test the validation error, if i create the schema without 'addedDate' then the schema-id comparison fails at service layer.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/BookingTest.kt
sudhakar-plr: done

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/BookingTest.kt
sudhakar-plr: done

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
davidatkinsuk: I agree with the end goal, but I don't think we should be capturing high level exceptions (like Exception or RuntimeException) because this will capturing things we probably should let propagate up (e.g. database errors, coding errors etc.). I think it's fine to capture something more specific e.g. ThrowableProblem and we can then also remove the Detekt suppression. I'll leave conversation unresolved by give approval for merge, up to you if you want to change this 👍

davidatkinsuk (approved):
One optional pending comment, but happy to approve 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
sudhakar-plr: I don't think we should propagate the error to high level in this case because:

1) Error while closing the assessment is not fatal error which we need to propagate and fail the transaction. The code is catching the exception and it knows what to do with it. You need to propagate to next level only when you dont know what to do with it.

2) We anyway catch the exception and raise alert if we get unexpected error for support team to investigate the issue.So we are raising  exception for investigation and not swolling it.

3) The main requirement is to try close the assessment automatically ,if we cant close it automatically then dont fail the booking. We have alternative option flow for the user to manually close the assessment .

PR #1467, opened 2024-02-08
Placement requests with bookings can be withdrawn

davidatkinsuk:
Relates to bullet point 1 on https://dsdmoj.atlassian.net/browse/APS-292

A placement request with a booking can be withdrawn, and with prioir changes this will automatically cascade to bookings.


edavey (approved):
Looks, good, the plan comes together!

PR #1468, opened 2024-02-08
Add isWithdrawn to PlacementRequest and PlacementApplication model

davidatkinsuk:
Relates to https://dsdmoj.atlassian.net/browse/APS-311


edavey (approved):
Thanks David, v clear.

Do you think that it would be a good idea (in the future!) to have each "withdrawable" entity record a `withdrawn_at` timestamp? 

That might make life easy and consistent when asking whether the thing `isWithdrawn()` and for use in Domain Events (I imagine that Timeline events will need to say something like "Placement request withdrawn by John Smith on 29 Feb 2024")

davidatkinsuk:
> Do you think that it would be a good idea (in the future!) to have each "withdrawable" entity record a `withdrawn_at` timestamp?
> 
> That might make life easy and consistent when asking whether the thing `isWithdrawn()` and for use in Domain Events (I imagine that Timeline events will need to say something like "Placement request withdrawn by John Smith on 29 Feb 2024")

Yes, i agree it would be good to capture this from the outset to support the timeline in the future.  I'll create a Story to do this but mark as low priority. Would be quite quick to add though.

PR #1469, opened 2024-02-08
Applicants shouldn't be able to withdraw bookings

davidatkinsuk:
Note - A majority of the change to stop applicants being able to remove bookings was already implemented via commit 0f15943e. But, additional changes were required for APS-294 specifically:

* Removing some permissions removal not captured by 0f15943e
* Removing some complexity in withdrawables service that is no longer required because withdrawals cascade from placement requests to bookings (i.e. an applicant withdrawing a placement request or 'parent' entity is sufficient for it to cascade down to a booking)
* Updating integration tests to ensure cascades happen from placement request to booking for applicants

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
@@ -1192,8 +1192,10 @@ class BookingService(
     else -> throw RuntimeException("Unknown premises type ${booking.premises::class.qualifiedName}")
   }
 
-  fun getCancelleableBookings(application: ApprovedPremisesApplicationEntity): List<BookingEntity> =
-    bookingRepository.findAllByApplication(application).filter { isCancellableCas1(it) }
+  fun getCancelleableCas1Bookings(user: UserEntity, application: ApprovedPremisesApplicationEntity): List<BookingEntity> =
+    bookingRepository.findAllByApplication(application).filter { booking ->
+      isCancellableCas1(booking) && userAccessService.userCanCancelBooking(user, booking)
+    }
 
   private fun isCancellableCas1(booking: BookingEntity) = (booking.cancellation == null) && (booking.arrivals.isEmpty())
 
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
@@ -1192,12 +1192,13 @@ class BookingService(
     else -> throw RuntimeException("Unknown premises type ${booking.premises::class.qualifiedName}")
   }
 
-  fun getCancelleableCas1Bookings(user: UserEntity, application: ApprovedPremisesApplicationEntity): List<BookingEntity> =
+  fun getCancelleableCas1BookingsForUser(user: UserEntity, application: ApprovedPremisesApplicationEntity): List<BookingEntity> =
     bookingRepository.findAllByApplication(application).filter { booking ->
-      isCancellableCas1(booking) && userAccessService.userCanCancelBooking(user, booking)
+      isInCancellableStateCas1(booking) && userAccessService.userCanCancelBooking(user, booking)
     }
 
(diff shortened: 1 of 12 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
```diff
@@ -1192,10 +1192,13 @@ class BookingService(
...
-  fun getCancelleableBookings(application: ApprovedPremisesApplicationEntity): List<BookingEntity> =
-    bookingRepository.findAllByApplication(application).filter { isCancellableCas1(it) }
+  fun getCancelleableCas1BookingsForUser(user: UserEntity, application: ApprovedPremisesApplicationEntity): List<BookingEntity> =
+    bookingRepository.findAllByApplication(application).filter { booking ->
+      isInCancellableStateCas1(booking) && userAccessService.userCanCancelBooking(user, booking)
+    }
 
-  private fun isCancellableCas1(booking: BookingEntity) = (booking.cancellation == null) && (booking.arrivals.isEmpty())
+  private fun isInCancellableStateCas1(booking: BookingEntity)
+    = (booking.cancellation == null) && (booking.arrivals.isEmpty())
```
edavey, line 1201: Not really part of this piece of work, but are we satisfied that "having no arrivals" is a sufficient guard against the possibility of cancelling a booking which has in fact been taken up?  Do we need to understand and document the process for restoring a booking which is in fact either i) currently under occupation or ii) was used in the past (e.g. in a ROTL plan)?

If we were to add an additional control to only permit cancellations of bookings which are in the future, I guess this is likely to mean than we would be unable to clean up crufty placements which ought to have been cancelled earlier but for "reasons" have remained uncancelled beyond their arrival date?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ApplicationTest.kt
```diff
@@ -3068,20 +3073,23 @@ class ApplicationTest : IntegrationTestBase() {
...
+          val updatedPlacementRequest2 = placementRequestRepository.findByIdOrNull(placementRequest2.id)!!
+          assertThat(updatedPlacementRequest2.isWithdrawn).isEqualTo(true)
+          assertThat(updatedPlacementRequest2.withdrawalReason).isEqualTo(PlacementRequestWithdrawalReason.WITHDRAWN_BY_PP)
+
+          val updatedBooking1 = bookingRepository.findByIdOrNull(booking1NoArrival.id)!!
+          assertThat(updatedBooking1.isCancelled).isTrue()
+          assertThat(updatedBooking1.cancellation!!.reason.name).isEqualTo("The probation practitioner requested it")
+
+          val updatedBooking2WithArrival = bookingRepository.findByIdOrNull(booking2HasArrival.id)!!
+          assertThat(updatedBooking2WithArrival.isCancelled).isFalse()
```
edavey, line 3092: do you think it's worth showing here that an `updatedBooking1NoArrival ` has not been cancelled/withdrawn?

edavey (approved):
Very clear (for such a complex refactoring!)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
davidatkinsuk: "Do we need to understand and document the process for restoring a booking which is in fact either i) currently under occupation or ii) was used in the past (e.g. in a ROTL plan)?"

I'm not aware of ROTL plans so can't answer to that, but i do think we could have a case where an arrival isn't recorded (e..g they just forget) and in the meantime the booking is cancelled. In which case you may need the ability to 'restore' a booking. Maybe this is all edge case stuff that is out of scope for the current solution? Adam could probably address it best i'd imagine

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ApplicationTest.kt
davidatkinsuk: Yes, absolutely agree. I actually have a PR ready to go with lots of tests for cascading that checks this and much more, so I think that'll cover it.

PR #1470, opened 2024-02-08
Allow Nomis users to add a Note to CAS2 Applications

Harriethw:
We would like CAS2 users to add notes to submitted applications. 

We are introducing this in parts - initially just enabling the ability for Nomis users to create notes on any applications they have submitted.

Screenshot 2024-02-09 at 15 48 09

In this PR we have

* Introduced `cas2_application_notes` new table, with foreign keys to `cas2_applications` and `nomis_users`. 
* Introduced new endpoint `submissions/{applicationId}/notes` restricted to users with POM role
* Added Service layer logic that restricts Note creation to submitted apps, and only if user created application

This is following a similar pattern in CAS1, but as our users and application models are different we've continued to use our own namespace/new tables. 

We will introduce the functionality to return Notes on a SubmittedApplication shortly. 

We will also very shortly be introducing the ability for External Users to create notes on any submitted application, which means some of the logic will need to change.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/SubmissionsController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/SubmissionsController.kt
@@ -5,12 +5,15 @@ import org.springframework.http.ResponseEntity
 import org.springframework.stereotype.Service
 import org.zalando.problem.AbstractThrowableProblem
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.cas2.SubmissionsCas2Delegate
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Cas2ApplicationNote
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Cas2ApplicationStatusUpdate
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Cas2SubmittedApplication
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Cas2SubmittedApplicationSummary
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.NewCas2ApplicationNote
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.SortDirection
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.SubmitCas2Application
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas2ApplicationEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas2ApplicationNoteEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas2ApplicationSummary
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas2StatusUpdateEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.BadRequestProblem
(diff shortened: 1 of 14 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/SubmissionsController.kt
```diff
@@ -120,9 +129,23 @@ class SubmissionsController(
...
+      .created(URI.create("/cas2/applications/$applicationId/notes/${note.id}"))
+      .body(
+        applicationNotesTransformer.transformJpaToApi(note),
+      )
+  }
+
+  private fun <EntityType> processAuthorisationFor(
     applicationId: UUID,
-    result: AuthorisableActionResult<ValidatableActionResult<Cas2StatusUpdateEntity>>,
+    result: AuthorisableActionResult<ValidatableActionResult<EntityType>>,
```
danhumphreys-moj, line 148: Presumably this change is required since `processAuthorisationFor` was previously only used when an application status was updated?  Therefore since now we are using it for adding an application note we need to refer to a common `EntityType` superclass?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/SubmissionsController.kt
```diff
@@ -131,12 +154,12 @@ class SubmissionsController(
     }
   }
 
-  private fun processValidation(validationResult: ValidatableActionResult<Cas2StatusUpdateEntity>) {
+  private fun <EntityType : Any> processValidation(validationResult: ValidatableActionResult<EntityType>): Any {
```
danhumphreys-moj, line 157: The same reason as the previous comment?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas2/ApplicationNoteServiceTest.kt
```diff
@@ -0,0 +1,157 @@
...
+        val createdNote = validatableActionResult.entity
+
+        Assertions.assertThat(createdNote).isEqualTo(noteEntity)
+      }
+    }
+
+    @Nested
+    inner class WhenUnsuccessful {
+      @Test
+      fun `returns Not Found when application not found`() {
```
danhumphreys-moj, line 90: I would imagine there is a lot of duplicated code for some of these tests, they must be quite common across all the CAS2 use cases.

danhumphreys-moj (approved):
Hi Harriet.  I have made some minor comments.  Based on what I understand this PR to do I'm happy to approve.  However you will no doubt want someone else to check too since I have only been here a few days.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas2/ApplicationNotesTransformer.kt
```diff
@@ -0,0 +1,23 @@
...
+class ApplicationNotesTransformer {
+
+  fun transformJpaToApi(
+    jpa: Cas2ApplicationNoteEntity,
+  ):
+    Cas2ApplicationNote {
+    return Cas2ApplicationNote(
+      id = jpa.id,
+      name = jpa.createdByNomisUser.name,
+      username = jpa.createdByNomisUser.nomisUsername,
```
tahb, line 17: Do we need to return the `nomisUsername` and `email` here? I'm not sure the frontend needs it for rendering the timeline, it just needs a name right? 

I also ask because that nomis username in particular is surely going to be a difference between nomis and external user objects. If we don't need to concern ourselves with it, I think we keep it simple and ignore it.

tahb (approved):
I wondered if this would be a good scenario for using polymorphism. 

Instead of explicitly adding and referencing each type of user, could we abstract it with `author` or `user`? This, I think, could give us a couple of benefits:

1. We could make a single database change here that supported by sides to this work (we could also make the service suitably generic instead of creating comments slightly differently for each type)
1. We could be open to adding new types of user in future (maybe notes from OpenHousing via the speculated integration or another user group)

I'm not super confident how we'd do it, just recognising a pattern, maybe something like this:

```
@Entity
class NomisUserEntity(
…
) : UserEntity()

@Entity
class ExternalUserEntity(
…
) : UserEntity()
```

And then:

```
    @ManyToOne
    @JoinColumn(name = "created_by_user_id")
    val createdByUser: UserEntity,
 ```
 
 I don't think it's essential but I do think it's worth taking a few minutes now to consider if this might be a better approach.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas2/ApplicationNotesTransformer.kt
Harriethw: External Users also have a `username` field, so they could be mapped to the same shape of data. But I think you're right that we're not using it on the frontend, so happy to take it out.

PR #1471, opened 2024-02-09
Placement Application Withdrawal Changes

davidatkinsuk:
Relates to [APS-293](https://dsdmoj.atlassian.net/browse/APS-293)

I've made the changes specified in the associated ticket, and at the same time i've made the withdrawal logic work in the same way for all 4 entity types, making things more predictable and, hopefully, easier to change. Specifically:
 
 - add a method to determine if entity is withdrawble for the user to all four entity services
 - calling code is now responsible for checking if user has permissions to withdraw (this required for cascading where this check is to be skipped)
 - withdrawing is always idempotent
 - withdrawing will cascade down to child types and log errors if those cascading withdrawals unexpectedly fail
   
Because of the cascading, withdrawals are now effectively depth-first traversing a tree made up of applications, placement applications, placement requests and bookings.

[APS-293]: https://dsdmoj.atlassian.net/browse/APS-293?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PlacementRequestsController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PlacementRequestsController.kt
@@ -33,6 +33,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.results.ValidatableActio
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.BookingService
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.OffenderService
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.PlacementRequestService
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.UserAccessService
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.UserService
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.BookingNotMadeTransformer
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.NewPlacementRequestBookingConfirmationTransformer
@@ -179,6 +180,7 @@ class PlacementRequestsController(
 
   override fun placementRequestsIdWithdrawalPost(id: UUID, body: WithdrawPlacementRequest?): ResponseEntity<Unit> {
     val user = userService.getUserForRequest()
+
     val reason = when (body?.reason) {
       WithdrawPlacementRequestReason.duplicatePlacementRequest -> PlacementRequestWithdrawalReason.DUPLICATE_PLACEMENT_REQUEST
       WithdrawPlacementRequestReason.alternativeProvisionIdentified -> PlacementRequestWithdrawalReason.ALTERNATIVE_PROVISION_IDENTIFIED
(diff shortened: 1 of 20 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserAccessService.kt
```diff
@@ -192,13 +193,28 @@ class UserAccessService(
     else -> false
   }
 
+  /**
+   * This function only checks if the user has the correct permissions to withdraw the given application.
+   *
+   * It doesn't consider if the application is in a withdrawable state
+   */
   fun userCanWithdrawApplication(user: UserEntity, application: ApplicationEntity): Boolean = when (application) {
```
edavey, line 202: just a tiny suggestion: would `userMayWithdrawApplication()` help convey that this is about permission rather than ability?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PlacementRequestsController.kt
```diff
@@ -192,7 +194,11 @@ class PlacementRequestsController(
     }
 
     val result = extractEntityFromAuthorisableActionResult(
-      placementRequestService.withdrawPlacementRequest(id, user, reason),
+      placementRequestService.withdrawPlacementRequest(
+        id,
+        user,
+        reason,
+        checkUserPermissions = true),
```
edavey, line 201: I wonder if it would be clearer about when and why it's necessary to check the user's permission if this optional parameter were named something like `requestedByUser` or `cascadedWithdrawal`?

edavey (approved):
This looks correct to me though as the work itself is so complex I can't pretend to have a full understanding! Thanks for structuring this PR so carefully.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PlacementRequestsController.kt
davidatkinsuk: This was actually on my TODO list, great minds think alike! :) In a subsequent PR i've added a WithdrawalContext that is passed to all withdrawal functions:

`data class WithdrawalContext(
  val triggeringUser: UserEntity?,
  val triggeringEntityType: WithdrawableEntityType,
)

enum class WithdrawableEntityType {
  Application,
  PlacementRequest,
  PlacementApplication,
  Booking,
}
`

The 'triggeringEntityType' can be used to determine if this was a 'direct' request from a user or a cascade. This means we can then remove the 'checkUserPermissions' boolean and use that instead. I'll make that change now and queue it up for review once this is merged 👍

davidatkinsuk:
> This looks correct to me though as the work itself is so complex I can't pretend to have a full understanding! Thanks for structuring this PR so carefully.

Yes I appreciate there's a lot going on here. I'm hopeful that once i've got through all the tickets this code will be a lot easier to read and consistent. You'll see in a subsequent PR there's some very readable integration tests i've introduced for cascading that should help.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserAccessService.kt
davidatkinsuk: Good idea. I'll do this on one of my subsequent PRs that will follow this as I have a lot stacked up so changing older ones is a bit painful!

PR #1472, opened 2024-02-09
Add Status Details to More Information

Harriethw:
These are the confirmed details for the More Information status.

Copy source to compare is here:
https://docs.google.com/document/d/17m4erVi3dtOxnWgYj_CcFQqYlBsL92yn-sI1vhXo3so/edit?pli=1


tahb (approved):
LGTM 👍 I think it's helpful that this is in it's own PR given the lack of clarity around approval of these statuses. I do think we could have targetted this branch into main as well as it's parent though, that way keep the option of releasing them separately.

PR #1473, opened 2024-02-09
Refactor appeals to have a mandatory link to an assessment

danlivings-dxw:
This PR modifies the `appeals` table and corresponding JPA `AppealEntity` so that the reference to an assessment cannot be null.

Following this change, an appeal now requires an assessment to be provided when it is created. This is currently set to the latest available assessment on an application when the `POST /applications/{applicationId}/appeals` method is called; a `409 Conflict` response is returned if no assessment exists on the application.


src/main/resources/db/migration/all/20240209155543__make_assessment_id_on_appeals_table_not_nullable.sql
```diff
@@ -0,0 +1,7 @@
+UPDATE appeals
+SET assessment_id = assessments.id
+FROM applications
+JOIN assessments ON applications.id = assessments.application_id
+WHERE applications.id = appeals.application_id;
```
davidatkinsuk, line 5: is it possible for an application to have multiple assessments? If yes ... would this SQL fail in those cases (or just pick a random assessment)? Maybe we could make it more deterministic by choosing the newest assessment?

davidatkinsuk (changes requested):
Looks good, just one comment from me

edavey (approved):
thanks Dan 👍

davidatkinsuk (approved):
LGTM 👍

PR #1476, opened 2024-02-12
Add ability to assert on emails in int-tests

davidatkinsuk:
The application withdrawal tests have been updated to demonstrate asserting that an email has been sent


edavey (approved):
This looks excellent, thanks David. Documenting the email confirmations sent for the various withdrawal journeys is really important.

PR #1478, opened 2024-02-12
Improve withdrawal test coverage

davidatkinsuk:
In this PR we introduce a new WithdrawalTest class that includes all cascading withdrawal tests, some moved from other existing integration tests. Moving these into one class make it easier to reason on the behaviour and ensure it's consistent across all entity types.


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/WithdrawalTest.kt
```diff
@@ -0,0 +1,749 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.integration
```
edavey, line 1: Bravo, that `ApplicationTest` is a real monster!

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/WithdrawalTest.kt
```diff
@@ -546,54 +546,72 @@ class WithdrawalTest : IntegrationTestBase() {
...
+    assertThat(updatedPlacementRequest.withdrawalReason).isEqualTo(reason)
+  }
+
+  private fun assertBookingWithdrawn(booking: BookingEntity, cancellationReason: String) {
+    val updatedBooking = bookingRepository.findByIdOrNull(booking.id)!!
+    assertThat(updatedBooking.isCancelled).isTrue()
+    assertThat(updatedBooking.cancellation!!.reason.name).isEqualTo(cancellationReason)
+  }
+
+  private fun assertBookingNotWithdrawn(booking: BookingEntity) {
```
edavey, line 805: very nice, thks for hiding away all the refreshing of the persisted things

edavey (approved):
Very clear, thanks for introducing this new `WithdrawalTest`

PR #1484, opened 2024-02-13
Add unique UUID's for seeded statusDetails

Harriethw:
These were previously not unique and causing
problems. There have been no Status
Details created by the frontend yet so I don't
think there's a problem changing these ID's now.


libuk (approved):
Thanks Harriet!

PR #1486, opened 2024-02-14
CAS2 Timeline Events on an Application

Harriethw:
https://dsdmoj.atlassian.net/jira/software/c/projects/CAS2/boards/1347?selectedIssue=CAS2-213

In order to show several different types of events in a timeline component on the CAS2 UI, we are creating a standard shape for them on the presentation/transformation layer of the API. We currently need to show - notes, status updates,
and when the application was submitted.

Screenshot 2024-02-14 at 12 21 01

([copy is slightly different to screenshot above](https://mojdt.slack.com/archives/C04T6SC31HA/p1707926701170309?thread_ts=1707926423.047549&cid=C04T6SC31HA))

We will follow an existing pattern set by CAS1 for an Application return a list of `TimelineEvents` components, but as our data has a different shape, and as we are already namespacing CAS2 work, we are creating a new `Cas2TimelineEvent` component.

We then introduce a `TimelineEventsTransformer` to take the relevant data on an Application and transform it into the generic TimelineEvent needed.

Screenshot 2024-02-14 at 12 24 01

## Applications versus Submissions Controller

I've had to return the events on both the /applications and /submissions end points, even though these events will only appear on submitted applications currently. This is because we have a division between 'referrer' and 'assessor' journeys.
The `/submissions` endpoint is queried by the frontend to retrieve all submitted applications and is only viewable by CAS2_Admin and CAS2_Assessor user roles. 

The `/applications` endpoint is queried by the frontend to retrieve all applications regardless of whether they have been submitted, and is used currently by only POM user roles. The frontend doesn't distinguish when querying for a specific application whether it is submitted or not - it just gets `/applications/{applicationId}`. 

I feel like this could be tidied up a bit, but would need work on the frontend as well/a bit of architectural thinking. A 'submitted application' is currently just an `Application` class with a `submittedAt` date, but it could be we need to move to having what the other CAS's have - a concept of an `Assessment` as we are introducing assessing functionality?

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas2/TimelineEventsTransformer.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas2/TimelineEventsTransformer.kt
@@ -1,0 +1,57 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.cas2
+
+import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Cas2TimelineEvent
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.TimelineEventType
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas2ApplicationEntity
+
+@Component("Cas2TimelineEventsTransformer")
+class TimelineEventsTransformer {
+  fun transformApplicationToTimelineEvents(jpa: Cas2ApplicationEntity): List<Cas2TimelineEvent> {
+    val timelineEvents: MutableList<Cas2TimelineEvent> = mutableListOf()
+
+    addSubmittedEvent(jpa, timelineEvents)
+
+    addStatusUpdateEvents(jpa, timelineEvents)
+
+    addNoteEvents(jpa, timelineEvents)
+
+    return timelineEvents.sortedByDescending { it.occurredAt }
+  }
+
+  private fun addNoteEvents(jpa: Cas2ApplicationEntity, timelineEvents: MutableList<Cas2TimelineEvent>) {
+    jpa.notes?.forEach {
+      timelineEvents += Cas2TimelineEvent(
+        type = TimelineEventType.cas2Note,
+        occurredAt = it.createdAt.toInstant(),
+        label = "Note",
+        createdByName = it.createdByNomisUser.name,
+        body = it.body,
+      )
+    }
+  }
+
+  private fun addStatusUpdateEvents(jpa: Cas2ApplicationEntity, timelineEvents: MutableList<Cas2TimelineEvent>) {
+    jpa.statusUpdates?.forEach {
+      timelineEvents += Cas2TimelineEvent(
+        type = TimelineEventType.cas2StatusUpdate,
+        occurredAt = it.createdAt?.toInstant()!!,
+        label = it.label,
+        createdByName = it.assessor.name,
+        body = it.statusUpdateDetails?.joinToString { detail -> detail.label },
+      )
+    }
+  }
+
+  private fun addSubmittedEvent(jpa: Cas2ApplicationEntity, timelineEvents: MutableList<Cas2TimelineEvent>) {
+    if (jpa.submittedAt !== null) {
+      val submittedAtEvent = Cas2TimelineEvent(
+        type = TimelineEventType.cas2ApplicationSubmitted,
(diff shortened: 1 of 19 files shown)


src/main/resources/static/codegen/built-api-spec.yml
```diff
@@ -2768,40 +2768,46 @@ paths:
         - Task data
       summary: List all tasks
       parameters:
-        - in: query
-          name: type
+        - name: type
+          in: query
```
Harriethw, line 2772: it looks like there's some leftover generation that is missing on `main`, should be in a separate commit/already in `main`

src/main/resources/static/codegen/built-cas2-api-spec.yml
```diff
@@ -3797,12 +3797,34 @@ components:
         - cas3_person_arrived
         - cas3_person_departed
         - application_timeline_note
+        - cas2_application_submitted
```
Harriethw, line 3809: re-using the enum here as it seems cas3 has done the same thing, an not sure we need our own namespaced enum 🤔

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ApplicationTest.kt
```diff
@@ -384,6 +385,65 @@ class Cas2ApplicationTest : IntegrationTestBase() {
...
+    @Test
+    fun `Get single submitted application returns 200 with timeline events`() {
+      `Given a CAS2 User` { userEntity, jwt ->
+        `Given an Offender` { offenderDetails, inmateDetails ->
+          cas2ApplicationJsonSchemaRepository.deleteAll()
+
+          val newestJsonSchema = cas2ApplicationJsonSchemaEntityFactory
+            .produceAndPersist {
+              withAddedAt(OffsetDateTime.parse("2022-09-21T12:45:00+01:00"))
+              withSchema(
```
tahb, line 403: This seems like a lot of generic test set up that, do we need it? Could we move it off into some set up helper. If we did we'd hopefully make this test more readable and help the next test follow along.

tahb: A ha! You did this in the very next commit, I'll leave this comment in as a woop!

tahb (approved):
LGTM!

src/main/resources/static/codegen/built-api-spec.yml
Harriethw: looks like it's gone away!

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ApplicationTest.kt
Harriethw: yah, there's a lot of boilerplate to get through and think it could be potentially abstracted more, but maybe next time I touch the code :)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas2/TimelineEventsTransformer.kt
```diff
@@ -0,0 +1,57 @@
...
+    addNoteEvents(jpa, timelineEvents)
+
+    return timelineEvents.sortedByDescending { it.occurredAt }
+  }
+
+  private fun addNoteEvents(jpa: Cas2ApplicationEntity, timelineEvents: MutableList<Cas2TimelineEvent>) {
+    jpa.notes?.forEach {
+      timelineEvents += Cas2TimelineEvent(
+        type = TimelineEventType.cas2Note,
+        occurredAt = it.createdAt.toInstant(),
```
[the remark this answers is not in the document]
Harriethw: hmm, good spot - I don't think `createdAt` should ever be null, because it's assigned when the entity is created (or should be!).
I think in the other examples:
* it's a mistake on the StatusUpdateEntity that createdAt is nullable
* the typing isn't clever enough to know that because we've checked `submittedAt` is not null, we don't need to guard against it in the `if` statement.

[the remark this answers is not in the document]
Harriethw: > it's a mistake on the StatusUpdateEntity that createdAt is nullable

this is quite a big oversight so I'm fixing, as otherwise we can't sort them!

PR #1487, opened 2024-02-14
Update application status on withdrawal

davidatkinsuk:
If an entity is withdrawn directly by a user (i.e. it's not a casacading withdrawal), we will update the application status in the following circumstances:

1. If a placement request is withdrawn and there are no other active placement requests, change the status to 'PENDING_PLACEMENT_REQUEST' (meaning a placement_application is required)
2. If a booking is withdrawn and there are no other active bookings, change the status to 'AWAITING_PLACEMENT'

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
@@ -1234,6 +1234,20 @@ class BookingService(
     updateBooking(booking)
     booking.cancellations += cancellationEntity
 
+    createPlacementRequestIfBookingAppealed(reason, booking)
+
+    val user = withdrawalContext.triggeringUser
+    if (shouldCreateDomainEventForBooking(booking, user)) {
+      createCas1CancellationDomainEvent(booking, user, cancelledAt, reason)
+    }
+
+    return success(cancellationEntity)
+  }
+
+  private fun createPlacementRequestIfBookingAppealed(
+    reason: CancellationReasonEntity,
+    booking: BookingEntity,
+  ) {
     if (reason.id == approvedPremisesBookingAppealedCancellationReasonId && booking.placementRequest != null) {
       val placementRequest = booking.placementRequest!!
       placementRequestService.createPlacementRequest(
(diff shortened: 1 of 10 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
```diff
@@ -1241,9 +1243,33 @@ class BookingService(
...
+      return
+    }
+
+    val application = booking.application!!
+    val bookings = bookingRepository.findAllByApplication(application)
+    val anyActiveBookings = bookings.any { it.isActive() }
+    if(!anyActiveBookings) {
+      applicationService.updateApprovedPremisesApplicationStatus(
+        application.id,
+        ApprovedPremisesApplicationStatus.AWAITING_PLACEMENT
```
edavey, line 1268: in the commit msg you talk about `ApprovedPremisesApplicationStatus.PLACEMENT_REQUIRED` which I think is imaginary...

edavey (approved):
Thanks, this looks correct, it took me a while to get my head around the state transitions. I think we're agreed that we need to consolidate our transitions into a more formal state machine.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/PlacementRequestService.kt
```diff
@@ -347,6 +352,23 @@ class PlacementRequestService(
...
+    val placementRequests = placementRequestRepository.findByApplication(application)
+    val anyActivePlacementRequests = placementRequests.any { it.isActive() }
+    if(!anyActivePlacementRequests) {
+      applicationService.updateApprovedPremisesApplicationStatus(
+        application.id,
+        ApprovedPremisesApplicationStatus.PENDING_PLACEMENT_REQUEST
+      )
+    }
+  }
+
```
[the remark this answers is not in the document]
davidatkinsuk: That's fantastic, i hadn't thought it through in so much detail but glad it works for the more complex uses cases

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
davidatkinsuk: Good spot, that's now fixed

PR #1492, opened 2024-02-14
Add withdrawal reason when returning Placement Request, Placement Application and Application Summary API Objects

davidatkinsuk:
Relates to https://dsdmoj.atlassian.net/browse/APS-328


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/ApplicationsTransformer.kt
```diff
@@ -144,6 +144,7 @@ class ApplicationsTransformer(
         status = getStatusFromSummary(domain),
         type = "CAS1",
         tier = domain.getTier(),
+        isWithdrawn = domain.getIsWithdrawn(),
```
edavey, line 147: It might be worth tweaking the commit title to be "Add '**is**Withdrawn property..." for added clarity and in case people are search the history with the property name in the future?

edavey (approved):
excellent, thks

PR #1494, opened 2024-02-14
Add IT regression tests for withdrawal emails

davidatkinsuk:
This relates to https://dsdmoj.atlassian.net/browse/APS-308

This is regression testing existing emails in code for withdrawal scenarios

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/notification/EmailNotificationAsserter.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/notification/EmailNotificationAsserter.kt
@@ -12,7 +12,7 @@ class EmailNotificationAsserter {
   private val requestedEmails = mutableListOf<EmailRequest>()
 
   @EventListener
-  fun emailRequested(emailRequested: SendEmailRequestedEvent) {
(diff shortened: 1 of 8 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/notification/EmailNotificationAsserter.kt
```diff
@@ -12,7 +12,7 @@ class EmailNotificationAsserter {
   private val requestedEmails = mutableListOf<EmailRequest>()
 
   @EventListener
-  fun emailRequested(emailRequested: SendEmailRequestedEvent) {
+  fun consumeEmailRequestedEvent(emailRequested: SendEmailRequestedEvent) {
```
edavey, line 17: where are we using this function?

edavey (approved):
thks for backfilling.

a couple of tiny comments.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/notification/EmailNotificationAsserter.kt
davidatkinsuk: This is called by spring when an event of type 'SendEmailRequesedEvent' is raised. I tried to rename it to better reflect that but still not sure it's 100% clear

PR #1495, opened 2024-02-14
[APS-289] Create a new assessment when an appeal is accepted

danlivings-dxw:
> See [APS-289 on the Approved Premises Jira board](https://dsdmoj.atlassian.net/browse/APS-289).

This PR updates the `AppealService.createAppeal` method to automatically create a new assessment for the application if the appeal was accepted.

The application will automatically be moved into the `"unallocatedAssessment"` status when this occurs due to the `@PrePersist` annotated method on the [`AssessmentListener` class](https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/main/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/listeners/AssessmentListener.kt), so no explicit update of the status is needed within this PR, although the integration tests assert that this expected status change occurs.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/AppealsTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/AppealsTest.kt
@@ -6,6 +6,8 @@ import org.junit.jupiter.api.Test
 import org.springframework.http.HttpStatus
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Appeal
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.AppealDecision
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ApprovedPremisesApplication
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ApprovedPremisesApplicationStatus
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.NewAppeal
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ServiceName
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.TimelineEvent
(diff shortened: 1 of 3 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/AppealsTest.kt
```diff
@@ -265,38 +269,94 @@ class AppealsTest : IntegrationTestBase() {
...
+            .uri("/applications/${application.id}")
+            .header("Authorization", "Bearer $jwt")
+            .exchange()
+            .expectStatus()
+            .isOk
+            .returnResult(ApprovedPremisesApplication::class.java)
+
+          val applicationBody = applicationResult.responseBody.blockFirst()!!
+
+          assertThat(applicationBody.status).isEqualTo(ApprovedPremisesApplicationStatus.unallocatedAssesment)
```
edavey, line 355: nicely done

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/AppealServiceTest.kt
```diff
@@ -308,6 +313,7 @@ class AppealServiceTest {
       val now = LocalDate.now()
 
       every { appealRepository.save(any()) } returnsArgument 0
+      every { assessmentService.createApprovedPremisesAssessment(any()) } returns mockk<ApprovedPremisesAssessmentEntity>()
```
edavey, line 316: probably as well as setting up these stubs we also want to verify whether / when the appeal service reaches out to the assessment service? e.g.

- when appeal is accepted
  - `assessmentService.createApprovedPremisesAssessment()` is called with the `application`

- when appeal is rejected
  - `assessmentService.createApprovedPremisesAssessment()` is not called

edavey (approved):
thanks Dan, this looks great. 

I do think we ought to add little more detail to the service's unit test to describe the logic around how/when it asks the assessment service to create a new assessment...

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/AppealServiceTest.kt
danlivings-dxw: I've added these assertions in new tests.

edavey (approved):
🚀  great stuff, thks!

PR #1498, opened 2024-02-15
Add option to filter on multiple types to /tasks

davidatkinsuk:
Relates to https://dsdmoj.atlassian.net/browse/APS-343

Previously the caller could get tasks of all types, of a specific type. The new ‘types’ param allow the caller to specify multiple types to be returned.

The existing ‘type’ param is deprecated in favour of the new more flexible ‘types’ param

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/TasksController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/TasksController.kt
@@ -134,12 +134,13 @@ class TasksController(
 
   override fun tasksGet(
     type: TaskType?,
+    types: List<TaskType>?,
     page: Int?,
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/TasksController.kt
```diff
@@ -134,30 +134,26 @@ class TasksController(
 
   override fun tasksGet(
     type: TaskType?,
+    types: List<TaskType>?,
```
edavey, line 137: This looks great, thks. I think there's a tiny typo in the commit message, which for clarity might be worth fixing up. I think you're meaning to say:

> Previously the caller could get tasks of all types, **or** of a specific type

edavey (approved):
🚀

PR #1500, opened 2024-02-15
CAS2 Assessors can create Notes on an Application

Harriethw:
https://dsdmoj.atlassian.net/jira/software/c/projects/CAS2/boards/1347?selectedIssue=CAS2-212 

We previously introduced Notes on an Application https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1470 that could be added a by a Nomis user (e.g. a Referrer/`POM`).

This PR allows users with the CAS2_ASSESSOR role to add notes to any submitted applications.

Screenshot 2024-02-19 at 12 17 27

This is the first time we've come across an entity that could have either an ExternalUser or a NomisUser belonging to it - we discussed [introducing polymorphism in the previous PR, but tabled it at the time.](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1470#pullrequestreview-1875714328).
In this PR, rather than representing the polymorphism on the database level, I've [introduced a Sealed Class](https://kotlinlang.org/docs/sealed-classes.html) for a `CAS2User` which gives us a couple of benefits:
 - to construct `when` blocks that the compiler will warn us about if we leave out any possibilities. 
 - add helper methods to the `CAS2NoteEntity` to assign a user and ensure it always has either a Nomis or ExternalUser. 

On the database level, there is a constraint on the database that a note has either a `nomis` or an `external` user ID foreign key.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas2ApplicationNoteEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas2ApplicationNoteEntity.kt
@@ -19,9 +19,8 @@ data class Cas2ApplicationNoteEntity(
   @Id
   val id: UUID,
 
-  @ManyToOne
-  @JoinColumn(name = "created_by_nomis_user_id")
-  val createdByNomisUser: NomisUserEntity,
+  @Transient
+  final val createdByUser: Cas2User,
 
   @ManyToOne
   @JoinColumn(name = "application_id")
@@ -31,5 +30,29 @@ data class Cas2ApplicationNoteEntity(
 
   var body: String,
 ) {
+
+  @ManyToOne
+  @JoinColumn(name = "created_by_nomis_user_id")
+  var createdByNomisUser: NomisUserEntity? = null
+
+  @ManyToOne
+  @JoinColumn(name = "created_by_external_user_id")
+  var createdByExternalUser: ExternalUserEntity? = null
+
+  init {
+    when (this.createdByUser) {
+      is NomisUserEntity -> this.createdByNomisUser = this.createdByUser
+      is ExternalUserEntity -> this.createdByExternalUser = this.createdByUser
+    }
+  }
+
+  fun getUser(): Cas2User {
+    return if (this.createdByNomisUser != null) {
+      this.createdByNomisUser!!
+    } else {
+      this.createdByExternalUser!!
+    }
+  }
+
   override fun toString() = "Cas2ApplicationNoteEntity: $id"
 }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas2User.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas2User.kt
@@ -1,0 +1,8 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity
+
+sealed interface User
+
+sealed interface Cas2User : User {
+  val name: String
+  val email: String?
+}
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ExternalUserEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ExternalUserEntity.kt
@@ -22,11 +22,11 @@ data class ExternalUserEntity(
   val username: String,
   var isEnabled: Boolean,
   var origin: String,
-  var name: String,
-  var email: String,
+  override var name: String,
+  override var email: String,
 
   @CreationTimestamp
   private val createdAt: OffsetDateTime? = null,
-) {
+) : Cas2User {
   override fun toString() = "External user $id"
 }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/ApplicationNoteService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/ApplicationNoteService.kt
@@ -32,12 +32,10 @@ class ApplicationNoteService(
       ?: return AuthorisableActionResult.NotFound()
 
     val isExternalUser = httpAuthService.getCas2AuthenticatedPrincipalOrThrow().isExternalUser()
-    val user = getExternalOrNomisUser(isExternalUser)
+    val user = getCas2User(isExternalUser)
 
-    if (!isExternalUser) {
-      if (!isApplicationCreatedByUser(application, user as NomisUserEntity)) {
-        return AuthorisableActionResult.Unauthorised()
-      }
+    if (!isExternalUser && !isApplicationCreatedByUser(application, user as NomisUserEntity)) {
+      return AuthorisableActionResult.Unauthorised()
     }
 
     if (application.submittedAt == null) {
@@ -55,7 +53,7 @@ class ApplicationNoteService(
     )
   }
 
-  private fun getExternalOrNomisUser(isExternalUser: Boolean): Cas2User {
+  private fun getCas2User(isExternalUser: Boolean): Cas2User {
     return if (isExternalUser) {
       externalUserService.getUserForRequest()
     } else {
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/ApplicationNoteService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/ApplicationNoteService.kt
@@ -3,11 +3,16 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas2
 import org.springframework.data.repository.findByIdOrNull
 import org.springframework.stereotype.Service
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.NewCas2ApplicationNote
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas2ApplicationEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas2ApplicationNoteEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas2ApplicationNoteRepository
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas2ApplicationRepository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas2User
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.NomisUserEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.results.AuthorisableActionResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.results.ValidatableActionResult
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.ExternalUserService
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.HttpAuthService
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.NomisUserService
 import java.time.OffsetDateTime
 import java.util.UUID
@@ -17,6 +22,8 @@ class ApplicationNoteService(
   private val applicationRepository: Cas2ApplicationRepository,
   private val applicationNoteRepository: Cas2ApplicationNoteRepository,
   private val userService: NomisUserService,
+  private val externalUserService: ExternalUserService,
+  private val httpAuthService: HttpAuthService,
 ) {
 
   fun createApplicationNote(applicationId: UUID, note: NewCas2ApplicationNote):
@@ -24,10 +31,13 @@ class ApplicationNoteService(
     val application = applicationRepository.findByIdOrNull(applicationId)
       ?: return AuthorisableActionResult.NotFound()
 
-    val user = userService.getUserForRequest()
+    val isExternalUser = httpAuthService.getCas2AuthenticatedPrincipalOrThrow().isExternalUser()
+    val user = getExternalOrNomisUser(isExternalUser)
 
-    if (application.createdByUser != user) {
-      return AuthorisableActionResult.Unauthorised()
+    if (!isExternalUser) {
+      if (!isApplicationCreatedByUser(application, user as NomisUserEntity)) {
+        return AuthorisableActionResult.Unauthorised()
+      }
     }
 
     if (application.submittedAt == null) {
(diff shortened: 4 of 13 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas2/ApplicationNotesTransformer.kt
```diff
@@ -11,10 +11,12 @@ class ApplicationNotesTransformer {
     jpa: Cas2ApplicationNoteEntity,
   ):
     Cas2ApplicationNote {
+    val name = jpa.createdByNomisUser?.name ?: jpa.createdByExternalUser?.name
```
Harriethw, line 14: I wonder if the `NoteEntity` should have a `getUser` function...

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ExternalUserEntity.kt
```diff
@@ -27,6 +27,6 @@ data class ExternalUserEntity(
 
   @CreationTimestamp
   private val createdAt: OffsetDateTime? = null,
-) {
+) : Cas2User() {
```
Harriethw, line 30: first time using sealed classes like this, interested in any feedback. It seemed neater than introducing e.g. a new table that joins nomis/external users table

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas2ApplicationNoteEntity.kt
```diff
@@ -19,9 +19,8 @@ data class Cas2ApplicationNoteEntity(
   @Id
   val id: UUID,
 
-  @ManyToOne
-  @JoinColumn(name = "created_by_nomis_user_id")
-  val createdByNomisUser: NomisUserEntity,
+  @Transient
+  final val createdByUser: Cas2User,
```
Harriethw, line 23: this seemed a good way to ensure that the Entity will always define either `createdByNomisUser` or `createdByExternalUser`... if a new type of CAS2User that is not able to create a note was added, it could throw an error in the `when` block below.

src/main/resources/db/migration/all/20240214170259__add_external_users_to_cas2_notes.sql
```diff
@@ -0,0 +1,23 @@
...
+        FOREIGN KEY(created_by_external_user_id)
+        	REFERENCES external_users(id);
+
+-- make nomis users nullable
+ALTER TABLE cas_2_application_notes
+    ALTER COLUMN created_by_nomis_user_id DROP NOT NULL;
+
+-- must have either nomis or external user
+ALTER TABLE cas_2_application_notes
+    ADD CONSTRAINT has_user
```
Harriethw, line 17: I'd like to be able to test this at the sql level, but I'm not sure how to?

Harriethw:
moving this back to in progress to figure out the sealed classes stuff

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas2ApplicationNoteEntity.kt
```diff
@@ -31,5 +30,29 @@ data class Cas2ApplicationNoteEntity(
...
+
+  init {
+    when (this.createdByUser) {
+      is NomisUserEntity -> this.createdByNomisUser = this.createdByUser
+      is ExternalUserEntity -> this.createdByExternalUser = this.createdByUser
+    }
+  }
+
+  fun getCreatedByNomisOrExternalUser(): Cas2User {
+    return if (this.createdByNomisUser !== null) {
```
davidatkinsuk, line 50: I don't think you need !== here, != should be fine if comparing to null

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/ApplicationNoteService.kt
```diff
@@ -17,17 +22,22 @@ class ApplicationNoteService(
...
 ) {
 
   fun createApplicationNote(applicationId: UUID, note: NewCas2ApplicationNote):
     AuthorisableActionResult<ValidatableActionResult<Cas2ApplicationNoteEntity>> {
     val application = applicationRepository.findByIdOrNull(applicationId)
       ?: return AuthorisableActionResult.NotFound()
 
-    val user = userService.getUserForRequest()
+    val isExternalUser = httpAuthService.getCas2AuthenticatedPrincipalOrThrow().isExternalUser()
+    val user = getExternalOrNomisUser(isExternalUser)
```
tahb, line 35: Just from reading this in order I wouldn't be sure what type of thing  `isExternalUser` becomes? Could be boolean from the 'is' prefix but then it's passed as a param which suggests it could be object?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas2ApplicationNoteEntity.kt
davidatkinsuk: I'm 50/50 on whether this should be done here (as you've done) or in thing building the Entity but I can't strongly argue it either way!

I like how both the constructor and getCreatedByNomisOrExternalUser deal with your Cas2User interface, hiding away the internal implementation details.

My only suggestion would be to pull the new init block up just below the constructor because it can be missed at the moment.

Use of @Transient looks good - coming from Java i'd normally always avoid @Transients but in this context i think it's perfect - you're telling JPA that this is 'just' a constructor arg and shouldn't be directly persisted into a column

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ExternalUserEntity.kt
davidatkinsuk: I agree, a join table would just add so much more complexity that this approach

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/ApplicationNoteService.kt
```diff
@@ -17,17 +22,22 @@ class ApplicationNoteService(
...
 
-    if (application.createdByUser != user) {
-      return AuthorisableActionResult.Unauthorised()
+    if (!isExternalUser) {
+      if (!isApplicationCreatedByUser(application, user as NomisUserEntity)) {
```
davidatkinsuk, line 38: this could be combined into a single if using &&

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas2ApplicationNoteEntity.kt
```diff
@@ -31,5 +30,29 @@ data class Cas2ApplicationNoteEntity(
...
+  var createdByExternalUser: ExternalUserEntity? = null
+
+  init {
+    when (this.createdByUser) {
+      is NomisUserEntity -> this.createdByNomisUser = this.createdByUser
+      is ExternalUserEntity -> this.createdByExternalUser = this.createdByUser
+    }
+  }
+
+  fun getCreatedByNomisOrExternalUser(): Cas2User {
```
davidatkinsuk, line 49: Have you considered calling this getCas2User() instead?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/ApplicationNoteService.kt
tahb: Would it make sense for the getExternalOrNomisUser method to also take responsibility for this logic too? Seems related to finding a user. I'm guessing that logic is here so that isExternalUser is available later on for the conditional?

```
val user = getExternalOrNomisUser
if (!user.isExternal){
…
}

  private fun getExternalOrNomisUser: Cas2User {
    val isExternalUser = httpAuthService.getCas2AuthenticatedPrincipalOrThrow().isExternalUser()
    
    return if (isExternalUser) {
      {
        externalUserService.getUserForRequest(),
        isExternal: true
      }
    } else {
      {
        userService.getUserForRequest(),
        isExternal: false
      }
    }
  }
 ```
 
I'm not sure exactly how we'd do it but it feels like a useful attribute of a Cas2User object to know internal/external more generally?

File: src/main/resources/db/migration/all/20240214170259__add_external_users_to_cas2_notes.sql (hunk 4 above)
Comment (davidatkinsuk) [reply src/main/resources/db/migration/all/20240214170259__add_external_users_to_cas2_notes.sql:17]: I don't think we have an tests for table constraints anywhere else, but I appreciate why we'd want to test this one as it's non trivial. I suspect it's probably not worth the effort as long as we have some integration tests that create an entity with both types of users?

Comment (tahb) [reply src/main/resources/db/migration/all/20240214170259__add_external_users_to_cas2_notes.sql:17]: @sudhakar-plr [has previously done this by adding a migration job which provides a way to wrap it with tests](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1252/commits/12f7b648c47b5f79665c9abb75b8322bfae68dc6). Worth checking out!

**Review by davidatkinsuk — CHANGES_REQUESTED** [review]
Hi @Harriethw,

I think the approach using sealed classes and encapsulating the implementation details in the entity is sound, all my comments are minor.

Thanks for the solid commit descriptions and diagrams, they're very helpful

File: src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas2/TimelineEventsTransformer.kt [hunk 9]
```diff
@@ -25,7 +25,7 @@ class TimelineEventsTransformer {
         type = TimelineEventType.cas2Note,
         occurredAt = it.createdAt.toInstant(),
         label = "Note",
-        createdByName = it.createdByNomisUser.name,
+        createdByName = it.getCreatedByNomisOrExternalUser().name,
```
tahb, line 28: I wonder if the timeline transformer needs to know about the underlying implementation? Would `getUser` suffice?

tahb (commented):
Lovely PR! The nice comments make it super easy to follow along, thank you.

I've never seen sealed class before so I'll just leave a comment now as I see David is looking at it too. I think it looks good, very interested to see if this is the way we do polymorphic things in kotlin.

Couple of other suggestions from me that I think might help.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas2User.kt
```diff
@@ -0,0 +1,8 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity
+
+sealed interface User
+
+sealed class Cas2User : User {
+  abstract val name: String
+  abstract val email: String?
+}
```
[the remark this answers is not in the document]
Harriethw: ooh thanks, this is great intel :)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas2ApplicationNoteEntity.kt
Harriethw: nice feedback, thank you!
> My only suggestion would be to pull the new init block up just below the constructor because it can be missed at the moment. 

I think it's not possible to put it above the `createdByNomisUser` and `createdByExternalUser` var's, as they need to be declared before the `init` function, hopefully as it's the first of the listed functions it should be clear enough?

davidatkinsuk: Yep you're right, it's as high as it can be 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas2ApplicationNoteEntity.kt
Harriethw: good shout

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/ApplicationNoteService.kt
Harriethw: I agree I think there's some tidying up/shifting around to do, but I'm not sure how exactly to do it either.

>  Could be boolean from the 'is' prefix but then it's passed as a param which suggests it could be object?
Is there something against passing a boolean as a param? not heard that before.

>  it feels like a useful attribute of a Cas2User object to know internal/external more generally?

I think the `isExternalUser` is quite a specific thing- it's based on the `auth_source` of the `jwt` so I think it makes sense for the `httpAuthService` to be the one that decides that (and we also then get to check that it's a valid auth source)... I don't know that the `Cas2User` should know about it, I think if it were a function or field on the `Cas2User` we'd also have to add it the classes that extend, which should definitely not know imo  🤔 

```
...
    return if (isExternalUser) {
      {
        externalUserService.getUserForRequest(),
        isExternal: true
      }
    } else {
      {
        userService.getUserForRequest(),
        isExternal: false
      }
    }
  }
...
```

This would be a good solution but I can't figure out how to do it in a Kotlin way that doesn't make it less readable 😢 

I've pushed a commit with a refactor to the file, I think it makes it more readable.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas2/TimelineEventsTransformer.kt
Harriethw: oh that was about a different method... yeah I think `getUser` would be better here.

src/main/resources/db/migration/all/20240214170259__add_external_users_to_cas2_notes.sql
Harriethw: >@sudhakar-plr [has previously done testing mirgations by adding a migration job which provides a way to wrap it with tests](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1252/commits/12f7b648c47b5f79665c9abb75b8322bfae68dc6). Worth checking out!

@tahb ooh wow that looks really interesting, so it's creating the migration in Kotlin? That seems a much more readable way of doing things than just adding a sql file :) 
It looks like it would be adding quite a lot to this PR but might be worth it in the long run - might timebox it till tomorrow morning to see how far I get with it

Harriethw: I've been grappling with adding a Migration Job to the Migration Service as in the PR above, and it would be a fair bit of work for me, and its a slightly different use case in terms of what we're doing to the db. 
@tahb I wonder if we say we have enough confidence for now that the functionality is tested, as we have integration tests for the whole process of saving, and unit tests around the service and Entity transformation? And in the future we'll look at using the Migration Service.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas2ApplicationNoteEntity.kt
Harriethw: I've gone with `getUser` as per @tahb's [suggestion below](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1500#discussion_r1494710694) - that perhaps those calling it don't need to know underlying implementation?

I originally wanted to call it `getCreatedByUser` because I'd prefer to be specific but got a Java(or Hibernate?) complaint that it clashed with the generated `get` method for the field `createdByUser` - I tried to look into best practices/solutions for overriding the generated one, and wasn't sure what to do, if you have any ideas?

davidatkinsuk: That makes sense. This is where my knowledge of kotlin is limited - I always get confused around constructors! I think there's a way to declare that the 'createdByUser' is _not_ a property and a constructor arg only (drop the 'val' keyword?), but i'm not sure if it would then be available to init. Maybe give that a quick go and if it doesn't work go with getUser() like suggested 👍

davidatkinsuk (approved):
I'm happy that all my comments have been discussed and/or resolved 👍

PR #1501, opened 2024-02-15
Enhance withdrawal cascade integration tests

davidatkinsuk:
Update tests to ensure that if a placement application is directly withdrawn, placement applications on the same applciation are not also withdrawn


edavey (approved):
bravo, thanks for filling in this important scenario.

I think there's a tiny typo in the commit msg and that you mean to say something like:

> Update tests to ensure that if a placement application is directly withdrawn, 
> **other** placement applications on the same applciation are not also withdrawn

PR #1508, opened 2024-02-20
(APS-368) Add an appeal URL to timeline events

pezholio:
This adds an appeal URL to timeline events (when an appeal ID is present). This will allow the timeline view to link directly to an appeal when one has occurred.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ApplicationService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ApplicationService.kt
@@ -526,7 +526,7 @@ class ApplicationService(
   }
 
   fun updateApprovedPremisesApplicationStatus(applicationId: UUID, status: ApprovedPremisesApplicationStatus) {
-    applicationRepository.updateStatus(applicationId,status)
+    applicationRepository.updateStatus(applicationId, status)
   }
 
   @Transactional
--- a/src/main/resources/static/_shared.yml
+++ b/src/main/resources/static/_shared.yml
@@ -3341,6 +3341,7 @@ components:
         - application
         - booking
         - assessment
+        - assessmentAppeal
     Cas2TimelineEvent:
       type: object
       properties:
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/transformer/ApplicationTimelineTransformerTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/transformer/ApplicationTimelineTransformerTest.kt
@@ -28,6 +28,7 @@ class ApplicationTimelineTransformerTest {
     UrlTemplate("http://somehost:3000/applications/#id"),
     UrlTemplate("http://somehost:3000/assessments/#id"),
     UrlTemplate("http://somehost:3000/premises/#premisesId/bookings/#bookingId"),
+    UrlTemplate("http://somehost:3000/applications/#applicationId/appeals/#appealId"),
     mockDomainEventDescriber,
   )
 
@@ -39,6 +40,7 @@ class ApplicationTimelineTransformerTest {
     override val assessmentId: UUID?,
     override val bookingId: UUID?,
     override val premisesId: UUID?,
+    override val appealId: UUID?,
   ) : DomainEventSummary
 
   @ParameterizedTest
@@ -53,6 +55,7 @@ class ApplicationTimelineTransformerTest {
       applicationId = null,
       assessmentId = null,
       premisesId = null,
+      appealId = null,
     )
 
     every { mockDomainEventDescriber.getDescription(domainEvent) } returns "Some event"
@@ -89,6 +92,7 @@ class ApplicationTimelineTransformerTest {
       applicationId = applicationId,
       assessmentId = null,
       premisesId = null,
+      appealId = null,
     )
 
     every { mockDomainEventDescriber.getDescription(domainEvent) } returns "Some event"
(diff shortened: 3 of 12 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ApplicationService.kt
```diff
@@ -1047,10 +1047,11 @@ class ApplicationService(
   fun getOfflineApplicationsForCrn(crn: String, serviceName: ServiceName) =
     offlineApplicationRepository.findAllByServiceAndCrn(serviceName.value, crn)
 
-  fun getApplicationTimeline(applicationId: UUID): List<TimelineEvent> {
+  fun getApplicationTimeline(applicationId: UUID, user: UserEntity): List<TimelineEvent> {
     val domainEvents = domainEventService.getAllDomainEventsForApplication(applicationId)
+    val canManageAppeals = user.hasRole(UserRole.CAS1_APPEALS_MANAGER)
```
pezholio, line 1052: I think it'd be good going forward if we could have one source of truth for what a given user can do (even if it is just based on one role)

src/main/resources/static/_shared.yml
```diff
@@ -3340,6 +3340,7 @@ components:
         - application
         - booking
         - assessment
+        - appeal
```
edavey, line 3343: would `assessmentAppeal` be clearer here? We've taken trouble elsewhere to  explicit that appeal relates to an assessment (see `APPROVED_PREMISES_ASSESSMENT_APPEALED` etc) as we expect that other things eg.  'bookings' are likely to be "appeal-able" in the future....

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/transformer/ApplicationTimelineTransformerTest.kt
```diff
@@ -106,6 +110,47 @@ class ApplicationTimelineTransformerTest {
...
+      assessmentId = null,
+      premisesId = null,
+      appealId = appealId,
+    )
+
+    every { mockDomainEventDescriber.getDescription(domainEvent) } returns "Some event"
+
+    val associatedUrls = if (domainEventType == DomainEventType.APPROVED_PREMISES_ASSESSMENT_APPEALED) {
+      listOf(
+        TimelineEventAssociatedUrl(TimelineEventUrlType.application, "http://somehost:3000/applications/$applicationId"),
```
edavey: In our review session yesterday I think we agreed that this (assessment appealed timeline event) will be clearer with just **one** url - that of the appeal.

edavey (approved):
Great stuff Stu: just a couple of tiny tweaks suggested here

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/transformer/ApplicationTimelineTransformerTest.kt
pezholio: I guess the question here would be - if someone is not an appeals manager, should they see no URLs at all against that event type?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ApplicationService.kt
edavey: hmm, I'm not aware of this requirement to limit access to the appeal page. Where does that come from? What's the information on that view which shouldn't be seen by applicants?

pezholio: Yeah, it's this line https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/e8c8d3c9bf64c70cbf456d538b4b4487180ec5e2/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AppealService.kt#L40

pezholio: It does seem a bit weird though - I'll remove the transformer requirement in this PR, and open another PR to remove the access limitation

pezholio: PR here: https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1510

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/transformer/ApplicationTimelineTransformerTest.kt
pezholio: Ignore that, as above, I think we'll remove that restriction

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/transformer/ApplicationTimelineTransformerTest.kt
```diff
@@ -106,6 +110,44 @@ class ApplicationTimelineTransformerTest {
     )
   }
 
+  @ParameterizedTest
+  @MethodSource("domainEventTypeArgs")
+  fun `transformDomainEventSummaryToTimelineEvent adds appealUrl for appeal events and the user can manage appeals`(args: Pair<DomainEventType, TimelineEventType>) {
```
edavey, line 115: I don't think this bit is relevant? The link to view the appeal page should be present for all users I believe?

>  and the user can manage appeals

edavey (approved):
thanks Stu, this looks tip-top! I just noticed what I think is a bit of cruft in the a test name referring to an appeal manager...

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/transformer/ApplicationTimelineTransformerTest.kt
pezholio: Good catch!

PR #1509, opened 2024-02-20
Add BookingMadeDate to App and PA Reports

davidatkinsuk:
Relates to https://dsdmoj.atlassian.net/browse/APS-372

Use the ‘booking made’ domain event’s creation timestamp to determine when a booking was made

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/reporting/model/PlacementApplicationReportRow.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/reporting/model/PlacementApplicationReportRow.kt
@@ -30,6 +30,7 @@ data class PlacementApplicationReportRow(
   val applicationWithdrawalDate: LocalDate?,
   val applicationWithdrawalReason: String?,
   val bookingID: String?,
+  val bookingMadeDate: LocalDate?,
   val bookingCancellationReason: String?,
   val bookingCancellationDate: LocalDate?,
   val expectedArrivalDate: LocalDate?,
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/reporting/model/PlacementApplicationReportRow.kt
```diff
@@ -30,6 +31,7 @@ data class PlacementApplicationReportRow(
   val applicationWithdrawalDate: LocalDate?,
   val applicationWithdrawalReason: String?,
   val bookingID: String?,
+  val bookingMadeDate: LocalDateTime?,
```
edavey, line 34: I think we've maybe got a mismatch between the name (date) and the format (timestamp)?

perhaps:
- If the report requires a date then we need to convert the timestamp to date
- if the report requires a timestamp then we need to name the field/property/column accordingly e.g. `bookingMadeAt` 

We've tried to stick to the naming convention of:

- `...At` for a timestamp, e.g. `bookingMadeAt`
- `..On` for a date, e.g. `bookingMadeOn`

edavey (approved):
looks good, though I think we have an issue with aligning the format and the name?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/reporting/model/PlacementApplicationReportRow.kt
davidatkinsuk: Good spot! looking at the existing report columns we are only returning a LocalDate (not LocalDateTime), so i'm going to change these occurrences to Date only.

As for naming, for these two reports we're using the 'Date' suffix everywhere LocalDate is used, so for consistency i'll stick with this naming

edavey: 👍 local conventions trump!

PR #1510, opened 2024-02-21
(Chore) Allow all users to see appeals data

pezholio:
We want to link to appeals in the timeline, so all users should be able to see data for an appeal.


edavey (approved):
thanks Stu, I should have noticed this issue earlier!

PR #1512, opened 2024-02-21
(APS 370) Surface that an assessment is an appealed assessment so that staff are aware

pezholio:
This adds a `createdFromAppeal` flag to Approved Premises assessments. If an appeal is accepted and a new assessment is created as a result, this flag is set to `true` in the database. This is then sent across in the API, so we can add the relevant flags on the frontend. It will also allow us to automatically allocate to CRU managers in a future PR.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/AssessmentEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/AssessmentEntity.kt
@@ -276,6 +276,7 @@ class ApprovedPremisesAssessmentEntity(
   referralHistoryNotes: MutableList<AssessmentReferralHistoryNoteEntity>,
   schemaUpToDate: Boolean,
   isWithdrawn: Boolean,
+  val createdFromAppeal: Boolean,
 ) : AssessmentEntity(
   id,
   application,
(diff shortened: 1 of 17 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/AssessmentEntity.kt
```diff
@@ -276,6 +276,7 @@ class ApprovedPremisesAssessmentEntity(
   referralHistoryNotes: MutableList<AssessmentReferralHistoryNoteEntity>,
   schemaUpToDate: Boolean,
   isWithdrawn: Boolean,
+  var createdFromAppeal: Boolean,
```
davidatkinsuk, line 279: I don't _think_ this needs to be a var?

src/main/resources/db/migration/all/20240221091158__add_created_from_appeal_to_assessments.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE approved_premises_assessments ADD COLUMN created_from_appeal BOOLEAN NOT NULL DEFAULT TRUE;
```
davidatkinsuk, line 1: shouldn't this be DEFAULT FALSE, otherwise all existing assessments will be 'created from appeal'?

davidatkinsuk (changes requested):
welcome back! :) A few comments from me

src/main/resources/db/migration/all/20240221091158__add_created_from_appeal_to_assessments.sql
pezholio: Good point!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/AssessmentEntity.kt
pezholio: I still can't get the difference straight in my head! All sorted now 👍

davidatkinsuk (approved):
LGTM 👍

PR #1298, opened 2024-01-05
[APS-67] Implement AP Delius Context API offender details data source

danlivings-dxw:
> See [APS-67 on the Approved Premises Jira board](https://dsdmoj.atlassian.net/browse/APS-67).

This PR provides the ability to use the AP Delius Context API as a data source for offender details. This is achieved by implementing the `ApDeliusContextApiOffenderDetailsDataSource` class.

The `ApDeliusContextApiOffenderDetailsDataSource` class performs transformations of the data in order to match the same data structure used by the Community API for backwards compatibility. These are enabled by various utility methods in the `OffenderDetailsUtils.kt` and `BaseHMPPSClient.kt` files.

This PR does not change the data source for offender details as this is down to the individual configuration for each environment.

PR #1513, opened 2024-02-21
[APS-427] Enable AP Delius Context API as the offender data source for test env

danlivings-dxw:
> See [APS-427 on the Approved Premises Jira board](https://dsdmoj.atlassian.net/browse/APS-427).

This PR updates the configuration to switch the data source for offender details and offender risks to the AP Delius Context API in the `test` environment. This will allow us to manually test that we get the information we expect from the AP Delius Context API and confirm that it is safe to disable the Community API as a source of this information.

In order to enable this change, several integration tests have been modified in order to mock/expect the AP Delius Context API as the source of this information. These changes include:
- Updating the API specification to prevent a `Person` with `UnknownPerson` incorrectly being deserialised as a `FullPerson` instance
- Modifying the WireMock stubs for the AP Delius Context API's `/users/access` and `/probation-cases/summaries` endpoints to support both bulk and individual fetch cases
- Modifying the `/users/access` WireMock stub to correctly handle the 'any username' case
- Introducing `Factory<T>.from(T)` as a method of creating a factory configured to produce a duplicate of an object (with optional further changes) in situations where the original factory is unavailable
- Modifying `Given an Offender` and `Given Some Offenders` to setup the necessary AP Delius Context API mocks
- Using `Given an Offender` in most cases where `OffenderDetailSummaryFactory` was explicitly constructed and used
- Explicitly setting up the required AP Delius Context API mocks to mimic the Community API mocks where the above was not suitable

See the PRs below for more context about this configuration change:
- #1200
- #1220 
- #1257 
- #1298 
- #1317 
- #1354 
- #1404

### DIFF
--- a/helm_deploy/values-test.yaml
+++ b/helm_deploy/values-test.yaml
@@ -72,8 +72,8 @@ generic-service:
     URL-TEMPLATES_API_CAS3_PERSON-DEPARTURE-UPDATED-EVENT-DETAIL: https://approved-premises-api-test.hmpps.service.justice.gov.uk/events/cas3/person-departure-updated/#eventId
     URL-TEMPLATES_API_CAS3_BOOKING-CANCELLED-UPDATED-EVENT-DETAIL: https://approved-premises-api-test.hmpps.service.justice.gov.uk/events/cas3/booking-cancelled-updated/#eventId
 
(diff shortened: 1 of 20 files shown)


nsmj63-dxw (approved):
Looks great man. A few, well-executed, clear changes. Shame the integration tests caused so much trouble.

PR #1515, opened 2024-02-21
(APS-382) Automatically allocate appealed assessments to CRU managers

pezholio:
This adds a new rule to automatically allocate "standard" appealed assessments directly to the person who created the appeal (the "arbitrator"). This will take precedence over all rules, so all appealed assessments will follow this rule.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/allocations/rules/AppealedAssessmentRule.kt
```diff
@@ -0,0 +1,42 @@
...
+
+@Component
+@ConditionalOnProperty(name = ["user-allocations.rules.appealed-assessments.enabled"])
+class AppealedAssessmentRule(
+  @Value("\${user-allocations.rules.appealed-assessments.priority:0}")
+  override val priority: Int,
+  private val appealRepository: AppealRepository,
+) : UserAllocatorRule {
+  override val name: String
+    get() = "Appeased assessments"
```
edavey, line 21: I think you mean "Appealed assessments" here? (Freudian slip for sure)

helm_deploy/values-dev.yaml
```diff
@@ -82,12 +82,14 @@ generic-service:
...
+    USER-ALLOCATIONS_RULES_ESAP-ASSESSMENTS_PRIORITY: 3
     USER-ALLOCATIONS_RULES_ESAP-ASSESSMENTS_ALLOCATE-TO-USER: # TODO: Determine appropriate user
     USER-ALLOCATIONS_RULES_EMERGENCY-ASSESSMENTS_ENABLED: true
-    USER-ALLOCATIONS_RULES_EMERGENCY-ASSESSMENTS_PRIORITY: 1
+    USER-ALLOCATIONS_RULES_EMERGENCY-ASSESSMENTS_PRIORITY: 2
     USER-ALLOCATIONS_RULES_FEMALE-ASSESSMENTS_ENABLED: true
-    USER-ALLOCATIONS_RULES_FEMALE-ASSESSMENTS_PRIORITY: 0
+    USER-ALLOCATIONS_RULES_FEMALE-ASSESSMENTS_PRIORITY: 1
+    USER-ALLOCATIONS_RULES_APPEALED-ASSESSMENTS_ENABLED: true
+    USER-ALLOCATIONS_RULES_APPEALED-ASSESSMENTS_PRIORITY: 0
```
edavey, line 92: Just to be clear on these priorities: we're saying that we apply this rule first? So  the new assessment is always allocated to the arbitrator?

pezholio: Yeah, exactly. Confirmed this with Adam too.

PR #1517, opened 2024-02-22
Workflow Manager should be able to withdraw placement applications

davidatkinsuk:
Relates to https://dsdmoj.atlassian.net/browse/APS-386


edavey (approved):
thanks David, and also for the refactorings included.

PR #1518, opened 2024-02-22
(APS-383) Make appeal manager the arbitrator of an appeal

pezholio:
This adds a `createdByUser` to the appeals response, rather than just the ID, allowing us to get more information about the user who created the appeal.

We also remove the `reviewer` field, as now the person who created the appeal is also considered to be the arbitrator of the appeal, so we no longer need this field.

### DIFF
--- a/src/main/resources/static/_shared.yml
+++ b/src/main/resources/static/_shared.yml
@@ -4281,8 +4281,6 @@ components:
           format: date
         appealDetail:
           type: string
-        reviewer:
-          type: string
         decision:
           $ref: '#/components/schemas/AppealDecision'
         decisionDetail:
@@ -4290,7 +4288,6 @@ components:
       required:
         - appealDate
         - appealDetail
-        - reviewer
         - decision
         - decisionDetail
     Appeal:
@@ -4303,8 +4300,6 @@ components:
           type: string
           format: date
         appealDetail:
-          type: string
-        reviewer:
           type: string
         decision:
           $ref: '#/components/schemas/AppealDecision'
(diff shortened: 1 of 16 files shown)


src/main/resources/static/_shared.yml
```diff
@@ -4319,9 +4319,8 @@ components:
         assessmentId:
           type: string
           format: uuid
-        createdByUserId:
```
davidatkinsuk, line 4322: sanity check - the ui isn't going to explode if this is missing?

davidatkinsuk (approved):
one sanity check from me (which i guess applies to other 'reviewer' removal too), but otherwise, LGTM

src/main/resources/static/_shared.yml
pezholio: It might, but I'm expecting we'll fix it fairly quickly

davidatkinsuk: ok I guess not an issue if only appearing on appeals pages that i guess aren't shown in prod anyway

src/main/resources/static/_shared.yml
```diff
@@ -4331,7 +4330,7 @@ components:
         - decisionDetail
         - createdAt
         - applicationId
-        - createdByUserId
+        - createdByUser
```
edavey, line 4327: I wonder if we should be more explicit here to make it clear that appeal was `arbitratedBy` the given user? That the person reporting the appeal decision is always and necessarily the person who ruled on the appeal.

If we use `createdBy` there is perhaps sufficient ambiguity that it may look like the this appeal decision is being reported by someone else? That can't be the case in reality.

pezholio: Yeah, I did consider this, but thought it probably wasn't worth unpicking all the database columns etc. Happy to do this in a different PR if you think it's worth doing

PR #1524, opened 2024-02-22
Return Assessment on CAS2 Applications

Harriethw:
https://dsdmoj.atlassian.net/jira/software/c/projects/CAS2/boards/1347?selectedIssue=CAS2-251

We previously created the `Assessment` model, with a view to expanding its use on CAS2 here https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1514

Now we want to return it on the `Application` so that the frontend can render it. This may be an interim solution until we have completed all the [phases of our suggested Application refactoring ](doc/architecture/decisions/0021-introduce-cas2-assessments.md)

In a previous commit we introduced functionality to [create an Assessment whenever a CAS2
application is submitted](https://github.com/ministryofjustice/hmpps-approved-premises-api/commit/e479df3f5556005cafb15598aed944668b67353e). This will have left any applications in Production that had already been submitted without an Assessment model. In order to allow the frontend to expect that all submitted applications have an assessment, and to ensure all our data is consistent, we introduce a migration here to add an Assessment to any Application that has been submitted and does not yet have one.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/SubmissionsController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/SubmissionsController.kt
@@ -30,20 +30,19 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas2.Application
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas2.OffenderService
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas2.StatusUpdateService
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.cas2.ApplicationNotesTransformer
-import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.cas2.SubmittedApplicationTransformer
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.cas2.SubmissionsTransformer
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.PageCriteria
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.cas2.getFullInfoForPersonOrThrow
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.cas2.getInfoForPersonOrThrowInternalServerError
 import java.net.URI
 import java.util.UUID
-import javax.persistence.metamodel.EntityType
 
 @Service("Cas2SubmissionsController")
 class SubmissionsController(
   private val httpAuthService: HttpAuthService,
   private val applicationService: ApplicationService,
   private val applicationNoteService: ApplicationNoteService,
-  private val applicationTransformer: SubmittedApplicationTransformer,
+  private val submissionsTransformer: SubmissionsTransformer,
   private val applicationNotesTransformer: ApplicationNotesTransformer,
   private val offenderService: OffenderService,
   private val externalUserService: ExternalUserService,
@@ -181,7 +180,7 @@ class SubmissionsController(
     Cas2SubmittedApplicationSummary {
     val personInfo = offenderService.getInfoForPersonOrThrowInternalServerError(application.getCrn())
 
-    return applicationTransformer.transformJpaSummaryToApiRepresentation(
+    return submissionsTransformer.transformJpaSummaryToApiRepresentation(
       application,
       personInfo,
     )
@@ -192,7 +191,7 @@ class SubmissionsController(
   ): Cas2SubmittedApplication {
     val personInfo = offenderService.getFullInfoForPersonOrThrow(application.crn)
 
-    return applicationTransformer.transformJpaToApiRepresentation(
+    return submissionsTransformer.transformJpaToApiRepresentation(
       application,
       personInfo,
     )
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/migration/Cas2AssessmentMigrationJob.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/migration/Cas2AssessmentMigrationJob.kt
@@ -1,0 +1,45 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.migration
+
+import org.slf4j.LoggerFactory
+import org.springframework.data.domain.Slice
+import org.springframework.transaction.support.TransactionTemplate
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas2ApplicationEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas2ApplicationRepository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas2AssessmentEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas2AssessmentRepository
+import java.time.OffsetDateTime
+import java.util.UUID
+
+class Cas2AssessmentMigrationJob(
+  private val assessmentRepository: Cas2AssessmentRepository,
+  private val applicationRepository: Cas2ApplicationRepository,
+  private val transactionTemplate: TransactionTemplate,
+) : MigrationJob() {
+  private val log = LoggerFactory.getLogger(this::class.java)
+  override val shouldRunInTransaction = true
+
+  override fun process() {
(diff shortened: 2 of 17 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/SubmissionsController.kt
```diff
@@ -30,20 +30,19 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas2.Application
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas2.OffenderService
```
tahb, line 30: There's a "[LINK TO PR]" still in the commit message here if you still feel there's value in including it.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/migration/Cas2AssessmentMigrationJob.kt
```diff
@@ -0,0 +1,47 @@
...
+  private val applicationRepository: Cas2ApplicationRepository,
+  private val transactionTemplate: TransactionTemplate,
+) : MigrationJob() {
+  private val log = LoggerFactory.getLogger(this::class.java)
+  override val shouldRunInTransaction = true
+
+  override fun process() {
+    log.info("Starting Cas2 Assessment Migration process...")
+
+    var page = 1
```
tahb, line 24: How does `page` get used?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2MigrateAssessmentsTest.kt
```diff
@@ -0,0 +1,89 @@
...
+
+      val submittedWithAssessment = createApplicationEntity(applicationSchema, userEntity, submittedAt)
+
+      cas2AssessmentEntityFactory.produceAndPersist {
+        withApplication(submittedWithAssessment)
+      }
+
+      migrationJobService.runMigrationJob(MigrationJobType.cas2ApplicationsWithAssessments, 1)
+
+      checkUnsubmittedDoesNotHaveAssessment(unsubmittedApp)
```
tahb, line 58: Nice 👍

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/factory/Cas2AssessmentEntityFactory.kt
```diff
@@ -0,0 +1,44 @@
...
+
+  fun withAssessorName(name: String) = apply {
+    this.assessorName = { name }
+  }
+
+  override fun produce(): Cas2AssessmentEntity = Cas2AssessmentEntity(
+    id = this.id(),
+    createdAt = this.createdAt(),
+    application = this.application(),
+    nacroReferralId = this.nacroReferralId(),
```
tahb, line 41: Though the ship has sailed I just wondered if introducing "referral" alongside "application" risks the reader thinking these are fundamentally different types of thing? Are they references to the same thing?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/MigrationJobService.kt
```diff
@@ -79,6 +82,12 @@ class MigrationJobService(
           applicationContext.getBean(TaskDeadlineService::class.java),
           pageSize,
         )
+
+        MigrationJobType.cas2ApplicationsWithAssessments -> Cas2AssessmentMigrationJob(
```
tahb, line 86: How do we make this run?

tahb (approved):
LGTM. Some non-blocking questions from me!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/migration/Cas2AssessmentMigrationJob.kt
Harriethw: ah yeah good spot, I don't think we need this and was copy/pasted from another different type of migration.
It did make me think that the logging could be more informative, so I've added `log.info("Saving assessment for ${application.id}")` before the transaction instead of the previous 'Getting page' log, sound good?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/MigrationJobService.kt
Harriethw: oh interesting... I assumed these ran alongside the other regular migrations but now I see they have to be triggered by a POST request if I understand correctly? 🤔  https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/418v

tahb: Ah yes. That PR has a doc that tells us how to run them https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/418/files#diff-53a8b6dceb125ee0bda5113152b27fcc0c72bbdc75867bfc3e0ed30bbca17a75R1. It sounds like there could be a post merge task here to run that on dev to double check it. Then to remember to run it on every environment before we're done?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/SubmissionsController.kt
Harriethw: good spot, link is in message now

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/SubmissionsController.kt
```diff
@@ -30,20 +30,19 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas2.Application
...
 import java.util.UUID
-import javax.persistence.metamodel.EntityType
 
 @Service("Cas2SubmissionsController")
 class SubmissionsController(
   private val httpAuthService: HttpAuthService,
   private val applicationService: ApplicationService,
   private val applicationNoteService: ApplicationNoteService,
-  private val applicationTransformer: SubmittedApplicationTransformer,
+  private val applicationTransformer: SubmissionsTransformer,
```
Gweaton, line 45: Is it worth renaming the variable here too?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/MigrationJobService.kt
Harriethw: hmm this has just made me think... as the dev database gets wiped on every deploy, we'll start [with the seeded data here ](https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/595f1e725d8db7c1aac9daac148ec81910cc32c8/src/main/resources/db/seed/local+dev+test/6__cas2_applications.csv)  - so we could test the migration in the dev environment because the seeded data has a few submitted applications without assessments. But then after we've tested we should ensure we create assessments in [the seed job here](https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/1a0e54e8d6543a0dd248c434c6d4ca7089bdfa9f/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas2/Cas2ApplicationsSeedJob.kt#L82 ) so that we don't have to keep running the migration every time we deploy.

Gweaton (approved):
This all makes sense to me, thanks for the detailed commit messages, they made it very easy to review as a newcomer to the project!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/MigrationJobService.kt
Harriethw: have created a task here so it doesn't get lost if it isn't done before I leave https://dsdmoj.atlassian.net/jira/software/c/projects/CAS2/boards/1347?selectedIssue=CAS2-301

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/SubmissionsController.kt
Harriethw: good point! have renamed

PR #1525, opened 2024-02-22
(APS-384) Send notification email on successful appeal

pezholio:
This sends notification emails to both the applicant and the appeal arbitrator if an appeal is upheld


edavey (approved):
👍 thanks, v neat

PR #1530, opened 2024-02-23
(APS-385) Send notification email on unsuccessful appeal

pezholio:
This updates the Appeals Email service to add an email to be sent when an appeal is _unsuccessful_

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AppealService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AppealService.kt
@@ -92,9 +92,14 @@ class AppealService(
 
         saveEvent(appeal)
 
-        if (decision == AppealDecision.accepted) {
-          assessmentService.createApprovedPremisesAssessment(application as ApprovedPremisesApplicationEntity, createdFromAppeal = true)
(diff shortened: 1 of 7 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AppealService.kt
```diff
@@ -92,9 +92,14 @@ class AppealService(
 
         saveEvent(appeal)
 
-        if (decision == AppealDecision.accepted) {
-          assessmentService.createApprovedPremisesAssessment(application as ApprovedPremisesApplicationEntity, createdFromAppeal = true)
-          cas1AppealEmailService.appealSuccess(application, appeal)
+        when (decision) {
```
davidatkinsuk, line 95: 'when' is 50% of reason why i'd use kotlin over java going forward, makes everything so much safer!

davidatkinsuk (approved):
LGTM 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AppealService.kt
pezholio: Yeah, it's ace!

PR #1531, opened 2024-02-23
Improve request for placement domain event description

davidatkinsuk:
This description now include the dates.

Screenshot 2024-02-23 at 11 21 18

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/domainevents/DomainEventDescriber.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/domainevents/DomainEventDescriber.kt
@@ -7,12 +7,16 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.DomainEventSummary
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.DomainEventService
 import java.time.LocalDate
 import java.time.ZoneOffset
+import java.time.format.DateTimeFormatter
 import java.util.UUID
 
 @Component
 class DomainEventDescriber(
   private val domainEventService: DomainEventService,
 ) {
+
+  val cas1UiExtendedDateFormat: DateTimeFormatter = DateTimeFormatter.ofPattern("EEEE dd MMMM yyyy")
+
   fun getDescription(domainEventSummary: DomainEventSummary): String? {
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1PlacementApplicationDomainEventService.kt
```diff
@@ -42,7 +42,13 @@ class Cas1PlacementApplicationDomainEventService(
       deliusEventNumber = application.eventNumber,
       withdrawnAt = eventOccurredAt,
       withdrawnBy = domainEventTransformer.toWithdrawnBy(user),
-      withdrawalReason = placementApplication.withdrawalReason!!.name
+      withdrawalReason = placementApplication.withdrawalReason!!.name,
+      placementDates = placementApplication.placementDates.map {
```
pezholio, line 44: Are we going to need to backfill these? Or are we not really concerned?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/domainevents/DomainEventDescriber.kt
```diff
@@ -71,6 +75,15 @@ class DomainEventDescriber(
     return event.describe { "The assessment was appealed and ${it.eventDetails.decision.value}. The reason was: ${it.eventDetails.decisionDetail}" }
   }
 
+  private fun buildPlacementApplicationWithdrawnDescription(domainEventSummary: DomainEventSummary): String? {
```
pezholio, line 78: It'd be nice if we could make the list more human with an `and` at the end of the last item, but I don't think it's a big issue

pezholio (approved):
Couple of comments, but LGTM on the whole

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1PlacementApplicationDomainEventService.kt
davidatkinsuk: These domain events have only been raised in pre-prod so happy to not bother in this case

PR #1533, opened 2024-02-23
Add Match Request Withdrawn Domain Event

davidatkinsuk:
Currently this is only raised for the Match Request that was created automatically when the application was assessed (i.e. the one created to fulfill the arrival date specified on the application). See the documentation added to Cas1PlacementRequestDomainEventService in this commit for more information


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/domainevents/DomainEventDescriberTest.kt
```diff
@@ -283,6 +285,26 @@ class DomainEventDescriberTest {
...
+        eventType = "approved-premises.match-request.withdrawn",
+        eventDetails = MatchRequestWithdrawnFactory()
+          .withDatePeriod(DatePeriod(LocalDate.of(2024, 1, 2), LocalDate.of(2024, 3, 4)))
+          .produce(),
+      )
+    }
+
+    val result = domainEventDescriber.getDescription(domainEventSummary)
+
+    assertThat(result).isEqualTo("A request for placement was withdrawn for dates Tuesday 02 January 2024 to Monday 04 March 2024")
```
pezholio, line 305: It'd be good if we could update `cas1UiExtendedDateFormat` to remove leading zeros

davidatkinsuk: ohhh, good call! done

PR #1534, opened 2024-02-26
Optionally include 'initial' request for placement when returning placement applications

davidatkinsuk:
Relates to https://dsdmoj.atlassian.net/browse/APS-395

The /applications/{id}/placement-applications endpoint has been enhanced to add an optional query param ‘includeInitialRequestForPlacement’, which is null/false by default. If set to true, an additional PlacementApplication of type ‘initial’ will be returned representing the request for placement implicit in the application submission (i.e. the one based upon the original applications’ arrival date).

This is actually modelled as a PlacementRequest in the database, so certain elements of this specific PlacementApplication should be used with care, specifically the ‘id’ field, which actually corresponds to a PlacementRequest ID. For more information on this approach, see the documentation on PlacementRequestEntity#isForApplicationsArrivalDate

### DIFF
--- a/src/main/resources/static/_shared.yml
+++ b/src/main/resources/static/_shared.yml
@@ -535,6 +535,15 @@ components:
           - RelatedApplicationWithdrawn
           - RelatedPlacementRequestWithdrawn
           - RelatedPlacementApplicationWithdrawn
+    PlacementApplicationType:
+      type: string
+      description: |
+        'Initial' means that the request for placement was created for the arrival date included on the original application. 
+        'Additional' means the request for placement was created after the application had been authorised
+        A given application should only have, at most, one request for placement of type 'Initial'
+      enum:
+        - Initial
+        - Additional
     Booking:
       allOf:
         - $ref: '#/components/schemas/BookingBody'
@@ -3463,6 +3472,7 @@ components:
           properties:
             id:
               type: string
+              description: If type is 'Additional', provides the PlacementApplication ID. If type is 'Initial' this field provides a PlacementRequest ID.
               format: uuid
             createdByUserId:
               type: string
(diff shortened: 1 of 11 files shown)


src/main/resources/static/_shared.yml
```diff
@@ -535,6 +535,15 @@ components:
           - RelatedApplicationWithdrawn
           - RelatedPlacementRequestWithdrawn
           - RelatedPlacementApplicationWithdrawn
+    PlacementApplicationType:
+      type: string
+      description: |
+        'Initial' means that the request for placement was created for the arrival date included on the original application. 
+        'Additional' means the request for placement was created after the application had been authorised.
```
edavey, line 542: suggest "accepted" or "assessed as suitable" rather than "authorised" here

src/main/resources/static/_shared.yml
```diff
@@ -3480,6 +3490,7 @@ components:
               format: date-time
             assessmentId:
               type: string
+              description: If type is 'Additional', provides the PlacementApplication ID. If type is 'Initial' this field shouldn't be used.
```
edavey, line 3493: is it worth noting here that this is a temporary complexity, as the separate PlacementApplication entity is being removed and we'll only have PlacementRequests "soon"?

edavey (approved):
Thanks for the careful surgery.

src/main/resources/static/_shared.yml
davidatkinsuk: I suspect this may be permanent as we'd still (presumably) want to differentiate between these two types even in the new data model (although we will probably do this using subtypes I guess!)

src/main/resources/static/_shared.yml
davidatkinsuk: Good idea, done

PR #1535, opened 2024-02-26
Placement Applications should only ever be persisted with one date pair

davidatkinsuk:
Relates to https://dsdmoj.atlassian.net/browse/APS-400

Previously, the code would use a single Placement Application for requests with multiple arrival dates. This created issues when managing withdrawals because we have a requirement to be able to withdrawal a single arrival date on a given Placement Application, something that this data model does not support.

To overcome this restriction we now create a Placement Application for each requested arrival date on submission by cloning the ‘baseline’ application. All Placement Applicaitons created, including the baseline, are linked by having the same value in ‘placement_applicaiton_submission_group_id’.

Ideally we would remove the concept of mulitple placement dates in the data model, but we would first have to retrofit exisitng Placement Applications in production to match this new model, which is a non-trivial task, especially whilst those Placement Applications are still active (i.e. being assessed, awaiting placement etc.). Therefore, this code continue to support the Placement Applications with multiple arrival dates, but only for existing applications.

This PR includes quite a few pre-requisite refactors and improvements before the above commit is made, each broken in to its own commit


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/PlacementApplicationService.kt
```diff
@@ -439,7 +442,7 @@ class PlacementApplicationService(
     return placementApplicationEntity
   }
 
-  private fun getApplicationForUpdateOrSubmit(id: UUID): Either<AuthorisableActionResult<ValidatableActionResult<PlacementApplicationEntity>>, PlacementApplicationEntity> {
+  private fun <T> getApplicationForUpdateOrSubmit(id: UUID): Either<AuthorisableActionResult<ValidatableActionResult<T>>, PlacementApplicationEntity> {
```
pezholio, line 483: Very neat!

pezholio (approved):
LGTM - some really nice refactors too!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/PlacementApplicationService.kt
```diff
@@ -341,6 +349,36 @@ class PlacementApplicationService(
 
     placementDateRepository.saveAll(placementDates)
     baselinePlacementApplication.placementDates = placementDates
```
edavey, line 351: is it worth adding some logging here so we can verify that this legacy behaviour is discontinued when we expect? 

e.g. 
`WARNING: Adding multiple dates to placement application for application {applicationId}. This behaviour is deprecated.`

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/PlacementApplicationServiceTest.kt
```diff
@@ -219,9 +254,50 @@ class PlacementApplicationServiceTest {
...
+      every { cas1PlacementApplicationEmailService.placementApplicationAllocated(placementApplication) } returns Unit
+
+      val result = placementApplicationServiceLegacyWithdrawalLogic.submitApplication(
+        placementApplication.id,
+        "translatedDocument",
+        PlacementType.releaseFollowingDecision,
+        listOf(
+          PlacementDates(expectedArrival = LocalDate.of(2024, 4, 1), duration = 5),
+          PlacementDates(expectedArrival = LocalDate.of(2024, 5, 2), duration = 10),
+          PlacementDates(expectedArrival = LocalDate.of(2024, 6, 3), duration = 15),
```
edavey, line 292: it's a thing of beauty, bravo!

edavey (approved):
Thanks, this _is_ complex (as promised) but thanks for breaking it down with narrative in the commits. 

Supporting the legacy and new behaviour is clearly described in the tests and code.

👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/PlacementApplicationService.kt
```diff
@@ -268,22 +269,11 @@ class PlacementApplicationService(
   ): AuthorisableActionResult<ValidatableActionResult<PlacementApplicationEntity>> {
     val placementApplicationAuthorisationResult = getApplicationForUpdateOrSubmit(id)
 
-    when (placementApplicationAuthorisationResult) {
-      is AuthorisableActionResult.NotFound -> return AuthorisableActionResult.NotFound()
-      is AuthorisableActionResult.Unauthorised -> return AuthorisableActionResult.Unauthorised()
-      is AuthorisableActionResult.Success -> Unit
+    if (placementApplicationAuthorisationResult is Either.Left) {
+      return placementApplicationAuthorisationResult.value
```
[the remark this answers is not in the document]
davidatkinsuk: > so if the result is either NotFound or Unauthorised, we return that here? (left is unhappy, right is happy I gather?)
> 
> I can see we've been using Either for quite a while in the code, but is it worth a short note in the commit message and maybe a link to docs? e.g. https://apidocs.arrow-kt.io/arrow-core/arrow.core/-either/index.html

Yes you're understanding is correct. Either describes itself as right-biased so success would be on the right. Me and Stu had a more general chat about how we handle error/success as I think we can come up with something a bit more fluid and flexible than our current approach https://mojdt.slack.com/archives/C03K0HB0LBE/p1708961128096279.

I've updated the commit

PR #1537, opened 2024-02-26
Add appeal data to application and placement reports

danlivings-dxw:
> See [APS-241 on the Approved Premises Jira board](https://dsdmoj.atlassian.net/browse/APS-241).

This PR introduces several new columns regarding appeals to the application and placement application reports.

The new columns are as follows:

| Column name | Type | Description |
| --- | --- | --- |
| `assessmentAppealCount` | `int` | This will show the amount of appeals that have been made against this application and assessment |
| `lastAssessmentAppealedDecision` | `String` | This will show the appeal decision of the latest appeal record (Upheld or Rejected) |
| `lastAssessmentAppealedDate` | `Date` | This will show the date of the last appeal |
| `assessmentAppealedFromStatus` | `String` | This will store and show the original assessment decision so that it’s possible to see why things are being appealed |

In addition, the following columns have been updated:

| Column name | Type | Description |
| --- | --- | --- |
| `assessmentDecision` | `String` | This column now shows the decision recorded on the assessment only if no appeals have been made. Otherwise, it shows the decision of the latest appeal. |

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/reporting/model/PlacementApplicationReportRow.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/reporting/model/PlacementApplicationReportRow.kt
@@ -44,4 +44,8 @@ data class PlacementApplicationReportRow(
   val notArrivedReason: String?,
   val placementRequestType: String?,
   val paroleDecisionDate: LocalDate?,
+  val assessmentAppealCount: Int?,
+  val lastAssessmentAppealedDecision: String?,
+  val lastAssessmentAppealedDate: LocalDate?,
+  val assessmentAppealedFromStatus: String?,
 )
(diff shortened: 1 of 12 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/reporting/model/PlacementApplicationReportRow.kt
```diff
@@ -44,4 +44,8 @@ data class PlacementApplicationReportRow(
   val notArrivedReason: String?,
   val placementRequestType: String?,
   val paroleDecisionDate: LocalDate?,
+  val assessmentAppealCount: Int?,
+  val lastAssessmentAppealedDecision: String?,
+  val lastAssessmentAppealedDate: LocalDate?,
+  val assessmentAppealedFromStatus: String?,
```
edavey, line 50: perhaps I'm misunderstanding this report, but my understanding is that only assessments with the `rejected` decision are ever appealed? i.e. that this column must be definition always contain `rejected`?

edavey (approved):
Nice work, v clear.

That is except for the complex SQL! But fixing that is a much bigger challenge. One might make the case that all the data analytics (aggregate data) belongs elsewhere and that the CAS1 reports ought to consist of only raw data from our domain events.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/reporting/model/PlacementApplicationReportRow.kt
danlivings-dxw: It always contains the original assessment decision, so when there's no appeal, it's the same as the `assessmentDecision` column. When an appeal gets made, `assessmentDecision` contains the decision of the most recent appeal for that assessment, and `assessmentAppealedFromStatus` will always be `"rejected"` in that case (for the reason you mentioned).

PR #1538, opened 2024-02-26
(APS-305) Add a `dueAt` field to tasks

pezholio:
This adds a new `TaskDeadlineService` to calculate a task's due date at the point of allocation. I've also added a migration job task to backfill old tasks. The next job will be to allow sorting by due date and returning the due date in the Transformer layer, but I think we need to ensure backfilling is nailed before we move on to that part.


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/WorkingDayCountServiceTest.kt
```diff
@@ -223,4 +224,19 @@ class WorkingDayCountServiceTest {
...
+    every {
+      mockGovUKBankHolidaysApiClient.getUKBankHolidays()
+    } returns emptyBankHolidays
+
+    val date = LocalDate.now()
+
+    assertThat(
+      workingDayCountService.nextWorkingDay(date),
+    ).isEqualTo(
+      date.getNextWorkingDay(workingDayCountService.bankHolidays),
```
edavey, line 239: Is it worth trying to be more literal here, e.g.

// Fri 1st March 2024
val date = LocalDate.parse(01-03-2024)

// Mon 4th March 2024
val expectedNextWorkingDate = LocalDate.parse(04-03-2024)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/TaskDeadlineService.kt
```diff
@@ -0,0 +1,64 @@
...
+    } else {
+      val nextWorkingDay = workingDayCountService.nextWorkingDay(assessment.createdAt.toLocalDate())
+      val nextWorkingDayAt11am = nextWorkingDay.atTime(11, 0)
+      Duration.between(nextWorkingDay, nextWorkingDayAt11am)
+    }
+
+    return assessment.createdAt.plus(duration).toInstant().atOffset(ZoneOffset.UTC)
+  }
+
+  companion object {
```
edavey, line 57: Just wondering if in the commit message we should be clear on whether we are saying "within x days" or "in no more than x days" e.g.

> If it is a short notice application, it is due in 2 days

If the application is submitted today Monday Feb 26th is its assessment due before the end of Weds 28th or is it due before the end of Thurs 29th? Maybe it's worth having some explicit examples like this so that everyone is clear on the rules?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AssessmentService.kt
```diff
@@ -216,6 +217,8 @@ class AssessmentService(
       dueAt = null,
     )
 
+    assessment.dueAt = taskDeadlineService.getDeadline(assessment)
```
edavey, line 220: we're also going to need to reset the countdown when the task (assessment, placement request, match request) is **reallocated** to a new person.  See https://dsdmoj.atlassian.net/browse/APS-168?focusedCommentId=427694

edavey (approved):
👍 very clear, thks!

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/WorkingDayCountServiceTest.kt
pezholio: I've fixed the tests to lean a bit less on mocks now, so hopefully they're more readable

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/TaskDeadlineService.kt
pezholio: Good plan, I've fixed the tests a bit more to be more clear

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AssessmentService.kt
pezholio: Ah nice, that's one less ticket to worry about!

PR #1540, opened 2024-02-26
Add placement dates to Placement Application API Model

davidatkinsuk:
Relates to https://dsdmoj.atlassian.net/browse/APS-401


edavey:
For the benefit of "future us", is it worth explaining why we're doing this in the commit message? 

I _think_ it's so that the UI can build a list of individual "match requests" which are grouped onto a single "placement request" -- which is an important feature in ROTL cases where a number of visits (sets of placement dates) are proposed on the same placement application?

edavey (approved):

davidatkinsuk:
> For the benefit of "future us", is it worth explaining why we're doing this in the commit message?
> 
> I _think_ it's so that the UI can build a list of individual "match requests" which are grouped onto a single "placement request" -- which is an important feature in ROTL cases where a number of visits (sets of placement dates) are proposed on the same placement application?

Hi Ed. This change is actually much more broad than that. Currently the UI accesses these dates from data JSON despite the API storing this information in first-class properties. Also, we need to add this to the API to allow a Match request masquerading as a PlacementApplication without having to generate the corresponding JSON on the backend.

RE: the use case mentioned. It'll be multiple Requests for Placements (placement_applications) being created for a single ROTL request from the user, one for each date the user provides. That'll be available in the next PR. These will each have their own Match Request, if approved.

PR #1541, opened 2024-02-27
Only Workflow Managers (CRU) can cancel bookings

davidatkinsuk:
Relates to https://dsdmoj.atlassian.net/browse/APS-396

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserAccessService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserAccessService.kt
@@ -79,8 +79,13 @@ class UserAccessService(
     else -> false
   }
 
+  /**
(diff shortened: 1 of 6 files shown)


pezholio (approved):
LGTM - we'll probably need to reflect this change in the UI too

davidatkinsuk:
> LGTM - we'll probably need to reflect this change in the UI too

I _think_ we're ok, at least functionality. I've checked and because the booking is not returned as a withdrawable there's just no option to withdraw. The only issue we may have is if the user can't withdraw anything the list of withdrawables is empty and the screen doesn't show anything to withdraw. Again, functionally good but maybe confusing for the user

pezholio:
Ah yeah, that's a good point, I was thinking back to the old withdraw flow when there's an option to withdraw a booking. As we're pivoting to the new flow and seeing the withdrawable options, I guess it doesn't matter

PR #1542, opened 2024-02-27
Allow Assessor to update assessment

Harriethw:
DO NOT MERGE until this PR has been merged: https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1542 

This is the step [following this PR](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1524 ) and that to be merged first, once it is then this PR can be pointed at `main` branch 

https://dsdmoj.atlassian.net/jira/software/c/projects/CAS2/boards/1347?selectedIssue=CAS2-254 

This allows Assessors to update an Assessment by a PUT request to `cas2/assessments/{id}` with these string fields:

- nacroReferralId
- assessorName 

Although an Assessment belongs to an Application, I've taken the choice here to make the end point only `assessments` and not e.g. `applications/{applicationId}/assessments` because of the work we hope to do to refactor assessments described in this PR https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1514 but interested in any feedback.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/OAuth2ResourceServerSecurityConfiguration.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/OAuth2ResourceServerSecurityConfiguration.kt
@@ -59,6 +59,7 @@ class OAuth2ResourceServerSecurityConfiguration {
         authorize(HttpMethod.DELETE, "/internal/room/*", permitAll)
         authorize(HttpMethod.GET, "/events/cas2/**", hasAuthority("ROLE_CAS2_EVENTS"))
         authorize(HttpMethod.GET, "/events/**", hasAuthority("ROLE_APPROVED_PREMISES_EVENTS"))
+        authorize(HttpMethod.PUT, "/cas2/assessments/**", hasRole("CAS2_ASSESSOR"))
         authorize(HttpMethod.POST, "/cas2/submissions/*/notes", hasAnyRole("POM", "CAS2_ASSESSOR"))
         authorize(HttpMethod.GET, "/cas2/submissions/**", hasAnyRole("CAS2_ASSESSOR", "CAS2_ADMIN"))
         authorize(HttpMethod.POST, "/cas2/submissions/*/status-updates", hasRole("CAS2_ASSESSOR"))
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/AssessmentsController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/AssessmentsController.kt
@@ -1,0 +1,46 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.controller.cas2
+
+import org.springframework.http.ResponseEntity
+import org.springframework.stereotype.Service
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.cas2.AssessmentsCas2Delegate
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Cas2Assessment
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.UpdateCas2Assessment
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.BadRequestProblem
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.ConflictProblem
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.ForbiddenProblem
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.NotFoundProblem
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.results.AuthorisableActionResult
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.results.ValidatableActionResult
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas2.AssessmentService
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.cas2.AssessmentsTransformer
+
+@Service("Cas2AssessmentsController")
+class AssessmentsController(
+  private val assessmentService: AssessmentService,
+  private val assessmentsTransformer: AssessmentsTransformer,
+) : AssessmentsCas2Delegate {
+
+  override fun assessmentsAssessmentIdPut(
+    assessmentId: java.util.UUID,
+    updateCas2Assessment: UpdateCas2Assessment,
+  ):
+    ResponseEntity<Cas2Assessment> {
+    val assessmentResult = assessmentService.updateAssessment(assessmentId, updateCas2Assessment)
+    val validationResult = when (assessmentResult) {
+      is AuthorisableActionResult.NotFound -> throw NotFoundProblem(assessmentId, "Assessment")
+      is AuthorisableActionResult.Unauthorised -> throw ForbiddenProblem()
+      is AuthorisableActionResult.Success -> assessmentResult.entity
+    }
+
+    val updatedAssessment = when (validationResult) {
+      is ValidatableActionResult.GeneralValidationError -> throw BadRequestProblem(errorDetail = validationResult.message)
+      is ValidatableActionResult.FieldValidationError -> throw BadRequestProblem(invalidParams = validationResult.validationMessages)
+      is ValidatableActionResult.ConflictError -> throw ConflictProblem(id = validationResult.conflictingEntityId, conflictReason = validationResult.message)
+      is ValidatableActionResult.Success -> validationResult.entity
+    }
+
+    return ResponseEntity.ok(
+      assessmentsTransformer.transformJpaToApiRepresentation(updatedAssessment),
+    )
+  }
+}
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/AssessmentService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/AssessmentService.kt
@@ -1,9 +1,13 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas2
 
+import org.springframework.data.repository.findByIdOrNull
 import org.springframework.stereotype.Service
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.UpdateCas2Assessment
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas2ApplicationEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas2AssessmentEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas2AssessmentRepository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.results.AuthorisableActionResult
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.results.ValidatableActionResult
 import java.time.OffsetDateTime
 import java.util.UUID
 import javax.transaction.Transactional
@@ -22,4 +26,20 @@ class AssessmentService(
         application = cas2ApplicationEntity,
       ),
     )
+
+  fun updateAssessment(assessmentId: UUID, newAssessment: UpdateCas2Assessment): AuthorisableActionResult<ValidatableActionResult<Cas2AssessmentEntity>> {
+    val assessmentEntity = assessmentRepository.findByIdOrNull(assessmentId)
+      ?: return AuthorisableActionResult.NotFound()
+
+    assessmentEntity.apply {
+      this.nacroReferralId = newAssessment.nacroReferralId
+      this.assessorName = newAssessment.assessorName
+    }
+
+    val savedAssessment = assessmentRepository.save(assessmentEntity)
+
+    return AuthorisableActionResult.Success(
+      ValidatableActionResult.Success(savedAssessment),
+    )
+  }
 }
--- a/src/main/resources/static/_shared.yml
+++ b/src/main/resources/static/_shared.yml
@@ -2453,10 +2453,25 @@ components:
     Cas2Assessment:
       type: object
       properties:
+        id:
+          type: string
+          format: uuid
         nacroReferralId:
           type: string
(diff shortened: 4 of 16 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2AssessmentTest.kt
```diff
@@ -0,0 +1,144 @@
...
+        .exchange()
+        .expectStatus()
+        .isForbidden
+    }
+  }
+
+  @Nested
+  inner class PutToUpdate {
+    @Test
+    fun `assessors create note returns 201`() {
```
danhumphreys-moj, line 89: Are there some unhappy path tests too?

validationResult NotFound - throw
validationResult Unauthorised - throw

updatedAssessment GeneralValidationError - throw
updatedAssessment FieldValidationError - throw
updatedAssessment ConflictError - throw

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/AssessmentService.kt
```diff
@@ -22,4 +26,20 @@ class AssessmentService(
         application = cas2ApplicationEntity,
       ),
     )
+
+  fun updateAssessment(assessmentId: UUID, newAssessment: UpdateCas2Assessment): AuthorisableActionResult<ValidatableActionResult<Cas2AssessmentEntity>> {
```
danhumphreys-moj, line 30: Is `assessmentId` a property of `UpdateCas2Assessment`?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/OAuth2ResourceServerSecurityConfiguration.kt
```diff
@@ -59,6 +59,7 @@ class OAuth2ResourceServerSecurityConfiguration {
         authorize(HttpMethod.DELETE, "/internal/room/*", permitAll)
         authorize(HttpMethod.GET, "/events/cas2/**", hasAuthority("ROLE_CAS2_EVENTS"))
         authorize(HttpMethod.GET, "/events/**", hasAuthority("ROLE_APPROVED_PREMISES_EVENTS"))
+        authorize(HttpMethod.PUT, "/cas2/assessments/**", hasRole("CAS2_ASSESSOR"))
```
danhumphreys-moj, line 62: This role is different to the `ROLE_CAS2_ADMIN` role tested in [this commit](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1542/commits/8e95b58b8ff260d09f2d81f20de24530bca62d82).  Should both roles be the same?

Ref: `fun 'updating an assessment is forbidden to external users who are not Assessors'()`

src/main/resources/static/_shared.yml
```diff
@@ -2444,12 +2444,27 @@ components:
...
+        id:
+          type: string
+          format: uuid
+        nacroReferralId:
+          type: string
+        assessorName:
+          type: string
+      required:
+        - id
+    UpdateCas2Assessment:
```
danhumphreys-moj, line 2458: Can we not get away with having one object to create and update an assessment?  i.e. the same object is passed to  `POST` and `PUT` if that makes sense?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/factory/Cas2AssessmentEntityFactory.kt
```diff
@@ -0,0 +1,44 @@
...
+
+class Cas2AssessmentEntityFactory : Factory<Cas2AssessmentEntity> {
+  private var id: Yielded<UUID> = { UUID.randomUUID() }
+  private var createdAt: Yielded<OffsetDateTime> = { OffsetDateTime.now() }
+  private var application: Yielded<Cas2ApplicationEntity> = {
+    Cas2ApplicationEntityFactory()
+      .withCreatedByUser(NomisUserEntityFactory().produce())
+      .produce()
+  }
+  private var nacroReferralId: Yielded<String?> = { null }
```
danhumphreys-moj, line 18: I might be wrong but I think we only need to use `Yielded` when we are generating [discrete values for every object](https://github.com/bluegroundltd/kfactory?tab=readme-ov-file#generating-dynamic-values-every-time), when they're the same for every object generated by the factory I think we won't need it.  But haven't checked this.

danhumphreys-moj (commented):
A couple of things about this work;
- how are we going to deal with creating assessent records for applications already assessed in the live environment.
- Kind of related but I recall from the work last week that the relationship between applications and assessments is 1:1, whereas it is initially 1:0, then 1:1 once the application is submitted.  If we enforce 1:1 in the prod database will we get database integrity errors for those submitted applications which don't have assessment records.  Can't recall the PK-FK relationships and if they will be enforced to be 1:1.

src/main/resources/static/_shared.yml
Harriethw: we will not have `POST` endpoint as they are created by the service when the Application is submitted rather than by an HTTP request. The model to Update currently doesn't have an `id`, and the fields for `nacroReferralId` and `assessorName` are required, whereas they are not for the `Cas2Assessment` model, and it's likely to diverge later, and it means we have an object with a single responsibility of updating an Assessment.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/OAuth2ResourceServerSecurityConfiguration.kt
Harriethw: not sure what  you mean here - CAS2_ASSESSORS are able to update assessments, but CAS2_ADMINs are not.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/AssessmentService.kt
Harriethw: nope, the `assessmentId` is a path param e.g. `cas2/assessments/{assessmentId}`

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2AssessmentTest.kt
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  @Nested
+  inner class ControlsOnExternalUsers {
+    @Test
+    fun `updating an assessment is forbidden to external users who are not Assessors`() {
+      val jwt = jwtAuthHelper.createClientCredentialsJwt(
+        username = "username",
+        authSource = "auth",
+        roles = listOf("ROLE_CAS2_ADMIN"),
```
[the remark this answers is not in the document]
Harriethw: nope, they have the role `CAS2_ASSESSORS` which is why are testing here that Admin's can't update an assessment

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/AssessmentsController.kt
```diff
@@ -0,0 +1,46 @@
...
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas2.AssessmentService
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.cas2.AssessmentsTransformer
+
+@Service("Cas2AssessmentsController")
+class AssessmentsController(
+  private val assessmentService: AssessmentService,
+  private val assessmentsTransformer: AssessmentsTransformer,
+) : AssessmentsCas2Delegate {
+
+  override fun assessmentsAssessmentIdPut(
```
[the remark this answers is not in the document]
Harriethw: these are generated function names that need to match in order for us to override them, see the `AssessmentsCas2Delegate`

Harriethw:
@danhumphreys-moj thanks Dan!
> how are we going to deal with creating assessent records for applications already assessed in the live environment.

In the parent PR to this one we have the migration that we'll be doing to ensure we can do this safely

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/AssessmentsController.kt
```diff
@@ -0,0 +1,46 @@
...
+    assessmentId: java.util.UUID,
+    updateCas2Assessment: UpdateCas2Assessment,
```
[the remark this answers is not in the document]
Harriethw: it's not yet been updated when it's passed in here, so I think I'd rather not use the past tense :)

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2AssessmentTest.kt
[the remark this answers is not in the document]
danhumphreys-moj: Ahh yes of course.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/OAuth2ResourceServerSecurityConfiguration.kt
danhumphreys-moj: Resolved above - my misunderstanding.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2AssessmentTest.kt
Harriethw: we've tested Unauthorised in the above tests, which returns Forbidden.
Good spot that `NotFound` isn't tested, but I think I will put that in the Service unit tests - I'm not entirely sure the others warrant an expensive integration test, from my understanding the validation/conflict errors are not actually being triggered by the Service method here, we just have to account for them to fulfil the `when` block.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/AssessmentsController.kt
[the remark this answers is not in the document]
danhumphreys-moj: Fair enough, `UpdateCas2Assessment` sounded more verb-like, more like a function.  Was thinking something to make it more noun-like to make it more like an object.

src/main/resources/static/_shared.yml
danhumphreys-moj: OK, yes `createAssessment()` is called after submitting the application.

I think whether you are creating or updating something, ultimately it could be the same object that is passed around; and that the object could have an mandatory `assessmentId` and nullable `nacroReferralId` and `assessorName` fields.  Perhaps this is because we are making fields mandatory in the front end and this is an Open API restriction.  I'll figure it out..

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/AssessmentService.kt
danhumphreys-moj: OK, trying to understand why `assessmentId` wouldn't be a property of the object we want to update.  Don't let me distract you though @Harriethw..

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/AssessmentsController.kt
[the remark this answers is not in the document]
Harriethw: I get what you mean but I think in this case as it is pointing to an `UpdateCas2Assessment` model, it makes sense that it would just be the camel-case version of it - we're also following the naming conventions in other controllers like the CAS1 Assessment Controller.
I think if there weren't the precedent to follow I would have called these types of models `AssessmentUpdate` or `ApplicationUpdate` but I think it makes more sense to follow the set patterns now.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/factory/Cas2AssessmentEntityFactory.kt
Harriethw: yep, good point! will update

Gweaton (approved):

src/main/resources/static/_shared.yml
```diff
@@ -2469,9 +2469,6 @@ components:
           type: string
         assessorName:
           type: string
-      required:
```
Gweaton: I'll probably squash this into the original commit where it was introduced, but @Harriethw is there a way we tend to test the non-requiredness of these fields or do you think this behaviour is covered by the fact we're using OpenAPI-generated code and the existing integration/unit tests?

Harriethw: hmm actually thinking about it today, I think we need to go a different approach and make the fields required, but set `nullable: true`.
When this model is used in the Service it treats it as a replacement for all the existing data and overwrites both fields, and if we make the fields not required I'm concerned that implies to the frontend if it doesn't supply one of the fields then it won't be updated?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/transformer/cas2/SubmissionsTransformerTest.kt
```diff
@@ -91,9 +98,6 @@ class SubmissionsTransformerTest {
         listOf(mockStatusUpdateApi, mockStatusUpdateApi),
       )
 
-      assertThat(transformation.assessment)
```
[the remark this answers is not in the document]
Gweaton: This is covered by the `every` as it would let us know if it wasn't called.

src/main/resources/static/_shared.yml
Gweaton: We've discovered this doesn't work as expected, so we've removed `nullable: true` but we're allowing submission of the assessment keys with an empty string value.

Gweaton:
Reviewed as a pair with Harriet - merging this!

PR #1543, opened 2024-02-27
[CAS-229]-CAS3-reports-Configure and generate based on configured number of months

sudhakar-plr:
Refer [CAS-195](https://dsdmoj.atlassian.net/browse/CAS-195) for more information.

This PR is to run the CAS3 reports for 3 months instead of 1 month to see the performance of the report(testing purpose only).

This change is just for testing purpose to make sure the service performance is acceptable and work without any issue before we change the service to search based on `startDate` and `endate` instead of searching only for 1 month.

[CAS-195]: https://dsdmoj.atlassian.net/browse/CAS-195?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/helm_deploy/values-preprod.yaml
+++ b/helm_deploy/values-preprod.yaml
@@ -112,3 +112,6 @@ generic-prometheus-alerts:
 
 env_details:
   production_env: false
+
+cas3-report:
+  end-date-override: 3
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/reporting/generator/BedUsageReportGenerator.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/reporting/generator/BedUsageReportGenerator.kt
@@ -1,5 +1,6 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.reporting.generator
 
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ServiceName.temporaryAccommodation
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.BedEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.BookingRepository
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.LostBedsRepository
@@ -18,6 +19,7 @@ class BedUsageReportGenerator(
   private val bookingRepository: BookingRepository,
   private val lostBedsRepository: LostBedsRepository,
   private val workingDayCountService: WorkingDayCountService,
+  private val cas3EndDateOverride: Int,
 ) : ReportGenerator<BedEntity, BedUsageReportRow, BedUsageReportProperties>(BedUsageReportRow::class) {
   override fun filter(properties: BedUsageReportProperties): (BedEntity) -> Boolean = {
     checkServiceType(properties.serviceName, it.room.premises) &&
@@ -26,7 +28,12 @@ class BedUsageReportGenerator(
 
   override val convert: BedEntity.(properties: BedUsageReportProperties) -> List<BedUsageReportRow> = { properties ->
     val startOfMonth = LocalDate.of(properties.year, properties.month, 1)
-    val endOfMonth = LocalDate.of(properties.year, properties.month, startOfMonth.month.length(startOfMonth.isLeapYear))
+
+    val endOfMonth = if (properties.serviceName == temporaryAccommodation && cas3EndDateOverride != 0) {
+      startOfMonth.plusMonths(cas3EndDateOverride.toLong())
+    } else {
+      LocalDate.of(properties.year, properties.month, startOfMonth.month.length(startOfMonth.isLeapYear))
+    }
 
     val bookings = bookingRepository.findAllByOverlappingDateForBed(startOfMonth, endOfMonth, this)
     val voids = lostBedsRepository.findAllByOverlappingDateForBed(startOfMonth, endOfMonth, this)
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/reporting/generator/BedUtilisationReportGenerator.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/reporting/generator/BedUtilisationReportGenerator.kt
@@ -1,5 +1,6 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.reporting.generator
 
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ServiceName.temporaryAccommodation
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.BedEntity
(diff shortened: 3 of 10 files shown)


helm_deploy/values-preprod.yaml
```diff
@@ -112,3 +112,6 @@ generic-prometheus-alerts:
 
 env_details:
   production_env: false
+
+cas3-report:
+  end-date-override: 3
```
davidatkinsuk, line 117: I could be wrong on this but if you want to provide config to spring I thought we had to:

1. Define the baseline value in application.yml
2. Define the helm files using the CAPITAL_STYLE e.g. USER-ALLOCATIONS_RULES_LEGACY-ALLOCATOR_ENABLED?

Looking at other lowercased entries in this file they don't appear to be used anywhere in spring as defined?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas3/ReportService.kt
```diff
@@ -18,13 +19,19 @@ class ReportService(
   private val offenderService: OffenderService,
   private val userService: UserService,
   private val transitionalAccommodationReferralReportRowRepository: TransitionalAccommodationReferralReportRepository,
+  @Value("\${cas3-report.end-date-override:0}") private val endDateOverride: Int,
```
davidatkinsuk, line 22: i'd much rather we used null to represent absence of value instead of 0, if at all possible

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/reporting/generator/BedUsageReportGenerator.kt
```diff
@@ -26,7 +28,11 @@ class BedUsageReportGenerator(
 
   override val convert: BedEntity.(properties: BedUsageReportProperties) -> List<BedUsageReportRow> = { properties ->
     val startOfMonth = LocalDate.of(properties.year, properties.month, 1)
-    val endOfMonth = LocalDate.of(properties.year, properties.month, startOfMonth.month.length(startOfMonth.isLeapYear))
+    var endOfMonth = LocalDate.of(properties.year, properties.month, startOfMonth.month.length(startOfMonth.isLeapYear))
```
davidatkinsuk, line 31: could you please use a val here and assign using if/else to avoid value mutation

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/reporting/generator/BedUtilisationReportGenerator.kt
```diff
@@ -26,7 +28,11 @@ class BedUtilisationReportGenerator(
 
   override val convert: BedEntity.(properties: BedUtilisationReportProperties) -> List<BedUtilisationReportRow> = { properties ->
     val startOfMonth = LocalDate.of(properties.year, properties.month, 1)
-    val endOfMonth = LocalDate.of(properties.year, properties.month, startOfMonth.month.length(startOfMonth.isLeapYear))
+    var endOfMonth = LocalDate.of(properties.year, properties.month, startOfMonth.month.length(startOfMonth.isLeapYear))
```
davidatkinsuk, line 31: could you please use a val here and assign using if/else to avoid value mutation

davidatkinsuk (changes requested):
A few comments from me, cheers!

helm_deploy/values-preprod.yaml
sudhakar-plr: Defining the properties in application.yaml is to set the property to all env as default and you override the value in different env files if you need. In this case we want to test the performance in only pre-prod env and its not applicable to Prod.
The UPPER case property in helm file is to add the property as env variable which you can change without making code change.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas3/ReportService.kt
sudhakar-plr: I defaulted to zero instead of null to avoid if anyone sets the property value to 0.

davidatkinsuk: I think that's addressing two different concerns though?

1. Shouldn't allow someone to set it to 0
2. Default behaviour when not set

IMO opinion it would be clearer if default was null but as this is presumably temp code can leave as is

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/reporting/generator/BedUsageReportGenerator.kt
sudhakar-plr: We cant use if/else here because the endmonth has to referenced in DB query. The purpose of var is to change the value of the variable and i couldnt see any issue in using it in this context.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/reporting/generator/BedUtilisationReportGenerator.kt
sudhakar-plr: We cant use if/else here because the endmonth has to referenced in DB query. The purpose of var is to change the value of the variable and i couldnt see any issue in using it in this context.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/reporting/generator/BedUsageReportGenerator.kt
davidatkinsuk: In kotlin if statements can return values so you can do like this:

`
val endOfMonth = if (properties.serviceName == temporaryAccommodation && cas3EndDateOverride != 0) {
      startOfMonth.plusMonths(cas3EndDateOverride.toLong())
} else {
      LocalDate.of(properties.year, properties.month, startOfMonth.month.length(startOfMonth.isLeapYear))
}
`

I prefer to avoid vars because mutation typically makes code more complex. Obviously sometimes it's easier to just use a var as code as can get contrived otherwise but in this case i think the if/else construct works nicely. I'll approve regardless and up to you if you want to change it.

davidatkinsuk (approved):
LGTM 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/reporting/generator/BedUsageReportGenerator.kt
sudhakar-plr: nice, still learning kotlin :-). Done changes as suggetsed

davidatkinsuk: awesome, looks good!

PR #1547, opened 2024-02-28
Seed Assessments on Submitted Applications for CAS2

Harriethw:
~~DO NOT MERGE until we have tested the migration job we are currently testing has passed.~~ this can be merged now as we successfully ran the `update_cas2_applications_with_assessments` job in all environments and checked that all non-seeded submitted applications have assessments ☺️ 

we [previously introduced a function](https://github.com/ministryofjustice/hmpps-approved-premises-api/commit/e479df3f5556005cafb15598aed944668b67353e) to create an Assessment when an Application is submitted. 

This commit ensures our seeded data matches what we would expect to see on real submitted applications.

I've used the `@SuppressWarnings("LongParameterList")` annotation because I don't see a way of reducing the parameters without making it less readable, any suggestions welcome!


danhumphreys-moj:
Given the short time I've been on this project - looks good to me!

I can see this job populates the database with assessment records for submitted applications.  Then interrogates the database to ensure that: submitted applications have equivalent records in the assessment table; and in-progress applications do not.

danhumphreys-moj (commented):
Given the short time I've been on this project - looks good to me!

I can see this job populates the database with assessment records for submitted applications. Then interrogates the database to ensure that: submitted applications have equivalent records in the assessment table; and in-progress applications do not.

tahb (approved):
LGTM 🚀 

I wondered if we ought to seed some assessments with values but given they're optional from the users point of view I don't think it matters.

Gweaton (approved):

Harriethw:
> I wondered if we ought to seed some assessments with values but given they're optional from the users point of view I don't think it matters.

@tahb ohh good idea. could be good to add in next time as I want to get this in before I go 🙈

PR #1548, opened 2024-02-28
Use new Placement Request Submitted email template when splitting requests into 'one per date'

davidatkinsuk:
Relates to https://dsdmoj.atlassian.net/browse/APS-431

If using the new withdrawal logic that splits requests for placements into an entity per date, use the new email template, allowing us to include more information on the email without breaking emails currently being sent in live


pezholio (approved):
LGTM 👍

PR #1549, opened 2024-02-28
Only submitted placement requests are withdrawable

davidatkinsuk:
Relates to https://dsdmoj.atlassian.net/browse/APS-432

Previously we were returning unsubmitted placement requests. Because these don’t neccessarily have dates associated with them, an error occurred when rendering the withdrawable on the UI. We should not be returning unsubmitted placement apps as they are not surfaced anywhere else on the UI once the user navigates away from the creation forms.


richpjames (approved):
Approving as this is time sensitive and it makes sense to me, but may be worth someone with better knowledge of the code base taking a look if we have time.

PR #1554, opened 2024-02-29
(APS-305) Show deadline in confirmation email

pezholio:
This alters the confirmation email copy to ensure that the deadline information is properly communicated to the user. It also sends a different email template if the assessment has been created as the result of an appeal.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1AssessmentEmailService.kt
```diff
@@ -0,0 +1,44 @@
...
+
+import org.springframework.beans.factory.annotation.Value
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.config.NotifyConfig
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.EmailNotifier
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.UrlTemplate
+import java.util.UUID
+
+@Service
+class Cas1AssessmentEmailService(
```
davidatkinsuk, line 17: love it!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/util/Dates.kt
```diff
@@ -81,6 +85,10 @@ fun LocalDate.getNextWorkingDay(bankHolidays: List<LocalDate>): LocalDate {
   return result
 }
 
+fun LocalDate.toUiFormat(): String = this.format(cas1UiExtendedDateFormat)
+
+fun OffsetDateTime.toUiFormattedTime(): String = this.format(cas1UiTimeFormat).lowercase()
```
davidatkinsuk, line 90: awesome

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/util/DatesTest.kt
```diff
@@ -101,4 +104,18 @@ class DatesTest {
...
+    val date = LocalDate.parse("2024-01-01")
+
+    assertThat(date.toUiFormat()).isEqualTo("Monday 1 January 2024")
+  }
+
+  @Test
+  fun `toUiFormattedTime formats a LocalDate correctly`() {
+    val dateTime = OffsetDateTime.parse("2024-01-01T11:15:00Z")
+
+    assertThat(dateTime.toUiFormattedTime()).isEqualTo("11am")
```
davidatkinsuk, line 119: Should this maybe be toUiFormattedHourOfDay, given it's lobbing off the 15 minutes? (I guess that's deliberate)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1AssessmentEmailService.kt
```diff
@@ -41,4 +49,38 @@ class Cas1AssessmentEmailService(
...
+      val deadlineDate = deadline.toLocalDate()
+      return if (deadlineDate.isEqual(LocalDate.now())) {
+        SAME_DAY_EMERGENCY_DEADLINE_COPY
+      } else {
+        NEXT_WORKING_DAY_EMERGENCY_DEADLINE_COPY.format(deadline.toUiFormattedTime(), deadlineDate.toUiFormat())
+      }
+    }
+
+    return STANDARD_DEADLINE_COPY.format(
+      workingDayCountService.getWorkingDaysCount(LocalDate.now(), deadline.toLocalDate()).toString(),
```
davidatkinsuk, line 87: above you're using toUiFormat for a date, but here using .toString(). Not sure if that's deliberate?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1AssessmentEmailService.kt
```diff
@@ -50,6 +50,25 @@ class Cas1AssessmentEmailService(
     }
   }
 
+  fun appealedAssessmentAllocated(allocatedUser: UserEntity, assessmentId: UUID, crn: String) {
+    allocatedUser.email?.let { email ->
```
davidatkinsuk, line 54: A wider point - we do this all over the place - is it expected that a user won't have an email? I wonder if we should be logging something if they don't (and come up with a clever way to do that everywhere). Maybe we should pass the user to the email service and let that do the ignore if null check and log if missing?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1AssessmentEmailService.kt
```diff
@@ -50,6 +50,25 @@ class Cas1AssessmentEmailService(
...
+        recipientEmailAddress = email,
+        templateId = notifyConfig.templates.appealedAssessmentAllocated,
+        personalisation = mapOf(
+          "name" to allocatedUser.name,
+          "assessmentUrl" to assessmentUrlTemplate.resolve("id", assessmentId.toString()),
+          "crn" to crn,
+          "deadlineCopy" to mapOf(
+            "name" to allocatedUser.name,
+            "assessmentUrl" to assessmentUrlTemplate.resolve("id", assessmentId.toString()),
+            "crn" to crn,
```
davidatkinsuk, line 65: this doesn't appear to align with the template in notify. Is deadline copy meant to be a map?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1AssessmentEmailService.kt
```diff
@@ -50,6 +50,25 @@ class Cas1AssessmentEmailService(
     }
   }
 
+  fun appealedAssessmentAllocated(allocatedUser: UserEntity, assessmentId: UUID, crn: String) {
```
davidatkinsuk, line 53: if is it worth updating an integration test to make sure this is sent in this circumstance?

davidatkinsuk (changes requested):
Some comments but nothing major

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1AssessmentEmailService.kt
pezholio: Yeah, `getWorkingDaysCount` returns an integer, so we're casting to a string to be used in the template.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1AssessmentEmailService.kt
pezholio: I'm not 100% sure, I think we do this because `email` is nullable. Looking at preprod, we only have 4 users without email addresses.

davidatkinsuk: ok, i'll keep pondering on it :)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1AssessmentEmailService.kt
pezholio: Whoops!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1AssessmentEmailService.kt
pezholio: Oh good call - I also realised I had the logic in the reallocation method, not the assessment creation method, so have fixed that now 😅

davidatkinsuk (approved):
LGTM 👍

PR #1555, opened 2024-02-29
[CAS-195]-Correcting the env config value setup to download report for 3 months

sudhakar-plr:
This PR is to correct the env property setup before as its not picking the value.


pezholio (approved):
LGTM 👍

PR #1558, opened 2024-03-01
Withdrawal Tree Spike to enable blocking withdrawl

davidatkinsuk:
This spike rewrites withdrawal cascading to be centrally managed, using a tree strucutre of related entities to determine what needs withdrawing when cascading.

This approach will make it simple to block ancestor elements being withdrawn when bookings have arrivals. This method on WithdrawalTreeNode can be used for this purpose:

fun isBlocked(): Boolean = status.blocking || children.any { it.isBlocked() }

This recursively drills down the tree to see if any descendents are blocking (to be tested!)

The tree is also used to get the initial list of withdrawables

There are various TODOs to improve things. It would be good if the top level withdrawal request was orchestrated through the tree too but this may be awkward because each entity has a diffierent type used to define the withdrawal reason. BUT - if we didn’t do this i’m not sure how we’d check if the ‘top level’ entity being withdrawn is blocked or not (although the user shouldn’t be requesting this in the first place as it will be marked as blocked when returning withdrawables, but it would be nice to be consistent just-in-case they somehow manage it)

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ApplicationsController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ApplicationsController.kt
@@ -540,17 +540,17 @@ class ApplicationsController(
       throw RuntimeException("Unsupported Application type: ${application::class.qualifiedName}")
     }
 
-    val withdrawables = withdrawableService.allWithdrawables(application, user)
-
-    val allWithdrawables =
-      if (withdrawables.application) {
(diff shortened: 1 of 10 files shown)


davidatkinsuk:
as per comment on commit i think we do need to run all withdrawals via the tree. that way we can do the top level user permissions check consistently by using the boolean flag we've added to the model. IT should be much easier to comprehend too. Just need to work out how to add the 'top level' specific withdrawal reason into the WithrawalContext - probably by using a sealed class for 'reason' within the model, one for each top-level target type. receiver will then have to cast it to get information. Not sure if we can type that to the triggeringEntityType - maybe use the sealed class itself to identify the type instead of an enum?

PR #1560, opened 2024-03-01
Cas2 114 extend status detail domain event

danhumphreys-moj:


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/factory/events/cas2/Cas2StatusFactory.kt
```diff
@@ -3,11 +3,17 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.factory.events.cas2
 import io.github.bluegroundltd.kfactory.Factory
 import io.github.bluegroundltd.kfactory.Yielded
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.cas2.model.Cas2Status
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.cas2.model.StatusDetailItem
 
 class Cas2StatusFactory : Factory<Cas2Status> {
   private var name: Yielded<String> = { "moreInfoRequested" }
   private var label: Yielded<String> = { "More information requested" }
   private var description: Yielded<String> = { "More information about the application has been requested" }
+  private var statusDetails = listOf(
```
Gweaton, line 12: I think it might be worth adding a bit more specificity to the commit message for this change to "Added status details to persisted domain event **factory object** so that application status update GET test ensures they are returned..." as I was initially expecting this commit to update the real persisted object (i.e. non-test data).

Gweaton (approved):
This looks good to me - thanks for breaking down into neat commits with descriptive messages - just one bit of feedback that might help other readers in future.

Might be worth waiting for an approval from Tom or Harriet on this as this is a bit new to me, but the technical approach looks good!

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/factory/events/cas2/Cas2StatusFactory.kt
danhumphreys-moj: Done this - thanks George.

danhumphreys-moj:
> This looks good to me - thanks for breaking down into neat commits with descriptive messages - just one bit of feedback that might help other readers in future.
> 
> Might be worth waiting for an approval from Tom or Harriet on this as this is a bit new to me, but the technical approach looks good!

Thanks for approving - no problem, will wait for @tahb's approval too.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/StatusUpdateService.kt
```diff
@@ -132,6 +135,7 @@ class StatusUpdateService(
               name = newStatus.name,
               description = newStatus.description,
               label = newStatus.label,
+              statusDetails = statusDetails.let { transformer.transformStatusDetailListToDetailItemList(it) },
```
danhumphreys-moj, line 138: No need for this to be a `let`; left over from when statusDetails had a null default value.  Removed and commit squashed.  @Gweaton conversant.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/factory/events/cas2/Cas2StatusFactory.kt
```diff
@@ -3,11 +3,17 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.factory.events.cas2
...
+    StatusDetailItem("exclusionZonesAndAreas", "Exclusion zones and preferred areas"),
+    StatusDetailItem("riskOfSeriousHarm", "Risk of serious harm"),
```
Harriethw, line 14: quite minor but I wonder as this field is potentially empty, if for the factory default we set it as an `emptylist()`, and then users of the factory can have a `withStatusDetails` function to add them if needed, otherwise we might be populating unecessary/unnoticed data

src/main/resources/static/cas2-domain-events-api.yml
```diff
@@ -185,10 +185,26 @@ components:
         description:
           type: string
           example: 'More information about the application has been requested from the POM (Prison Offender Manager).'
+        statusDetails:
+          type: array
+          items:
+            $ref: '#/components/schemas/StatusDetailItem'
       required:
         - name
```
Harriethw, line 193: should `statusDetails` be added as a required field here? We should always return the field even if it's an empty list? (not sure myself)

src/main/resources/static/cas2-domain-events-api.yml
```diff
@@ -185,10 +185,26 @@ components:
...
         - label
         - description
+    StatusDetailItem:
```
Harriethw, line 196: I wonder if a better name would just be `StatusDetail` or `Cas2StatusDetail` to following existing naming convention. I don't think `Item` is needed?

Harriethw (commented):
Looks good Dan!
I just had a question about whether `statusDetails` should be a required field on the `Cas2Status` component which I think would be good to confirm before this is merged...

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/factory/events/cas2/Cas2StatusFactory.kt
danhumphreys-moj: Changed as per your comment in a more recent commit.

src/main/resources/static/cas2-domain-events-api.yml
danhumphreys-moj: `statusDetails` has been made required in a recent commit.

src/main/resources/static/cas2-domain-events-api.yml
danhumphreys-moj: Changed the name of this property from `StatusDetailItem` to `Cas2StatusDetail` in a more recent commit.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas2/StatusUpdateServiceTest.kt
```diff
@@ -167,9 +168,10 @@ class StatusUpdateServiceTest {
                 noms = "NOMSABC",
               ) &&
                 it.data.eventDetails.newStatus == Cas2Status(
-                name = "moreInfoRequested",
-                label = "More information requested",
-                description = "The prison offender manager (POM) must provide information requested for the application to progress.",
+                  name = "moreInfoRequested",
```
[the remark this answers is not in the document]
danhumphreys-moj: Strange, I think there are linting issues in main, running `./gradlew ktlintCheck` on main branch shows `BUILD FAILED`.  Running `./gradlew ktlintFormat` fixes lots of files and then when `./gradlew ktlintCheck` is run after the changes we get `BUILD SUCCESSFUL`.

Which leads to why the CI build stage isn't falling over if linting fails?  Unless this is per design?

Harriethw (approved):
thanks Dan! happy to take a look at fixing up commits and handling conflicts next time :)

PR #1564, opened 2024-03-01
Seed Assessments on AutoScripted CAS2 Applications

Gweaton:
We [recently introduced a seed job](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/15470) for creating Assessments when creating seeded Applications. 

This worked for the 6 Applications created by the seed job, but we were unaware of the existence of many other Applications that were created by the AutoScript method.

This PR ensures Assessments are seeded for AutoScripted Applications, and adds a "how-to" guide for CAS2 seeded data to hopefully make this clearer in future, as we spent a lot of time debugging the issue without realising there was this separate way of seeding applications.


doc/how-to/seed_cas2_data_in_non_production_environments.md
```diff
@@ -0,0 +1,50 @@
+# How to seed CAS2 data in non-production environments
+
+The seeding of `Cas2Application` and `Cas2Assessment` data is mostly done automatically for developers when the API
```
tahb, line 3: I think there might also be value in talking a little about the way seeded data is cleared using these SQL scripts. As in, seeded data is meant to applied afresh on every app restart. This means that seeded data should likely be paired with a relevant `TRUNCATE` found in the `resources/db/migration/<env>.sql`.

doc/how-to/seed_cas2_data_in_non_production_environments.md
```diff
@@ -0,0 +1,50 @@
...
+server is started up on the local, dev and test environments, creating a number of Applications and Assessments that can
+be used to test out the API or UI.
+
+Due to some confusion between the two ways of seeding data, this how-to aims to clarify the use case for each type of
+seeding.
+
+## Seed jobs
```
tahb, line 10: For applications there's that gotcha around seeding also setting `created_at` so checking your DB directly to ensure seeds are run is not the way to test it.

tahb (dismissed):
Very nice!

Some minor suggestions for the new documentation that don't prevent a merge 👍

doc/how-to/seed_cas2_data_in_non_production_environments.md
Gweaton: Yeah, might be useful to prevent people getting confused like we did! Have added.

tahb (approved):

PR #1565, opened 2024-03-03
Orchestrate all CAS1 Withdrawals via the Withdrawal Service

davidatkinsuk:
This PR builds upon https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1563, refactoring the code to make it possible to block entities from being withdrawn if they relate a booking that has arrivals.

In this PR we refactor the code to trigger all CAS1 withdrawals using the Withdrawal Service (instead of controllers going directly to their corresponding Entity's service).

Using the Withdrawal Service to orchestrate the complete withdrawal process has the following benefits:

1. It makes it very easy to implement the 'blocking' functionality which is the ultimate goal of this refactoring.
2. It allows for all permission checking related to withdrawals (e.g. can the user directly withdraw this entity, is this entity actually withdrawable) to be performed in one place, using the Withdrawable Tree to determine this in a standardised manner.
3. It removes duplicate functionality from entity-specific services related to permissions checking and triggering cascade withdrawals
4. Entity services no longer need to be aware of the Withdrawal Service and they don't need to call it to manage cascading withdrawals (this logic was duplicated across all entity services). This decoupling also breaks the circular dependencies between services (we can remove @Lazy), which is nice!


edavey (approved):
Thanks for doing this refactoring to make the withdrawals clearer and more maintainable 🏅

PR #1568, opened 2024-03-05
Bug/cas 243 assessment search multiple status

sudhakar-plr:
This PR is to fix a bug in CAS3 assessment search where multiple assessment status are passed in as a query param.


froddd (approved):
LGTM

PR #1569, opened 2024-03-05
Bookings with arrivals blocks ancestor withdrawals

davidatkinsuk:
If a booking has an arrival, any related entity (match request, request for placement, application) cannot be withdrawn and are marked as blocked. Such entities will not be returned in the withdrawable list, and any attempt to withdraw one of these entities will result in a 400 error.

As part of this commit the WithdrawalTest has been effectively rewritten as so many existing tests broke. This class should now be more concise and each test has a complete description of the entity tree.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1WithdrawableService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1WithdrawableService.kt
@@ -43,6 +43,7 @@ class WithdrawableService(
 
     return rootNode.flatten()
       .filter { it.status.withdrawable }
+      .filter { !it.isBlocked() }
       .filter { it.status.userMayDirectlyWithdraw }
       .map {
         WithdrawableEntity(
@@ -171,6 +172,10 @@ class WithdrawableService(
       return CasResult.GeneralValidationError("${context.triggeringEntityType.label} is not in a withdrawable state")
     }
 
+    if (rootNode.isBlocked()) {
+      return CasResult.GeneralValidationError("${context.triggeringEntityType.label} withdrawal is blocked")
+    }
+
     val withdrawalResult = withdrawRootEntityFunction.invoke()
     if (withdrawalResult is CasResult.Success) {
       cas1WithdrawableTreeOperations.withdrawDescendantsOfRootNode(rootNode, context)
@@ -219,6 +224,13 @@ data class WithdrawableState(
    * This doesn't consider if the entity itself is in a withdrawable state.
    */
   val userMayDirectlyWithdraw: Boolean,
+  /**
+   * If true, not only is this entity not withdrawable, but any ancestor
+   * of this entity is also not withdrawable. For example, bookings with
+   * arrivals will block the withdrawal of any associated request for
+   * placement and application
+   */
+  val blockAncestorWithdrawals: Boolean = false,
 )
 
 data class WithdrawableDatePeriod(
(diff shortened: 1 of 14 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1WithdrawableService.kt
```diff
@@ -219,6 +224,13 @@ data class WithdrawableState(
    * This doesn't consider if the entity itself is in a withdrawable state.
    */
   val userMayDirectlyWithdraw: Boolean,
+  /**
+   * If true, not only is this entity not withdrawable, but any ancestor
+   * of this entity is also not withdrawable. For example, bookings with
+   * arrivals will block the withdrawal of any associated request for
+   * placement and application
+   */
+  val blockAncestorWithdrawals: Boolean = false,
```
edavey, line 233: are we still able to handle:

> - **scenario 1**: part way through a series of ROTL placements John commits a serious offence and will not visit an AP again for some time, and not until a full fresh  risk assessment (application) has been made

In this scenario, we need to withdraw the application even though one or more associated placements have taken place...

davidatkinsuk: That's not going to work unfortunately. A few options:

* Allow managers to CRU withdraw applications even if bookings are in place ( but i think they need to be warned  that bookings with arrivals exist in this case )
* When the applicant creates a new placement request, they're asked if there are any significant changes. Presumably in this case the answer should be yes, so it would be covered by that process?

Probably a question for the users? The ask for this ticket was to block all applications if bookings have arrivals

edavey: ok, thks, I've left a comment on https://dsdmoj.atlassian.net/browse/APS-445

davidatkinsuk:
This is on hold until we've resolved the question raised by Ed on APS-445 (and more generally, determining what problem this change is solving and why)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1WithdrawableService.kt
edavey: I've reviewed this and in fact the "blocking" behaviour makes sense as it leaves the door open to reusing the existing application/assessment and control with the applicant and the assessor who will judge whether a fresh application is needed. Two scenarios where a series of ROTL placements is interrupted midway:

## Withdrawals during ROTL series

### Scenario 1: fresh application required

> Part way through a series of ROTL placements John commits a serious offence and will not visit an AP again for some time, and not until a full fresh  risk assessment (application) has been made. 

In this scenario, because there are placements in the "arrived" state, only the future placements can be withdrawn. The existing application and placement request will not be withdrawable. 

The applicant *can* make a new placement request on the back of the existing application/assessment explaining why they believe the change in circumstances does not require a full fresh application. The reviewer of the PR (who conducts the mini-assessment) will then **reject** the placement request, requiring the applicant to make a fresh application.

Note that the control of risk remains in the hands of the applicant and the reviewer of the placement request. Because a placement in the arrived state exists the application can not be "withdrawn".
  
### Scenario 2: fresh application not required

> Part way through a series of ROTLs John has been recalled due to a minor infraction. Placement 3 and 4 need to be cancelled. In a few weeks we expect ROTL visits to continue.

In this scenario the applicant withdraws placement requests 3 and 4. 

Using the existing application/assessment the applicant creates new placement requests explaining that the changes in circumstances do not require a full fresh application. In this case the reviewer of the PR **accepts** the placement request and the PR proceeds to "match" and book the placement in for the dates requested.

edavey (approved):
thanks for helping to thinking this through and verify that it's the desired behaviour

PR #1571, opened 2024-03-05
[CAS-245]-CAS3-Report-Throttle offender service call to get person info

sudhakar-plr:
This PR is to throttle the existing service call `offenderService.getOffenderSummariesByCrns()` with configured number of crns in each request.

This change is to fix the problem in CAS3 report where calling the service with more than 500 crns is failing as there is a restriction in number crn in each request by integration team. Please find the details here : https://mojdt.slack.com/archives/C03KVQUFSER/p1709284991241579

The change is to split the number of crns into subset of 400 crns and then call the `offenderService.getOffenderSummariesByCrns` service sequentially with 400 crns in each request and collect the response.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ReportService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ReportService.kt
@@ -1,6 +1,7 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.service
 
 import com.fasterxml.jackson.databind.ObjectMapper
+import org.apache.commons.collections4.ListUtils
 import org.apache.poi.ss.usermodel.WorkbookFactory
 import org.jetbrains.kotlinx.dataframe.io.writeExcel
 import org.springframework.beans.factory.annotation.Value
@@ -43,6 +44,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.BookingTrans
 import java.io.OutputStream
 import java.time.LocalDate
 import java.time.temporal.TemporalAdjusters
+import java.util.stream.Collectors
 
 @Service
 class ReportService(
@@ -62,6 +64,7 @@ class ReportService(
   private val placementApplicationEntityReportRowRepository: PlacementApplicationEntityReportRowRepository,
   private val objectMapper: ObjectMapper,
   @Value("\${cas3-report.end-date-override:0}") private val cas3EndDateOverride: Int,
+  @Value("\${cas3-report.crn-search-limit:400}") private val numberOfCrn: Int,
 ) {
   fun createBookingsReport(properties: BookingsReportProperties, outputStream: OutputStream) {
     val startOfMonth = LocalDate.of(properties.year, properties.month, 1)
@@ -79,8 +82,7 @@ class ReportService(
     )
 
     val crns = bookingsInScope.map { it.crn }.distinct().sorted()
-    val personInfos = offenderService.getOffenderSummariesByCrns(crns, userService.getUserForRequest().deliusUsername)
-      .associateBy { it.crn }
+    val personInfos = splitAndRetrievePersonInfo(crns.toSet())
 
     val reportData = bookingsInScope.map {
       val personInfo = personInfos[it.crn] ?: PersonSummaryInfoResult.Unknown(it.crn)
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ReportService.kt
```diff
@@ -195,4 +197,15 @@ class ReportService(
         WorkbookFactory.create(true)
       }
   }
+
+  private fun splitAndRetrievePersonInfo(crns: Set<String>): Map<String, PersonSummaryInfoResult> {
+    val deliusUsername = userService.getUserForRequest().deliusUsername
+
+    val crnMap = ListUtils.partition(crns.toList(), numberOfCrn)
+      .stream().map { crns ->
+        offenderService.getOffenderSummariesByCrns(crns.toSet(), deliusUsername).associateBy { it.crn }
```
davidatkinsuk, line 206: If 500 is the hard limit for this API, wouldn't it make more sense to do this partitioning in ApDeliusContextApiOffenderDetailsDataSource.getOffenderDetailSummaries? This will transparently protect us from this issue in other places and i think would mostly just be a case of shifting this partition logic into that method

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas3/ReportService.kt
```diff
@@ -55,4 +54,15 @@ class ReportService(
...
+  private fun splitAndRetrievePersonInfo(crns: Set<String>): Map<String, PersonSummaryInfoResult> {
+    val deliusUsername = userService.getUserForRequest().deliusUsername
+
+    val crnMap = ListUtils.partition(crns.toList(), numberOfCrn)
+      .stream().map { crns ->
+        offenderService.getOffenderSummariesByCrns(crns.toSet(), deliusUsername).associateBy { it.crn }
+      }.collect(Collectors.toList())
+
+    return crnMap.flatMap { it.toList() }.toMap()
+  }
```
davidatkinsuk, line 67: as per previous comment, moving this logic into ApDeliusContextApiOffenderDetailsDataSource would remove duplication between these two services

davidatkinsuk (changes requested):
I think we should move the partitioning into the client code, please see my comments for more info.

FYI - in production we're not currently using the bulk endpoint (in prod helm it's DATA-SOURCES_OFFENDER-DETAILS: community_api). Dan Livings is currently working through the pre-reqs to get this enabled in prod.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ReportService.kt
sudhakar-plr: Its a good idea, i want to try this fix to see if it resolve the current problem in reporting. 

Most of the cases we call this endpoint with single crn, only couple of places we called with more than 500 crns
I would create card to refactor this logic into common place once it resolves the problem.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas3/ReportService.kt
sudhakar-plr: Since the fix is isolated into single place now would give confident of the fix before we can move into common place if needed.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ReportService.kt
davidatkinsuk: Sure, could you please create that now and link here. ta.

davidatkinsuk: actually don't worry let's see it fixed first. will approve

davidatkinsuk (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ReportService.kt
sudhakar-plr: Card to refactor this logic in common client level after we confident with the solution: https://dsdmoj.atlassian.net/browse/CAS-246

PR #1573, opened 2024-03-06
Update CAS2 seed data

patrickjfl:
Updates CAS2 seed data after changing the application schema to move HDC dates into their own task at the top of the task list.

### DIFF
--- a/src/main/resources/db/seed/local+dev+test/cas2_application_data/data_A1234AI.json
+++ b/src/main/resources/db/seed/local+dev+test/cas2_application_data/data_A1234AI.json
@@ -13,8 +13,23 @@
       "consentDate-day": "1"
     }
   },
+  "hdc-licence-dates": {
+    "hdc-licence-dates": {
+      "hdcEligibilityDate": "2024-02-28",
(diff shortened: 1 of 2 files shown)


Harriethw (approved):
nice one!

Gweaton (approved):

src/main/resources/db/seed/local+dev+test/cas2_application_data/data_A1234AI.json
```diff
@@ -13,8 +13,23 @@
       "consentDate-day": "1"
     }
   },
+  "hdc-licence-dates": {
+    "hdc-licence-dates": {
```
Gweaton, line 17: just a thought - do we need this double nesting here? `hdc-licence-dates` inside `hdc-licence-dates`?

patrickjfl: good spot but no, the outer label refers to the task and the inner label to the page. Because it's a task with a single page, the two may as well be the same

PR #1574, opened 2024-03-06
(APS-455) Fix off by one error in emails

pezholio:
The email notifications when people were allocated assessments were off by one, because we were counting the number of working days including today when stating how many days they have to complete the assessment (inclusive), instead of the difference in working days between the due date and today's date (exclusive).

This adds a new `getDifferenceInWorkingDays` function and uses it in the email service to let the user know how many TOTAL working days they have to complete an assessment.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/WorkingDayCountService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/WorkingDayCountService.kt
@@ -3,6 +3,7 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.service
 import org.springframework.stereotype.Service
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.ClientResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.GovUKBankHolidaysApiClient
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.getDaysUntilExclusiveEnd
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.getDaysUntilInclusive
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.getNextWorkingDay
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.isWorkingDay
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/WorkingDayCountService.kt
```diff
@@ -24,6 +25,10 @@ class WorkingDayCountService(
     return from.getDaysUntilInclusive(to).filter { it.isWorkingDay(bankHolidays) }.size
   }
 
+  fun getDifferenceInWorkingDays(from: LocalDate, to: LocalDate): Int {
+    return from.getDaysUntilExclusiveEnd(to).filter { it.isWorkingDay(bankHolidays) }.size
+  }
```
davidatkinsuk, line 30: Is getDifferenceInWorkingDays only ever going to be used for today's date? If yes, maybe we change it to reflect that and call it something like:

getCompleteWorkingDaysFromNowUntil(to: LocalDate)?

I think it would be clearer what it's doing then? My concern is that it's not obvious what difference is between the two functions we have.

davidatkinsuk: Or could always call it getCompleteWorkingDatesCount() if you want to pass in both arguments? and in this case, update 'from' to be 'fromExclusive' and 'to' to be 'toInclusive'? Then update argument names on other function use same pattern?

davidatkinsuk (changes requested):
I have thoughts :)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/WorkingDayCountService.kt
edavey: yes I agree that adding the term "complete" to the function name will probably help explain the purpose of this function which rounds down the count of days remaining. Personally I'm not sure the `fromExclusive` and `toInclusive` arg names would help my understanding...

pezholio: Good call, I've gone with `getCompleteWorkingDaysFromNowUntil` 👍

davidatkinsuk (approved):
LGTM 👍

PR #1576, opened 2024-03-06
Cas2 115 extend status updates report with status details

danhumphreys-moj:
Reference these JIRA tickets for information: [CAS2-115](https://dsdmoj.atlassian.net/browse/CAS2-115) and [CAS2-114](https://dsdmoj.atlassian.net/browse/CAS2-114)

**Introduction**

The JSON structure of the data field in the domain events table for a record with an event type `CAS2_APPLICATION_STATUS_UPDATED` is as follows:

```JSON
{
   "eventType":"applications.cas2.application.status-updated",
   "eventDetails":{
      "applicationId":"xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx",
      "applicationUrl":"http://localhost:3000/applications/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx",
      "personReference":{
         "noms":"XXXXXXX",
         "crn":"XXXXXXX"
      },
      "newStatus":{
         "name":"moreInfoRequested",
         "label":"More information requested",
         "description":"The prison offender manager (POM) must provide information requested for the application to progress.",
         "statusDetails":[
            {
               "name":"personalInformation",
               "label":"Personal information"
            },
            {
               "name":"healthNeeds",
               "label":"Health needs"
            },
            {
               "name":"other",
               "label":"Other"
            }
         ]
      },
      "updatedBy":{
         "username":"CAS2_ASSESSOR_USER",
         "name":"CAS2 Assessor",
         "email":"assessor@example.com",
         "origin":"XXXXX"
      },
      "updatedAt":"2024-02-03T16:28:54Z"
   },
   "id":"xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx",
   "timestamp":"2024-02-03T16:28:54Z",
   "eventType":"applications.cas2.application.status-updated"
}
```

The requirement is to add an additional column in the 'Application Status Updated' Excel report which contains the secondary detail information for each status update, these are included in the `statusDetail` element and need to be incuded in the Excel report, specifically the `name` values.

There may be zero, one or more details for each status update, and they should be pipe delimited and therefore the SQL query to generate the report needs to be modified to extract the value of `name` for each element within the `statusDetails` array.  Due to the JSON structure, the array needs to be unnested using `json_array_elements` and aggregated using `string_agg`.

**SQL Query Logic**

To illustrate the logic a simplified example is used to show how the query is built up.  There is also a corresponding [db-fiddle example](https://www.db-fiddle.com/f/gkd4jrYihKQjRaZnMJw3S9/6).

Start with a SQL table and insert an `id` and some JSON `data` where the first record contains no details; the second record contains one detail; and the third contains three:

```SQL
create table events (
  id INT PRIMARY KEY NOT NULL,
  data JSON
);

insert into events (id, data) values

(1, '{"status":{"name":"place withdrawn","statusDetails":[]}}'),

(2, '{"status":{"name":"placeOffered","statusDetails":[{"name": "awaiting confirmation"}]}}'),

(3, '{"status":{"name":"informationRequired","statusDetails":[{"name": "personalInformation"}, {"name": "healthNeeds"}, {"name": "other"}]}}');
```

**Query 1: Select all data from events**

```SQL
select id, data from events;
```

Results:

image

**Query 2: Select id and details using LATERAL join**

```SQL
select id, details from events, lateral json_array_elements(events.data -> 'status' -> 'statusDetails') as details;
```

`LATERAL` with `json_array_elements` expands the statusDetails array elements, creates new rows for each element and joins it back with the id.  More information on Postgres LATERAL can be found [here](https://www.postgresql.org/docs/current/queries-table-expressions.html#QUERIES-LATERAL).  The `lateral` keyword can be omitted from this query since it is being used with a set function but has been left for clarity.

image

**Query 3: Select the values of all 'name' keys for each row**

```SQL
select id, elements ->> 'name' from events, lateral json_array_elements(events.data -> 'status' -> 'statusDetails') as elements;
```

image

**Query 4: Aggregate the name values, delimit by '|' and group them by id**

```SQL
select id, string_agg(elements ->> 'name', '|') as statusDetails from events, lateral json_array_elements(events.data -> 'status' -> 'statusDetails') as elements group by id;
```

image

**Query 5: use a left join so all rows are returned from events, regardless of whether they have details or not**

This means that we include rows where statusDetails is null.  `left join lateral` requires `on true`.

```SQL
select id, string_agg(elements ->> 'name', '|') as statusDetails from events left join lateral json_array_elements(events.data -> 'status' -> 'statusDetails') as elements on true group by id;
```

image

**Query 6: use COALESCE to return a blank string for null statusDetails**

```SQL
select id, coalesce(string_agg(elements ->> 'name', '|'), '') as statusDetails from events left join lateral json_array_elements(events.data -> 'status' -> 'statusDetails') as elements on true group by id;
```

image

[CAS2-115]: https://dsdmoj.atlassian.net/browse/CAS2-115?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CAS2-114]: https://dsdmoj.atlassian.net/browse/CAS2-114?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ReportsTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ReportsTest.kt
@@ -10,9 +10,11 @@ import org.junit.jupiter.api.Nested
 import org.junit.jupiter.api.Test
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.cas2.model.Cas2ApplicationStatusUpdatedEvent
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.cas2.model.Cas2ApplicationSubmittedEvent
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.cas2.model.Cas2StatusDetail
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.cas2.model.EventType
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.factory.events.cas2.Cas2ApplicationStatusUpdatedEventDetailsFactory
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.factory.events.cas2.Cas2ApplicationSubmittedEventDetailsFactory
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.factory.events.cas2.Cas2StatusFactory
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.integration.IntegrationTestBase
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.DomainEventType
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.reporting.model.Cas2ExampleMetricsRow
@@ -296,10 +298,29 @@ class Cas2ReportsTest : IntegrationTestBase() {
       val newer = Instant.now().minusSeconds(daysInSeconds(100))
       val tooOld = Instant.now().minusSeconds(daysInSeconds(366))
 
+      val event1StatusDetails = listOf(
+        Cas2StatusDetail("personalInformation", "Personal information"),
+        Cas2StatusDetail("riskOfSeriousHarm", "Risk of serious harm"),
+        Cas2StatusDetail("hdcAndCpp", "HDC licence and CPP details"),
+      )
+
+      val event1Status = Cas2StatusFactory()
+        .withStatusDetails(event1StatusDetails)
+        .produce()
+
       val event1Details = Cas2ApplicationStatusUpdatedEventDetailsFactory()
+        .withStatus(event1Status)
         .withUpdatedAt(old)
         .produce()
+
+      val event2StatusDetails = emptyList<Cas2StatusDetail>()
+
+      val event2Status = Cas2StatusFactory()
+        .withStatusDetails(event2StatusDetails)
+        .produce()
+
       val event2Details = Cas2ApplicationStatusUpdatedEventDetailsFactory()
+        .withStatus(event2Status)
         .withUpdatedAt(newer)
         .produce()
       val event3Details = Cas2ApplicationStatusUpdatedEventDetailsFactory()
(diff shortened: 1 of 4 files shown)


Harriethw (approved):
thanks for the very detailed PR message Dan!
It does seem incredibly complicated just to concatenate some strings, but after a brief look myself I also don't see another way of making it happen. 
I think in retrospect we probably should have created the Domain Events to be in the shape we needed here, e.g. have the detailed as a delimited string rather than an object, but I know we've already put the wheels in progress for the integration of those. c'est la vie!

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ReportsTest.kt
```diff
@@ -368,6 +390,7 @@ class Cas2ReportsTest : IntegrationTestBase() {
           newStatus = event1Details.newStatus.name,
           updatedAt = event1Details.updatedAt.toString().split(".").first(),
           updatedBy = event1Details.updatedBy.username,
+          statusDetails = event1Details.newStatus.statusDetails.joinToString("|") { it.name },
```
davidatkinsuk, line 393: i appreciate you're following the already established testing pattern here for assertions, but I think it would be really helpful if this had the hardcoded expectation as a string

davidatkinsuk (changes requested):
Lovely breakdown of how you ended up with the query you did. I've got one request which i think may help readability of test (specifically because the SQL you've used for this is quite novel)

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ReportsTest.kt
danhumphreys-moj: Thanks David - I've changed this [here](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1576/commits/cb3775f8635e96e88063f80f606be4cdb0caa2bd).

Ok to approve?  Thanks.

davidatkinsuk (approved):
looks great, thanks!

PR #1577, opened 2024-03-06
(CAS2-259) Add CAS2-specific reply-to email address

Gweaton:
## Context

[Jira ticket 259](https://dsdmoj.atlassian.net/jira/software/c/projects/CAS2/boards/1347?selectedIssue=CAS2-259)

Up until now, we've been sending emails in the service when an application is submitted using the default `reply-to` email for CAS1 specific in GOV.UK Notify, meaning if a user replies to an email they receive, it will go to the CAS1 mailbox.

This isn't right for CAS2, and we should be supplying our own `reply-to` here.

## Changes in this PR

Here we add an optional parameter to the `EmailService`'s `sendEmail` function to allow us to pass in the CAS2 reply-to email's Notify ID, and configure the CAS2 `ApplicationService` to send emails with the CAS2 mailbox's `reply-to` email address.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/EmailNotificationService.kt
```diff
@@ -19,8 +19,13 @@ class EmailNotificationService(
 ) : EmailNotifier {
   var log: Logger = LoggerFactory.getLogger(this::class.java)
 
-  override fun sendEmail(recipientEmailAddress: String, templateId: String, personalisation: Map<String, *>) {
-    applicationEventPublisher.publishEvent(SendEmailRequestedEvent(EmailRequest(recipientEmailAddress, templateId, personalisation)))
+  override fun sendEmail(
+    recipientEmailAddress: String,
+    templateId: String,
+    personalisation: Map<String, *>,
+    replyToEmailId: String?,
```
tahb, line 26: As we're adding a new reply-to email address into Notify, does this break the default behaviour for CAS1? Does Notify know to use the first/oldest one as the default when we don't pass an ID?

If it's not clear, do we need to add the replyToID in explicitly for CAS1 emails to avoid breaking their reply to?

tahb (approved):
LGTM! One question we should make sure we're happy isn't an issue before merging.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/EmailNotificationService.kt
Harriethw: it (hopefully) doesn't break the default behaviour for CA1, as the original email is set as the default one in Notify, and the default is used unless another is specified. We've also set the code here to make the param optional so we don't need to touch the other CAS methods that are sending emails using the default address - we've already added in the new email address on Notify and emails are working currently, so think it should be safe?

PR #1578, opened 2024-03-06
Remove use of mockkStatic for test stability

davidatkinsuk:
We suspect that the introduction of mockkStatisc was causing test instability issues in circle. This change uses the TimeService in WorkingDayCountService to retrieve the current LocalDate, which can then be mocked in the unit test without mockkStatic


pezholio (approved):
LGTM 👍

PR #1579, opened 2024-03-06
[CAS-235] -  Add end_date column to beds table for Cas3 premises

sudhakar-plr:
Refer [CAS-235](https://dsdmoj.atlassian.net/browse/CAS-235) for more information

This PR is to add an `endDate` to the CAS3 beds table to indicate the availability of the particular room. Similar to archive property but using `endDate`.  

I have added optional `endDate` to to the API which creates room and change the service to store the endDate in the CAS3 specific 'beds` table. 

This PR contains below changes:
1. Modify existing POST API `/premises/{premisesID}/rooms` by adding optional `endDate`
2. Modify existing GET API `/premises/{id}/rooms` by adding optional `endDate`
3.  Modify existing GET API `/premises/{id}/rooms/{roomId}` by adding optional `endDate`
4. Change the service to store the endDate into `beds` table and retrieve the same.

Note:We will have separate story to modify and use this `endDate` in search and book process

[CAS-235]: https://dsdmoj.atlassian.net/browse/CAS-235?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/resources/static/codegen/built-api-spec.yml
+++ b/src/main/resources/static/codegen/built-api-spec.yml
@@ -7514,6 +7514,11 @@ components:
           items:
             type: string
             format: uuid
+        endDate:
+          type: string
+          format: date
+          example: 2024-03-30
+          description: End date of the room availability, open for availability if not specified.
       required:
         - name
         - characteristicIds
@@ -7567,6 +7572,11 @@ components:
         code:
           type: string
           example: NEABC04
+        endDate:
+          type: string
+          format: date
+          example: 2024-03-30
+          description: End date of the room availability, open for availability if not specified
       required:
         - id
         - name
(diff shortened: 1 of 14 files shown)


src/main/resources/static/codegen/built-api-spec.yml
```diff
@@ -7520,6 +7520,11 @@ components:
           items:
             type: string
             format: uuid
+        endDate:
```
davidatkinsuk, line 7523: as discussed can we please change name (and description) to reflect that this applies to the default bed, not the room itself

davidatkinsuk (approved):
Assuming the name change we discussed is done, approved 👍

pezholio:
This is VERY useful for CAS1 too, we've had a lot of users requesting that we "delete" beds, which isn't possible

PR #1580, opened 2024-03-06
Email Referrer when an Assessor adds a Note

Harriethw:
JIRA https://dsdmoj.atlassian.net/jira/software/c/projects/CAS2/boards/1347?selectedIssue=CAS2-8

_As a referrer
I want to know when a note has been added to a referral I have made
So that I can reply in a timely manner to progress the referral_

This uses the `EmailNotificationService` to send an email to the referrer when a Note is added by an ExternalUser - I don't think we need to specify the role of the External User in this case, because the only other users able to leave a note are Assessors currently, and presumably the referrer should be notified whenever someone who is not a POM leaves a note.

I've also added some CAS2-specific date formatting to the date service, as we display dates differently than the other CAS projects.

## unhappy path

For the edge case that a Nomis user does not have an email,
we want to alert the team that the email has failed for this note, but
not let an error bubble up to the user.

We found that all of our existing 200+ users have emails associated
with their accounts, so we are hoping that we are unlikely to encounter
this.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/ApplicationNoteService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/ApplicationNoteService.kt
@@ -1,8 +1,10 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas2
 
+import org.springframework.beans.factory.annotation.Value
 import org.springframework.data.repository.findByIdOrNull
 import org.springframework.stereotype.Service
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.NewCas2ApplicationNote
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.config.NotifyConfig
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas2ApplicationEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas2ApplicationNoteEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas2ApplicationNoteRepository
@@ -11,9 +13,12 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas2User
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.NomisUserEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.results.AuthorisableActionResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.results.ValidatableActionResult
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.EmailNotificationService
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.ExternalUserService
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.HttpAuthService
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.NomisUserService
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.toCas2UiFormat
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.toCas2UiFormattedHourOfDay
 import java.time.OffsetDateTime
 import java.util.UUID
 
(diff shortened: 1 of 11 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/ApplicationNoteService.kt
```diff
@@ -26,6 +27,7 @@ class ApplicationNoteService(
   private val httpAuthService: HttpAuthService,
 ) {
 
+  @Transactional
```
Gweaton, line 35: I'm not sure we need this to be wrapped in a transaction, as the only thing we're saving here is a `note` (rather than, say, a note and an application), but I could be misunderstanding...

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/ApplicationNoteService.kt
```diff
@@ -48,13 +53,32 @@ class ApplicationNoteService(
 
     val savedNote = saveNote(application, note.note, user)
 
+    if (isExternalUser) {
```
Gweaton, line 56: Is this something I'd need to check on my branch too (emailing the referrer when the status has updated on an application)?

Gweaton (commented):
Approach looks good to me, there's definitely a lot I can re-use on my branch, thanks for doing the hard work!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/ApplicationNoteService.kt
Harriethw: hmm I don't think so currently, because only an assessor has permission to update the status, so `isExternalUser` should always be true

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/ApplicationNoteService.kt
Harriethw: ah yeah that's a good point, I think I was thinking we should add it to any methods that do something to the database, but there's no point if it's just one transaction as if it fails there's nothing else to roll back?

tahb (approved):
LGTM 🚀 

I wonder if we ought to do something about the possibility of a missing email address for an author. As we add two-way communication to the service this feels it should now become required? I think throwing an error is fine for now, especially as we've checked existing users. Is there another bit of work here to investigate making this email address  mandatory on `user` create? We'd probably have to throw an error on sign in and let them know that they need an email address in nomis before they can continue?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/ApplicationNoteService.kt
```diff
@@ -46,13 +53,32 @@ class ApplicationNoteService(
...
     return AuthorisableActionResult.Success(
       ValidatableActionResult.Success(
         savedNote,
       ),
     )
   }
 
+  private fun sendEmailToReferrer(application: Cas2ApplicationEntity, savedNote: Cas2ApplicationNoteEntity) {
+    emailNotificationService.sendEmail(
+      // TODO should we check if email exists? Throw error if it doesn't
```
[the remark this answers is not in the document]
Harriethw: Yep that's a good point that it should be handled further up stream...
It might be a bit tricky to do it when they first sign in because of the way we integrate with HMPPS Auth - but CAS1 have a solution where the user can update their details via the application form which we could use. 

It's a bit complicated as I think we would  have to emit a Domain Event as well, so that Nomis could be updated as well, but would be possible. I've made a ticket in the backlog here https://dsdmoj.atlassian.net/browse/CAS2-323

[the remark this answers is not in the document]
Harriethw: I've said on the card that it might be worth asking the Nomis team if any POMs in their service do have missing emails... I wonder if the nullable nature of it is just a technicality, and in practice they have to have an email to sign up to Nomis.

Harriethw:
This PR does not yet make use of the [just-merged `replyToEmailID`](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1577) for us to add a CAS2-specific email. As this feature is still behind a feature flag,I think it's safe to merge, and I will pick up the replyToEmail in the next PR.

PR #1582, opened 2024-03-07
Added linting check stage to Github Actions.

danhumphreys-moj:


davidatkinsuk (approved):
Looks really good to me, maybe someone with a better understanding of our circle job defs may want to review too.

I suspect most of the issues you've had to fix in the second commit are from me ... sorry ... but this will save me from myself going forward :)

pezholio (approved):
LGTM - I'm surprised we don't have this already! We'll probs need to add this to the required build steps in the settings too

.circleci/config.yml
```diff
@@ -339,6 +352,9 @@ workflows:
       - static_code_analysis_with_detekt:
           requires:
             - build
+      - code_linting_check_with_ktlint:
```
sudhakar-plr, line 355: only minor suggestion would be run the auto-correction task before code analysis. I mean the order of execution would be 
1.  Build
2. ./gradlew ktlintCheck
3. ./gradlew detekt
4. Unit test
5. Integration test

danhumphreys-moj: Makes sense, have done this.

danhumphreys-moj: @sudhakar-plr this isn't an autocorrection task though, if linting fails this pipeline stage will fail.  It will be down to the developers to check their branches using`./gradlew ktlintCheck` locally, and then running `./gradlew ktlintFormat` to fix any issues (ideally) before pushing.

I've added a couple of lines to the README to explain this.

PR #1583, opened 2024-03-07
Remove CAS1 User Details from App Update API

davidatkinsuk:
These details are only used once an application is submitted, so capturing them before submission doesn’t add any value


pezholio (approved):
LGTM 👍

PR #1584, opened 2024-03-07
Add CAS1 Placement Matching Outcomes Report

davidatkinsuk:
Relates to https://dsdmoj.atlassian.net/browse/APS-472


edavey (approved):
this looks very clear (considering the SQL!), thks

I do have big and fundamental reservations about how we are providing metrics, CAS-wide. I think we need to have a x-team chat on this topic and consider different approaches and architectures which may be more maintainable, robust and easy to comprehend. But that's another discussion, for the long-term...

davidatkinsuk:
> I do have big and fundamental reservations about how we are providing metrics, CAS-wide. I think we need to have a x-team chat on this topic and consider different approaches and architectures which may be more maintainable, robust and easy to comprehend. But that's another discussion, for the long-term...

Yes, i'm inclined to agree, especially with the move to only provide 'raw' data and let those ingesting it do the heavy lifting

PR #1590, opened 2024-03-11
(APS-338) Allow sorting tasks by person and allocated user

pezholio:
This adds two new fields to sort by:

- `person` to sort by the offender’s name; and
- `allocatedTo` to sort by the allocated user’s name


edavey (approved):
looks great, the sorting pattern seems to extend easily! 👍

PR #1593, opened 2024-03-12
CAS2-313, PersonInfoResult API error, ClassCastException

danhumphreys-moj:
An error has been observed in the DEV environment that occurs when the offender details API is not responsive (or times out) which causes `OffenderService.getInfoForPerson` to return `PersonInfoResult.Unknown`.  Since this is not caught in PersonUtils.kt this results in a `ClassCastException` when `PersonInfoResult.Unknown` is attempted to be converted to `PersonInfoResult.Success.Full`.

This has been resolved by changing the conditional `if` statements to a `when` and including a check to return `NotFoundProblem` when the person info is `PersonInfoResult.Unknown`.

For CAS2, `OffenderService.getInfoForPerson()` will never return `PersonInfoResult.Success.Restricted`, however the check has been left in `PersonUtils.kt`.  Therefore there is also no requirement to add a unit test to test this condition (since it can't be tested since it's never returned).

`PersonUtilsTest` extends `OffenderServiceTest` which requires that `OffenderServiceTest` is declared `open`, also inherited properties are made `protected` so that they are available to sub-classes only.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/util/cas2/PersonUtils.kt
```diff
@@ -17,8 +17,9 @@ fun OffenderService.getInfoForPersonOrThrowInternalServerError(crn: String):
 fun OffenderService.getFullInfoForPersonOrThrow(crn: String):
   PersonInfoResult.Success.Full {
   val personInfo = this.getInfoForPerson(crn)
-  if (personInfo is PersonInfoResult.NotFound) throw NotFoundProblem(crn, "Offender")
-  if (personInfo is PersonInfoResult.Success.Restricted) throw ForbiddenProblem()
-
-  return personInfo as PersonInfoResult.Success.Full
+  when (personInfo) {
+    is PersonInfoResult.NotFound, is PersonInfoResult.Unknown -> throw NotFoundProblem(crn, "Offender")
+    is PersonInfoResult.Success.Restricted -> throw ForbiddenProblem()
```
danhumphreys-moj, line 22: For CAS2 personInfo will never be PersonInfoResult.Restricted, however the check has been left here in PersonUtils.kt.  Therefore there is also no requirement to add a unit test to test this condition (since it can't be tested since it's never returned).

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas2/OffenderServiceTest.kt
```diff
@@ -36,19 +36,19 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas2.OffenderSer
 import java.time.LocalDate
 import java.time.OffsetDateTime
 
-class OffenderServiceTest {
-  private val mockPrisonsApiClient = mockk<PrisonsApiClient>()
+open class OffenderServiceTest {
```
Gweaton, line 39: Why has this class changed to open and its values to `protected`? I wonder if it's slipped in by mistake. If it's important for fixing the bug, it would be really helpful to document the reasoning in the commit message for future readers (and me!)

Gweaton (commented):
Looks good to me - just one question about the class changes.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas2/OffenderServiceTest.kt
danhumphreys-moj: `PersonUtilsTest` extends `OffenderServiceTest` which requires that `OffenderServiceTest` is declared `open`, also inherited properties are made protected so that they are available to sub-classes only.

Added extra details to commit message and description details above.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas2/PersonUtilsTest.kt
```diff
@@ -0,0 +1,79 @@
...
+import org.springframework.http.HttpMethod
+import org.springframework.http.HttpStatus
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.ClientResult
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.factory.InmateDetailFactory
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.factory.OffenderDetailsSummaryFactory
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.PersonInfoResult
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.NotFoundProblem
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.cas2.getFullInfoForPersonOrThrow
+
+class PersonUtilsTest : OffenderServiceTest() {
```
Gweaton, line 16: ah sorry I didn't see this was extending the other test class. This is a tricky one. I'd probably have opted for not extending the tests and doing some setup in this file, just so it's easier for readers to not have to click through files to see e.g. where `mockOffenderDetailsDataSource` is defined. We also don't tend to have many other test files extending test files elsewhere on the project (other than the `BaseIntegrationTest`).

However, this defining of a function on the `OffenderService` in a separate utils file is a bit new to me too, so it's an unusual scenario!

I'd still probably opt for the former to help with readability, but could be swayed.

Any thoughts @Harriethw ?

danhumphreys-moj: Yes I don't know why we have two functions that extend the CAS2 `OffenderService` in the `PersonUtil` module either really.  If they were in the same module we could remove `PersonUtils` and include the tests in `OffenderServiceTest` and therefore no requirement for a dedicated `PersonUtilsTest` any more.

Harriethw: Looking at the git history, I wonder if `PersonUtil` was a legacy of splitting off from CAS1, or just a first attempt at breaking out some things from the overly-long Offender Service... if it's causing confusion, happy for the functions to be brought back into the `OffenderService` and then just tested within `OffenderServiceTest` .

I think regardless of whether we remove `PersonUtil` we could test it without inheritance - as George suggests, by bringing in the setup to the `PersonUtilsTest` file rather than inheriting.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas2/PersonUtilsTest.kt
```diff
@@ -0,0 +1,79 @@
...
+
+  @Test
+  fun `throws NotFoundProblem when offender is PersonInfoResult-NotFound, status 404`() {
+    val crn = "ABC123"
+    every { mockOffenderDetailsDataSource.getOffenderDetailSummary(crn) } returns ClientResult
```
Harriethw, line 21: could we be even more precise about what's under test here, and mock the return of the `getInfoForPerson` method in the Offender Service?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas2/PersonUtilsTest.kt
```diff
@@ -0,0 +1,79 @@
...
+    every { mockOffenderDetailsDataSource.getOffenderDetailSummary(crn) } returns ClientResult.Success(
+      status = HttpStatus.OK,
+      body = offenderDetails,
+    )
+
+    val inmateDetail = InmateDetailFactory()
+      .withOffenderNo(nomsNumber)
+      .produce()
+
+    every { mockPrisonsApiClient.getInmateDetailsWithWait(nomsNumber) } returns ClientResult.Success(
```
Harriethw, line 67: as above, I don't think we ever need to get to this level in this test - we're already testing the behaviour of the Offender Service in `OffenderServiceTest`, so these unit tests can keep the focus on how `getFullInfoForPersonOrThrow` behaves depending on what `getInfoForPerson` returns

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas2/PersonUtilsTest.kt
```diff
@@ -0,0 +1,79 @@
...
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.factory.InmateDetailFactory
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.factory.OffenderDetailsSummaryFactory
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.PersonInfoResult
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.NotFoundProblem
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.cas2.getFullInfoForPersonOrThrow
+
+class PersonUtilsTest : OffenderServiceTest() {
+
+  @Test
+  fun `throws NotFoundProblem when offender is PersonInfoResult-NotFound, status 404`() {
```
Harriethw, line 19: It might help with readability to add a `@Nested` class around these two tests to show what function is under test, e.g. 
```
  @Nested
  inner class GetFullInfoForPersonOrThrow {
  ...
  }

File: src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas2/PersonUtilsTest.kt (hunk 3 above)
Comment (danhumphreys-moj) [reply src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas2/PersonUtilsTest.kt]: PersonUtils.kt has gone in the most recent commit and the extension functions in it have now been moved to OffenderService.kt.

All tests are now in OffenderServiceTest.kt.

Nice and tidy.

File: src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas2/PersonUtilsTest.kt (hunk 6 above)
Comment (danhumphreys-moj) [reply src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas2/PersonUtilsTest.kt]: Incorporated this in OffenderServiceTest.

File: src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas2/PersonUtilsTest.kt (hunk 4 above)
Comment (danhumphreys-moj) [reply src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas2/PersonUtilsTest.kt]: With the recent changes I think this means doing something like:

```kotlin
   val mockOffenderService = mockk<OffenderService>()

    @Test
    fun `throws NotFoundProblem when offender is PersonInfoResult-NotFound, status 404`() {
      val crn = "ABC123"
      every {mockOffenderService.getInfoForPerson(crn)} returns PersonInfoResult.NotFound(crn)
      assertThrows<NotFoundProblem> { offenderService.getFullInfoForPersonOrThrow(crn) }
    }
```
But this results in a mockk exception, presumably because I am mocking OffenderService and calling OffenderService.getFullInfoForPersonOrThrow on it.  Possibly there might be a way to do this (I had a chat with @Gweaton about this too), but I don't currently know what it is.

OK as we are @Harriethw?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/OffenderService.kt
```diff
@@ -108,6 +111,24 @@ class OffenderService(
     )
   }
 
+  fun getInfoForPersonOrThrowInternalServerError(crn: String):
+    PersonInfoResult.Success {
+    val personInfo = getInfoForPerson(crn)
```
Gweaton, line 116: I wonder if we're using `getInfoForPerson` anywhere here - it's quite hard to tell from a quick search. If we're not, we _could_ make it `private` and remove the tests for it, as the behaviour would be tested in the new tests. This could potentially be a refactor on top of this PR though.

danhumphreys-moj: I have made `getInfoForPerson()` `private` (it's only used in `OffenderService` by the looks of it).  Also removed `GetInfoForPerson` tests since as you say they are superseded by `GetFullInfoForPersonOrThrow` tests now.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas2/PersonUtilsTest.kt
danhumphreys-moj: This should be ok now?  All tests are now in `OffenderServiceTest`, `GetInfoForPerson` tests have been removed and now superseded by `GetFullInfoForPersonOrThrow` tests.

Harriethw (approved):
Nice one!

PR #1594, opened 2024-03-12
Aps 482 update withdrawal reasons on bookings withdrawal screen

pezholio:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
```diff
@@ -1188,6 +1189,10 @@ class BookingService(
       return CasResult.FieldValidationError(mapOf("$.reason" to "incorrectCancellationReasonServiceScope"))
     }
 
+    if (reason.name == "Other" && otherReason == null) {
```
davidatkinsuk, line 1192: !otherReason.isNullOrEmpty()  may be more robust?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
```diff
@@ -1348,7 +1353,7 @@ class BookingService(
   fun createCas3Cancellation(
     booking: BookingEntity,
     cancelledAt: LocalDate,
-    reasonId: UUID,
+    reasonId: UUID?,
```
davidatkinsuk, line 1356: Not sure why this became optional?

davidatkinsuk (approved):
Looking good, a few minor comments from me.

Do we want to consider adding this other reason to the timeline (via the domain event?)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
pezholio: Yeah, that was a hangover from another rabbit hole I went down!

PR #1595, opened 2024-03-12
[CAS-199]-CAS3-Consider bedspace enddate in bespace search functionality

sudhakar-plr:
This PR is to consider the newly added `bedspace enddate` in bedSpace search to filter out the beds which are not archived by the endate.


pezholio (approved):
LGTM 👍

PR #1597, opened 2024-03-12
Remove placement columns from misc. reports

davidatkinsuk:
Relates to https://dsdmoj.atlassian.net/browse/APS-479

Placement columns have removed from the applications, referrals and placement application reports.

These are now available from the new placement matching outcome reports.

For context:

“The raw metrics reports should only contain information that is relevant to the period in which they are being downloaded for. If they contain too much information then users may try to create metrics based on datasets that are incomplete.

Matching and Booking data should be removed from reports until the user is ready to report on those”

The reports will now only ever return a single row per application


edavey (approved):
It's good to see reports getting smaller rather than bigger!

PR #1598, opened 2024-03-12
Enable Notify in `test`

Harriethw:
As we are planning to use the `test` environment
for user acceptance/sandbox testing[1], because
it's an environment that can be easily frozen,
then it should have parity with the dev
environment, and we will need to test emails in
it.

This turns Notify into
`TEST_AND_GUEST_LIST` mode, which means it will
only send emails to our guestlist and use the test API key (see the `EmailNotificationService`).

I've also included the two API key secrets - both
are needed although the GUESTLIST one is the only one used, otherwise the app fails to compile.

[1]https://mojdt.slack.com/archives/C03K0HB0LBE/p1709818574878819

BEFORE MERGING:
[ ] add the API keys to the test environment in kubectl


edavey (approved):
looks good, thks!

Gweaton (approved):

PR #1600, opened 2024-03-13
Capture triggering user in domain events and show on Timeline

davidatkinsuk:
Relates to https://dsdmoj.atlassian.net/browse/APS-490

This PR adds 'triggering_user_id' to the common domain_events table and automatically populates it with the calling user's user id. For consistency with other entries in the table, no FK constraint is defined against the user, and a left join is used when retrieving the user to be included in timeline entries, as shown below:

Screenshot 2024-03-13 at 19 43 16

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/DomainEventSummary.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/DomainEventSummary.kt
@@ -1,16 +1,19 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.model
 
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.DomainEventType
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserEntity
 import java.sql.Timestamp
+import java.time.OffsetDateTime
 import java.util.UUID
 
 interface DomainEventSummary {
   val id: String
   val type: DomainEventType
-  val occurredAt: Timestamp
+  val occurredAt: OffsetDateTime
   val applicationId: UUID?
   val assessmentId: UUID?
   val bookingId: UUID?
   val premisesId: UUID?
   val appealId: UUID?
+  val triggeredByUser: UserEntity?
 }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/DomainEventService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/DomainEventService.kt
@@ -19,6 +19,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.domainevent.SnsEve
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.domainevent.SnsEventAdditionalInformation
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.domainevent.SnsEventPersonReference
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.domainevent.SnsEventPersonReferenceCollection
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.UserService
 import uk.gov.justice.hmpps.sqs.HmppsQueueService
 import uk.gov.justice.hmpps.sqs.MissingTopicException
 import java.time.OffsetDateTime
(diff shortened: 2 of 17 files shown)


pezholio (approved):
LGTM 👍

danhumphreys-moj:
Is the table `ALTER` statement (or equivalent) in a different PR David?

@Harriethw @Gweaton tagging for visibility, this PR makes some minor changes to the CAS2 `DomainEventService` - do we want to approve too?

davidatkinsuk:
> Is the table `ALTER` statement (or equivalent) in a different PR David?

It's here, but a one line so easy to miss - https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1600/files#diff-7a5728fbafb6299956166f05f97db164cdf7a84567bc6ca9995fa5eab875dfe4

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/DomainEventSummary.kt
```diff
@@ -1,16 +1,19 @@
...
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserEntity
 import java.sql.Timestamp
+import java.time.OffsetDateTime
 import java.util.UUID
 
 interface DomainEventSummary {
   val id: String
   val type: DomainEventType
-  val occurredAt: Timestamp
+  val occurredAt: OffsetDateTime
```
edavey, line 12: This looks like a big benefit to using JPA QL rather than raw SQL. Should we be aiming to re-write out SQL functions generally into JPA QL? It would be good to understand whether this is one of our "tech debt" tasks to chip away at...

edavey (approved):
Great stuff, thks

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/DomainEventSummary.kt
davidatkinsuk: I think it's something we need to decide on a case by case basis. If we're finding that our Native SQL is replicating what JPA can do for us (e.g. populating entity objects), we should probably use JPA QL. But - i find it harder to write personally and it can result in some cumbersome/heavy SQL if we're not careful. Definitely something to keep in mind!

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas2/DomainEventServiceTest.kt
```diff
@@ -34,6 +37,8 @@ import java.util.UUID
 class DomainEventServiceTest {
   private val domainEventRepositoryMock = mockk<DomainEventRepository>()
   private val hmppsQueueServiceMock = mockk<HmppsQueueService>()
+  private val userService = mockk<UserService>()
+  private val user = UserEntityFactory().withDefaultProbationRegion().produce()
```
Harriethw, line 42: hmm..I think for CAS2 it would be better to set the user id to `null`, or we need to implement our own function to find the user ID.
We have a different set of users on CAS2, they will be Nomis users or External users rather than Delius, so the current `getUserForRequestOrNull` used will always return null.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/DomainEventService.kt
```diff
@@ -107,6 +109,7 @@ class DomainEventService(
         createdAt = OffsetDateTime.now(),
         data = objectMapper.writeValueAsString(domainEvent.data),
         service = "CAS2",
+        triggeredByUserId = userService.getUserForRequestOrNull()?.id,
```
Harriethw, line 112: as mentioned in other comment, this will always return `null` for CAS2.
We'd need to implement something like the `getCas2User` in `ApplicationNoteService` 

https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/78389046ca19c0ce72761a60cac89d81996076f9/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/ApplicationNoteService.kt#L150 

I wonder if for now so that we aren't blocking you, we just set it to `null`? We could add it to our backlog, or I guess if we find we need it we would come around to fixing it naturally...

davidatkinsuk: That makes sense, i'll null it out for CAS2/3 for now to keep things simple 👍

PR #1601, opened 2024-03-13
[CAS3] - Fix issue in archive assessments pagination for multiple status

sudhakar-plr:
PR to fix bug in SQL for archive assessments .


froddd (approved):
LGTM

PR #1602, opened 2024-03-13
Don’t link adhoc placements to match requests

davidatkinsuk:
Before this commit an adhoc placement would be linked to the latest ‘non placed’ match request on an application with the same CRN. This could result in placements being linked to match requests with completely different dates.

As the link to match requests is not neccessary, this functionality has been removed.


pezholio (approved):
LGTM 👍

PR #1603, opened 2024-03-13
(APS 337) Filter tasks on task allocation by qualification and CRN / name

pezholio:
This adds some additional filtering to the `/tasks` endpoint to allow tasks to be filtered by required qualification as well as CRN/Name.

I've removed a bunch of the retired endpoints that we no longer use, as well as refactored the tests to make them easier to grok. In an ideal world, I'd like to share the database context across all these subtests, but I haven't had time to investigate so far. I'll keep digging!


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/TasksController.kt
```diff
@@ -64,53 +65,45 @@ class TasksController(
   private val taskService: TaskService,
 ) : TasksApiDelegate {
 
-  @Deprecated("Use tasksGet")
```
davidatkinsuk, line 67: +1,292 −4,452  love it!

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/TasksTest.kt
```diff
@@ -735,6 +736,128 @@ class TasksTest : IntegrationTestBase() {
...
+          }
+        }
+      }
+
+      @ParameterizedTest
+      @CsvSource(
+        "assessment,WOMENS", "assessment,PIPE", "assessment,ESAP", "assessment,EMERGENCY",
+        "placementRequest,WOMENS", "placementRequest,PIPE", "placementRequest,ESAP", "placementRequest,EMERGENCY",
+        "placementApplication,WOMENS", "placementApplication,PIPE", "placementApplication,ESAP", "placementApplication,EMERGENCY",
+      )
```
davidatkinsuk, line 817: i do love how readable this CsvSource stuff is for describing lots of permutations

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/TaskEntity.kt
```diff
@@ -29,6 +29,15 @@ interface TaskRepository : JpaRepository<Task, UUID> {
       )
     """
 
+    private const val CRN_OR_NAME_QUERY = """
+      :crnOrName IS NULL OR 
+      (
+          ((SELECT COUNT(1) FROM applications a WHERE a.id = application.id AND a.crn = UPPER(:crnOrName)) = 1)
+          OR
+          ((SELECT COUNT(1) FROM approved_premises_applications apa WHERE apa.id = application.id AND apa.name LIKE UPPER('%' || :crnOrName || '%')) = 1)
```
davidatkinsuk, line 37: would it be more efficient to ensure all three queries have a join on 'application' and 'approved_premises_application' and then reference here without the need for an inline query?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/TaskEntity.kt
```diff
@@ -29,6 +29,15 @@ interface TaskRepository : JpaRepository<Task, UUID> {
       )
     """
 
+    private const val CRN_OR_NAME_QUERY = """
+      :crnOrName IS NULL OR 
+      (
+          ((SELECT COUNT(1) FROM applications a WHERE a.id = application.id AND a.crn = UPPER(:crnOrName)) = 1)
```
davidatkinsuk, line 35: couldn't we just do 'a.crn = UPPER(:crnOrName)'. I feel like i'm missing something :)

src/test/resources/application-test.yml
```diff
@@ -207,4 +209,17 @@ notify:
...
+    org:
+      hibernate:
+        SQL: DEBUG
+        type:
+          descriptor:
+            sql:
+              BasicBinder: TRACE
+    org.springframework.jdbc.core:
+      JdbcTemplate: DEBUG
+      StatementCreatorUtils: TRACE
```
davidatkinsuk, line 225: did you mean to leave this enabled? Suspect logs could get quite big?

src/test/resources/application-test.yml
```diff
@@ -14,10 +14,12 @@ spring:
   flyway:
     locations: classpath:db/migration/all,classpath:db/migration/integration
   jpa:
+    show-sql: true
```
davidatkinsuk, line 17: do you mean to leave this enabled? I think it's helpful but wonder if it'll slow down tests in circle?

davidatkinsuk (changes requested):
there's nothing more satisfying than deleting a tonne of code :) 

A few comments from me

src/test/resources/application-test.yml
pezholio: Whoops!

src/test/resources/application-test.yml
pezholio: Also whoops!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/TaskEntity.kt
pezholio: Yeah, I nabbed this query from elsewhere without really thinking - have sorted 👍

davidatkinsuk (approved):
LGTM 👍

PR #1605, opened 2024-03-14
Adhoc bookings are exempt from cascade withdrawals

davidatkinsuk:
Relates to https://dsdmoj.atlassian.net/browse/APS-475

Several legacy adhoc bookings are linked to potentially unrelated placement requests. This behaviour was removed in commit 2d96a4d37567fde3f91a7a9172b459a91fb30625.  To avoid such adhoc placements being unintentionally withdrawn when (unrelated) ancestor elements are withdrawn, we mark them as exempt from cascade.

Exempting such makes sense in the general context because when creating an adhoc booking the user doesn’t explicitly link it to an application, it is automatically assigned to an application with the required CRN (if such an applciation exists). Therefore, it’s unexpected for an adhoc to be withdrawn automatically if an application withdraws the application.

The withdrawal tree toString has been updated to highlight when placements are exempt from cascade. The simple example below shows a booking with a single adhoc booking

```
2024-03-14 08:11:21.614 DEBUG 6138 --- [io-8080-exec-10] u.g.j.d.h.a.s.cas1.WithdrawableService   : Tree for withdrawing Application with id 15b5300d-c7d0-4e5d-8849-f94d5a42966f is 

  Application(15b), withdrawable:Y, mayDirectlyWithdraw:Y,  
  ---> Booking(498), withdrawable:Y, mayDirectlyWithdraw:Y,  EXEMPT FROM CASCADE
```

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1WithdrawableTreeBuilder.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1WithdrawableTreeBuilder.kt
@@ -23,6 +23,9 @@ import java.util.UUID
  * ---> Match Request (For initial application dates)
  * ------> Placement
  * ---> Placement (For adhoc placements)
+ *
+ * Note that whilst assessments are automatically withdrawn when
+ * an application is withdrawn, that is not managed by the tree
  **/
 @Component
 class WithdrawableTreeBuilder(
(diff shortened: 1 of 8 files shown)


pezholio (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1WithdrawableTreeBuilder.kt
```diff
@@ -124,15 +127,33 @@ data class WithdrawableTreeNode(
...
   private fun blockingDescription() = if (status.blockAncestorWithdrawals) {
     "BLOCKING"
   } else if (isBlocked()) {
     "BLOCKED"
   } else {
     ""
   }
+
+  private fun exemptFromCascadeDescription() = if (status.exemptFromCascade) {
+    "EXEMPT FROM CASCADE"
```
edavey, line 155: As (currently) the only reason to be "exempt from cascade" is by being an ad-hoc booking, would it be clearer if we reported:

```
EXEMPT FROM CASCADE (AD-HOC BOOKING)
```

In the future if there are other special cases we could pass a particular exemption reason in?

edavey (commented):
Thanks for address this 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1WithdrawableTreeBuilder.kt
davidatkinsuk: good idea, i'll push as a separate PR

PR #1607, opened 2024-03-14
User Service Tidy Up

davidatkinsuk:
Removing an unused function and improve naming of getUserForUsername to better reflect side effects


edavey (approved):
thanks for making this area clearer and less crufty

PR #1608, opened 2024-03-14
Enable AP Delius Context API as the offender data source for production

danlivings-dxw:
This PR updates the configuration to switch the data source for offender details and offender risks to the AP Delius Context API in the production environment.

Following the deployment of this change to `test` (#1513) as well as `dev` and `preprod` (#1404), we now have confidence that this change will work as expected and deliver some performance improvements.

This change still needs to be enabled locally (through WireMock and/or [ap-tools](https://github.com/ministryofjustice/hmpps-approved-premises-tools)), but this doesn't prevent deployment to production.

See the PRs below for more context about this configuration change:
- #1200
- #1220 
- #1257 
- #1298 
- #1317 
- #1354 
- #1404 
- #1513


nsmj63-dxw (approved):
Let's get this ball rolling!

PR #1613, opened 2024-03-14
Allow many ‘initial’ match requests to be returned

davidatkinsuk:
Whilst there should only ever be 0 or 1 placement requests returned when requesting getPlacementRequestForInitialApplicationDates, due to data inconsistencies there are a small number of applications that will return multiple placement requests.

We know that only one of these placement requests relate to the initial application dates, and the rest relate to placement_applications, but we have no programmatic way of 'fixing' this link for those placement requests.


pezholio (approved):

edavey (approved):
Maybe we can start planning to implement some of the planned changes to the data model?

PR #1615, opened 2024-03-14
placement.adhoc backfill now supports true & false

davidatkinsuk:
Previously the adhoc backfill job would only set the adhoc field to false. The job now takes a spreadsheet that can be used to set the value to either true or false, allowing us to selectively populate this field for large datasets


edavey (approved):
Thanks for getting this in place so quickly.

PR #1616, opened 2024-03-14
`INNER JOIN` approved_premises_applications when listing tasks

pezholio:
We’ve been getting an error in certain circumstances when querying for tasks. My suspicion is that doing a `LEFT JOIN` instead of an `INNER JOIN` is causing non-AP assessments to be included. I’ve created a temporary accommodation Assessment in some of the task lisiting tests to prove this theory and it seems to ring true.


davidatkinsuk (approved):
LGTM 👍

PR #1618, opened 2024-03-14
[APS-473] Add support for Mental Health Approved Premises when updating or submitting an application

danlivings-dxw:
> See [APS-473 on the Approved Premises Jira board](https://dsdmoj.atlassian.net/browse/APS-473).

This PR adds support for Mental Health Approved Premises (MHAP) within CAS1 applications. There are two ways to specify that an application is MHAP when updating or submitting an application:
- Setting `isMhapApplication` to `true`
- Setting `apType` to `"mhap"`

Using the `apType` enum is preferred for future usage, but `isMhapApplication` has been introduced for backwards compatibility by analogy with the current usage of `isPipeApplication` and `isEsapApplication`. Only one method can be used; for example, setting both `isMhapApplication` to `true` and `apType` to `"mhap"` will return a validation error. This is to avoid having to decide precedence or assign arbitrarily when the values disagree. If `apType` is used, all three of `isPipeApplication`, `isEsapApplication`, and `isMhapApplication` must be `null`; likewise, if any of those three are non-null, `apType` must be `null`.

Internally, the `approved_premises_applications` table has been remodelled to remove the `is_pipe_application` and `is_esap_application` in favour of an `ap_type` column, allowing for greater flexibility in adding new AP types in future.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApplicationEntityReportRow.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApplicationEntityReportRow.kt
@@ -30,13 +30,7 @@ interface ApplicationEntityReportRowRepository : JpaRepository<ApplicationEntity
       submission_event.data -> 'eventDetails' ->> 'mappa' as mappa,
       submission_event.data -> 'eventDetails' ->> 'offenceId' as offenceId,
       submission_event.data -> 'eventDetails' -> 'personReference' ->> 'noms' as noms,
-      (
-        CASE
-          WHEN apa.is_pipe_application THEN 'pipe'
-          WHEN apa.is_esap_application THEN 'esap'
-          ELSE 'normal'
-        END
-      ) as premisesType,
+      lower(apa.ap_type) as premisesType,
       application.data -> 'basic-information' -> 'sentence-type' ->> 'sentenceType' as sentenceType,
       submission_event.data -> 'eventDetails' ->> 'releaseType' as releaseType,
       cast(submission_event.data -> 'eventDetails' ->> 'submittedAt' as date) as applicationSubmissionDate,
@@ -119,13 +113,7 @@ interface ApplicationEntityReportRowRepository : JpaRepository<ApplicationEntity
       submission_event.data -> 'eventDetails' ->> 'mappa' as mappa,
       submission_event.data -> 'eventDetails' ->> 'offenceId' as offenceId,
       submission_event.data -> 'eventDetails' -> 'personReference' ->> 'noms' as noms,
-      (
-        CASE
-          WHEN apa.is_pipe_application THEN 'pipe'
-          WHEN apa.is_esap_application THEN 'esap'
-          ELSE 'normal'
-        END
-      ) as premisesType,
+      lower(apa.ap_type) as premisesType,
       application.data -> 'basic-information' -> 'sentence-type' ->> 'sentenceType' as sentenceType,
       submission_event.data -> 'eventDetails' ->> 'releaseType' as releaseType,
       cast(
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/ApprovedPremisesType.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/ApprovedPremisesType.kt
@@ -1,0 +1,10 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.model
+
+enum class ApprovedPremisesType {
+  NORMAL,
+  PIPE,
+  ESAP,
+  RFAP,
+  MHAP_ST_JOSEPHS,
+  MHAP_ELLIOTT_HOUSE,
+}
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/ApprovedPremisesType.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/ApprovedPremisesType.kt
@@ -1,5 +1,7 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.model
 
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ApType
+
 enum class ApprovedPremisesType {
   NORMAL,
   PIPE,
@@ -7,4 +9,13 @@ enum class ApprovedPremisesType {
   RFAP,
   MHAP_ST_JOSEPHS,
   MHAP_ELLIOTT_HOUSE,
+}
+
+fun ApType.asApprovedPremisesType() = when (this) {
+  ApType.normal -> ApprovedPremisesType.NORMAL
+  ApType.pipe -> ApprovedPremisesType.PIPE
+  ApType.esap -> ApprovedPremisesType.ESAP
+  ApType.rfap -> ApprovedPremisesType.RFAP
+  ApType.mhapStJosephs -> ApprovedPremisesType.MHAP_ST_JOSEPHS
+  ApType.mhapElliottHouse -> ApprovedPremisesType.MHAP_ELLIOTT_HOUSE
 }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ApplicationService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ApplicationService.kt
@@ -7,6 +7,7 @@ import org.springframework.stereotype.Service
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.ApplicationWithdrawn
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.ApplicationWithdrawnEnvelope
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.PersonReference
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ApType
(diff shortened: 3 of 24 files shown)


yndajas (commented):
I think we might also need `isMHAPElliottHouse` and `isMHAPStJosephs` adding to the placement criteria so that I can add these to the matching information screen without breaking anything. Would it be possible to add this in this or another PR?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApplicationEntityReportRow.kt
```diff
@@ -30,13 +30,7 @@ interface ApplicationEntityReportRowRepository : JpaRepository<ApplicationEntity
...
       submission_event.data -> 'eventDetails' ->> 'offenceId' as offenceId,
       submission_event.data -> 'eventDetails' -> 'personReference' ->> 'noms' as noms,
-      (
-        CASE
-          WHEN apa.is_pipe_application THEN 'pipe'
-          WHEN apa.is_esap_application THEN 'esap'
-          ELSE 'normal'
-        END
-      ) as premisesType,
+      lower(apa.ap_type) as premisesType,
```
davidatkinsuk, line 33: nice!

src/main/resources/db/migration/all/20240311103635__make_ap_type_column_on_approved_premises_applications_table.sql
```diff
@@ -0,0 +1,13 @@
+ALTER TABLE approved_premises_applications ADD COLUMN ap_type TEXT NOT NULL DEFAULT 'NORMAL';
```
davidatkinsuk, line 1: it may be worth splitting this into two migrations. one to add the column and one to do the backfill. Otherwise if backfill fails for some reason (not that it should) we don't have to manually remove the column to re-run this migration once fixed?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/factory/ApprovedPremisesApplicationEntityFactory.kt
```diff
@@ -103,7 +103,10 @@ class ApprovedPremisesApplicationEntityFactory : Factory<ApprovedPremisesApplica
   }
 
   fun withIsPipeApplication(isPipeApplication: Boolean) = apply {
```
davidatkinsuk, line 105: I was going to suggest deprecating this method but I think it's probably worth just removing it and updating the callers to set status instead. Particularly because calling this with 'false' doesn't actually do anything, which could be confusing?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/factory/ApprovedPremisesApplicationEntityFactory.kt
```diff
@@ -171,7 +174,14 @@ class ApprovedPremisesApplicationEntityFactory : Factory<ApprovedPremisesApplica
   }
 
   fun withIsEsapApplication(isEsapApplication: Boolean?) = apply {
```
davidatkinsuk, line 176: I was going to suggest deprecating this method but I think it's probably worth just removing it and updating the callers to set status instead. Particularly because calling this with 'false' doesn't actually do anything, which could be confusing?

src/main/resources/db/migration/all/20240311103635__make_ap_type_column_on_approved_premises_applications_table.sql
```diff
@@ -0,0 +1,13 @@
...
+SET ap_type = 'PIPE'
+WHERE is_pipe_application = true;
+
+UPDATE approved_premises_applications
+SET ap_type = 'ESAP'
+WHERE is_esap_application = true;
+
+ALTER TABLE approved_premises_applications ALTER COLUMN ap_type DROP DEFAULT;
+ALTER TABLE approved_premises_applications DROP COLUMN is_pipe_application;
+ALTER TABLE approved_premises_applications DROP COLUMN is_esap_application;
```
davidatkinsuk, line 13: This is probably me being paranoid but i wonder if it's worth keeping this around for a few weeks just in case we've missed something re: backfill? I'm probably worrying too much though!

src/main/resources/static/_shared.yml
```diff
@@ -1608,6 +1608,8 @@ components:
               type: boolean
             isEsapApplication:
               type: boolean
+            isMhapApplication:
```
davidatkinsuk, line 1611: Do we absolutely need to add this given the new status? Is this a UI requirement to help migrate over?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ApplicationService.kt
```diff
@@ -500,6 +504,25 @@ class ApplicationService(
       )
     }
 
+    if (updateFields.isUsingLegacyApTypeFields && updateFields.isUsingNewApTypeField) {
+      return AuthorisableActionResult.Success(
+        ValidatableActionResult.GeneralValidationError(
+          "`isPipeApplication`/`isEsapApplication`/`isMhapApplication` should not be used in conjunction with `apType`",
```
davidatkinsuk, line 510: i like this!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ApplicationService.kt
```diff
@@ -792,6 +819,7 @@ class ApplicationService(
       apType = when {
```
davidatkinsuk, line 819: i may be missing it, but are we using SubmitApprovedPremisesApplication.apType if defined like we do in the update function?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/ApplicationServiceTest.kt
```diff
@@ -1040,6 +1049,39 @@ class ApplicationServiceTest {
       assertThat(validatableActionResult.message).isEqualTo("This application has already been submitted")
     }
 
+    @Test
+    fun `updateApprovedPremisesApplication returns GeneralValidationError when application has AP type specified in multiple ways`() {
```
davidatkinsuk, line 1053: i can see lots of good tests for when defining AP Type in multiple ways, but do we have any when defining apType enum instead of booleans?

davidatkinsuk (changes requested):
Looking good, this is moving us towards a much simpler model.

A few comments from me.

src/main/resources/db/migration/all/20240311103635__make_ap_type_column_on_approved_premises_applications_table.sql
danlivings-dxw: I've split this into multiple migrations (including one for the `DROP`ping of columns).

src/main/resources/db/migration/all/20240311103635__make_ap_type_column_on_approved_premises_applications_table.sql
danlivings-dxw: This is its own separate migration now.

src/main/resources/static/_shared.yml
danlivings-dxw: You're right, I've removed this as it would be easier to just use `apType` instead of introducing an intermediate step.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ApplicationService.kt
danlivings-dxw: This was an oversight I think, I've changed this.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/ApprovedPremisesType.kt
```diff
@@ -0,0 +1,19 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.model
+
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ApType
+
+enum class ApprovedPremisesType {
+  NORMAL,
+  PIPE,
+  ESAP,
+  RFAP,
+  MHAP,
```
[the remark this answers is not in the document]
danlivings-dxw: This has been done.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/factory/ApprovedPremisesApplicationEntityFactory.kt
danlivings-dxw: This has been removed.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/factory/ApprovedPremisesApplicationEntityFactory.kt
danlivings-dxw: This has been removed.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/ApplicationServiceTest.kt
danlivings-dxw: I've added tests for submitting and updating the application using the `ApType` enum.

davidatkinsuk (approved):
LGTM 👍

PR #1619, opened 2024-03-15
revert use of delius_context_api for offender info (do not merge)

davidatkinsuk:
An issue has been found with this integration, causing an exception in production, as detailed on APS-509.

This commit reverts back to the non-bulk endpoint whilst the issue is being resolved, if we decide reversion is required in the meantime


davidatkinsuk:
Closing without merge as reversion not required.

PR #1620, opened 2024-03-15
Block withdrawal cascade between placement requests and adhoc bookings only

davidatkinsuk:
Relates to [APS-508](https://dsdmoj.atlassian.net/browse/APS-508)

We currently block all cascading from placement requests to adhoc bookings to avoid incorrectly linked placement requests cascading down to the placement. But, it’s acceptable for cascade withdrawal to happen from the application. 

This PR removes the previous behaviour for adhoc bookings, and ensures cascade withdrawals are only blocked from requests for placements to bookings. This means that when an application is withdrawn, any associate adhoc booking will be withdrawn.

This is achieved by ensuring request for placement to adhoc booking relationships are not included when building the withdrawal tree.

[APS-508]: https://dsdmoj.atlassian.net/browse/APS-508?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1WithdrawableTreeBuilder.kt
```diff
@@ -124,17 +124,18 @@ data class WithdrawableTreeNode(
   }
 
   @SuppressWarnings("MagicNumber")
-  private fun render(depth: Int): String {
-    val padding = "  " + if (depth > 0) { "-".repeat(3 * depth) + "> " } else { "" }
-    val abbreviatedId = entityId.toString().substring(0, 3)
+  fun render(depth: Int, includeIds: Boolean = true): String {
+    val padding = if (depth > 0) { "-".repeat(3 * depth) + "> " } else { "" }
+    val abbreviatedId = if (includeIds) { entityId.toString().substring(0, 3) } else ""
```
edavey, line 140: thanks for adding this logging

edavey (approved):
Great stuff, thanks for spotting this improved way to distinguish between adhoc bookings.

PR #1621, opened 2024-03-15
Move `personName`, `crn` and `nomsNumber` to top level `Cas2SubmittedApplicationsSummaries`

Harriethw:
We have an 'assessor's dashboard' page on CAS2, which queries the `/submissions` endpoint to retrieve paged results of all submitted applications across all time.

Screenshot 2024-03-15 at 15 28 03

The response times for this are very slow https://dsdmoj.atlassian.net/browse/CAS2-318 

Part of the reason for this could be that we are making two calls to integration APIs in order to retrieve a `FullPerson` type.

1. Here when we get the `offenderSummary` 

https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/938a6dda5c012a3b80080ef109d80de6f63fdeee/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/OffenderService.kt#L86 

2. And then to get the `inmateDetail` 

https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/938a6dda5c012a3b80080ef109d80de6f63fdeee/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/OffenderService.kt#L101

There's three phases of work I suggest to resolve this:

### 1. remove the second API call to get inmate details (this PR)
For this dashboard, we only need the person's name from the integration APIs, which means that only the first call is needed. We already have the `nomsNumber` and `crn` on the Application Entity. 
Let's just return the person's name on the summary and not bother to return a full `Person` type. This should significantly reduce the response time as we'd go from 1 to 2 external API calls.
Cons: We might need to show inmate details on a dashboard in future, but that could be introduced again if needed, and maybe solved by phase 2.
This would be done in stages, in order not to break the UI.
A. (API) Introduce the top level fields
B. (FE) Use the new fields on the UI (PR here https://github.com/ministryofjustice/hmpps-community-accommodation-tier-2-ui/pull/583) 
c. (API) Remove the old `Person` field and extraneous API call.

### 2. use bulk search for CRNs (near-future work)
We use the `getOffenderDetailSummaries` on the `OffenderDetailsDataSource` to bulk search for the offender summaries, rather than making a call per CRN.

### 3. save offender/inmate details on the Application (far-future work)
The wider CAS teams are now agreeing to pull in offender data and save in our db, so [we do not have to make these integration calls](https://mojdt.slack.com/archives/C03K0HB0LBE/p1710429149171069).
This would mean we save the prisoner name when an application is created on the Application table, and remove the need for this API call altogether.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/SubmissionsController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/SubmissionsController.kt
@@ -177,10 +177,12 @@ class SubmissionsController(
   ):
     Cas2SubmittedApplicationSummary {
     val personInfo = offenderService.getInfoForPersonOrThrowInternalServerError(application.getCrn())
+    val personName = offenderService.getOffenderNameOrPlaceholder(personInfo)
 
     return submissionsTransformer.transformJpaSummaryToApiRepresentation(
       application,
       personInfo,
+      personName,
     )
   }
 
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas2ApplicationEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas2ApplicationEntity.kt
@@ -46,6 +46,7 @@ WHERE a.created_by_user_id = :userId
 SELECT
     CAST(a.id AS TEXT) as id,
     a.crn,
+    a.noms_number as nomsNumber,
     CAST(a.created_by_user_id AS TEXT) as createdByUserId,
     a.created_at as createdAt,
     a.submitted_at as submittedAt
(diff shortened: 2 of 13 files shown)


Gweaton:
This approach makes sense to me - let's chat about it next week.

One other thing, is could we use [this](https://probation-offender-search-dev.hmpps.service.justice.gov.uk/swagger-ui/index.html#/offender-search/findByIds) bulk endpoint to search for all CRNs in one go, rather than individually? This could potentially be something we do in parallel with your changes here for maximum efficiency. Not sure if this is the same endpoint in Dan's other branch

Harriethw:
>One other thing, is could we use [this](https://probation-offender-search-dev.hmpps.service.justice.gov.uk/swagger-ui/index.html#/offender-search/findByIds) bulk endpoint to search for all CRNs in one go, rather than individually? This could potentially be something we do in parallel with your changes here for maximum efficiency. Not sure if this is the same endpoint in Dan's other branch

@Gweaton that's a good shout, I was thinking it might be a bit too complex to use but looking at the code in `splitAndRetrievePersonInfo` for the CAS1 `ReportService` it could be something we do. It could also be really helpful for any migrations we do in the future. I'll add it as a phase...

Harriethw:
Or maybe we should do the bulk search first 🤔 One for next week!

Gweaton (approved):
This looks good to me, nice one 👍

PR #1622, opened 2024-03-15
Handle missing CRNs returned from bulk endpoints

davidatkinsuk:
Relates to https://dsdmoj.atlassian.net/browse/APS-509

/probation-cases/summaries and /users/access will not return a CRN if information isn’t found for that CRN.

Before this commit, the code was (implicitly) assuming that if a CRN wasn’t found, a 404 would be returned.

This commit changes the logic calling these two endpoints to return an individual ‘Not Found’ status code for any CRN not included in the result.

The code isn’t ideal as we need to include the method/path when building the not found responses, which the OffenderDetailsDataSource shouldn’t really need to know. Ideally we would refactor this code to stop returning such a complex response hierarchy from the updated functions.


pezholio (approved):
LGTM - I agree it'd be good if we can decouple the HTTP responses from the actual process of building up offender responses, but I guess this will be easier now we're migrating towards the bulk approach

PR #1624, opened 2024-03-17
[CAS-244] - Update occupancy rate calculation in Occupancy Report based on Bedspace end date

muhammad-elabdulla:
This PR [CAS-244](https://dsdmoj.atlassian.net/browse/CAS-244) is to change the calculation of occupancy rate in Occupancy Report based on Bedspace start and end date. This will include:

1. The Occupancy Rate in the Occupancy Report will be calculated as follows:
occupancy percentage = bookedDaysActiveAndClosed / bedspaceOnlineDays
2. Add new columns to the report:
- bedspaceStartDate - set to the Creation Date of the Bedspace
- bedspaceEndDate - set if it not null
- bedspaceOnlineDays this will be calculated as the number of days between these dates:
      - bedspaceStartDate or report start date - whichever is LATER
      - bedspaceEndDate or report end date - whichever is EARLIER
4. Remove the column totalDaysInTheMonth

[CAS-244]: https://dsdmoj.atlassian.net/browse/CAS-244?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/reporting/generator/BedUtilisationReportGenerator.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/reporting/generator/BedUtilisationReportGenerator.kt
@@ -13,7 +13,6 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.earliestDateOf
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.getDaysUntilInclusive
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.latestDateOf
 import java.time.LocalDate
-import java.time.YearMonth
 
 class BedUtilisationReportGenerator(
   private val bookingRepository: BookingRepository,
@@ -83,7 +82,12 @@ class BedUtilisationReportGenerator(
     }
 
     val totalBookedDays = bookedDaysActiveAndClosed
-    val daysInMonth = YearMonth.of(properties.year, properties.month).lengthOfMonth()
+    val bedspaceOnlineDaysEndDate =
+      if (this.endDate == null) endOfMonth else earliestDateOf(this.endDate!!, endOfMonth)
+
+    val bedspaceOnlineDays = latestDateOf(this.createdAt.toLocalDate(), startOfMonth)
+      .getDaysUntilInclusive(bedspaceOnlineDaysEndDate)
+      .count()
 
     val temporaryAccommodationPremisesEntity = premises as? TemporaryAccommodationPremisesEntity
     listOf(
@@ -103,8 +107,10 @@ class BedUtilisationReportGenerator(
         effectiveTurnaroundDays = effectiveTurnaroundDays,
         voidDays = voidDays,
         totalBookedDays = totalBookedDays,
-        totalDaysInTheMonth = daysInMonth,
-        occupancyRate = totalBookedDays.toDouble() / daysInMonth,
+        bedspaceStartDate = this.createdAt.toLocalDate(),
+        bedspaceEndDate = this.endDate,
+        bedspaceOnlineDays = bedspaceOnlineDays,
+        occupancyRate = totalBookedDays.toDouble() / bedspaceOnlineDays,
         uniquePropertyRef = premises.id.toShortBase58(),
         uniqueBedspaceRef = this.room.id.toShortBase58(),
       ),
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/reporting/generator/BedUtilisationReportGeneratorTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/reporting/generator/BedUtilisationReportGeneratorTest.kt
@@ -32,6 +32,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.WorkingDayCountS
 import java.time.LocalDate
 import java.time.OffsetDateTime
 
+@SuppressWarnings("LargeClass")
 class BedUtilisationReportGeneratorTest {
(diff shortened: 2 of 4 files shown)


detekt-baseline.xml
```diff
@@ -688,5 +688,6 @@
     <ID>UnusedPrivateProperty:UserService.kt$UserService$private val userTransformer: UserTransformer</ID>
     <ID>UnusedPrivateProperty:WorkingDayCountService.kt$WorkingDayCountService$i</ID>
     <ID>VariableNaming:ApplicationServiceTest.kt$ApplicationServiceTest.SubmitApplication$val _schema = application.schemaVersion as Cas2ApplicationJsonSchemaEntity</ID>
+    <ID>LargeClass:BedUtilisationReportGeneratorTest.kt$BedUtilisationReportGeneratorTest</ID>
```
davidatkinsuk, line 691: Going forward (don't sorry about it now), can we please use @SuppressWarning instead of adding entries into this file? This makes the exclusions more visible and (hopefully) more likely to be fixed

davidatkinsuk, line 691: Going forward (don't sorry about it now), can we please use @SuppressWarning instead of adding entries into this file? This makes the exclusions more visible and (hopefully) more likely to be fixed

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/reporting/generator/BedUtilisationReportGenerator.kt
```diff
@@ -103,8 +107,10 @@ class BedUtilisationReportGenerator(
         effectiveTurnaroundDays = effectiveTurnaroundDays,
         voidDays = voidDays,
         totalBookedDays = totalBookedDays,
-        totalDaysInTheMonth = daysInMonth,
-        occupancyRate = totalBookedDays.toDouble() / daysInMonth,
+        bedspaceStartDate = this.createdAt.toLocalDate(),
+        bedspaceEndDate = this.endDate,
+        bedspaceOnlineDays,
+        occupancyRate = totalBookedDays.toDouble() / bedspaceOnlineDays,
```
davidatkinsuk, line 113: the ticket mentions the following report filter changes. I couldn't square them with the existing query but i may be misunderstanding the requirement. Does this need doing?

```The report needs to only include Bedspaces where the following is true:

Bedspace Start Date (i.e. Creation Date) BEFORE Report End Date

AND (Bedspace End Date is AFTER Report Start Date OR Bedspace End Date is <blank>)

```

davidatkinsuk (changes requested):
Hi,

Functionally looks good for the modified occupancy rate.

Can you please rebase to remove the merge commits and remove the standalone linting fix commit?

Thanks!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/reporting/generator/BedUtilisationReportGenerator.kt
muhammad-elabdulla: This is the criteria to calculate bedspaceOnlineDays calculate the  which is in line 88

detekt-baseline.xml
muhammad-elabdulla: Thanks good to know for the future

davidatkinsuk (changes requested):
Changes look good. Can you please rebase into a single commit?

davidatkinsuk (approved):
LGTM 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/reporting/generator/BedUtilisationReportGenerator.kt
```diff
@@ -103,8 +107,10 @@ class BedUtilisationReportGenerator(
         effectiveTurnaroundDays = effectiveTurnaroundDays,
         voidDays = voidDays,
         totalBookedDays = totalBookedDays,
-        totalDaysInTheMonth = daysInMonth,
-        occupancyRate = totalBookedDays.toDouble() / daysInMonth,
+        bedspaceStartDate = this.createdAt.toLocalDate(),
+        bedspaceEndDate = this.endDate,
+        bedspaceOnlineDays,
```
sudhakar-plr, line 112: would be useful to explicitly use 'bedspaceOnlineDays' attributes and assign the value into the constructor like`bedspaceOnlineDays = bedspaceOnlineDays` . To avoid any issue when adding additional attribute to the bean.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/reporting/generator/BedUtilisationReportGeneratorTest.kt
```diff
@@ -877,7 +879,678 @@ class BedUtilisationReportGeneratorTest {
...
+    assertThat(result[0][BedUtilisationReportRow::bedspaceStartDate]).isEqualTo(LocalDate.parse("2024-01-16"))
+  }
+
+  @Test
+  fun `bedspaceEndDate show bedspace end date when it is not null`() {
+    val apArea = ApAreaEntityFactory()
+      .produce()
+
+    val probationRegion1 = ProbationRegionEntityFactory()
+      .withApArea(apArea)
```
sudhakar-plr, line 941: If possible, Data setup for bed can be refactored into reusable method, so that we can avoid duplicate code.

muhammad-elabdulla: Its not a big win as the premises will be needed later

PR #1625, opened 2024-03-18
(APS-450) Return distinct domain events in timeline

pezholio:
We’ve noticed when making multiple appeals, the timeline shows multiple events because we’re carrying out multiple JOINs on the query. I’ve gone through a number of permutations of using the `DISTINCT` operator, but nothing seems to work as we expect, so I’ve opted for calling `distinctBy` in the service method, which seems to do the trick. As the number of events is unlikely to get unmanageable, we’re not asking the system to do too much extra work (either on the application or database side)


davidatkinsuk (approved):
I agree with PR comment, not the ideal fix but pragmatic and solves the issue with minimal performance issues

PR #1626, opened 2024-03-18
Fix bug that would withdraw unrelated placements

davidatkinsuk:
The findAllAdhocOrUnknownByApplication was not filtering out bookings not related to the application.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BookingEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BookingEntity.kt
@@ -133,7 +133,7 @@ interface BookingRepository : JpaRepository<BookingEntity, UUID> {
   fun findAllByCrn(crn: String): List<BookingEntity>
 
   @Query(
-    "SELECT b From BookingEntity b WHERE b.adhoc IS TRUE or b.adhoc IS NULL",
+    "SELECT b From BookingEntity b WHERE b.application = :application AND (b.adhoc IS TRUE or b.adhoc IS NULL)",
   )
   fun findAllAdhocOrUnknownByApplication(application: ApplicationEntity): List<BookingEntity>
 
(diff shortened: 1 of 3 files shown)


edavey (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BookingEntity.kt
```diff
@@ -133,7 +133,7 @@ interface BookingRepository : JpaRepository<BookingEntity, UUID> {
   fun findAllByCrn(crn: String): List<BookingEntity>
 
   @Query(
-    "SELECT b From BookingEntity b WHERE b.adhoc IS TRUE or b.adhoc IS NULL",
+    "SELECT b From BookingEntity b WHERE b.application = :application AND b.adhoc IS TRUE or b.adhoc IS NULL",
```
pezholio, line 136: Should we have brackets here? i.e:

```suggestion
    "SELECT b From BookingEntity b WHERE b.application = :application AND (b.adhoc IS TRUE or b.adhoc IS NULL)",
```

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/WithdrawalTest.kt
```diff
@@ -628,6 +638,16 @@ class WithdrawalTest : IntegrationTestBase() {
               adhoc = null,
             )
 
+            // regression test to ensure other application's
```
pezholio, line 641: Could we create a booking with a `null` adhoc too?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BookingEntity.kt
pezholio: Or maybe `IS NOT FALSE` would cover both bases?

pezholio (changes requested):
I still think this will return all bookings where `adhoc` is not present

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BookingEntity.kt
davidatkinsuk: great spot, should be fixed now

pezholio (approved):

PR #1627, opened 2024-03-18
Check to ensure withdrawal is limited to same app

davidatkinsuk:
We previously had a bug in the code where withdrawing an application would inadvertently cascade to adhoc bookings from other applications.

Whilst this bug has been fixed and specific regression tests added for the specific scenario, we want to add additional guard rails in the cascade withdrawal logic to ensure is not possible for cascades of any entity types. This is especially important as withdrawal cascades are somewhat opaque on the UI as the user it not explicitly told which entities will be withdrawn on cascade.

This commit adds a check in the withdrawal logic to ensure that all entities being withdrawn belong to the same application as the root entity being withdrawn.

Note that this implementation isn’t perfect as the check only occurs after the root entity has already been withdrawn (ideally this would be blocked). This is acceptable because this scenario should never occur, and the guardrails are only in place to avoid very damaging side affects, which _would_ be blocked.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1WithdrawableTreeBuilder.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1WithdrawableTreeBuilder.kt
@@ -50,6 +50,7 @@ class Cas1WithdrawableTreeBuilder(
     }
 
     return WithdrawableTreeNode(
+      applicationId = application.id,
       entityType = WithdrawableEntityType.Application,
       entityId = application.id,
       status = applicationService.getWithdrawableState(application, user),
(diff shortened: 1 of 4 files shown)


edavey (approved):
Thanks, this protection is a valuable safeguard. 

Some other ideas for managing these cascades:

- show the user a complete list of what will and won't be withdrawn in a proposed cascade BEFORE they confirm
- show the user a list of what withdrawn AFTER withdrawal on the "success" page
- record a new "withdrawal-cascade" domain event which records the details of the cascade which was executed. Having a point-in-time record of the entities which were judged withdrawable and not-withdrawable in their tree could be va aluable debugging artefact and an important record where entities need to be reinstated  or recreated.

davidatkinsuk:
> Thanks, this protection is a valuable safeguard.
> 
> Some other ideas for managing these cascades:
> 
> * show the user a complete list of what will and won't be withdrawn in a proposed cascade BEFORE they confirm
> * show the user a list of what withdrawn AFTER withdrawal on the "success" page
> * record a new "withdrawal-cascade" domain event which records the details of the cascade which was executed. Having a point-in-time record of the entities which were judged withdrawable and not-withdrawable in their tree could be va aluable debugging artefact and an important record where entities need to be reinstated  or recreated.

I agree with all of this, maybe we should try and make some of it a priority. I think the third point can be dealt to a degree with using existing domain events as we have a special withdrawal type when cascading occurs, but it's a little clunky!

PR #1628, opened 2024-03-18
Block cascade withdrawls if > 100 entities

davidatkinsuk:


edavey (approved):
Thanks for this additional safeguard. Just a tiny point: in the commit message you mention a max of 50, but in the code we set `MAX_WITHDRAWAL_COUNT` to 100...

davidatkinsuk:
> Thanks for this additional safeguard. Just a tiny point: in the commit message you mention a max of 50, but in the code we set `MAX_WITHDRAWAL_COUNT` to 100...

Good spot! I initially used 50 but i quickly rationalised some scenarios where that would be legitimate. i've fixed the commit

PR #1631, opened 2024-03-18
Added a new private function to remove XSS characters from the applic…

danhumphreys-moj:
…ation.data json string.

Added a new test to test that the characters are removed and the application is saved when the data string contains these characters.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/ApplicationService.kt
```diff
@@ -165,7 +165,7 @@ class ApplicationService(
       this.data = data
     }
 
-    val savedApplication = applicationRepository.save(application)
+    val savedApplication = applicationRepository.save(removeXssCharacters(application))
```
Harriethw, line 168: I'm wondering if it would be a bit better to be more specific about what we are doing here, by passing in just the `data` blob to this function before or at the same time as the `apply` above on line 164? 
I think passing the whole `application` in makes it read as if more fields than just the `data` field are being modified...

danhumphreys-moj: I can look at this Harriet.  There aren't any other application fields we need to apply this sanitisation logic to are there?

Harriethw: There's no other user-generated content except the `data` field currently so I don't think so

danhumphreys-moj: Changed this now.  No need for a dedicated function since we only do this here.  https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1631/files

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/ApplicationService.kt
```diff
@@ -162,7 +162,14 @@ class ApplicationService(
     }
 
     application.apply {
-      this.data = data
+      this.data = data?.replace("<", "")
+        ?.replace("＜", "")
```
Harriethw, line 166: ah sorry might not have been clear in my previous comment, I think we would still benefit from a utility function, but it can just take the data string as an argument. something like:

```
  private fun removeXssCharacters(data: String?): String? {
    if (data != null) {
      val xssCharacters = setOf('<', '＜', '〈', '〈', '>', '＞', '〉', '〉')
      var sanitisedData = ""
      xssCharacters.forEach { character ->
        sanitisedData = data.replace(character.toString(), "")
      }
      return sanitisedData
    }
    return null
  }
 ```

which could be called directly from the service

```
    application.apply {
      this.data = removeXssCharacters(application.data)
    }
```

Comment (danhumphreys-moj) [reply src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/ApplicationService.kt]: Right, so have made these changes Harriet, apart from:

```kotlin
var sanitisedData = data
```
```kotlin
   sanitisedData = sanitisedData?.replace(character.toString(), "")
```
And,
```kotlin
application.apply {
   this.data = removeXssCharacters(data)
}
```
Also changed the parametrised test to test a list of strings (previously chars) and included a test for `<＜〈〈>＞〉〉`:
```kotlin
@ValueSource(strings = ["<", "＜", "〈", "〈", ">", "＞", "〉", "〉", "<＜〈〈>＞〉〉"])
```

Harriethw (approved):

PR #1633, opened 2024-03-19
Placement Application to Placement Request Migration Job Improvements

davidatkinsuk:
Relates to https://dsdmoj.atlassian.net/browse/APS-438

* Logging improvements
* Don't disregard applications with one or more placement apps with decisionMadeAt Set. Only disregard them if they all have it set
* Add manual fixes into migration for cases where we couldn't programmatically determine lines, but could when reviewing applications manually
* Update migration logic to ignore/leave alone placement applications already linked to placement requests


edavey (approved):
Thanks for detailing these changes so clearly.

PR #1635, opened 2024-03-19
(APS-514) Add notice type field to applications

pezholio:
This adds a new `noticeType` field to applications, to allow us to get/set the notice type of an application via the frontend and eventually replace the `isEmergencyApplicaiton` field.  The values are:

- `emergency` - less than 7 days between when the application was created and the arrival date
- `shortNotice` - less than 28 days between when the application was created and the arrival date
- `standard` - more than 28 days between when the application was created and the arrival date

There is also a backfill task to backfill the status of all applications. Once the work is implemented in the frontend to send the `noticeType` directly, we can simplify some of the short notice and emergency code and have one source of truth for an application's notice type, as well as assign short notice applications to emergency assessors.

### DIFF
--- a/src/main/resources/static/_shared.yml
+++ b/src/main/resources/static/_shared.yml
@@ -3512,6 +3512,7 @@ components:
         - update_cas2_applications_with_assessments
         - update_cas1_user_details
         - update_cas1_fix_placement_app_links
+        - update_cas1_notice_types
     PlacementDates:
       type: object
       properties:
--- a/src/main/resources/static/_shared.yml
+++ b/src/main/resources/static/_shared.yml
@@ -2202,6 +2202,8 @@ components:
               type: boolean
             caseManagerUserDetails:
               $ref: '#/components/schemas/Cas1ApplicationUserDetails'
+            noticeType:
+              $ref: '#/components/schemas/Cas1ApplicationTimelinessCategory'
           required:
             - isPipeApplication
             - isWomensApplication
@@ -2299,6 +2301,12 @@ components:
         - residencyManagement
         - bailAssessment
         - bailSentence
+    Cas1ApplicationTimelinessCategory:
+      type: string
+      enum:
+        - standard
+        - emergency
+        - shortNotice
     PrisonCaseNote:
       type: object
       properties:
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApplicationEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApplicationEntity.kt
@@ -9,6 +9,7 @@ import org.springframework.data.jpa.repository.Lock
 import org.springframework.data.jpa.repository.Modifying
 import org.springframework.data.jpa.repository.Query
 import org.springframework.stereotype.Repository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Cas1ApplicationTimelinessCategory
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.listeners.ApplicationListener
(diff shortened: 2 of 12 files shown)


src/main/resources/static/_shared.yml
```diff
@@ -2301,6 +2303,12 @@ components:
         - residencyManagement
         - bailAssessment
         - bailSentence
+    Cas1ApplicationNoticeType:
```
edavey, line 2306: just wondering if we could be a bit clearer with the terminology here? Would:

- `Cas1ApplicationTimelinessCategory` or
- `Cas1ApplicationTimelinessRating`

be clearer? 

It wasn't immediately clear to me what "notice" would represent here.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/migration/NoticeTypeSeedJob.kt
```diff
@@ -0,0 +1,71 @@
...
+import org.springframework.data.jpa.repository.Modifying
+import org.springframework.data.jpa.repository.Query
+import org.springframework.stereotype.Repository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Cas1ApplicationNoticeType
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApplicationEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApprovedPremisesApplicationEntity
+import java.util.UUID
+
+@Repository
+interface NoticeTypeSeedApplicationRepository : JpaRepository<ApplicationEntity, UUID> {
```
davidatkinsuk, line 17: Given we have a concept of Seed Jobs (that take CSVs), would it be ok to remove the word 'Seed' from this (maybe replace with Migration) to disambiguate?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/migration/NoticeTypeSeedJob.kt
```diff
@@ -0,0 +1,71 @@
...
+  @Query(
+    "UPDATE ApprovedPremisesApplicationEntity ap set " +
+      "ap.noticeType = 'emergency' " +
+      "where ap.isEmergencyApplication = true",
+  )
+  fun updateEmergencyApplicationStatus()
+
+  @Modifying
+  @Query("UPDATE ApprovedPremisesApplicationEntity ap set ap.noticeType = :status where ap.id in :ids")
+  fun updateApplicationStatus(ids: List<UUID>, status: Cas1ApplicationNoticeType)
```
davidatkinsuk, line 28: can we please rename this to updateApplicationNoticeType for clarity (and also change arg to noticeType)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/migration/NoticeTypeSeedJob.kt
```diff
@@ -0,0 +1,71 @@
...
+
+  @Query("SELECT ap FROM ApprovedPremisesApplicationEntity ap where ap.isEmergencyApplication = false")
+  fun getApplicationsThatRequireStatus(pageable: Pageable): Slice<ApprovedPremisesApplicationEntity>
```
davidatkinsuk, line 31: can we please rename this to getApplicationsThatRequireNoticeTypeUpdating for clarity

edavey (approved):
Capturing the (emergency | short notice | standard) timeliness rating in a structured form looks great. Do you think we could name the enum / column more clearly however? e.g. `timeliness_rating`?

davidatkinsuk (approved):
Just a few proposals from me around naming for readability

src/main/resources/static/_shared.yml
pezholio: 👍

PR #1638, opened 2024-03-19
[CAS-251] Add PDU to database and Referrals report

muhammad-elabdulla:
This PR [CAS-251](https://dsdmoj.atlassian.net/browse/CAS-251) includes:

1. Capture and store pdu as first class field in CAS3 application.
2. Populate the stored pdu into CAS3's referral report

[CAS-251]: https://dsdmoj.atlassian.net/browse/CAS-251?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/resources/db/migration/all/20240319141046__add_pdu_to_cas3_application.sql
+++ b/src/main/resources/db/migration/all/20240319141046__add_pdu_to_cas3_application.sql
@@ -1,0 +1,1 @@
+ALTER TABLE temporary_accommodation_applications ADD COLUMN pdu TEXT NULL;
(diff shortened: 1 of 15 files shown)


src/main/resources/db/migration/all/20240319141046__add_pdu_to_cas3_application.sql
```diff
@@ -0,0 +1,2 @@
+-- Alter table temporary_accommodation_applications by adding new optional column pdu
```
davidatkinsuk, line 1: i don't think this comment is required as DDL is fairly self explanatory?

davidatkinsuk (approved):
Looks good, one very minor comment from me

If you make a similar change in the future I propose that it's split into two commits. One adding the PDU field and populating it, and another updating the report.

PR #1640, opened 2024-03-20
(Fix) Carry out one update task per page of data on Notice Type migration job

pezholio:
Storing all those UUIDs in two arrays gave us memory issues


davidatkinsuk (approved):
LGTM 👍

PR #1641, opened 2024-03-20
(APS-514) Set notice type when updating

pezholio:
We do this with most of the other first-class fields, so we should do it here too for consistency.


davidatkinsuk (approved):
LGTM

PR #1643, opened 2024-03-20
APS-528 Populate bookingId property on booking changed evt

davidatkinsuk:
This will ensure the link to the booking then appears on the timeline in the UI

Screenshot 2024-03-20 at 10 26 28


pezholio (approved):
Nice spot!

PR #1646, opened 2024-03-20
APS-521 show application withdrawal reason in timeline events

davidatkinsuk:
Screenshot 2024-03-20 at 12 14 04


edavey (approved):
very neat, and thks for refactoring for improved clarity.

PR #1647, opened 2024-03-20
APS-404 Add new method to populate user's ap area and a migration job to test it

davidatkinsuk:


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/UserMappingServiceTest.kt
```diff
@@ -0,0 +1,167 @@
...
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApAreaRepository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.UserMappingService
+
+class UserMappingServiceTest {
+
+  private val apAreaRepository = mockk<ApAreaRepository>()
+
+  private val service = UserMappingService(apAreaRepository)
+
+  private val usersProbationRegionArea = ApAreaEntityFactory().produce()
```
edavey, line 24: would something like `usersApAreaThroughProbationRegion` or `usersProbationRegionApArea` be clearer here? I think using "Area" on it's own is inherently confusing as we have Delius "Probation area" and CAS1 "AP Area"...

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/UserMappingServiceTest.kt
```diff
@@ -0,0 +1,167 @@
...
+  private val usersProbationRegion = ProbationRegionEntityFactory()
+    .withDefaults()
+    .withApArea(usersProbationRegionArea)
+    .produce()
+
+  @Nested
+  inner class DetermineApArea {
+
+    @Test
+    fun `determineApArea when user not in national probation area, return probation's associated ap area`() {
```
edavey, line 35: I think this should be "return probation **region**'s associated ap area"?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/UserMappingServiceTest.kt
```diff
@@ -0,0 +1,167 @@
...
+      "XX, N43SCE, SEE",
+      "XX, N43WSW, SWSC",
+      "XX, N43OMD, Mids",
+      "XX, N43OLK, LON",
+      "XX, N43ONM, NW",
+      "XX, N43OSE, SEE",
+      "XX, N43OWS, SWSC",
+      "XX, N41CRU, NE",
+    )
+    fun `determineApArea when user in national probation area and mapping exists for team`(
```
edavey, line 96: would something like 

> determine ApArea **from team code** when user in national probation area and mapping exists for team

be more descriptive?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/UserMappingServiceTest.kt
```diff
@@ -0,0 +1,167 @@
...
+        .produce()
+
+      assertThatThrownBy {
+        service.determineApArea(
+          usersProbationRegion,
+          staffUserDetails,
+        )
+      }.hasMessage(
+        "Internal Server Error: Could not find a delius team mapping for delius user J_ALUCARD " +
+          "with delius probation area code N43 and teams []",
```
edavey, line 138: lovely, thks

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/UserMappingServiceTest.kt
```diff
@@ -0,0 +1,167 @@
...
+      )
+    }
+
+    @Test
+    fun `determineApArea when user in national probation area but mapping is not available for any of their teams`() {
```
edavey, line 143: would something like:

> throws error and logs details when user in national probation area but mapping is not available for any of their teams

be clearer when scanning the test file?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/migration/Cas1ValidateApAreaUserMapping.kt
```diff
@@ -0,0 +1,39 @@
...
+  private val userRepository: UserRepository,
+  private val userMappingService: UserMappingService,
+  private val communityApiClient: CommunityApiClient,
+) : MigrationJob() {
+  private val log = LoggerFactory.getLogger(this::class.java)
+  override val shouldRunInTransaction = false
+
+  @SuppressWarnings("MagicNumber", "TooGenericExceptionCaught")
+  override fun process() {
+    userRepository.findAll().forEach { user ->
```
edavey, line 19: I'll be interested to know how this behaves for CAS3 users. It doesn't make sense to record an AP Area for them, but I'm not sure we can distinguish between them from their db records once created. 

This is going to be a consideration in `UserService.getExistingUserOrCreate()` also -- as we'll want to set `User.ap_area_id` only for CAS1 users I think?

edavey (approved):
Looks good, but I'm wondering how we're going to distinguish CAS1 users as AP Area isn't meaningful for other users...

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/migration/Cas1ValidateApAreaUserMapping.kt
davidatkinsuk: As discussed, we'll only attempt to set/update the user.ap_area field if the call originated from the CAS1 UI (as we have no other way to identify if a given user will be using CAS1).

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/UserMappingServiceTest.kt
davidatkinsuk: Thanks for the excellent naming suggestions, it's much easier to understand now

PR #1648, opened 2024-03-20
(Fix) Set createdAt in UpdateApplicationCas1 test

pezholio:
Now we’re relying on the created at date (at least temporarily) to get the noticeType, we need to ensure that the date the application is created is consistent (rather than being a random date in the past) to avoid flaky tests. This sets that `createdAt` date to now, so when we set the arrivalDate as 10 days or 7 months from now, we get a predictable response for the noticeType


pezholio:
Sorry for letting this sneak in!

davidatkinsuk:
> Sorry for letting this sneak in!

no worries, it's pretty subtle! (also i did a very similar thing 2 days ago)

PR #1649, opened 2024-03-20
CAS2-338: Bulk search CRNs for CAS2 Submissions 

Harriethw:
https://dsdmoj.atlassian.net/browse/CAS2-338

Following investigating the slow response times of the /submissions endpoint, we think we can increase the efficiency  by bulk searching for CRN's. This will also help when we introduce the Prison dashboard.

Previously we iterated through each application and [make an individual call for the OffenderSummary here](https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/938a6dda5c012a3b80080ef109d80de6f63fdeee/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/OffenderService.kt#L86 )

This PR

* uses the `offenderDetailsDataSource.getOffenderDetailSummaries` method to search for a list of CRN's, and then return a map of `crn` -> `personName`. 
* The map can then be used by the `SubmissionsController` to map back to the relevant `applications`
* I then introduced a `partition` on a max search number set by CAS3, to protect against the off chance that this is called without pagination, and to future-proof if we want to use for reports or a dashboard.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/OffenderService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/OffenderService.kt
@@ -1,6 +1,8 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas2
 
+import org.apache.commons.collections4.ListUtils
 import org.slf4j.LoggerFactory
+import org.springframework.beans.factory.annotation.Value
 import org.springframework.http.HttpStatus
 import org.springframework.stereotype.Service
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.ApOASysContextApiClient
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/OffenderService.kt
```diff
@@ -106,11 +110,18 @@ class OffenderService(
     return offenderDetailsDataSource.getOffenderDetailSummaries(crns.toList())
   }
 
-  fun getOffenderNamesOrPlaceholder(crns: Set<String>): Map<String, String> {
-    // get all offender summaries mapped to crns
+  fun getMapOfPersonNamesAndCrns(crns: List<String>): Map<String, String> {
+    val personNamesListOfMaps = ListUtils.partition(crns.toList(), numberOfCrn).stream()
```
Gweaton, line 114: 👍

Gweaton (approved):
This looks good to me

PR #1650, opened 2024-03-20
(APS-514) Allocate short notice assessments to emergency assessors

pezholio:
This allocates any short notice assessments to emergency assessors, in addition to any emergency assessments. I've also removed a lot of the references to the `isEmergencyApplication` boolean to give us confidence that the `noticeType` field is now the source of truth


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/TaskEntity.kt
```diff
@@ -22,7 +22,7 @@ interface TaskRepository : JpaRepository<Task, UUID> {
         CASE
           WHEN :requiredQualification = 'womens' THEN apa.is_womens_application = true
           WHEN :requiredQualification = 'pipe' THEN apa.is_pipe_application = true
-          WHEN :requiredQualification = 'emergency' THEN apa.is_emergency_application = true
+          WHEN :requiredQualification = 'emergency' THEN (apa.notice_type = 'emergency' OR apa.notice_type = 'shortNotice')
```
davidatkinsuk, line 25: I wish i'd spotted this before but we're using API generated enums in the entity model. I'd much rather we had used an internal enum. I suspect it would be a pain to change this now?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApplicationEntity.kt
```diff
@@ -356,7 +356,7 @@ class ApprovedPremisesApplicationEntity(
       requiredQualifications += UserQualification.WOMENS
     }
 
-    if (isEmergencyApplication == true) {
+    if (noticeType == Cas1ApplicationTimelinessCategory.emergency || noticeType == Cas1ApplicationTimelinessCategory.shortNotice) {
```
davidatkinsuk, line 359: I propose we @Deprecate these two booleans as it looks like you've fixed all current usages?

davidatkinsuk, line 359: Is it correct that shortNotice results in an EMERGENCY user qualification? If yes, maybe the EMERGENCY qualification needs renaming?

davidatkinsuk (approved):
A few comments from me but generally looking really good

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApplicationEntity.kt
pezholio: Yeah, I did think about that, but we don't _really_ use this at the mo, as allocations are done via config

pezholio: 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/TaskEntity.kt
davidatkinsuk: No worries we can always change it later if it becomes a problem, it's not too difficult

PR #1655, opened 2024-03-20
APS-343 Use CAS1 Placement Application Email Service for all related emails

davidatkinsuk:
This is in preparation for making changes to these emails


pezholio (approved):
Nice - good to see we're moving more email logic towards these services

PR #1656, opened 2024-03-20
Move App Withdrawn email into standalone service

davidatkinsuk:
This is in preparation for changes to this email.

As part of the refactoring, i’ve made a few changes

1. Previously there were many personalisation attributes being included for the email, but only one is used in the template (crn). All other redundant attributes have been removed
2. The email was previously sent to the person triggering the withdrawal. This is incorrect and the email should be sent to the applicant owner/creator. This accounts for scenarios where the CRU withdraws the application


pezholio (approved):
LGTM - assuming that failing integration test is just flakiness

PR #1661, opened 2024-03-21
[CAS-227]Archive cas3 bedspace when premises is archived

sudhakar-plr:
PR to archive CAS3 bedSpace when CAS3's premises is archived. The bedSpace is archived by adding the `endDate` to currentDate.


davidatkinsuk (approved):
LGTM 👍

PR #1662, opened 2024-03-21
(APS-524) Attempt to fix missing documents

pezholio:
We've had reports from users that documents are missing when they try to attach documents to their application. I have a hunch that this may be down to the fact that not all documents have IDs applied to them. In https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/792 we attempted a short-term fix by filtering out any documents without IDs. Rather than just do this, here we autogenerate unique IDs for documents without an ID, so they still show up.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/community/GroupedDocuments.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/community/GroupedDocuments.kt
@@ -1,6 +1,8 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.model.community
 
+import com.fasterxml.jackson.annotation.JsonProperty
 import java.time.LocalDateTime
+import java.util.UUID
 
 data class GroupedDocuments(
   val documents: List<Document>,
@@ -10,7 +12,8 @@ data class GroupedDocuments(
 }
 
 data class Document(
-  val id: String?,
+  @JsonProperty("id")
+  private val _id: String?,
   val documentName: String,
   val author: String?,
   val type: DocumentType,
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/community/GroupedDocuments.kt
```diff
@@ -10,15 +11,20 @@ data class GroupedDocuments(
...
   val author: String?,
   val type: DocumentType,
   val extendedDescription: String?,
   val createdAt: LocalDateTime,
   val lastModifiedAt: LocalDateTime?,
   val parentPrimaryKeyId: Long?,
-)
+) {
+  val id: String
+    get() = this._id ?: UUID.nameUUIDFromBytes(
```
edavey, line 24: this is neat!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/community/GroupedDocuments.kt
```diff
@@ -10,15 +11,20 @@ data class GroupedDocuments(
...
+      listOf(this.documentName, this.createdAt).joinToString("").toByteArray(),
```
davidatkinsuk, line 25: as discussed, we should check to see if documentName is immutable to ensure the ID being generated is consistent on subsequent retrieval

davidatkinsuk (approved):
Looking good, hopefully this solves it!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/community/GroupedDocuments.kt
davidatkinsuk: Do we know why the ID would be null?

edavey (approved):
This looks like a pragmatic solution to handle these docs thks! Do we know if the exclusion zone document referred to on ticket 524 was in this category of "missing ID"? I'd check myself but I'm having a temporary auth problem!

pezholio:
I'm going to close this. As neat as the solution is, I don't think it solves the underlying issue, and I'm worried about introducing other issues.

PR #1663, opened 2024-03-21
APS-404 Cas1ValidateApAreaUserMapping job improvements

davidatkinsuk:
Reduced delay (integration teams suggested no delay way fine)

Updated user mapping service to ensure that an area will always be returned and an exception thrown otherwise

Renamed UserMappingService to Cas1UserMappingService as it’s returning a CAS1 specific entity (ApAreaEntity)


edavey (approved):
So we _could_ set AP Area on every User (CAS3 included) without error, but when we implement the code to set the association, we'll only go through with it if the request originates from CAS1?

davidatkinsuk:
> So we _could_ set AP Area on every User (CAS3 included) without error, but when we implement the code to set the association, we'll only go through with it if the request originates from CAS1?

Yes I think that's the safest approach as to not lock out a non-CAS1 user incase something changes

PR #1664, opened 2024-03-21
CAS2-345: Calculate HDC dates for seed data

Harriethw:
https://dsdmoj.atlassian.net/browse/CAS2-305

As the frontend work is now completed to throw validation errors on HDC/CRD dates that don't meet the requirements below, we need to ensure our seeded data for in-progress applications has valid dates, otherwise the task will not show as 'complete'. 

There is a risk that if we change or remove the `hdc-licence-dates` task on the frontend then this function will become out of date, but we have the same problem with the JSON we  build the applications from currently so I think it's bearable?

In this PR is a utility function that creates dates to this criteria:

* HDC cannot be more than 6 months before CR, but must be before the CRD
* CRD must be more than 21 days in future

And is then used in the auto script and local seeding job.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/Cas2AutoScriptTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/Cas2AutoScriptTest.kt
@@ -2,6 +2,7 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.unit.service
 
 import io.mockk.every
 import io.mockk.mockk
+import io.mockk.mockkStatic
 import io.mockk.verify
 import org.junit.jupiter.api.BeforeEach
 import org.junit.jupiter.api.Test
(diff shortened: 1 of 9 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/Cas2AutoScriptTest.kt
```diff
@@ -101,13 +103,16 @@ class Cas2AutoScriptTest {
 
     every { mockApplicationService.createCas2ApplicationSubmittedEvent(any()) } answers { }
     every { mockStatusUpdateService.createStatusUpdatedDomainEvent(any()) } answers { }
+    mockkConstructor(Cas2SeedUtils::class)
```
Harriethw, line 106: I'm not sure why I had to do this to mock the `insertHdcDates` - I would have thought a regular `mockk<Cas2SeedUtils>()` would have worked but it wasn't picking up the mock. Maybe because it's just a regular class that gets instantiated, whereas the other things we mock are annotated beans??

PR #1665, opened 2024-03-21
APS-530 Add additional fields to CAS1 withdrawal emails

davidatkinsuk:
Relates to https://dsdmoj.atlassian.net/browse/APS-530

The following emails have been updated to include a link to the application timeline, and the name of the person who triggered the withdrawal:

* application withdrawn
* assessment withdrawn
* request for placement withdrawm
* match request withdrawn
* placement withdrawn

This is controlled by a feature flag that is disabled in prod by default

### DIFF
--- a/helm_deploy/values-dev.yaml
+++ b/helm_deploy/values-dev.yaml
@@ -46,6 +46,7 @@ generic-service:
 
     URL-TEMPLATES_FRONTEND_APPLICATION: https://approved-premises-dev.hmpps.service.justice.gov.uk/applications/#id
     URL-TEMPLATES_FRONTEND_APPLICATION-APPEAL: https://approved-premises-dev.hmpps.service.justice.gov.uk/applications/#applicationId/appeals/#appealId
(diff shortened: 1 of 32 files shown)


edavey (approved):
Thanks, this is very clear.

But I'm not sure why we need a feature flag here? Probably I'm missing some complexity, but would it be adequate and simpler to:

- introduce the new template at Notify
- change the code to use that new template
- remove the old template at Notify once the new code has been deployed to production

davidatkinsuk:
> Thanks, this is very clear.
> 
> But I'm not sure why we need a feature flag here? Probably I'm missing some complexity, but would it be adequate and simpler to:
> 
> * introduce the new template at Notify
> * change the code to use that new template
> * remove the old template at Notify once the new code has been deployed to production

Using this approach we have the potential for new emails to be used in production before we've had a chance to test them in pre-prod. Note that we don't have a way to test these emails until we get to pre-prod

I think we need a discussion about email templates in general. If we rendered the template 'in code' and used notify as a glorified SMTP relay that would avoid this issue.

PR #1667, opened 2024-03-22
APS-404 Update and use user's ap area property for CAS1 (Do not merge)

davidatkinsuk:
There are two parts to this PR

1. Update AP Area and Team Codes on the user entity whenever the user is updated (we only update ap area if originating request is for CAS1)
2. Use user.apArea when determining which ap area the user belongs to

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ProbationRegionEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ProbationRegionEntity.kt
@@ -21,6 +21,9 @@ data class ProbationRegionEntity(
   @Id
   val id: UUID,
   val name: String,
+  /**
+   * If the AP Area for a user is required, instead use [UserEntity.apArea]
+   */
   @ManyToOne
   @JoinColumn(name = "ap_area_id")
   val apArea: ApAreaEntity,
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
@@ -1,5 +1,6 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.service
 
+import org.apache.commons.collections4.CollectionUtils
 import org.slf4j.LoggerFactory
 import org.springframework.beans.factory.annotation.Value
 import org.springframework.data.domain.Sort
@@ -32,6 +33,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.community.StaffUse
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.BadRequestProblem
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.NotFoundProblem
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.results.AuthorisableActionResult
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas1.Cas1UserMappingService
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.AllocationType
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.getMetadata
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.getPageable
@@ -53,6 +55,7 @@ class UserService(
   private val userQualificationAssignmentRepository: UserQualificationAssignmentRepository,
   private val probationRegionRepository: ProbationRegionRepository,
   private val probationAreaProbationRegionMappingRepository: ProbationAreaProbationRegionMappingRepository,
+  private val cas1UserMappingService: Cas1UserMappingService,
 ) {
   private val log = LoggerFactory.getLogger(this::class.java)
 
@@ -63,10 +66,11 @@ class UserService(
   fun getUserForRequest(): UserEntity {
     val deliusPrincipal = httpAuthService.getDeliusPrincipalOrThrow()
     val username = deliusPrincipal.name
+    val serviceForRequest = requestContextService.getServiceForRequest()
 
-    val user = getExistingUserOrCreate(username)
+    val user = getExistingUserOrCreate(username, serviceForRequest)
 
-    if (requestContextService.getServiceForRequest() == ServiceName.temporaryAccommodation) {
+    if (serviceForRequest == ServiceName.temporaryAccommodation) {
       if (!user.hasAnyRole(*UserRole.getAllRolesForService(ServiceName.temporaryAccommodation).toTypedArray())) {
         user.roles += userRoleAssignmentRepository.save(
           UserRoleAssignmentEntity(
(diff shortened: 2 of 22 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
```diff
@@ -333,6 +346,10 @@ class UserService(
   }
 
   private fun userHasChanged(user: UserEntity, deliusUser: StaffUserDetails): Boolean {
-    return (deliusUser.email !== user.email) || (deliusUser.telephoneNumber !== user.telephoneNumber) || (deliusUser.staff.fullName != user.name) || (deliusUser.staffCode != user.deliusStaffCode) || (deliusUser.probationArea.code != user.probationRegion.deliusCode)
+    return (deliusUser.email !== user.email) ||
+      (deliusUser.telephoneNumber !== user.telephoneNumber) ||
+      (deliusUser.staff.fullName != user.name) ||
+      (deliusUser.staffCode != user.deliusStaffCode) ||
+      (deliusUser.probationArea.code != user.probationRegion.deliusCode)
```
davidatkinsuk, line 353: This won't apply if the user's team codes change (which could lead to a change in ap area). We'd have to persist these locally to determine this.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
```diff
@@ -62,10 +64,11 @@ class UserService(
   fun getUserForRequest(): UserEntity {
     val deliusPrincipal = httpAuthService.getDeliusPrincipalOrThrow()
     val username = deliusPrincipal.name
+    val serviceForRequest = requestContextService.getServiceForRequest()
```
davidatkinsuk, line 67: originally i threw an exception here if this couldn't be resolved, but 500+ integration tests broke because they're not setting the service name header, so for now i've allowed it to be nullable

davidatkinsuk:
Lots of integration tests are currently failing because the users being created in integration tests do not have a value for user.apArea. Previously this was resolved via user.region.apArea, which was fine because user.region was always populated.

We need to consider how best to resolve this. We could populate user.apArea automatically in the factories, but this would then result in apArea being populated for non-CAS1 users, which doesn't accurately reflect user state during CAS2/CAS3 integration tests

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
davidatkinsuk: Have agreed with Ed to persist team codes going forward (should consider backfilling these along with ap area)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
davidatkinsuk: Resolving this as i've added this behaviour back in, i'm looking at fixing tests

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
```diff
@@ -303,9 +290,17 @@ class UserService(
...
+  private fun findProbationRegionFromArea(deliusUser: StaffUserDetails): ProbationRegionEntity {
+    val probationArea = deliusUser.probationArea
+    val probationRegion = probationAreaProbationRegionMappingRepository
       .findByProbationAreaDeliusCode(probationArea.code)?.probationRegion
+
+    return probationRegion ?: if (assignDefaultRegionToUsersWithUnknownRegion) {
+      log.warn("Unknown probation region code '${probationArea.code}' for user '${deliusUser.username}', assigning a default region of 'North West'.")
+      return probationRegionRepository.findByName("North West")!!
+    } else {
+      throw BadRequestProblem(errorDetail = "Unknown probation region code '${probationArea.code}' for user '${deliusUser.username}'")
```
davidatkinsuk, line 302: we now throw this on update of existing users, previously we would just ignore an unknown probation region and leave as-is. Ed, do you think there's any change that this'll cause an issue for CAS2/3?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ProbationRegionEntity.kt
```diff
@@ -21,6 +21,9 @@ data class ProbationRegionEntity(
   @Id
   val id: UUID,
   val name: String,
+  /**
+   * If the AP Area for a user is required, instead use [UserEntity.apArea]
+   */
```
edavey, line 26: thks for adding this comment/warning

src/main/resources/db/migration/local+dev+test/R__1_create_dev_users.sql
```diff
@@ -2,15 +2,15 @@
...
-    ('9807de9d-05b3-49e2-84b8-529790a299cc', 'C. Load-Tester', 'CAS-LOAD-TESTER', 2500041004, 'db82d408-d440-4eb5-960b-119cb33427cd', 'STAFFCODE'), -- Kent, Surrey & Sussex
-    ('0621c5b0-0028-40b7-87fb-53ec65704314', 'T. Assessor', 'TEMPORARY-ACCOMMODATION-E2E-TESTER', 2500041003, 'db82d408-d440-4eb5-960b-119cb33427cd', 'STAFFCODE'), -- Kent, Surrey & Sussex
-    ('7e36a89e-c69e-48b1-a5cf-a7c8949f432a', 'T. Referrer', 'TEMPORARY-ACCOMMODATION-E2E-REFERRER', 2500510725, 'db82d408-d440-4eb5-960b-119cb33427cd', 'STAFFCODE'), -- Kent, Surrey & Sussex
-    ('b5825da0-1553-4398-90ac-6a8e0c8a4cae', 'Tester Testy', 'TESTER.TESTY', 2500043547, 'c5acff6c-d0d2-4b89-9f4d-89a15cfa3891', 'STAFFCODE'), -- North East
-    ('695ba399-c407-4b66-aafc-e8835d72b8a7', 'Bernard Beaks', 'BERNARD.BEAKS', 2500057096, 'ca979718-b15d-4318-9944-69aaff281cad', 'STAFFCODE'), -- East of England
-    ('045b71d3-9845-49b3-a79b-c7799a6bc7bc', 'Panesar Jaspal', 'PANESAR.JASPAL', 2500054544, 'afee0696-8df3-4d9f-9d0c-268f17772e2c', 'STAFFCODE') -- Wales
+INSERT INTO "users" (id, name, delius_username, delius_staff_identifier, probation_region_id, delius_staff_code, ap_area_id) VALUES
```
edavey, line 5: should we add an explanatory note in the commit message to explain what we're changing here and why?

edavey (approved):
👍

src/main/resources/db/migration/local+dev+test/R__1_create_dev_users.sql
davidatkinsuk: done

PR #1668, opened 2024-03-22
(APS-534) Add jira_update to environment deploys

pezholio:
This adds a status to Jira tickets when a branch with a ticket ID is deployed, so we can see at a glance the status of a given ticket.


davidatkinsuk (approved):
LGTM 👍

PR #1670, opened 2024-03-22
CAS2-217: prevent submissions when there is no Prison Code

Harriethw:
More detailed context in Jira ticket https://dsdmoj.atlassian.net/jira/software/c/projects/CAS2/boards/1347?selectedIssue=CAS2-200

We had an incident in the dev environment where an application was submitted (a 'submitted at' date was given to it in the database) but the Domain Event failed to emit, because the prison code was missing from the Prison API. 

Here we ensure that this won't happen again by:
- If the API is down/unsuccessful: catching any `UpstreamApiExceptions` thrown by the Prison API (e.g. Not Found) when we try to `apply` the prison code to the Application Entity, and transforming the error into a General Validation Error to be handled by the UI. 
- If the API call succeeds but the prison code is missing: throwing an `UpstreamApiException` error with appropriate message, which then gets turned into a General Validation Error and the UI can display the message.

This is a bit of a band-aid over the issue rather than a solution - although we hope this should never happen in production data (and has not occurred yet). 

We should go on to decide whether we should allow an application to be created if the prison code is missing, or what we can prompt the user to do if the prison code is missing.

We should also consider moving this prison code onto the application data itself so we don't have to continually make potentially-breakable API calls. follow up card here https://dsdmoj.atlassian.net/browse/CAS2-367

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/SubmissionsController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/SubmissionsController.kt
@@ -34,6 +34,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.cas2.Submiss
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.PageCriteria
 import java.net.URI
 import java.util.UUID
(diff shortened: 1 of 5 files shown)


Gweaton (approved):
Well done for figuring this out - looks like a fair bit of complex test wrangling was required.

I'll get this one merged and deployed today.

PR #1672, opened 2024-03-22
(APS-526) Hide retired beds in calendar view

pezholio:
The calendar view is only used in the North East, but we’ve had a report that “deleted” beds are still showing in the booking calendar. We will probably be revisiting all this functionality soon, but for now let’s update the calendar view to remove all beds with a end_date before the start_date of the calendar view’s query. This way, we can still see historical bookings if necessary, but remove all old beds for current and future views.


davidatkinsuk (approved):
LGTM 👍

PR #1674, opened 2024-03-25
CAS2-333 add optional page parameter

danhumphreys-moj:
JIRA ticket: https://dsdmoj.atlassian.net/browse/CAS2-333

**User Story**

With page param:

As a consumer of the CAS API
When I make an index request to /applications?page=x
The I should receive the correct page of results
And any relevant metadata (e.g. total count)

Without page param:

When I make an index request to /applications
The I should receive all the correct applications

**Technical implementation**

The existing pattern recently introduced in this PR to the Submissions controller has been reused:

https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1413

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/ApplicationsController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/ApplicationsController.kt
@@ -7,9 +7,9 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.cas2.ApplicationsCas
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Application
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ApplicationSummary
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.NewApplication
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.SortDirection
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.UpdateApplication
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas2ApplicationEntity
-import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas2ApplicationSummary
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.NomisUserEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.BadRequestProblem
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.ConflictProblem
@@ -22,6 +22,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.NomisUserService
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas2.ApplicationService
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas2.OffenderService
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.cas2.ApplicationsTransformer
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.PageCriteria
 import java.net.URI
 import java.util.UUID
 import javax.transaction.Transactional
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/ApplicationsController.kt
```diff
@@ -39,16 +40,20 @@ class ApplicationsController(
...
 
-  override fun applicationsGet(isSubmitted: Boolean?): ResponseEntity<List<ApplicationSummary>> {
+  override fun applicationsGet(isSubmitted: Boolean?, page: Int?): ResponseEntity<List<ApplicationSummary>> {
     val user = userService.getUserForRequest()
 
-    val applications = when (isSubmitted) {
-      true -> applicationService.getSubmittedApplicationsForUser(user)
-      false -> applicationService.getUnsubmittedApplicationsForUser(user)
-      null -> applicationService.getAllApplicationsForUser(user)
+    val pageCriteria = PageCriteria("created_at", SortDirection.desc, page)
```
danhumphreys-moj, line 46: Changing the sort order from `desc` to `asc` results in test `Get all applications with pagination returns 200 with correct body and header` still passing.  Bug [CAS2-374](https://dsdmoj.atlassian.net/browse/CAS2-374) raised in backlog.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/ApplicationService.kt
```diff
@@ -52,16 +52,22 @@ class ApplicationService(
   @Value("\${url-templates.frontend.cas2.submitted-application-overview}") private val submittedApplicationUrlTemplate: String,
 ) {
 
-  fun getAllApplicationsForUser(user: NomisUserEntity): List<Cas2ApplicationSummary> {
-    return applicationRepository.findAllCas2ApplicationSummariesCreatedByUser(user.id)
+  fun getAllApplicationsForUser(user: NomisUserEntity, pageCriteria: PageCriteria<String>): Pair<MutableList<Cas2ApplicationSummary>, PaginationMetadata?> {
```
Gweaton, line 55: Can we squash this commit into the previous one? Ideally we'd have all tests passing on each commit, and I don't think the first commit works without this change.

Gweaton (commented):
Approach looks good, a couple of small things!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/ApplicationService.kt
danhumphreys-moj: Ahh yes, hadn't thought of structuring the commits in a way that tests always pass after very commit.  Good reasoning.

Have done this.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ApplicationTest.kt
```diff
@@ -304,6 +373,30 @@ class Cas2ApplicationTest : IntegrationTestBase() {
...
+     * If two dates are ascending the difference will be negative (which is set to 0).
+     *
+     * For a list of dates, the cumulative multiple of these signs will be 1 if all
+     * dates in the range are descending (= 1 x 1 x 1 etc.).
+     *
+     * If any dates are ascending the multiple will be 0 ( = 1 x 1 x 0 etc.).
+     *
+     * If all dates are ascending the multiple will also be 0 ( = 1 x 0 x 0 etc.).
+     */
+    private fun isCreatedAtDescending(responseBody: List<Cas2ApplicationSummary>): Boolean {
```
danhumphreys-moj, line 392: @Gweaton is this better?  Added a private function and some comments.  I slightly changed the logic, but generally this is what I would have used to check something like this.  Let me know if this kind of pattern is out of sorts with what's normally done.

src/main/resources/static/cas2-api.yml
```diff
@@ -62,6 +67,15 @@ paths:
                 type: array
                 items:
                   $ref: '_shared.yml#/components/schemas/ApplicationSummary'
+          headers:
+            X-Pagination-CurrentPage:
+              $ref: '#/components/headers/X-Pagination-CurrentPage'
```
Gweaton, line 72: Does this work for you if you try and view the OpenAPI spec in IntelliJ (top right of your editor, see screenshot)? My branch highlighted the other occurrence of this as an error so I had to make this `_shared.yml#/components/headers/X-Pagination-CurrentPage` - not sure if it's just a me thing or if you're seeing any issues?
image

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ApplicationTest.kt
Gweaton: I think encapsulating this in a function is much clearer, thanks. I'd probably call it `isOrderedByCreatedAtDescending` to be even more explicit to future readers - does that sound good to you?

Gweaton (approved):
Thanks for making those changes - a couple of small things that would be good to tweak/investigate but otherwise good to go!

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ApplicationTest.kt
danhumphreys-moj: Yes I agree.  Makes it clearer.  Changed this now.

src/main/resources/static/cas2-api.yml
danhumphreys-moj: I am yes.  Missed this earlier.  I've added a more recent commit to address this issue in 8 places (4 for /applications and 4 for /submissions).  Would you mind re-approving George then I will release this morning?

@Harriethw tagged for info

Gweaton (approved):
Thanks for the updates!

PR #1675, opened 2024-03-25
[CAS-225] Populate Release Date in Temporary Accommodation Application table from json

muhammad-elabdulla:
This PR [CAS-225](https://dsdmoj.atlassian.net/browse/CAS-225) will populate the release date field in temporary accommodation application from data json field

[CAS-225]: https://dsdmoj.atlassian.net/browse/CAS-225?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/resources/db/migration/all/20240325170535__backfill_new_release_date_field_on_cas3_referrals.sql
+++ b/src/main/resources/db/migration/all/20240325170535__backfill_new_release_date_field_on_cas3_referrals.sql
@@ -1,0 +1,7 @@
+UPDATE temporary_accommodation_applications AS ta
+SET person_release_date = (
+    SELECT TO_DATE(app.data->'eligibility'->'release-date'->>'releaseDate','YYYY-MM-DD')
+    FROM applications AS app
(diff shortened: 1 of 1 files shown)


sudhakar-plr (commented):
Need to check 'release-date' format stored in the json are correct or not. I recently come across a referral where the release-date is stored in the json as `24-04-19` instead of `2024-04-09`. So not sure the sql script conversion will fail to parse to `yyyy-mm-dd` format

muhammad-elabdulla:
There are only one case in PreProd when the date is not well formatted in data field json
The release date is 24 -03-08 so it will be casted as 0024-03-08

sudhakar-plr (approved):
Approving,  since only 1 case exists(human error) with date 24 -03-08  and it can be transformed to 0024-03-08 without error. Hence its not going to fail the db migration.

PR #1676, opened 2024-03-25
APS-538 Add job to fix broken ‘further info’ assessments

davidatkinsuk:
There is a known issue that affects assessments where a user has requested further information, and goes back to edit the respons, selecting ‘no’ beacause they don’t want to trigger a re-request. This inconsistent state leads to an infinite loop when the UI is attempting to validate the answers.

This commit introduces a seed job the updates the ‘sufficient information’ response to ‘yes’ to break the infinite loop. It can be used to ‘fix’ assessments in this inconsistent state

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/AssessmentEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/AssessmentEntity.kt
@@ -216,6 +216,10 @@ interface AssessmentRepository : JpaRepository<AssessmentEntity, UUID> {
   @Modifying
   @Query("UPDATE ApprovedPremisesAssessmentEntity a SET a.dueAt = :dueAt WHERE a.id = :id")
   fun updateDueAt(id: UUID, dueAt: OffsetDateTime?)
+
+  @Modifying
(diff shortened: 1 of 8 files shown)


edavey (approved):
Thanks for adding the clear tests to demonstrate what's changed and what's left untouched.

I haven't investigated the cause of the bug but it might be of interest to note to our UI team that CAS2 have a `deleteOrphaned{Thing}` pattern for tidying up related `application.data` when the answer to a question with side-effects is changed. See https://github.com/ministryofjustice/hmpps-community-accommodation-tier-2-ui/blob/0a4d02a86285b50e20fb748dec58963e53ffc69f/server/utils/applications/deleteOrphanedData.ts

davidatkinsuk:
> Thanks for adding the clear tests to demonstrate what's changed and what's left untouched.
> 
> I haven't investigated the cause of the bug but it might be of interest to note to our UI team that CAS2 have a `deleteOrphaned{Thing}` pattern for tidying up related `application.data` when the answer to a question with side-effects is changed. See https://github.com/ministryofjustice/hmpps-community-accommodation-tier-2-ui/blob/0a4d02a86285b50e20fb748dec58963e53ffc69f/server/utils/applications/deleteOrphanedData.ts

Thanks Ed, that's good to know. I think if we have another issue like this we can make the seed job a bit more generic.

PR #1677, opened 2024-03-26
(APS-473) Fix CAS1 migration

yndajas:
This needs updating to reflect the
`is_esap_application`/`is_pipe_application` to `ap_type` change, Without this, running the Tilt stack with `--local-api` will fail


danlivings-dxw (approved):
Thanks, I should have included this in #1618 but I must have missed it.

PR #1678, opened 2024-03-26
APS-549 Use placement_app submission date for due at calcs

davidatkinsuk:
This commit changes the placement application due at calculation to use the submission date instead of the creation date.

The previous behaviour isn’t generally an issue because users can’t start creating a request for placement on the UI and then return to it later unless they bookmark it, which is unlikely.


pezholio (approved):
LGTM 👍

PR #1679, opened 2024-03-26
[APS-523] Allow sorting CAS1 assessments by `dueAt` field

danlivings-dxw:
> See [APS-523 on the Approved Premises Jira board](https://dsdmoj.atlassian.net/browse/APS-523).

This PR introduces the ability to sort CAS1 assessments using the `dueAt` field.

As CAS1 assessments do not have a due date, the `GET /assessments` endpoint will return a `400 Bad Request` response if this sort field is used for the CAS1 service.


pezholio (approved):
LGTM 👍

PR #1680, opened 2024-03-26
APS-544 Update CAS1 booking cancellation reasons

davidatkinsuk:
This PR introduces a sort order column to booking cancellation_reasons. Cancellation reasons are sorted by sort order and then name, where lowest sort order numbers (0) are shown first.

Screenshot 2024-04-02 at 16 00 35


edavey (approved):
neat, thanks.

PR #1685, opened 2024-03-27
Tidy up user update on change tests

davidatkinsuk:
Before making any more changes to the update user logic i’ve given the tests a tidy up and added a test for each field that triggers an entity update


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/UserServiceTest.kt
```diff
@@ -451,8 +451,6 @@ class UserServiceTest {
       assertUserUpdated(user, deliusUser, newProbationRegion)
     }
 
-    // probation area code
-
```
edavey: I think you can squash the 2 commits to lose this tidying commit?

edavey (approved):
👍

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/UserServiceTest.kt
davidatkinsuk: oops!

PR #1687, opened 2024-03-27
Move CAS1 app submitted email into email service

davidatkinsuk:
Tidy up of application service, moving email logic into cas1 email service


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/DomainEventService.kt
```diff
@@ -184,14 +184,15 @@ class DomainEventService(
     )
 
   @Transactional
-  fun saveApplicationWithdrawnEvent(domainEvent: DomainEvent<ApplicationWithdrawnEnvelope>) =
+  fun saveApplicationWithdrawnEvent(domainEvent: DomainEvent<ApplicationWithdrawnEnvelope>, emit: Boolean) =
```
edavey, line 187: I'm thinking that maybe the behaviour we ultimately want is:

- applicant or system "abandons" rather than "withdraws" an unsubmitted application
- we save and emit an `application.abandoned` event

i.e. we won't need this logic to selectively emit/not-emit a particular domain event. 

Once we start creating events which we know are purely internal perhaps we would include or mix in a "broadcast-able" or "emit-able" behaviour?

edavey (approved):
Very clear and thorough, thank you.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/DomainEventService.kt
davidatkinsuk: Yes agreed. I think the whole domain event patterns in code are bit clunky at the moment (and leads to lots of unit tests) so I think we can stream things considerable, tagging things are 'emit-able' would be one step towards streamlining it.

PR #1690, opened 2024-03-28
CAS2-344 enable bulk CRN search

danhumphreys-moj:
JIRA ticket: https://dsdmoj.atlassian.net/browse/CAS2-344

Following the implementation of bulk searching CRN’s in [this PR](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1649) for /submissions we should do the same on /applications.

Replaces `getPersonDetailAndTransformToSummary` with `getPeopleDetailsAndTransformToSummaries` that uses the `getOffenderSummariesByCrns` function in the CAS2 offenderService.

Once we have a map of all the offender detail summaries this is used to build the PersonInfoResult.

The call to `offenderDetailsDataSource.getOffenderDetailSummary` in `getInfoForPerson` is replaced with data collected in bulk.

So:

1. Do the search to get map of offender summaries with crn as key
2. Iterate through each application and for each application
-- use the crn to get the relevant offender summary from the map in step 1
-- use the nomsNumber to go get the inmateDetails from the inmate API
-- build the person info result and then use that to pass into `transformJpaSummaryToSummary`

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/ApplicationsController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/ApplicationsController.kt
@@ -10,7 +10,9 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.NewApplication
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.SortDirection
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.UpdateApplication
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas2ApplicationEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas2ApplicationSummary
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.NomisUserEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.PersonInfoResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.BadRequestProblem
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.ConflictProblem
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.ForbiddenProblem
@@ -53,7 +55,7 @@ class ApplicationsController(
 
     return ResponseEntity.ok().headers(
       metadata?.toHeaders(),
-    ).body(applications.map { getPersonDetailAndTransformToSummary(it, user) })
+    ).body(getPeopleDetailsAndTransformToSummaries(applications, user))
   }
 
   override fun applicationsApplicationIdGet(applicationId: UUID):
@@ -134,15 +136,29 @@ class ApplicationsController(
     return ResponseEntity.ok(getPersonDetailAndTransform(updatedApplication, user))
   }
 
-  private fun getPersonDetailAndTransformToSummary(
-    application: uk.gov.justice.digital
-    .hmpps.approvedpremisesapi.jpa.entity.Cas2ApplicationSummary,
-    user: NomisUserEntity,
-  ):
-    ApplicationSummary {
-    val personInfo = offenderService.getInfoForPersonOrThrowInternalServerError(application.getCrn())
+  private fun getPeopleDetailsAndTransformToSummaries(applications: List<Cas2ApplicationSummary>, user: NomisUserEntity): List<ApplicationSummary> {
+    val peopleInfo = offenderService.getOffenderSummariesByCrns(
+      applications.map { it.getCrn() }.toSet(),
+    )
 
-    return applicationsTransformer.transformJpaSummaryToSummary(application, personInfo)
+    val summaries = mutableListOf<ApplicationSummary>()
+
+    for (application in applications) {
+      val crn = application.getCrn()
+      val personInfo = when (val offenderDetailSummary = peopleInfo[crn]) {
+        null -> PersonInfoResult.Success.Restricted(crn, application.getNomsNumber())
+        else -> offenderService.getInfoForOffender(application.getCrn(), offenderDetailSummary)
+      }
+
+      summaries.add(
+        applicationsTransformer.transformJpaSummaryToSummary(
+          application,
+          personInfo as PersonInfoResult.Success,
+        ),
+      )
+    }
+
+    return summaries
   }
 
   private fun getPersonDetailAndTransform(
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ApplicationTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ApplicationTest.kt
@@ -5,6 +5,8 @@ import com.fasterxml.jackson.databind.JsonNode
 import com.fasterxml.jackson.databind.node.NullNode
 import com.ninjasquad.springmockk.SpykBean
 import io.mockk.clearMocks
+import io.mockk.every
+import io.mockk.mockk
 import org.assertj.core.api.Assertions
 import org.junit.jupiter.api.AfterEach
 import org.junit.jupiter.api.BeforeEach
@@ -29,6 +31,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.integration.httpmocks.Pr
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApplicationRepository
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas2ApplicationEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.NomisUserEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas2.OffenderService
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.randomDateTimeBefore
 import java.time.OffsetDateTime
 import java.util.UUID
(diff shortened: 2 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/ApplicationsController.kt
```diff
@@ -134,15 +136,28 @@ class ApplicationsController(
...
+  private fun getPeopleDetailsAndTransformToSummaries(applications: List<Cas2ApplicationSummary>, user: NomisUserEntity): List<ApplicationSummary> {
+    val peopleInfo = offenderService.getOffenderSummariesByCrns(
+      applications.map { it.getCrn() }.toSet(),
+    )
+
+    val summaries = mutableListOf<ApplicationSummary>()
+
+    for (application in applications) {
+      val crn = application.getCrn()
+      val offenderDetailSummary = peopleInfo[crn] ?: throw NotFoundProblem(crn, "No offender details found")
```
Harriethw, line 148: I think we can follow our previously established pattern of not throwing an error if the offender details haven't been found https://github.com/ministryofjustice/hmpps-approved-premises-api/commit/5360a6cd22ffd663df506dbcd4a59603a62bd72f

I don't think we want the whole request to error if the offender details haven't been found, but just to return with placeholder copy - if it's because of upstream api being down, that will be fixed shortly, if it's because person is genuinely not found then the user still needs to be able to access the application.

Harriethw: also means we will have to update the test `Get list of applications returns 500 when a person cannot be found` because we no longer want to return a 500

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/ApplicationsController.kt
```diff
@@ -134,15 +136,28 @@ class ApplicationsController(
...
-  private fun getPersonDetailAndTransformToSummary(
-    application: uk.gov.justice.digital
-    .hmpps.approvedpremisesapi.jpa.entity.Cas2ApplicationSummary,
-    user: NomisUserEntity,
-  ):
-    ApplicationSummary {
-    val personInfo = offenderService.getInfoForPersonOrThrowInternalServerError(application.getCrn())
+  private fun getPeopleDetailsAndTransformToSummaries(applications: List<Cas2ApplicationSummary>, user: NomisUserEntity): List<ApplicationSummary> {
+    val peopleInfo = offenderService.getOffenderSummariesByCrns(
+      applications.map { it.getCrn() }.toSet(),
```
Harriethw, line 141: if possible we could partition the number of crns it's possible to search by, just to be over-cautious, we've done this elsewhere https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/0f8d9d5e678ddf2e0fe6abd3efd22632f4c0ac40/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/OffenderService.kt#L114

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/ApplicationsController.kt
```diff
@@ -134,15 +136,29 @@ class ApplicationsController(
...
+    )
+
+    val summaries = mutableListOf<ApplicationSummary>()
+
+    for (application in applications) {
+      val crn = application.getCrn()
+      val personInfo = when (val offenderDetailSummary = peopleInfo[crn]) {
+        null -> PersonInfoResult.Success.Restricted(crn, application.getNomsNumber())
```
danhumphreys-moj, line 149: @Harriethw I'm not sure creating a `Restricted` PersonInfoResult is the right approach here but it allows me to cast to PersonInfoResult.Success later.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ApplicationTest.kt
```diff
@@ -309,22 +312,22 @@ class Cas2ApplicationTest : IntegrationTestBase() {
...
 
         webTestClient.get()
           .uri("/cas2/applications")
           .header("Authorization", "Bearer $jwt")
           .exchange()
           .expectStatus()
-          .is5xxServerError
-          .expectBody()
-          .jsonPath("$.detail").isEqualTo("Unable to get Person via crn: $crn")
+          .isOk
```
danhumphreys-moj, line 330: This test is failing due to a `ClassCastException` but I'm not sure why since we should be setting personInfo to an emptyMap() and getting an object from a map using a non-existing key should return null.

danhumphreys-moj: Also would be good to talk over the difference between this test and `Get list of applications returns successfully when the person cannot be fetched from the prisons API()`.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/ApplicationsController.kt
```diff
@@ -134,15 +136,29 @@ class ApplicationsController(
...
+        else -> offenderService.getInfoForOffender(application.getCrn(), offenderDetailSummary)
+      }
+
+      summaries.add(
+        applicationsTransformer.transformJpaSummaryToSummary(
```
Harriethw, line 154: I think this could be simplified by re-using what we have elsewhere, and just passing in the person name to the transformer because the whole PersonInfoResult is not needed except for the name?
Did you take at what we've done in the Submissions Controller? Could we  follow the same pattern and re-use `getMapOfPersonNamesAndCrns` from the OffenderService?

https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/3c8ba93099d7f5e0f25efb253dedc78c78f4ac48/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/SubmissionsController.kt#L181

Harriethw: Ah but that would break the frontend wouldn't it...

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/ApplicationsController.kt
Harriethw: I think we don't need to worry about above rn, as this will be used as paged endpoint and we have plans to refactor...

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/ApplicationsController.kt
```diff
@@ -134,15 +136,29 @@ class ApplicationsController(
...
   }
 
-  private fun getPersonDetailAndTransformToSummary(
-    application: uk.gov.justice.digital
-    .hmpps.approvedpremisesapi.jpa.entity.Cas2ApplicationSummary,
-    user: NomisUserEntity,
-  ):
-    ApplicationSummary {
-    val personInfo = offenderService.getInfoForPersonOrThrowInternalServerError(application.getCrn())
+  private fun getPeopleDetailsAndTransformToSummaries(applications: List<Cas2ApplicationSummary>, user: NomisUserEntity): List<ApplicationSummary> {
```
Harriethw, line 139: I think you could simplify this function a lot, as below:

```
  private fun getPeopleDetailsAndTransformToSummaries(applications: List<Cas2ApplicationSummary>): List<ApplicationSummary> {
    val offenderSummaries = offenderService.getOffenderSummariesByCrns(
      applications.map { it.getCrn() }.toSet(),
    )

    return applications.map { application ->
      val crn = application.getCrn()
      val personInfo = offenderService.getInfoForOffender(application.getCrn(), offenderSummaries[crn]!!)

      applicationsTransformer.transformJpaSummaryToSummary(application, personInfo as PersonInfoResult.Success)
    }
  }
 ```

File: src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ApplicationTest.kt (hunk 4 above)
Comment (Harriethw) [reply src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ApplicationTest.kt:330]: > Also would be good to talk over the difference between this test and Get list of applications returns successfully when the person cannot be fetched from the prisons API().

The test for 'when the person cannot be fetched from the prisons API' is regarding the `inmateDetails` we get from the `PrisonApi` whereas we get the `offenderDetails` from the `OffenderDetailsDataSource` / aka AP_DELIUS_CONTEXT...
you can see in the OffenderService when we make a call to get the inmate details, if it wasn't a success we just set the details to null
```
    val inmateDetails = offender.otherIds.nomsNumber?.let { nomsNumber ->
      when (val inmateDetailsResult = getInmateDetailByNomsNumber(offender.otherIds.crn, nomsNumber)) {
        is AuthorisableActionResult.Success -> inmateDetailsResult.entity
        else -> null
      }
    }
 ```

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ApplicationTest.kt
```diff
@@ -309,22 +312,22 @@ class Cas2ApplicationTest : IntegrationTestBase() {
     }
 
     @Test
-    fun `Get list of applications returns 500 when a person cannot be found`() {
+    fun `Get list of applications returns successfully when a person cannot be found`() {
```
Harriethw, line 315: You can set this test back to the way it was - apologies, I sent you down the wrong path with suggesting we don't error. We should still throw a 500...and I don't think we need anymore integration tests for this PR, as we haven't changed the end result of the code, just how we reach it

Harriethw: actually ignore the above! I've realised as we are changing where we get the CRNs from we will have to change the stubbed API for integration tests 🙃

PR #1692, opened 2024-03-28
Enable cas3 arrival,departure and update domain events in dev env

sudhakar-plr:
PR to enable CAS3 domain event in `dev` env
person.arrived.updated
person.arrived
person.departed
person.departed.updated


froddd (approved):
LGTM

PR #1696, opened 2024-04-02
Fix: update email AP email address when re-running seeding

edavey:
The `emailAddress` was missing from the "update" process when seeding Approved Premises.


davidatkinsuk (approved):
LGTM 👍

PR #1698, opened 2024-04-02
[CAS-60]-CAS3-Updated the error message related to bed-enddate

sudhakar-plr:
This PR is to update the error message thrown from bed-space end date validation to support UI.


froddd (approved):
LGTM

PR #1699, opened 2024-04-02
[APS-496] Introduce 'assessment allocated' domain event type

danlivings-dxw:
> See [APS-496 on the Approved Premises Jira board](https://dsdmoj.atlassian.net/browse/APS-496).

This PR introduces the `approved-premises.assessment.allocated` domain event type. This is emitted in two scenarios:

1. when an application is submitted and the assessment is first generated; or
2. when the assessment is reallocated.

Additionally, the event contains information about whether the assessment was assigned automatically.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/domainevents/DomainEventDescriber.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/domainevents/DomainEventDescriber.kt
@@ -1,6 +1,7 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.domainevents
 
 import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.StaffMember
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.DomainEventType
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.DomainEvent
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.DomainEventSummary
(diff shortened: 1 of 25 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/domainevents/DomainEventDescriber.kt
```diff
@@ -88,6 +90,15 @@ class DomainEventDescriber(
     return event.describe { "The assessment was appealed and ${it.eventDetails.decision.value}. The reason was: ${it.eventDetails.decisionDetail}" }
   }
 
+  private fun buildAssessmentAllocatedDescription(domainEventSummary: DomainEventSummary): String? {
+    val event = domainEventService.getAssessmentAllocatedEvent(domainEventSummary.id())
+    return event.describe { ev ->
+      val allocatedTo = ev.eventDetails.allocatedTo?.name ?: "an unknown user"
+      val allocatedDesc = ev.eventDetails.allocatedBy?.let { "by ${it.name}" } ?: "automatically"
```
pezholio, line 97: 👍

pezholio (approved):
LGTM - thanks for picking this up! 👍

src/main/resources/static/domain-events-api.yml
```diff
@@ -1334,6 +1370,48 @@ components:
...
+        - timestamp
+        - eventType
+        - eventDetails
+    AssessmentAllocated:
+      type: object
+      properties:
+        assessmentId:
+          $ref: '#/components/schemas/AssessmentId'
+        assessmentUrl:
+          $ref: '#/components/schemas/AssessmentUrl'
```
edavey, line 1398: It's good to include the ID and URL of the assessment being allocated, but I think we always need to include the ID and URL of the *application* also in our domain events. 

The application is the entity from which all related entities flow. For example when a Delius integration consumes a domain event it typically uses `applicationUrl` to provide a link back to the CAS1 service. 

See https://github.com/ministryofjustice/hmpps-probation-integration-services/blob/d59771747e6f2263e5f94509f7cbdba1f249993d/projects/approved-premises-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/approvedpremises/EventDetails.kt#L49-L71

src/main/resources/static/domain-events-api.yml
```diff
@@ -1410,7 +1410,6 @@ components:
         - assessmentId
         - assessmentUrl
         - personReference
-        - allocatedBy
```
edavey: Is this removed by mistake?

edavey (approved):
👍 looks good -- but I think we're missing the essential:

- `applicationId` and
- `applicationUrl`

properties?

src/main/resources/static/domain-events-api.yml
danlivings-dxw: I'd say that an allocation that happens automatically isn't "by" anyone, so would best be represented by `null`.

src/main/resources/static/domain-events-api.yml
danlivings-dxw: This has been done.

PR #1701, opened 2024-04-03
(CAS2-377) Fix submitted time shown for CAS2 submitted report

Harriethw:
We were previously using the `Instant` created on
the event data, which did not take into account
timezones/british summer time.

Using the `occurred_at` timestamp means we are
using an OffsetDatetime/ `TIMESTAMP WITH TIMEZONE` which is the same format used for
`application.created_at` and means it should
reflect the correct local time at which the event
happened. It also means we don't have to dig into
the JSON data to find the the timestamp.

This change is happening in the SQL query itself - we don't have unit tests for these queries, and
it's difficult to test in integration tests, see
this commit for information
https://github.com/ministryofjustice/hmpps-approved-premises-api/commit/260b1c6ded66c71d8495eca6512d9663be0dfe9e

But I think we have enough confidence by checking
that the `createdAt` of the report matches the
`occurredAt` of the event entity.

I've also tidied up the tests to use less verbose
means of getting timestamp.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ReportsTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ReportsTest.kt
@@ -98,13 +98,13 @@ class Cas2ReportsTest : IntegrationTestBase() {
       val event2Id = UUID.randomUUID()
       val event3Id = UUID.randomUUID()
 
-      val oldSubmitted = Instant.now().minusSeconds(daysInSeconds(365))
-      val oldCreated = oldSubmitted.minusSeconds(daysInSeconds(7))
+      val oldSubmitted = OffsetDateTime.now().minusDays(365)
+      val oldCreated = oldSubmitted.minusDays(7)
 
-      val newerSubmitted = Instant.now().minusSeconds(daysInSeconds(100))
-      val newerCreated = newerSubmitted.minusSeconds(daysInSeconds(7))
+      val newerSubmitted = OffsetDateTime.now().minusDays(100)
+      val newerCreated = newerSubmitted.minusDays(7)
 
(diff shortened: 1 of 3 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ReportsTest.kt
```diff
@@ -141,29 +141,29 @@ class Cas2ReportsTest : IntegrationTestBase() {
...
-        withCreatedAt(tooOldCreated.atOffset(ZoneOffset.ofHoursMinutes(0, 0)))
+        withCreatedAt(tooOldCreated)
         withData("{}")
-        withSubmittedAt(tooOldSubmitted.atOffset(ZoneOffset.ofHoursMinutes(0, 0)))
+        withSubmittedAt(tooOldSubmitted)
       }
 
       val event1Details = Cas2ApplicationSubmittedEventDetailsFactory()
-        .withSubmittedAt(oldSubmitted)
+        .withSubmittedAt(oldSubmitted.toInstant())
```
tahb, line 160: If the Postgres takes the time with zone and converts it to a utc time, do we also need to do this before we pass it in? [I figured that's what we were doing above](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1701/files#diff-5321e95ac99d42dcb5127d2aaffd665804f3aff8da93b397ca4bb75783ec7e07R133).

tahb (approved):
This looks reasonable to me.

It would indeed be nice to test this though so when we go back to UTC we can be sure there aren't failures, or any regressions between now and then. This might be old ground now but I wonder if we could [set a default timezone during test set up](https://www.baeldung.com/java-jvm-time-zone#3-setting-the-time-zone-from-the-application) `TimeZone.setDefault(TimeZone.getTimeZone(""));` and then set it back afterwards?

PR #1704, opened 2024-04-03
APS-594 Log in sentry if email has missing personalisation

davidatkinsuk:
Whilst some exceptions if notify can be ignored (e.g. user not on guest list for test environments, invalid email address), we should be raising alerts if the message sent to notify does not include all of the personalisation expected of the corresponding template


edavey (approved):
👍 thks

PR #1711, opened 2024-04-04
DRAFT: Cas2 390 nomis active agency

Harriethw:


tahb (approved):
LGTM! Really enjoyed seeing the refactorings.

To answer your question: We don't imagine users to be changing their prison very often at all right? So leveraging the sign in process as the trigger to check and update this seems proportionate. 

At worst the user submits a support request saying their seeing content from their old prison and we can tell them to sign in and out.

In future, or when we get more support requests than we are happy with, we could try and find and subscribe to a user updated event if that existed. 

The alternatives are, as you suggest, trying to do a mass synchronisation over night which would be expensive to do for every user. Even then if after the update is done (say 9am) an update is made, the user will be blocked from using seeing their new prison until the next day unless we also had a mechanism for kicking off the sync task.

PR #1715, opened 2024-04-05
APS-620 Allow for null NOMS Number in appeal domain events

davidatkinsuk:
This relates to https://dsdmoj.atlassian.net/browse/APS-620 where the user is getting a null pointer when attempting to create an appear.

This PR also moves the appeals domain event logic into a standalone class


edavey (approved):
👍 did you review other domain events to verify that Noms number is always optional? Or should we capture that as a separate task?

davidatkinsuk:
> 👍 did you review other domain events to verify that Noms number is always optional? Or should we capture that as a separate task?

Good idea. I've just checked and there is no other domain event using `nomsNumber!!`

PR #1717, opened 2024-04-08
APS-621 Make approved_premises.qcode unique

davidatkinsuk:
We had an error in our seeding data that used the same q_code for two approved premises. This change ensures data like this will be refused during any future seeding.


pezholio (approved):
Yikes. Good catch!

PR #1718, opened 2024-04-08
(APS-287) Swap Jagdeep and Vicky back

pezholio:
Jagdeep and Vicky had swapped regions for 8 weeks. They need to be switched back, so Vicky is allocated North East assessments and Jag is allocated Midlands assessments


davidatkinsuk (approved):
LGTM

PR #1724, opened 2024-04-08
(APS-602) Show completed tasks in task allocation section

pezholio:
This allows us to optionally return completed tasks instead of incomplete tasks (which is the default).

This makes the SQL a tad more complex. When we have time we should really think about refactoring to have actual Task entities which map to the relevant Assessment / Placement Application / Placement Request, but this is probably OK for now.

### DIFF
--- a/src/main/resources/static/api.yml
+++ b/src/main/resources/static/api.yml
@@ -2847,6 +2847,11 @@ paths:
           description: search by CRN or name
           schema:
             type: string
+        - name: isCompleted
+          in: query
+          description: filter by if the tasks are completed (defaults to `false`)
+          schema:
+            type: boolean
       responses:
         200:
           description: successfully retrieved tasks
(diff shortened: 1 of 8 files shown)


src/main/resources/static/api.yml
```diff
@@ -2847,6 +2847,11 @@ paths:
           description: search by CRN or name
           schema:
             type: string
+        - name: isCompleted
+          in: query
+          description: filter by if the tasks are completed (defaults to `true`)
```
davidatkinsuk, line 2852: i think default is false in this case?

davidatkinsuk (approved):
Looks good, just one minor thing from me

src/main/resources/static/api.yml
pezholio: Whoops!

PR #1725, opened 2024-04-08
APS-617: Add qualifications for Recovery-focused and Mental Health premises

edavey:
See Jira [APS-617: Add mental health and recovery focused as user qualifications](https://dsdmoj.atlassian.net/browse/APS-617)

In this PR we add two new user qualifications for handling tasks related to particular types of premises:
  
- `recovery-focused` for applications with the `RFAP` type
- `mental-health-specialist` for application with either `MHAP_ST_JOSEPHS` or `MHAP_ELLIOTT_HOUSE` type

The mapping between types of application and the user qualifications are made in `ApprovedPremisesApplicationEntity.getRequiredQualifications()`.

Through the `TaskEntity` it's now possible to filter tasks according to these "required qualifications" as they related to types of premises.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/UsersController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/UsersController.kt
@@ -163,5 +163,7 @@ class UsersController(
       UserQualification.lao -> uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserQualification.LAO
       UserQualification.emergency -> uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserQualification.EMERGENCY
       UserQualification.esap -> uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserQualification.ESAP
+      UserQualification.mentalHealthSpecialist -> uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserQualification.MENTAL_HEALTH_SPECIALIST
+      UserQualification.recoveryFocused -> uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserQualification.RECOVERY_FOCUSED
     }
 }
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/TasksTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/TasksTest.kt
@@ -821,6 +821,8 @@ class TasksTest : IntegrationTestBase() {
                 UserQualification.WOMENS,
                 UserQualification.ESAP,
                 UserQualification.PIPE,
+                UserQualification.RECOVERY_FOCUSED,
+                UserQualification.MENTAL_HEALTH_SPECIALIST,
               ).forEach { qualification ->
                 assessmentTasks[qualification] = listOf(
                   createAssessmentTask(qualification),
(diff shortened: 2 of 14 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/TasksTest.kt
```diff
@@ -858,9 +860,26 @@ class TasksTest : IntegrationTestBase() {
...
+
+        "placementApplication,WOMENS",
+        "placementApplication,PIPE",
+        "placementApplication,ESAP",
+        "placementApplication,EMERGENCY",
+        "placementApplication,RECOVERY_FOCUSED",
+        "placementApplication,MENTAL_HEALTH_SPECIALIST",
       )
       fun `Get all tasks filters by task type and required qualification`(taskType: TaskType, qualification: UserQualification) {
         val url = "/tasks?type=${taskType.value}&requiredQualification=${qualification.name.lowercase()}"
```
davidatkinsuk, line 885: are there corresponding tasks in the tasks collection for the new types?

I can see at line 820 (in the old code) something like this...

              listOf(
                UserQualification.WOMENS,
                UserQualification.ESAP,
                UserQualification.PIPE,
              ).forEach { qualification ->

Do we need to add to this? Would it be good to add an assertion at line 887 to make sure expected tasks is not empty? Or am i misreading this :)

davidatkinsuk (approved):
Thanks for splitting it up into very easy to read commits. One comment from me

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/UsersController.kt
```diff
@@ -163,5 +163,7 @@ class UsersController(
       UserQualification.lao -> uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserQualification.LAO
       UserQualification.emergency -> uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserQualification.EMERGENCY
       UserQualification.esap -> uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserQualification.ESAP
+      UserQualification.mentalHealthSpecialist -> uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserQualification.MENTAL_HEALTH_SPECIALIST
```
pezholio, line 166: There seem to be a lot of converting back and forth of enums. I recently made a tweak to `PlacementApplicationDecision` to keep the conversion logic in the enum itself, which @davidatkinsuk made some suggestions to tweak (which I forgot to implement). We could do something similar for `UserQualification` like so:

```kotlin
enum class UserQualification(val apiValue: uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.UserQualification) {
  WOMENS(uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.UserQualification.womens),
  PIPE(uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.UserQualification.pipe),
  LAO(uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.UserQualification.lao),
  ESAP(uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.UserQualification.esap),
  EMERGENCY(uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.UserQualification.emergency);

  companion object {
    fun valueOf(apiValue: uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.UserQualification): UserQualification? {
      for (e in entries) {
        if (e.apiValue == apiValue) {
          return e
        }
      }
      return null
    }
  }
}
```

That way we could do (for example):

```kotlin
UserQualification.mentalHealthSpecialist.apiValue //=> UserQualification.mentalHealthSpecialist -> uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserQualification.MENTAL_HEALTH_SPECIALIST
```

And:

```kotlin
UserQualification.valueOf(uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserQualification.MENTAL_HEALTH_SPECIALIST) // => UserQualification.mentalHealthSpecialist
```

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/TasksTest.kt
edavey: 👍 yes I think is covered in https://github.com/ministryofjustice/approved-premises-api/blob/e343d80d274904f6dc6bdb9f6e50ba8fc6421eed/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/TasksTest.kt#L820-L900

There's a bunch of iteration which builds up to define the `tasks` fixture:

```kt
    tasks = mapOf(
                TaskType.assessment to assessmentTasks,
                TaskType.placementRequest to placementRequestTasks,
                TaskType.placementApplication to placementApplicationTasks,
              )
```
and the parameterised tests then assert against elements of that tasks fixture, e.g. `tasks['placementRequest']['RECOVERY_FOCUSED']`

Probably we could make this easier to read!

davidatkinsuk: Perfect, i completely missed that. Thanks.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/UsersController.kt
edavey: This looks very neat, but I think I'm missing part of the puzzle, as the enum is defined in `package uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.UserQualification` which is auto-generated from the OpenApi spec at https://github.com/ministryofjustice/approved-premises-api/blob/e343d80d274904f6dc6bdb9f6e50ba8fc6421eed/src/main/resources/static/_shared.yml#L2956-L2965
So does your example `enum class UserQualification` over-ride that built definition? If so how?

I'm planning to merge this work as-is -- and then circle back "later" to make this improvement...

pezholio: Nah, so we have two `UserQualification` enums - an API one, and a JPA one, we spend a lot of time translating between the two, so it makes sense to have one source of truth for both. I did a similar piece of work here https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1733

PR #1726, opened 2024-04-08
APS-585 add request for placement created domain event

davidatkinsuk:
This PR adds a new 'Request for Placement Created' domain event. This is not emitted to SNS as it's only (current) purpose is to show on the timeline when request for placements are created in an application.

The domain event is raised when an application with arrival date/duration is assessed (at which point the placement_request is created), and when the user creates requests for placements after application assessment

The screenshot below shows the domain event rendered on the timeline for all variations:

Screenshot 2024-04-09 at 15 31 38

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/util/Dates.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/util/Dates.kt
@@ -103,3 +103,31 @@ fun latestDateOf(date1: LocalDate, date2: LocalDate): LocalDate {
 
   return date2
 }
+
+@SuppressWarnings("MagicNumber")
+fun toWeekAndDayDurationString(durationInDays: Int): String {
+  if (durationInDays < 7) {
+    return toDaysString(durationInDays)
+  } else if (durationInDays % 7 == 0) {
+    val weeks = durationInDays / 7
+    return toWeeksString(weeks)
+  } else {
+    val weeks = durationInDays / 7
+    val days = durationInDays % 7
+    return "${toWeeksString(weeks)} and ${toDaysString(days)}"
+  }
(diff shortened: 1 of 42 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/util/Dates.kt
```diff
@@ -103,3 +103,31 @@ fun latestDateOf(date1: LocalDate, date2: LocalDate): LocalDate {
 
   return date2
 }
+
+@SuppressWarnings("MagicNumber")
+fun toWeekAndDayDurationString(durationInDays: Int): String {
```
pezholio, line 108: 🥇

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/util/DatesTest.kt
```diff
@@ -118,4 +122,33 @@ class DatesTest {
 
     assertThat(dateTime.toUiFormattedHourOfDay()).isEqualTo("11am")
   }
+
+  @Nested
+  inner class ToWeekAndDayDurationString {
+
+    @ParameterizedTest
+    @CsvSource(
```
pezholio, line 130: Loving `CsvSource` ❤️

pezholio (approved):
Great stuff 👍

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/util/DatesTest.kt
davidatkinsuk: Yeah, it really is a game changer!

PR #1730, opened 2024-04-09
(APS-602) Return task outcome and recorded at date

pezholio:
Now we’re returning completed tasks, we need to record what the outcome was and when that outcome was recorded.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PlacementApplicationEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PlacementApplicationEntity.kt
@@ -19,6 +19,8 @@ import javax.persistence.JoinColumn
 import javax.persistence.ManyToOne
 import javax.persistence.OneToMany
 import javax.persistence.Table
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.PlacementApplicationDecision as ApiPlacementApplicationDecision
+
 @Service
 @Suppress("FunctionNaming")
 @Repository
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PlacementApplicationEntity.kt
```diff
@@ -153,6 +155,14 @@ enum class PlacementApplicationDecision {
   REJECTED,
   WITHDRAW,
   WITHDRAWN_BY_PP,
+  ;
+
+  fun convertToApi() = when (this) {
```
davidatkinsuk, line 160: nice

davidatkinsuk (approved):
LGTM 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PlacementApplicationEntity.kt
pezholio: I think we could do this with a lot of our enum classes

davidatkinsuk: I think we could do something like this and then add a function to do the inverse mapping if required (possibly as an extension function with kotlin enums can implement interfaces. Java is inflexible around enums

```
enum class PlacementApplicationDecision(val ApiPlacementApplicationDecision apiDecision) {
  ACCEPTED(accepted),
  REJECTED(rejected),
  WITHDRAW(withdraw),
  WITHDRAWN_BY_PP(withdrawnByPp),
}
```

PR #1732, opened 2024-04-09
APS-628 Raise error if allocator rule returns invalid user

davidatkinsuk:
Whilst such issues are logged at warn level, these are only useful once we know there is an issue (i.e. the user has informed us). This commit also raises a message in sentry so we can immediately investigate and rectify such issues.


pezholio (approved):
Great idea! 💡

PR #1733, opened 2024-04-09
(Chore) Improve enum conversion

pezholio:
This updates the changes made recently to `PlacementApplicationDecision` to enable easy conversion from JPA enum to API enum and vice-versa. Hopefully this is a pattern we can replicate elsewhere.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PlacementApplicationEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PlacementApplicationEntity.kt
@@ -150,18 +150,22 @@ enum class PlacementType {
   ADDITIONAL_PLACEMENT,
 }
 
-enum class PlacementApplicationDecision {
-  ACCEPTED,
-  REJECTED,
-  WITHDRAW,
-  WITHDRAWN_BY_PP,
+enum class PlacementApplicationDecision(val apiValue: ApiPlacementApplicationDecision) {
+  ACCEPTED(ApiPlacementApplicationDecision.accepted),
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PlacementApplicationEntity.kt
```diff
@@ -150,18 +150,22 @@ enum class PlacementType {
...
-    WITHDRAW -> ApiPlacementApplicationDecision.withdraw
-    WITHDRAWN_BY_PP -> ApiPlacementApplicationDecision.withdrawnByPp
+  companion object {
+    fun valueOf(apiValue: ApiPlacementApplicationDecision): PlacementApplicationDecision? {
+      for (e in entries) {
+        if (e.apiValue == apiValue) {
+          return e
+        }
+      }
+      return null
```
davidatkinsuk, line 167: i was playing around with interfaces, enums and extension functions to see if this could be done in a generic way and i couldn't get it to work (with extension functions at least). But maybe we could at least delegate to another functions for the for loop bit. i'll give it a go next time we add an enum

davidatkinsuk (approved):
looks great! 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PlacementApplicationEntity.kt
pezholio: Yeah, that would be good. It's definitely a pattern that will be useful in a ton of places!

PR #1735, opened 2024-04-10
CAS2-392: Allow referrer's to see and add notes to submitted applications within their own prison

Harriethw:
Screenshot 2024-04-10 at 11 38 41

_As a referrer (POM)
When I view a submitted application
For an offender with the same prison code as my active caseload
Then I should be able to leave a Note_

Apologies for a bit of a chunky PR, and quite a lot of moving about of tests.
This should allow a Nomis user to see and leave notes for applications with following criteria:

* Application is submitted
* User did not create the application
* The referringPrisonCode is the same as the user's activeCaseloadId (aka the prison code)

We are here using the `referringPrisonCode` saved
on our Application - this is currently read from the inmateDetails 
from the Prison API and saved at
the time when the application is submitted. 

There may be requirement in the future to update
this `referringPrisonCode` if, for example, the
offender has been moved between prisons after the
application has been submitted. For now though, we
are using the 'snapshot' of the code taken at the
time of submission.

## questions

Lots of branching if statements 🤯  - I tried to make it as readable as possible, open to any suggestions to streamline.

E2E UI test is here https://github.com/ministryofjustice/hmpps-community-accommodation-tier-2-ui/compare/cas2-392-view-prison-applications?expand=1


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/ApplicationNoteService.kt
```diff
@@ -155,8 +156,14 @@ class ApplicationNoteService(
     }
   }
 
-  private fun isApplicationCreatedByUser(application: Cas2ApplicationEntity, user: NomisUserEntity) =
-    application.createdByUser.id == user.id
+  private fun nomisUserCanAddNote(application: Cas2ApplicationEntity, user: NomisUserEntity):
```
Harriethw, line 159: hmm I wonder if this should all be abstracted away to the `UserAccessService` ?

PR #1736, opened 2024-04-10
(APS-603) Return releaseType in application summaries

pezholio:
We need this when displaying applications that are pending a placement request


davidatkinsuk (approved):
LGTM 👍

edavey (approved):

PR #1737, opened 2024-04-10
[CAS-316]-Bugfix to allow booking departure date is same as bedspace enddate

sudhakar-plr:
This PR is to allow booking departure date can be same as bedspace's enddate instead of stoping the booking.

Currently the booking validation stops the booking to be completed if the departure date is same as bedSpace endDate.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/BookingTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/BookingTest.kt
@@ -1152,7 +1152,7 @@ class BookingTest : IntegrationTestBase() {
   }
 
   @Test
-  fun `Create Temporary Accommodation Booking returns 409 Conflict when bed archived date is exactly on departure date`() {
+  fun `Create Temporary Accommodation Booking returns OK response when bed archived date is exactly on departure date`() {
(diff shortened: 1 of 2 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/BookingTest.kt
```diff
@@ -1215,12 +1215,26 @@ class BookingTest : IntegrationTestBase() {
           )
           .exchange()
           .expectStatus()
-          .is4xxClientError
+          .isOk
```
froddd, line 1218: Test title should be updated to `... returns OK...` instead of `... returns 409 Conflict...` ([line 1155](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1737/files#diff-814a441d05bc78abafaaa017a1339bc6150666556a526da2ae6665e187a2b9f7R1155))

froddd (approved):
Test title should be updated to reflect change of assertion, LGTM other than that

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/BookingTest.kt
sudhakar-plr: Its my fault, fixed it.

PR #1738, opened 2024-04-11
APS-623 send placement application withdrawn email to app creator

davidatkinsuk:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1EmailUtils.kt
```diff
@@ -12,6 +12,7 @@ fun ApprovedPremisesApplicationEntity.interestedPartiesEmailAddresses(): Set<Str
 
 fun PlacementApplicationEntity.interestedPartiesEmailAddresses(): Set<String> {
```
pezholio, line 13: I love the idea of keeping all this context in one place 👍

pezholio (approved):

PR #1741, opened 2024-04-11
APS-633 Use a suppression list for email errors

davidatkinsuk:


pezholio (approved):
LGTM 👍 Nice to keep the noise down in Sentry

PR #1744, opened 2024-04-11
(APS-603) Allow applications to be filtered by ap area

pezholio:
Now we want to show users applications that are still awaiting a placement request, we need to be able to filter applications by AP Area (I imagine this will be a requirement more generally anyway, now we've rolled out nationally).

As well as adding the AP Area filtering I've also made the long-wished for (at least by me!) ability to share database context between tests. Previously testing the database has meant we've had to have a ton of tests that have setup/teardown steps that run before every test and mean adding new functionality will add more latency to an already slow test suite. 

I haven't yet deleted the old application search tests (we'll still need some as the new tests run directly on the service), but I will do that assuming we're happy with this new approach.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApplicationEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApplicationEntity.kt
@@ -74,6 +74,9 @@ AND (
 AND (
     :status IS NULL OR (apa.status = :#{#status?.toString()})
 )
+AND (
+    (CAST(:apAreaId AS uuid) IS NULL) OR (apa.ap_area_id = :apAreaId)
+)
 """,
     countQuery = """
     SELECT COUNT(*)
@@ -89,6 +92,9 @@ AND (
       AND (
           :status IS NULL OR (apa.status = :#{#status?.toString()})
       )
+      AND (
+          (CAST(:apAreaId AS uuid) IS NULL) OR (apa.ap_area_id = :apAreaId)
+      )
     """,
     nativeQuery = true,
   )
@@ -96,6 +102,7 @@ AND (
     pageable: Pageable?,
     crnOrName: String?,
     status: ApprovedPremisesApplicationStatus?,
+    apAreaId: UUID?,
   ): Page<ApprovedPremisesApplicationSummary>
 
   @Query("SELECT a FROM ApplicationEntity a WHERE TYPE(a) = :type AND a.createdByUser.id = :id")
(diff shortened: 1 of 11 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/IntegrationTestBase.kt
```diff
@@ -932,3 +948,17 @@ abstract class IntegrationTestBase {
   fun loadPreemptiveCacheForOffenderDetails(crn: String) = communityApiClient.getOffenderDetailSummaryWithCall(crn)
   fun loadPreemptiveCacheForInmateDetails(nomsNumber: String) = prisonsApiClient.getInmateDetailsWithCall(nomsNumber)
 }
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+@Tag("isPerClass")
+abstract class PerClassIntegrationTestBase : IntegrationTestBase() {
```
davidatkinsuk, line 954: it may be more descriptive to call this something like 'InitialiseDatabasePerClassTestBase'?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/service/GetAllApprovedPremisesApplicationsTest.kt
```diff
@@ -0,0 +1,263 @@
...
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.RiskTier
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.RiskWithStatus
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.ApplicationService
+import java.sql.Timestamp
+import java.time.Instant
+import java.time.LocalDate
+import java.time.OffsetDateTime
+import kotlin.random.Random
+
+class GetAllApprovedPremisesApplicationsTest : PerClassIntegrationTestBase() {
```
davidatkinsuk, line 29: nice!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApplicationEntity.kt
```diff
@@ -74,6 +74,9 @@ AND (
 AND (
     :status IS NULL OR (apa.status = :#{#status?.toString()})
 )
+AND (
+    (CAST(:apAreaId AS pg_catalog.uuid) IS NULL) OR (apa.ap_area_id = :apAreaId)
```
davidatkinsuk, line 78: I think i introduced using `AS pg_catalog.uuid` but looking elsewhere in the code `AS uuid` should be fine i think?

davidatkinsuk (approved):
LGTM, just some minor comments 👍

PR #1746, opened 2024-04-12
APS-573 Add seed job to remove assessment details

davidatkinsuk:
https://dsdmoj.atlassian.net/browse/APS-573

This is required for a specific assessment in production where we want to remove all elements of the assessment data/document except for ‘sufficient-information-confirm’

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApplicationTimelineNoteEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApplicationTimelineNoteEntity.kt
@@ -22,7 +22,7 @@ data class ApplicationTimelineNoteEntity(
   val applicationId: UUID,
   @ManyToOne
   @JoinColumn(name = "created_by_user_id")
-  val createdBy: UserEntity,
+  val createdBy: UserEntity?,
   val createdAt: OffsetDateTime,
   val body: String,
 )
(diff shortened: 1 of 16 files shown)


edavey:
👍 do you think it would be prudent to include in this job the creation of an `ApplicationTimelineNote` on the association application to record something along the line of "Assessment details redacted". That would provide an adequate audit event with a timestamp?

edavey (approved):

davidatkinsuk:
> 👍 do you think it would be prudent to include in this job the creation of an `ApplicationTimelineNote` on the association application to record something along the line of "Assessment details redacted". That would provide an adequate audit event with a timestamp?

I think if we find ourselves using this a lot we could do this? I think there's already a note on the application in this case? I can add if you think we need it?

edavey:
> > 👍 do you think it would be prudent to include in this job the creation of an `ApplicationTimelineNote` on the association application to record something along the line of "Assessment details redacted". That would provide an adequate audit event with a timestamp?
> 
> I think if we find ourselves using this a lot we could do this? I think there's already a note on the application in this case? I can add if you think we need it?

I do feel that these data migrations ought to include some narrative in the associated timeline(s) to aid future / further investigation and debugging.

davidatkinsuk:
> > > 👍 do you think it would be prudent to include in this job the creation of an `ApplicationTimelineNote` on the association application to record something along the line of "Assessment details redacted". That would provide an adequate audit event with a timestamp?
> > 
> > 
> > I think if we find ourselves using this a lot we could do this? I think there's already a note on the application in this case? I can add if you think we need it?
> 
> I do feel that these data migrations ought to include some narrative in the associated timeline(s) to aid future / further investigation and debugging.

OK. Given what you've just said on the standup i'm going to rename the seed job to RedactAssessment too

edavey (approved):
👍 nice, thks!

PR #1747, opened 2024-04-12
CAS-317 Migrate PDU code to populate it in the database

muhammad-elabdulla:
Migrate the nDelius PDU codes to CAS3 database. 

- Add new column PDU code to the existing PDU table
- Update the PDU codes in probation_delivery_units from nDelius


src/main/resources/db/migration/all/20240411164054__add_pdu_code.sql
```diff
@@ -0,0 +1,110 @@
+ALTER TABLE probation_delivery_units ADD COLUMN pdu_delius_code TEXT;
```
sudhakar-plr, line 1: I would name it as 'delius_code' instead of 'pdu_delius_code'. Because the the `pdu` is covered in the table name `probation_delivery_unit`.

muhammad-elabdulla: Done 
I thought about it but to make it clear I added prefix of pdu_

src/main/resources/db/migration/all/20240411164054__add_pdu_code.sql
```diff
@@ -0,0 +1,110 @@
...
+
+Update probation_delivery_units Set delius_code = 'N57EKNT' Where id = 'f97a944d-72d4-4423-8b15-2db8e57bb012';
+Update probation_delivery_units Set delius_code = 'N57KECL' Where id = '1fb21be8-6f2b-483d-918f-9aad2e6e2838';
+Update probation_delivery_units Set delius_code = 'N57SUR' Where id = '25fb6114-ed80-460b-bd88-631b49b27304';
+Update probation_delivery_units Set delius_code = 'N57SUS' Where id = '81ec3d88-eecd-49d7-ad90-5dff98080dfb';
+Update probation_delivery_units Set delius_code = 'N57WSUS' Where id = '0cfab9ff-0547-41e7-af3d-20d040337f88';
+Update probation_delivery_units Set delius_code = 'N56ESXN' Where id = '6f410070-e848-4f4f-a285-6b6fa36dc539';
+Update probation_delivery_units Set delius_code = 'N56ESXS' Where id = 'f6fd8e4d-50a4-438d-83b8-335c538b8b66';
+Update probation_delivery_units Set delius_code = 'N56SUFF' Where id = '3fa6a349-5f48-4c42-8d96-938bda727c35';
+Update probation_delivery_units Set delius_code = 'N56NOSU' Where id = 'e5feabec-7f00-48ea-a9c0-27afdd492586';
```
sudhakar-plr, line 11: Just to make sure the ndelius-code is same in all env? if not then you may need to add these codes in test/dev envs in separate sql

muhammad-elabdulla: Its the same I already checked

PR #1751, opened 2024-04-15
(APS-661) Ensure task cant be reallocated multiple times

pezholio:
If a user reallocates a task and they attempt to reallocate a previously reallocated task again, this results in duplicate tasks, which has caused confusion in the past. 

This throws an error if the task has already been reallocated and the user attempts to reallocate it again.


davidatkinsuk (approved):
LGTM and thanks for the test tidy up 👍

PR #1752, opened 2024-04-15
fix returning latest status update on all applicatons

Harriethw:
Previous work was done to add latest status updates to the submitted applications query [1].

This query would not work if pagination was
required, I think because it wasn't
able to work out the left join. This rewrites the
psql query to only do the left join after the
applications have been selected - I think what
happens now is that the query goes and gets the
first page of applications as it's sub SELECT, and only then tries to do the LEFT JOIN.

I've also now added the query to the method that
returns all applications for a user.

[1] https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1710


Harriethw:
closing for a different solution https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1760

PR #1753, opened 2024-04-16
SUPERSEDED CAS2-401 refactor application summaries

danhumphreys-moj:
Ref: https://dsdmoj.atlassian.net/browse/CAS2-401

We currently return for ApplicationSummaries a full Person result, which includes a lot of data, and causes issues with how we handle whether the person has been found/is restricted.

I propose that we follow the standard set for /submissions and just return a personName which will be ‘Unknown’ if person not found or restricted.

This will mean we can then do a bulk search with the CRNs more easily, and just get the person’s name.
 
Tasks:
* API: add personName as field to Cas2ApplicationSummary
* UI: start using personName on the applications dashboard
* API: remove the Person field on application summaries and use the bulk search for getMapOfPersonNamesAndCrns to get CRNs


danhumphreys-moj:
Would be good to have a chat to confirm this PR contains sufficient integration test coverage.

danhumphreys-moj:
Marked as draft to avoid merging since this (as it stands) this will break the front end.

PR #1754, opened 2024-04-16
Increase use of InitialiseDatabasePerClassTestBase

davidatkinsuk:
I’ve picked some ‘low hanging fruit’ to migrate to the new InitialiseDatabasePerClassTestBase, where it’s clear there won’t be any issues retaining the database state between tests. These are typically tests where all but one tests are checking calls fail if permisisons are missing.


pezholio (commented):
LGTM 👍

pezholio (approved):

PR #1755, opened 2024-04-16
(Chore) Improve logging of database cleanup / restore

pezholio:
While digging into the running time of the tests, I saw a bunch of SQL queries that I didn’t expect to be running at the start of each test run. While digging into it I noticed that these weren’t actually SQL queries, but we were outputting an SQL query for each individual database table when attempting to restore the database to its original state. I’ve made a couple of tweaks to improve the logging so it’s clearer as to what’s going on and also only running `TableResults#insert` if there are any rows to insert.

I don’t think this will help performance much, but it’ll at least make it clearer as to what’s going on.


davidatkinsuk (approved):
LGTM 👍

PR #1756, opened 2024-04-16
(Chore) Save test results correctly

pezholio:
This should hopefully allow us to keep timing data correctly and allow the node balancing on CircleCI to split the tests more effectively.


pezholio:
Closing this, because I think we're doing the correct thing anyway

PR #1758, opened 2024-04-16
CAS-319 Check and update user PDU when logging in the portal

muhammad-elabdulla:
This PR [CAS-319](https://dsdmoj.atlassian.net/browse/CAS-319) includes:

- Add a new column PDU to the users table

- When the user logged in the portal part of the process is to get the user details from nDelius we will check the PDU assigned to this user. 
   - If there is no PDU assigned to the user then it will updated with the PDU from nDelius
   - If the user have a PDU but is different from the PDU in nDelius will be updated with the PDU from nDelius

[CAS-319]: https://dsdmoj.atlassian.net/browse/CAS-319?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
@@ -15,6 +15,8 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.UserSortField
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.ClientResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.CommunityApiClient
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ProbationAreaProbationRegionMappingRepository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ProbationDeliveryUnitEntity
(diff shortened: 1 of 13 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
```diff
@@ -284,6 +292,9 @@ class UserService(
       }
     }
 
+    val staffProbationDeliveryUnit =
+      staffUserDetails.teams?.firstOrNull()?.borough?.let { probationDeliveryUnitRepository.findByDeliusCode(it.code) }
```
davidatkinsuk, line 296: i appreciate you're following an established pattern but i think it may be good to introduce a private function that returns a nullable ProbationDeliveryUnitEntity given StaffUserDetails and then use that in the create/update methods to avoid code duplication

src/main/resources/db/migration/all/20240412134527__add_pdu_id_to_users.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE users
+ADD COLUMN probation_delivery_unit_id UUID;
```
davidatkinsuk, line 2: I think this should have a FK constraint?

davidatkinsuk (approved):
A few minor comments from me.

When this gets deployed into pre-prod could you please push into prod ASAP. We have an issue with the job that copies data from prod into pre-prod that means we loose all test user's permissions in pre-prod if a column exists in the users table in preprod, but not in prod

PR #1760, opened 2024-04-16
CAS2-407:  fix status updates

Harriethw:
This contains two main changes/fixes:

1. The `findSubmittedCas2ApplicationSummariesCreatedByUser` query was failing when a pageable object was passed in - from debugging the hibernate/sql going on, it looks like it was failing
when trying to do the count query, when there was a LEFT JOIN in place.
This adds a [custom countQuery](https://docs.spring.io/spring-data/jpa/docs/current/api/org/springframework/data/jpa/repository/Query.html#countQuery()) that is much simpler than
the select query. 
I've added in custom `countQueries` wherever I think they could simplify things.

2. Return `latestStatusUpdate` on `findAllCas2ApplicationSummariesByPrison` and `findAllCas2ApplicationSummariesCreatedByUser` - we should return these fields on any query that potentially returns submitted applications


tahb (approved):
🚀

PR #1761, opened 2024-04-16
(Chore) Tidyup premises tests

pezholio:
The premises tests are probably one of the worst-performing tests in CI. This seperates them out to make them easier to read, as well as sharing some of the database context in some of the tests, which should hopefully shave a minute or so off the total run time 🤞 

There's definitely more scope for improvement here, but I didn't want to sink too much time into it 😅


davidatkinsuk (approved):
Looks good! 👍

PR #1765, opened 2024-04-17
(APS-677) Add a `hasRequestsForPlacement` field to an Application

pezholio:
This adds a boolean to indicate is an application has any associated placement requests or applications. To test this, I’ve created a placement application and placement request for two applications and added their IDs to a list - `applicationIdsWithRequestsForPlacement`

When validating if a returned summary matches the application, we can then check if the application’s ID is contained in the `applicationIdsWithRequestsForPlacement` list - if it does, we know the expected value should be true, otherwise it’s false.


davidatkinsuk (approved):
LGTM 👍

PR #1767, opened 2024-04-17
CAS2-401 refactor application summaries (add personName)

danhumphreys-moj:
This PR contains the work required for the first bullet point of [CAS2-401](https://dsdmoj.atlassian.net/browse/CAS2-401)

Tasks:
* API: add personName as field to Cas2ApplicationSummary

We currently return for ApplicationSummaries a full Person result, which includes a lot of data, and causes issues with how we handle whether the person has been found/is restricted.

We propose that we follow the standard set for /submissions and just return a personName which will be ‘Unknown’ if person not found or restricted.

This will mean we can then do a bulk search with the CRNs more easily, and just get the person’s name.

[CAS2-401]: https://dsdmoj.atlassian.net/browse/CAS2-401?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/ApplicationsController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/ApplicationsController.kt
@@ -5,7 +5,6 @@ import org.springframework.http.ResponseEntity
 import org.springframework.stereotype.Service
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.cas2.ApplicationsCas2Delegate
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Application
-import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ApplicationSummary
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.NewApplication
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.SortDirection
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.UpdateApplication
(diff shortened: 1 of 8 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/ApplicationsController.kt
```diff
@@ -141,7 +140,7 @@ class ApplicationsController(
     .hmpps.approvedpremisesapi.jpa.entity.Cas2ApplicationSummary,
     user: NomisUserEntity,
   ):
-    ApplicationSummary {
+    uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Cas2ApplicationSummary {
```
Harriethw, line 143: minor: could this import path be moved to the top of the file? can see we have other models up there

Harriethw (approved):
approving as just minor changes to make :)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/ApplicationsController.kt
danhumphreys-moj: IntelliJ makes this an explicit import here because we already import the JPA Cas2ApplicationSummary there need it to avoid the ambiguity.  I've moved the import to the top and aliased model Cas2ApplicationSummary as `ModelCas2ApplicationSummary`, hopefully a reasonable compromise.

danhumphreys-moj:
Added crn and noms number and also unit test coverage for these (as recommended in Slack conversation).

PR #1768, opened 2024-04-17
(APS-643) Add an `info_requested` task status

pezholio:
This allows CRU managers to see at a glance what assessments are awaiting information requests from PPs.


davidatkinsuk (approved):
LGTM, presumably UI will be updated to handle the new status value before this goes live? Or will it handle it gracefully if not known?

pezholio:
I'm on it now 👍

PR #1769, opened 2024-04-17
CAS2-408: return author's name on ApplicationSummaries

Harriethw:
Our dashboard for displaying lists of applications to referrers needs the author's name - this PR gets the name from the `nomis_users` table in the sql queries that return `ApplicationSummaries` and adds it to the API model.

Note: the ApplicationSummary is shared across the models on `/applications` and `/submissions` [(we are aware this is a bit of tech debt we need to address) ](https://dsdmoj.atlassian.net/browse/CAS2-325) - the submissions end point does not need to see this author name atm, so I'm not currently returning it in the SubmissionsTransformer.

### DIFF
--- a/src/main/resources/static/codegen/built-cas2-api-spec.yml
+++ b/src/main/resources/static/codegen/built-cas2-api-spec.yml
@@ -2443,6 +2443,8 @@ components:
             createdByUserId:
               type: string
               format: uuid
+            createdByUserName:
+              type: string
             status:
               $ref: '#/components/schemas/ApplicationStatus'
             latestStatusUpdate:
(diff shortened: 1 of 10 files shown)


src/main/resources/static/codegen/built-cas2-api-spec.yml
```diff
@@ -2443,6 +2443,8 @@ components:
             createdByUserId:
               type: string
               format: uuid
+            createdByUserName:
```
Harriethw, line 2446: just realised maybe userName could easily be confused with username 🤔

danhumphreys-moj (approved):

PR #1773, opened 2024-04-18
(Chore) Attempt to improve build times

pezholio:
This is a grabbag of attempts to _try_ and improve build times. 

I started by changing the gradle.properties (which I think will only help us locally). There's a degree of cargo-culting there (see https://stackoverflow.com/questions/45327617/very-very-slow-gradle-build-on-android-studio), but it all seems to make sense.

I then noticed that the build was running twice, so have attempted to fix that, which shaves about a minute off the build time. I also noticed that the `UserAllocationsTest` had some expensive setup, so have shared the database context there.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -136,7 +136,7 @@ jobs:
             - gradle-
       - run:
           name: Build project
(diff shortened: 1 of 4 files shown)


davidatkinsuk (approved):
LGTM 👍

PR #1774, opened 2024-04-18
(APS-266) Remove Community API data sources

pezholio:
We no longer use these in any environment, so these are safe to remove.


davidatkinsuk (approved):
LGTM 👍 

i'm not sure if there's any functions we can/want to get rid of in the communityApiClient?

pezholio:
I think there might be, but let's keep them around for now

PR #1775, opened 2024-04-18
(Chore) Share database context in TasksTest

pezholio:
We are doing A LOT of shared setup / teardown here, especially in the filtering tests. This seems to be a significant contributor to our slow build times. I’m nested all the tests at one level now and made the filtering/pagination tests inherit from our `InitialiseDatabasePerClassTestBase` class, and things seem a lot quicker.

We could probs improve this further, but I think this addresses the vast majority of the problematic tests.


davidatkinsuk (approved):
lovely stuff 👍

PR #1778, opened 2024-04-18
(Chore) Return event dates in UI format

pezholio:
At the moment these get returned in ISO8601 format, which I love but it’s not very friendly for end users.


davidatkinsuk (approved):
LGTM 👍

PR #1781, opened 2024-04-18
[APS-587] Add API endpoints for Requests for Placement

danlivings-dxw:
> See [APS-587 on the Approved Premises Jira board](https://dsdmoj.atlassian.net/browse/APS-587).

This PR introduces the `GET /applications/{applicationId}/requests-for-placement` and `GET /applications/{applicationId}/requests-for-placement/{requestForPlacementId}` endpoints, which can be used to retrieve a list of requests for placement and a specific request respectively.

These endpoints use the newly-introduced `RequestForPlacement` schema, which provides a unified abstraction over both placement applications and placement requests. In the future both of these will be fully merged into the request for placement concept.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PlacementApplicationEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PlacementApplicationEntity.kt
@@ -8,6 +8,7 @@ import org.springframework.data.jpa.repository.Modifying
 import org.springframework.data.jpa.repository.Query
 import org.springframework.stereotype.Repository
 import org.springframework.stereotype.Service
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.WithdrawPlacementRequestReason
 import java.time.OffsetDateTime
 import java.util.UUID
 import javax.persistence.Column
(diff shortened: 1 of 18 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PlacementApplicationEntity.kt
```diff
@@ -169,15 +170,21 @@ enum class PlacementApplicationDecision(val apiValue: ApiPlacementApplicationDec
...
-  ALTERNATIVE_PROVISION_IDENTIFIED,
-  WITHDRAWN_BY_PP,
-  CHANGE_IN_CIRCUMSTANCES,
-  CHANGE_IN_RELEASE_DECISION,
-  NO_CAPACITY_DUE_TO_LOST_BED,
-  NO_CAPACITY_DUE_TO_PLACEMENT_PRIORITISATION,
-  NO_CAPACITY,
-  ERROR_IN_PLACEMENT_REQUEST,
-  RELATED_APPLICATION_WITHDRAWN,
+enum class PlacementApplicationWithdrawalReason(val apiValue: WithdrawPlacementRequestReason) {
```
pezholio, line 175: 🙌

pezholio (approved):
LGTM - nice work 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/RequestForPlacementService.kt
```diff
@@ -0,0 +1,63 @@
...
+  private val placementApplicationService: PlacementApplicationService,
+  private val placementRequestService: PlacementRequestService,
+  private val requestForPlacementTransformer: RequestForPlacementTransformer,
+) {
+  fun getRequestsForPlacementByApplication(applicationId: UUID): AuthorisableActionResult<List<RequestForPlacement>> {
+    if (applicationService.getApplication(applicationId) == null) {
+      return AuthorisableActionResult.NotFound("Application", applicationId.toString())
+    }
+
+    val placementApplications = placementApplicationService.getAllPlacementApplicationsForApplicationId(applicationId)
```
davidatkinsuk, line 22: Assuming the UI isn't filer out any results, I think this is going to return unexpected entities as it will return unsubmitted and reallocated placement apps

We have slightly strange behaviour at the moment where the UI will save unsubmitted placement applications in the backend, but there's no way for it to retrieve them, so they effectively become orphaned if the user navigates away. I don't think we should be returning these applications in this list as i don't think the UI knows what to do with them (although it would be nice to have the ability to managed unsubmitted placement_applications in the future)

Reallocated placement applications shouldn't be returned because they're effectively duplicates of the copy made when reallocation occurs. Ideally we'd handle this with a separate task entity instead of doing this, but that's for the future

I think you should be using the same method as the 'old' endpoint which is now called getAllActivePlacementApplicationsForApplicationId as we know this was returning the correct set of placement applications. Note that this backs onto a repository method that claims to filter out withdrawn placement_applications but that isn't the case as we do see withdrawn request for placements on the current UI. We probably need to tidy up the name on that

davidatkinsuk: If you take a look at ApplicaitonTest.GetPlacementApplications there's a bunch of test cases for these scenarios you can probably lift and shift to test the new endpoint

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/RequestsForPlacementTest.kt
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    @Test
+    fun `Get a Request for Placement for an application returns a 200 response with the expected placement application`() {
+      `Given a User` { user, jwt ->
+        val schema = approvedPremisesPlacementApplicationJsonSchemaEntityFactory.produceAndPersist()
+
+        `Given a Placement Application`(createdByUser = user, schema = schema) { placementApplication ->
+          webTestClient.get()
+            .uri("/applications/${placementApplication.application.id}/requests-for-placement/${placementApplication.id}")
```
edavey, line 199: are we still planning to implement a composite ID e.g.

- `.../requests-for-placement/placement_application:bc12b7e8-a404-4216-97ad-1b18199e0602`
- `.../requests-for-placement/placement_request:49f282ad-ff82-4747-8a7d-9c30407c594e`

I believe we thought that this would be helpful:

- to the controller or service handling the request (perhaps `RequestForPlacementService. getRequestForPlacement()`) and also, perhaps especially
- to humans when debugging / developing

See https://dsdmoj.atlassian.net/browse/APS-587?focusedCommentId=444903

edavey (commented):
👍 nice work! Just a question about the "composite" ID which we discussed at the outset...

PR #1783, opened 2024-04-19
(APS-691) Return information requests in timeline

pezholio:
A user has reported that once they have made an information request, it's impossible to see what the initial request was. I've added the ability to review it in the assess screen in https://github.com/ministryofjustice/hmpps-approved-premises-ui/pull/1742, but I think it'll be useful to show this information in the timeline too for audit purposes.


davidatkinsuk (approved):
LGTM 👍 

Do we know if this will formatted with line breaks when rendered on the UI?

pezholio:
There's some code in the UI PR that will do this 👍

PR #1785, opened 2024-04-19
(Chore) Share workspace more effectively in CI

pezholio:
When we build the project in the CircleCI build step, we throw away all the generated build info in the `~/.gradle` directory, meaning when we start our test steps, we have to initialise Gradle all over again, adding a couple of minutes to each test step.

This updates the build step to share everything in the `~/.gradle` home directory with the test steps, reducing the amount of time Gradle takes to spin up. We also remove the `checkout` step as we don't need this as the workspace is shared.

Now the build step is more effective, we can increase the number of parallel containers used in our integration tests and this will have more effect as we're spending less time in each container repeating the setup process.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -207,7 +207,7 @@ jobs:
       - store_artifacts:
           path: build/reports/tests
   integration_test:
-    parallelism: 6
+    parallelism: 7
     environment:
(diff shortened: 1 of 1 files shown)


davidatkinsuk (approved):
LGTM 👍

PR #1787, opened 2024-04-19
APS-638 add /people/{crn}/offences/v2 endpoint

davidatkinsuk:


src/main/resources/static/codegen/built-api-spec.yml
```diff
@@ -1485,6 +1486,37 @@ paths:
                 $ref: '#/components/schemas/Problem'
         500:
           $ref: '#/components/responses/500Response'
+  /people/{crn}/offences/v2:
```
pezholio, line 1489: I wonder if we can version using headers instead of a URL, so if we make a request with, for example:

```
Content-Type: application/json; version=2
```

We get the new behaviour. That way once we retire version 1 of the API, we're not left with a `*/v2` version of the URL we have to keep around

davidatkinsuk: That should be possible - presumably the UI would be OK dealing with sending custom mime types in the Accept header? If yes, I'll give it a go

pezholio: Yep, we should be good with that. Would also allow us to potentially configure it via feature flags, which could help

PR #1788, opened 2024-04-22
APS-687 Withdrawal emails are now sent if triggered by seed jobs

davidatkinsuk:
To address APS-687 we added a seed job to withdraw match requests (placement_requests) via a seed job.

When tested in pre-prod we saw an error sending the email due to:

`Status code: 400 {"errors":[{"error":"BadRequestError","message":"Missing personalisation: withdrawnBy"}],"status_code":400}`

This PR refactors how the Withdrawal Context model to provide a more consistent approach to identifying who or what triggered withdrawal of the entity. This is then subsequently leveraged to ensure we can always a determine a 'withdrawn by' value for withdrawal emails, regardless of whether it was triggered by a user or a seed job.


edavey (approved):
Very nice, thanks for these improvements

PR #1795, opened 2024-04-24
(Chore) Add Configuration class for Domain Event URLs

pezholio:
This adds a new `DomainEventUrlConfig` class that allows us to get url templates for a domain event type directly, instead of having to pass them as `@Value` annotated strings in our service classes, which is getting unweildy. It also allows us to easily map domain events to URL templates elsewhere in our code.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/DomainEventUrlConfig.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/DomainEventUrlConfig.kt
@@ -1,0 +1,49 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.config
+
+import org.springframework.boot.context.properties.ConfigurationProperties
+import org.springframework.context.annotation.Configuration
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.DomainEventType
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.UrlTemplate
+import java.util.UUID
+
+@Configuration
+@ConfigurationProperties(prefix = "url-templates.api")
+class DomainEventUrlConfig {
+  lateinit var cas1: Map<String, UrlTemplate>
+  lateinit var cas2: Map<String, UrlTemplate>
+  lateinit var cas3: Map<String, UrlTemplate>
+
+  fun getUrlForDomainEventId(domainEventType: DomainEventType, eventId: UUID): String {
+    val template = when (domainEventType) {
+      DomainEventType.APPROVED_PREMISES_APPLICATION_SUBMITTED -> cas1["application-submitted-event-detail"]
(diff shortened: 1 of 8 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/DomainEventUrlConfig.kt
```diff
@@ -0,0 +1,45 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.config
+
+import org.springframework.boot.context.properties.ConfigurationProperties
+import org.springframework.context.annotation.Configuration
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.DomainEventType
+
+@Configuration
+@ConfigurationProperties(prefix = "url-templates.api")
+class DomainEventUrlConfig {
+  lateinit var cas1: Map<String, String>
```
davidatkinsuk, line 10: this may be a headache but it would be really nice to use UrlTemplate instead of String for the map value? I think spring should be able to do the mapping automatically

davidatkinsuk (approved):
Funnily enough I was just doing something similar to simplify replaying of domain events. I prefer your solution though

LGTM 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/DomainEventUrlConfig.kt
```diff
@@ -0,0 +1,45 @@
...
+  lateinit var cas2: Map<String, String>
+  lateinit var cas3: Map<String, String>
+
+  fun getUrlForDomainEventType(domainEventType: DomainEventType): String {
```
davidatkinsuk, line 14: If all templates use eventId as the param name, could this take in the event ID and return the rendered url instead of the template? This would help ensure new templates use that param in the future too

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/DomainEventUrlConfig.kt
```diff
@@ -0,0 +1,45 @@
...
+    return when (domainEventType) {
+      DomainEventType.APPROVED_PREMISES_APPLICATION_SUBMITTED -> cas1["application-submitted-event-detail"]
```
davidatkinsuk, line 16: could we put the template ids into the enum ... this may be overkill :) i guess in hindsight it would have been good to use the enum name to build the template name

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/DomainEventUrlConfig.kt
pezholio: Nice idea!

PR #1796, opened 2024-04-24
(APS-673) Add a has domain event field to clarification notes

pezholio:
We will shortly be introducing a proper domain event for when information requests are created. To ensure we don't double up events in the timeline, this adds a `hasDomainEvent` boolean to the `AssessmentClarificationNoteEntity`, which defaults to `false`. We then filter out any entities which has this set to `true`. This will allow us to create new domain events without timelines doubling up, and also will allow us to find all notes which don't have a domain event when we're backfilling.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ApplicationTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ApplicationTest.kt
@@ -2797,6 +2797,9 @@ class ApplicationTest : IntegrationTestBase() {
           )
         }
 
+        // Create a clarification note that already has a domain event
+        createAssessmentClarificationNote(assessment, user, 3, true)
+
         val allSummaries = summaries + informationRequestSummaries
         val expectedJson = objectMapper.writeValueAsString(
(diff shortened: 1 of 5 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ApplicationTest.kt
```diff
@@ -2797,6 +2797,9 @@ class ApplicationTest : IntegrationTestBase() {
           )
         }
 
+        // Create a clarification note that already has a domain event
+        createAssessmentClarificationNote(assessment, user, 3, true)
```
davidatkinsuk, line 2801: if we used named args is the comment still required?

davidatkinsuk (approved):
LGTM 👍

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ApplicationTest.kt
pezholio: I don't think it's worth worrying too much - this code will be deleted soon :)

PR #1797, opened 2024-04-24
CAS2 Gatling tests

tahb:
## Changes

* We start with some refactoring of the old CAS3 tests to allow new CAS2 tests to clearly slot in 
* We add new CAS2 application tests
* These tests only work locally due to problems with the Nomis Roles API. There are three proposals to amend that service to support us [1](https://github.com/ministryofjustice/nomis-user-roles-api/pull/272), [2](https://github.com/ministryofjustice/nomis-user-roles-api/pull/270), [3](https://github.com/ministryofjustice/nomis-user-roles-api/pull/269). If these were merged into main we could be able to try running this on our real dev environment. Until then we are limited to using a patched local instance of Nomis Roles API with these changes applied. Notes will be included at the bottom for how to do that.

## Results

These results are limited to running on our local machines. This is not perfect but our hope is that it will give us an indication into the performance of some of our least efficient pages, highlighting any inefficient SQL queries.

Screenshot 2024-04-24 at 12-25-33 Gatling Stats - Global Information

Screenshot 2024-04-24 at 12-25-37 Gatling Stats - Create CAS2 Application

Screenshot 2024-04-24 at 12-30-00 Gatling Stats - Update Application

Screenshot 2024-04-24 at 12-30-04 Gatling Stats - Submit Application

Screenshot 2024-04-24 at 12-30-08 Gatling Stats - Get Applications

We can see that 3957 applications for the same user were created. The 'Get Applications' endpoint is not paginated and is able to respond with a reassuring 99th percentile of 2358ms. We don't imagine a single referrer or group of referrers belonging to the same prison to have the number for a long time. Pagination is planned in to happen soon too so good news aside from the caveat this isn't run on a real environment.

## Running a patched Nomis Roles API

Why do we have to patch this? There is more detail in the linked pull requests but essentially we get a basic authentication JWT from HMPPS Auth which we use in our downstream requests. Outside the context of Gatling this JWT is expanded with additional data via the callback mechanism. Given we are in the context of Gatling, without a frontend that can receive a callback we cannot replicate it entirely. Nomis Roles API requires data to include a client_id for telemetery and a user_name (where it could look at the `sub`). The same strategy works for CAS3 so there is some hope we could find a way to get those proposals merged.

### Patching to run locally <a href="#user-content-patching" id="patching">#</a>

* Clone [the Nomis Roles API](https://github.com/ministryofjustice/nomis-user-roles-api)
* Apply the changes of each of these changes [1](https://github.com/ministryofjustice/nomis-user-roles-api/pull/272), [2](https://github.com/ministryofjustice/nomis-user-roles-api/pull/270), [3](https://github.com/ministryofjustice/nomis-user-roles-api/pull/269) to your branch 
* Edit your AP Tools docker-compose.yml file to use this version of Nomis Roles API with:
     ```
       nomis-user-roles-api:
       # image: quay.io/hmpps/nomis-user-roles-api:latest
       build:
         context: ../nomis-user-roles-api
         dockerfile: Dockerfile
     ```
* Start AP tools as normal
* Run Gatling tests from the CAS API `./gradlew gatlingRun-uk.gov.justice.digital.hmpps.approvedpremisesapi.gatling.simulations.cas2.ApplyJourneyStressSimulation`

### DIFF
--- a/doc/how-to/run_performance_tests.md
+++ b/doc/how-to/run_performance_tests.md
@@ -1,12 +1,15 @@
 # How to run performance tests
+
 Performance tests use [Gatling](https://gatling.io), which provides a Gradle plugin.
 This bundles up Gatling in a convenient way so that no additional tools need to be downloaded.
 
 ## Running against a local instance
+
 This requires the [ap-tools](https://github.com/ministryofjustice/hmpps-approved-premises-tools)
 utility to be running.
 
 Once the API has started locally, just run:
+
 ```shell
 ./gradlew gatlingRun
 ```
@@ -18,7 +21,9 @@ To run a specific test define the FQN of the simulation:
 ```
 
 ## Running against a remote environment
+
 This requires some configuration, which can be done through environment variables:
+
 ```shell
 GATLING_BASE_URL="https://approved-premises-api-{env}.hmpps.service.justice.gov.uk" \
 GATLING_CAS3_USERNAME="{your username}" \
@@ -28,5 +33,6 @@ GATLING_HMPPS_AUTH_BASE_URL="https://sign-in-{env}.hmpps.service.justice.gov.uk"
 ```
 
 ## Results
+
 Once Gatling has finished, it will output reports into the `build/reports/gatling` folder.
 It will also provide a link in the terminal to open the report directly in the browser.
--- a/doc/how-to/run_performance_tests.md
+++ b/doc/how-to/run_performance_tests.md
@@ -1,11 +1,13 @@
 # How to run performance tests
 
-Performance tests use [Gatling](https://gatling.io), which provides a Gradle plugin.
-This bundles up Gatling in a convenient way so that no additional tools need to be downloaded.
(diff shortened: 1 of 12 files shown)


tahb:
@Harriethw gave this a review locally and [verified](https://mojdt.slack.com/archives/C05GEPLBRN3/p1716461253538099) that this and the steps get this working locally, work. This is a good thing should we want to run this locally again.

However, I've changed my mind about the proposals we'd need to make to the `nomis-user-roles-api` to get this working locally.

I don't think asking that project to tinker with auth in order to support our hack to enable Gatling is justified.

Without those dependent changes this proposal only works with the those manually patched dependencies and changes to the local set up, which is very inaccessible. 

I'm not convinced we should merge this in. Instead I think we ought to explore end-to-end load tests (which include the frontend) via something like https://k6.io/. This pull request remains open as a reference should the new team want to take this on.

[In addition to this we have failed to make Gatling work on CI for CAS3.](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1237)

PR #1798, opened 2024-04-24
[APS-703] Allow filtering and sorting applications by release type

danlivings-dxw:
> See [APS-703 on the Approved Premises Jira board](https://dsdmoj.atlassian.net/browse/APS-703).

This PR updates the `GET /applications/all` endpoint to allow for sorting and filtering by release type:
- The `sortBy` query parameter now accepts `releaseType` as a value. Both ascending and descending sorting is supported.
- An optional `releaseType` query parameter has been introduced. If provided, it should have a value corresponding to a variant of the `ReleaseTypeOption` enum, and will restrict the list of returned results to only those of the given release type.


pezholio (approved):
Nice work 👍

PR #1799, opened 2024-04-24
(APS-673) Add further information requested domain event

pezholio:
This adds a new `APPROVED_PREMISES_ASSESSMENT_INFO_REQUEST` domain event, which is triggered when an assessor creates a clarification note when assessing an application and presents them within the timeline. Currently the timeline already transforms information requests to timeline events, but to rationalise things, we're transferring this to the domain event approach. Once we've backfilled all the Information request clarification notes into domain events, we can remove this logic.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/DomainEventUrlConfig.kt
```diff
@@ -31,6 +31,7 @@ class DomainEventUrlConfig {
       DomainEventType.APPROVED_PREMISES_PLACEMENT_APPLICATION_ALLOCATED -> cas1["placement-application-allocated-event-detail"]
       DomainEventType.APPROVED_PREMISES_MATCH_REQUEST_WITHDRAWN -> cas1["match-request-withdrawn-event-detail"]
       DomainEventType.APPROVED_PREMISES_REQUEST_FOR_PLACEMENT_CREATED -> cas1["request-for-placement-created-event-detail"]
+      DomainEventType.APPROVED_PREMISES_ASSESSMENT_INFO_REQUEST -> cas1["assessment-info-request-detail"]
```
edavey, line 34: I think the form we're using for our domain events is to put the action in the past tense (e.g. `application.submitted`, `booking.made`) so I think we probably ought to be sticking with that form for these new domain events throughout? :

- `assessment.info-requested`
- `assessment.info-received`

does it sound sensible to stay consistent?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/domainevents/DomainEventDescriber.kt
```diff
@@ -42,10 +45,25 @@ class DomainEventDescriber(
...
+  private fun buildInfoRequestDescription(domainEventSummary: DomainEventSummary): String? {
+    val event = domainEventService.getFurtherInformationRequestMadeEvent(domainEventSummary.id())
+
+    return event.describe { data ->
+      val assessmentClarificationNote = assessmentClarificationNoteRepository.findByIdOrNull(data.eventDetails.requestId)
+        ?: throw RuntimeException("Clarification note with ID ${data.eventDetails.requestId} not found")
+
+      """
+        A further information request was made to the applicant:
+        "${assessmentClarificationNote.query}"
```
edavey, line 62: so we're planning on keeping the details of the request for further info out of the domain event, but retrieving it when we build the timeline. This seems like a good solution!

edavey (approved):
This looks great!

I just noted a couple of places where the domain event was referred to as `info-request` rather than `info-requested`...

PR #1800, opened 2024-04-25
APS-712 Add domain_events.noms_number to support replay

davidatkinsuk:
To replay a domain event (i.e. adding the necessary information in an SNS message) we need its corresponding noms number, if defined.

Currently the noms number is only stored in the domain event JSON, which is difficult to extract programmatically. This is because we don’t have a common hierarchy for the ‘domain event envelope’ classes as they’re generated from open api specs. This means we’d need ‘type specific’ code to extract the noms number for each domain event type which would be cumbersome.

For that reason, this commit introduces a nullable ‘first class property’ for the noms number on the DomainEventEntity, populating wherever there is a noms number relevant to the domain event for CAS1, 2 and 3

Evidence from local testing showing noms number being populated, and assessment id/booking id unaffected by refactoring:

Screenshot 2024-04-25 at 11 06 12

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/DomainEventService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/DomainEventService.kt
@@ -76,7 +76,6 @@ class DomainEventService(
     saveAndEmit(
       domainEvent = domainEvent,
       eventType = DomainEventType.APPROVED_PREMISES_APPLICATION_SUBMITTED,
-      nomsNumber = domainEvent.data.eventDetails.personReference.noms,
     )
 
   @Transactional
@@ -84,7 +83,6 @@ class DomainEventService(
     saveAndEmit(
       domainEvent = domainEvent,
       eventType = DomainEventType.APPROVED_PREMISES_APPLICATION_ASSESSED,
-      nomsNumber = domainEvent.data.eventDetails.personReference.noms,
     )
 
   @Transactional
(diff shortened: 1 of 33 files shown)


Harriethw (approved):
makes sense to me, thanks for clear structure and documentation. if others are happy 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/DomainEventService.kt
```diff
@@ -127,31 +121,27 @@ class DomainEventService(
     saveAndEmit(
       domainEvent = domainEvent,
       eventType = DomainEventType.APPROVED_PREMISES_BOOKING_NOT_MADE,
-      nomsNumber = domainEvent.data.eventDetails.personReference.noms,
```
pezholio, line 141: I do wonder now if we can remove the event type specific calls now and just delegate to `saveAndEmit` and work out the eventType from there?

pezholio (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/DomainEventService.kt
davidatkinsuk: I guess using a reified generic somehow? sounds like a good idea

davidatkinsuk: actually probably easiest to just add type into DomainEvent :) Yes, then can remove a bunch of functions.

PR #1805, opened 2024-04-25
(APS-643) Set information requested status correctly

pezholio:
We’ve noticed that assessments with a status of `REQUESTED_FURTHER_INFORMATION` were not showing correctly. After some investigation I discovered that the status was getting set, but then immediately being set back to `ASSESSMENT_IN_PROGRESS` with a subsequent update, which is obviously not what we want.

I’ve updated the AssessmentListener to immediately short circuit if the application’s status is `REQUESTED_FURTHER_INFORMATION`, as well as updated the tests to give a more realistic scenario.

I’ve also added a simple SQL query that gets all assessments with pending clarification notes, then updates their associated application’s status to `REQUESTED_FURTHER_INFORMATION`. I could have added a seed job, but this seemed like overkill.

### DIFF
--- a/src/main/resources/db/migration/all/20240425143644__update_information_request_statuses.sql
+++ b/src/main/resources/db/migration/all/20240425143644__update_information_request_statuses.sql
@@ -1,0 +1,20 @@
+UPDATE
+    approved_premises_applications
+SET
+    status = 'REQUESTED_FURTHER_INFORMATION'
+WHERE
+        id IN (
+        SELECT
+            application_id
+        from
+            assessments
+        where
+                id IN (
+                SELECT
+                    assessment_id
+                from
+                    assessment_clarification_notes
+                where
+                    response is null
+            )
+    );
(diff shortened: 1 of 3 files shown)


src/main/resources/db/migration/all/20240425143644__update_information_request_statuses.sql
```diff
@@ -0,0 +1,20 @@
...
+        where
+                id IN (
+                SELECT
+                    assessment_id
+                from
+                    assessment_clarification_notes
+                where
+                    response is null
+            )
+    );
```
davidatkinsuk, line 20: to help my understanding, i think this query is the equivalent of this:

```
UPDATE
    approved_premises_applications
SET
    status = 'REQUESTED_FURTHER_INFORMATION'
WHERE
        id IN (
              select application_id from assessments assess
              inner join assessment_clarification_notes acn on acn.assessment_id = assess.id
              where acn.response is null
        )
);
```

davidatkinsuk (approved):
assuming this is well tested i think this is good enough to fix the issue for now.

i think we've talked about this before, but we should really move status change logic into something more deterministic instead of listeners, maybe triggered by spring events (or even domain events wrapped in spring events?)

src/main/resources/db/migration/all/20240425143644__update_information_request_statuses.sql
pezholio: Yeah, that's right 👍

pezholio:
Yeah, listeners definitely seem like the thing to do here

PR #1806, opened 2024-04-25
Catch RuntimeExceptions in InformationReceived job

pezholio:
We hit a RuntimeException when running this in preprod, because a user doesn’t exist. We want to cleanly recover from this and continue.

I’m hoping that this issue won’t come up in prod, but I want to run this again in preprod and see how many errors we get. If there are only a handful, I think this will be safe to run in prod. If we get any errors in prod, we can note down which ones have come up and decide how to proceed from there.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/migration/Cas1InformationReceivedMigrationJob.kt
```diff
@@ -35,6 +36,9 @@ class Cas1InformationReceivedMigrationJob(
             clarificationNoteRepository.updateHasDomainEvent(clarificationNote.id)
           } catch (e: IllegalStateException) {
             migrationLogger.error(e.message ?: "An unknown error occurred")
+          } catch (e: RuntimeException) {
+            val errorMessage = e.message ?: "An unknown error occurred"
+            migrationLogger.error("Can't create domain event for ${clarificationNote.id} - $errorMessage")
```
davidatkinsuk, line 41: can you pass the exception into the logger please a second arg, then we'll get the stack trace

davidatkinsuk (approved):
one thing ...

PR #1807, opened 2024-04-25
CAS2-423: give LICENCE_CA create/update permissions

Harriethw:
JIRA: https://dsdmoj.atlassian.net/jira/software/c/projects/CAS2/boards/1347/backlog?selectedIssue=CAS2-423

We are adding a new user role to CAS2 - LICENCE_CA. The current thinking is that this is
the more correct role for our user base than POM, and so we will replace POM with LICENCE_CA in a phased approach. 

Previously we allowed LICENCE_CA to read applications from their prison, now we are giving them all the same permissions as POM.
Since our only role-based logic is in `OAuth2ResourceServerSecurityConfiguration` we
just need to make the change there, and then check in integration tests that everything is working.

https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/4450091020/User+roles+in+CAS-2


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/OAuth2ResourceServerSecurityConfiguration.kt
```diff
@@ -59,15 +59,14 @@ class OAuth2ResourceServerSecurityConfiguration {
         authorize(HttpMethod.DELETE, "/internal/room/*", permitAll)
         authorize(HttpMethod.GET, "/events/cas2/**", hasAuthority("ROLE_CAS2_EVENTS"))
         authorize(HttpMethod.GET, "/events/**", hasAuthority("ROLE_APPROVED_PREMISES_EVENTS"))
-        authorize(HttpMethod.GET, "/cas2/applications/**", hasAnyRole("POM", "LICENCE_CA"))
         authorize(HttpMethod.PUT, "/cas2/assessments/**", hasRole("CAS2_ASSESSOR"))
         authorize(HttpMethod.GET, "/cas2/assessments/**", hasAnyRole("CAS2_ASSESSOR", "CAS2_ADMIN"))
         authorize(HttpMethod.POST, "/cas2/submissions/*/notes", hasAnyRole("POM", "CAS2_ASSESSOR"))
         authorize(HttpMethod.GET, "/cas2/submissions/**", hasAnyRole("CAS2_ASSESSOR", "CAS2_ADMIN"))
         authorize(HttpMethod.POST, "/cas2/submissions/*/status-updates", hasRole("CAS2_ASSESSOR"))
         authorize(HttpMethod.GET, "/cas2/reference-data/**", hasAnyRole("CAS2_ASSESSOR", "POM"))
```
tahb, line 67: Similar thing here, should we be adding the new role in here too?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/OAuth2ResourceServerSecurityConfiguration.kt
```diff
@@ -59,15 +59,14 @@ class OAuth2ResourceServerSecurityConfiguration {
         authorize(HttpMethod.DELETE, "/internal/room/*", permitAll)
         authorize(HttpMethod.GET, "/events/cas2/**", hasAuthority("ROLE_CAS2_EVENTS"))
         authorize(HttpMethod.GET, "/events/**", hasAuthority("ROLE_APPROVED_PREMISES_EVENTS"))
-        authorize(HttpMethod.GET, "/cas2/applications/**", hasAnyRole("POM", "LICENCE_CA"))
         authorize(HttpMethod.PUT, "/cas2/assessments/**", hasRole("CAS2_ASSESSOR"))
         authorize(HttpMethod.GET, "/cas2/assessments/**", hasAnyRole("CAS2_ASSESSOR", "CAS2_ADMIN"))
         authorize(HttpMethod.POST, "/cas2/submissions/*/notes", hasAnyRole("POM", "CAS2_ASSESSOR"))
```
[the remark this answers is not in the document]
Harriethw: ah yep good point for both this and reference data

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/OAuth2ResourceServerSecurityConfiguration.kt
Harriethw: hmm actually I don't think the referrers should have access to this endpoint currently - it just returns the possible application status updates, which referrers are not allowed to add to an application

tahb (approved):
LGTM 🚀

PR #1808, opened 2024-04-25
Fix ignored tests

pezholio:
We missed a `@Nested` annotation from the inner class, so these tests weren't running


davidatkinsuk (approved):
LGTM

PR #1810, opened 2024-04-25
Catch Kotlin RuntimeException

pezholio:
The RuntimeException that gets thrown when we can’t get things upstream is the default Kotlin `RuntimeException`, not the imported Java one


davidatkinsuk (approved):
LGMT 👍

PR #1812, opened 2024-04-26
Rename ignoreLao to ignoreLaoRestrictions

davidatkinsuk:
In several places we have an ignoreLao boolean parameter in the code. The intention of this is to indicate that restrictions on viewing LAO offenders can be ignored (otherwise an unathorised access problem will be returned). I often misread this as ‘don’t return offenders with LAO’, which is incorrect. I’ve renamed this variable for clarity


edavey (approved):
thank you, "literate programming" FTW

PR #1814, opened 2024-04-26
APS-723 - Add logging when staff member not found for qcode

davidatkinsuk:
in the context of retreiving a staff member ‘not found’ could refer to the qCode not being found, or the staff code not existing in qCode result. This commit adds a warning message to clarify when the later issue occurs


edavey (approved):
thanks for adding this clarifying distinction

PR #1817, opened 2024-04-26
APS-724 Increase web client buffer size

davidatkinsuk:
We have an issue in production retreiving alerts from the prison API where the response exceeds our limit of 750000. This commit doubles that limit to 1.5MB

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/WebClientConfiguration.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/WebClientConfiguration.kt
@@ -1,6 +1,7 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.config
 
 import io.netty.channel.ChannelOption
+import org.slf4j.LoggerFactory
 import org.springframework.beans.factory.annotation.Value
 import org.springframework.context.annotation.Bean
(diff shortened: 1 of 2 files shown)


tahb:
[We have this rather old PR which looked at streaming all requests to get around this](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1125). It felt overkill to do the overhead on all requests (rather than just the troublesome ones) at the time but perhaps the need for this change makes it more worthwhile.

tahb (approved):
LGTM. 

Presumably there is something very unusual about this request we're only now seeing in prod. I wonder if we can flag it with the relevant team in case this is unexpected?

See comment re previous alternative approach to doubling which might be more sustainable 👍

davidatkinsuk:
> LGTM.
> 
> Presumably there is something very unusual about this request we're only now seeing in prod. I wonder if we can flag it with the relevant team in case this is unexpected?
> 
> See comment re previous alternative approach to doubling which might be more sustainable 👍

My hunch is that this particular offender has slightly more alerts than the buffer is allowing, but it's hard to tell with the error we're getting, and the response doesn't have a response size header (or one isn't shown in app insights).

There's an urgency to get this resolved but appreciate constantly increasing the buffer isn't really sustainable. I've changed the config to only apply to prisons API so we're not 'stuck' with this config across all APIs

PR #1821, opened 2024-04-29
(APS-728) Allow a booking to be viewed if key worker no longer at premises

pezholio:
If we can't find a keyworker for a specific booking, we want to fail silently rather than blowing up. As this is still somewhat of an issue, we still want to alert Sentry. Going forward, we'll want to handle when keyworkers leave a premises as well as be able to handle transfers etc.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
@@ -81,6 +81,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.ApprovedPremisesAp
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.DomainEvent
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.PersonInfoResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.validated
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.ForbiddenProblem
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.InternalServerErrorProblem
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.results.AuthorisableActionResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.results.CasResult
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/StaffMemberService.kt
```diff
@@ -26,13 +26,17 @@ class StaffMemberService(private val apDeliusContextApiClient: ApDeliusContextAp
       AuthorisableActionResult.Success(staffMember)
     } else {
       log.warn("Whilst a result was returning for qCode $qCode, no staff member with code $code was found in it")
-      AuthorisableActionResult.NotFound()
+      AuthorisableActionResult.NotFound("Staff Code", code)
```
davidatkinsuk, line 29: love it, much better than my 'warn' solution

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
```diff
@@ -165,13 +166,20 @@ class BookingService(
       // Bookings will need to be specialised in a similar way to Premises so that TA Bookings do not have a keyWorkerStaffCode field
       check(premises is ApprovedPremisesEntity) { "Booking has a Key Worker specified but Premises is not an ApprovedPremises" }
 
-      val staffMemberResult = staffMemberService.getStaffMemberByCode(keyWorkerStaffCode, premises.qCode)
-
-      if (staffMemberResult !is AuthorisableActionResult.Success) {
-        throw InternalServerErrorProblem("Unable to get Key Worker via Staff Code: $keyWorkerStaffCode / Q Code: ${premises.qCode}")
+      when (val staffMemberResult = staffMemberService.getStaffMemberByCode(keyWorkerStaffCode, premises.qCode)) {
```
davidatkinsuk, line 169: i think we may need to do a similar thing for PremisesController#premisesPremisesIdBookingsGet? although i'm not sure if that endpoint is used?

davidatkinsuk (approved):
Looking great, just need to check the other endpoint mentioned in my comments

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
pezholio: Nah, we don't use that anymore, as we have the premisesSummary stuff to return a cleaner summary of bookings and premises info

PR #1822, opened 2024-04-29
(APS-673) Tidy up info request domain work

pezholio:
Now the work has been done to populate info request domain events, we can removed the `hasDomainEvent` field, and also remove the logic that fetched information requests for an application and transformed these into timeline events.


davidatkinsuk (approved):
LGTM, thanks for taking the time to remove all this 'dead' code 👍

PR #1827, opened 2024-04-30
(APS-748) Remove assessment from further info Timeline event

pezholio:
This is not particularly useful and in some cases can give assessors access to complete an assessment they are not allocated to, so let’s remove it.


davidatkinsuk (approved):
LGTM 👍

PR #1829, opened 2024-04-30
APS-524 Don't filter docs on conviction for CAS1

davidatkinsuk:
We have found that filtering on conviction ID leads to users not being able to access document they’d like to link to the document. We've used a feature flag so we can disable this is if there are any unforeseen side effects

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/DocumentTransformer.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/DocumentTransformer.kt
@@ -4,6 +4,7 @@ import org.slf4j.LoggerFactory
 import org.springframework.stereotype.Component
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Document
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.DocumentLevel
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.community.ConvictionDocuments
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/DocumentTransformer.kt
```diff
@@ -26,23 +26,21 @@ class DocumentTransformer {
       )
     }
 
-    val documentsForConvictionId = groupedDocuments.convictions
-      .firstOrNull { it.convictionId == convictionId.toString() }
+    val documentsForConvictionId = groupedDocuments
+      .convictions
+      .filter { convictionId == null || it.convictionId == convictionId.toString() }
```
edavey, line 31: filtering on `convictionId == null` seems like a confusing mechanism!

Would things be easier if we used two different transformer functions, eg. 

- `DocumentTransformer.toApiFiltered(documents, convictionId)` and
- `DocumentTransformer.toApiUnfiltered(documents)`

then the controller could avoid setting `convictionId` to null -- which doesn't help us understand that we're actually interested in **all** convictions!

edavey (approved):
👍 but I wonder if we should refactor a little to make things simpler rather than more complex?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/DocumentTransformer.kt
davidatkinsuk: Yes, I completely agree, also the variable being called documentsForConvictionId is quite confusing. I'll make the proposed change before merging

PR #1832, opened 2024-05-01
APS-558: Seed AP staff users

edavey:
We have a list of 1,248 AP Staff who are set up correctly in Delius. We're going to seed them.

They don't need any internal roles or qualifications. Some of the AP staff being seeding will already 
be present in the service and some of those will have existing Roles and Qualifications.

We need to be confident that we aren't clearing those Roles and Qualifications.

In fact this is why we are introducing the new `ApStaffUsersSeedJob` rather than
using the `UsersSeedJob` -- which does clear those roles and qualifications.


davidatkinsuk (approved):
LGTM 

Do you think it's worth adding Javadoc to the existing UsersSeedJob to make it clear that existing roles/qualifications will be 'reset' and if that's not desired the new seed job should be used instead?

PR #1834, opened 2024-05-01
APS-753 add isDirectlyWithdrawable flag to request for placements

davidatkinsuk:
Currently the code in the Cas1WithdrawalService.isDirectlyWithdrawable function retrieves the complete set of WithdrawableEntities for the whole application, despite only being concerned with WithdrawableEntities for the specific placement request/placement application. This will become a performance issue if/when we start calling external APIs to determine if an entity is withdrawable (as is planned). Before any such change is made we should modify this methods to test against a subset of the tree (i.e. where the root node is placement application/placement request)

### DIFF
--- a/src/main/resources/static/_shared.yml
+++ b/src/main/resources/static/_shared.yml
@@ -3842,6 +3842,11 @@ components:
           format: date-time
         document:
           $ref: '#/components/schemas/AnyValue'
+        canBeDirectlyWithdrawn:
+          description: |
+            If true, the user making this request can withdraw this request for placement. 
+            If false, it may still be possible to indirectly withdraw this request for placement by withdrawing the application.
+          type: boolean
         isWithdrawn:
           type: boolean
         withdrawalReason:
@@ -3858,6 +3863,7 @@ components:
         - id
         - createdByUserId
         - createdAt
+        - canBeDirectlyWithdrawn
         - isWithdrawn
         - type
         - placementDates
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/RequestsForPlacementTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/RequestsForPlacementTest.kt
@@ -82,8 +82,16 @@ class RequestsForPlacementTest : IntegrationTestBase() {
             .json(
               objectMapper.writeValueAsString(
                 listOf(
-                  requestForPlacementTransformer.transformPlacementApplicationEntityToApi(placementApplication),
-                  requestForPlacementTransformer.transformPlacementRequestEntityToApi(placementRequest),
+                  // user created the placement application, but it's not submitted, so they can't withdraw
+                  requestForPlacementTransformer.transformPlacementApplicationEntityToApi(
+                    placementApplication,
+                    canBeDirectlyWithdrawn = false,
+                  ),
+                  // user created associated application, so they can withdraw the placement request
+                  requestForPlacementTransformer.transformPlacementRequestEntityToApi(
+                    placementRequest,
+                    canBeDirectlyWithdrawn = true,
+                  ),
                 ),
               ),
             )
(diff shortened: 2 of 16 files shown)


src/main/resources/static/_shared.yml
```diff
@@ -3842,6 +3842,11 @@ components:
           format: date-time
         document:
           $ref: '#/components/schemas/AnyValue'
+        canBeDirectlyWithdrawn:
+          description: |
+            If true, the user making this request can withdraw this request for placement. 
+            If false, it may still be possible to indirectly withdraw this request for placement by withdrawing the application.
```
edavey, line 3848: Can you remind me (and maybe include in the commit msg) in which scenario a RfP is not individually withdraw-able but can be withdrawn as part of a withdrawal cascade?

-> I think I found the answer in the integration test:

> user created the placement application, but it's not submitted, so they can't withdraw

I've left another comment about whether this logic should be more encapsulated and thereby perhaps easier to find?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/RequestsForPlacementTest.kt
```diff
@@ -182,7 +190,11 @@ class RequestsForPlacementTest : IntegrationTestBase() {
             .expectBody()
             .json(
               objectMapper.writeValueAsString(
-                requestForPlacementTransformer.transformPlacementRequestEntityToApi(placementRequest),
+                requestForPlacementTransformer.transformPlacementRequestEntityToApi(
+                  placementRequest,
+                  // user created associated application, so they can withdraw the placement request
+                  canBeDirectlyWithdrawn = true,
+                ),
```
edavey, line 197: hmm, maybe I'm missing it, but does it feel like this logic around the withdrawal rules should to be laid out in a unit test rather than in an integration test? i.e. that a class, e.g. `RequestForPlacementWithdrawal(rfp, user): boolean` should have responsibility for making this decision, or is it ok to have in the RfPTransformer?

edavey (approved):
👍 Just a minor question about where the responsibility for handling the logic should live

src/main/resources/static/_shared.yml
davidatkinsuk: Yes, that's a good example. See other comment for notes on encapsulation

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/RequestsForPlacementTest.kt
davidatkinsuk: In this case because we're testing the entire JSON response i was forced to determine whether it was withdrawable or not, and it's not immediately obvious from the test data so I added a comment to explain why this is the case. I think we'll always have this challenge in integration tests where we're returning withdrawable status and test the whole JSON

As for where we test withdrawal logic, it's partially split across entity-specific service unit testing  (e.g. PlacementRequestService.getWithdrawableState), the Cas1WithdrawableTreeBuilderTest , and then WithdrawalTest which provides some high level integration tests and 'fills the gap' of testing where the state of one entity in the tree affects some other entity (e.g. booking blocks a placement request withdrawal)

On completing the PR I decided I didn't like passing the list of WithdrawableEntities into the transformer, especially as i've added a function in the transformer that takes a boolean in instead. I was going to change that in my next PR that calls the delius endpoint, but i think i'll change that now.

davidatkinsuk:
@edavey I've rewritten this. There's now functions on the Cas1WithdrawableService to determine if PA/PRs are withdrawable, which is much nicer coupling. Thanks for the nudge!

`  fun isDirectlyWithdrawable(placementRequest: PlacementRequestEntity, user: UserEntity): Boolean {
    return allDirectlyWithdrawables(
      application = placementRequest.application,
      user = user,
    )
      .any { it.type == WithdrawableEntityType.PlacementRequest && it.id == placementRequest.id }
  }

  fun isDirectlyWithdrawable(placementApplication: PlacementApplicationEntity, user: UserEntity): Boolean {
    return allDirectlyWithdrawables(
      application = placementApplication.application,
      user = user,
    )
      .any { it.type == WithdrawableEntityType.PlacementApplication && it.id == placementApplication.id }
  }`

PR #1835, opened 2024-05-01
[APS-563] Account for afternoon and non-working-day submissions in deadline calculations

danlivings-dxw:
> See [APS-563 on the Approved Premises Jira board](https://dsdmoj.atlassian.net/browse/APS-563).

This PR makes sure that all applications that are submitted after 1pm on a working day or on a non-working day (weekends or bank holidays) have their effective submission date set to 9am on the next working day before performing the deadline calculation. This prevents time from being effectively lost due to out-of-hours submissions.

This working pattern adjustment is timezone-aware: it correctly accounts for both winter (GMT/UTC) time and summer time (BST/UTC+1), as well as the weekends where the clocks change.


davidatkinsuk (approved):
This is really great and it's really nice to see all the timezone considerations in the implementation and tests. LGTM 👍

PR #1837, opened 2024-05-01
Update report date range

marcocolluramoj:
Throw an error if report date range is greater than or equal to the maximum duration in months.


froddd (approved):
LGTM

PR #1839, opened 2024-05-01
CAS2-380 added conditional release date filter

danhumphreys-moj:
Ref: https://dsdmoj.atlassian.net/browse/CAS2-380

Soft delete submitted applications once their CRD date has been reached.

Referrer dashboards only.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ApplicationTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ApplicationTest.kt
@@ -34,6 +34,8 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApplicationRe
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas2ApplicationEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ExternalUserEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.NomisUserEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.randomDateAfter
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.randomDateBefore
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.randomDateTimeBefore
 import java.time.LocalDate
 import java.time.OffsetDateTime
@@ -189,6 +191,7 @@ class Cas2ApplicationTest : IntegrationTestBase() {
               withId(UUID.randomUUID())
             }
 
+            // unsubmitted application
             val firstApplicationEntity = cas2ApplicationEntityFactory.produceAndPersist {
               withApplicationSchema(applicationSchema)
               withCreatedByUser(userEntity)
@@ -198,6 +201,7 @@ class Cas2ApplicationTest : IntegrationTestBase() {
               withHdcEligibilityDate(LocalDate.now().plusMonths(3))
             }
 
+            // submitted application, CRD >= today so should be returned
             val secondApplicationEntity = cas2ApplicationEntityFactory.produceAndPersist {
               withApplicationSchema(applicationSchema)
               withCreatedByUser(userEntity)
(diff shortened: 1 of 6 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ApplicationTest.kt
```diff
@@ -746,10 +867,38 @@ class Cas2ApplicationTest : IntegrationTestBase() {
...
                     withData("{}")
                     withSubmittedAt(OffsetDateTime.now().minusDays(it.toLong()))
+                    withConditionalReleaseDate(LocalDate.now().randomDateAfter())
+                  }.id,
+                )
+              }
+
+              // create 2 x submitted applications for this user
+              // with most recent first and conditional release dates of today
+              repeat(2) {
```
Harriethw, line 877: quite minor but might shave off some milliseconds - do you think we need to create two each time to test that applications with CRD of today are returned?

Harriethw:
I think you might need to update our seeded data so that we don't hide what gets seeded  - we currently hardcode the conditional release to in the past! Here
https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/258cc251a98a3eb851c91f608a262446622e7103/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas2/Cas2AutoScript.kt#L100  and here https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/8eab5722ba3915dceb68885dec26532675fda992/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas2/Cas2ApplicationsSeedJob.kt#L103

Harriethw (approved):
looks good! approving as I think the code is fine, just would be worth updating our seeded data so it isn't hidden.

danhumphreys-moj:
> I think you might need to update our seeded data so that we don't hide what gets seeded - we currently hardcode the conditional release to in the past! Here
> 
> https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/258cc251a98a3eb851c91f608a262446622e7103/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas2/Cas2AutoScript.kt#L100
> 
> and here
> https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/8eab5722ba3915dceb68885dec26532675fda992/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas2/Cas2ApplicationsSeedJob.kt#L103

We seed data in the database in the dev and test environments so that we have data to run the unit tests against right? 

So the change would be to use `LocalDate.now()` in the two places linked?

danhumphreys-moj:
> > I think you might need to update our seeded data so that we don't hide what gets seeded - we currently hardcode the conditional release to in the past! Here
> > https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/258cc251a98a3eb851c91f608a262446622e7103/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas2/Cas2AutoScript.kt#L100
> > 
> > and here
> > https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/8eab5722ba3915dceb68885dec26532675fda992/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas2/Cas2ApplicationsSeedJob.kt#L103
> 
> We seed data in the database in the dev and test environments so that we have data to run the unit tests against right? Just checking my understanding.
> 
> So the change would be to use `LocalDate.now()` in the two places linked?

Quoting answer from @Harriethw:

> data gets seeded on both local and dev/test - the primary reason is to setup the environment for users so we don't have to manually go in and create applications/notes/status updates etc. So much time would be wasted either doing that manually or via the e2es, and every single user would have to do it every time the data gets wipes. It also then creates a basis for what the data should look like.
> 
> There's more info here on the difference between the two ways of seeding [https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/a79ec78ba993[…]815/doc/how-to/seed_cas2_data_in_non_production_environments.md](https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/a79ec78ba9934332ba28ef3e0d2a877509c91815/doc/how-to/seed_cas2_data_in_non_production_environments.md)

PR #1842, opened 2024-05-01
APS-558: Simplify seeding outcome report

edavey:
Tweak to the AP Staff User seeding.

It turns out that 9,224 of our 10,137 users are missing the created_at property, as this was added quite recently. However, every record has an `updated_at` property set.

As it seems to be tricky to set `User.updatedAt` in our test fixtures, I've simplified the way the outcome is reported. We no longer attempt to use the timestamp to derive whether the record was created or previously existed. We now make do with saying when the record was created or updated.


davidatkinsuk (approved):
LGTM 👍

PR #1843, opened 2024-05-02
APS-735 Optimise how we determine if PA/PR withdrawable

davidatkinsuk:
Update the ‘isDirectlyWithdrawable’ functions to only build a withdrawable tree for the entity of interest, instead of the whole application. This is a pre-requisite to adding in calls to delius to determine booking state, ensuring we don’t make any uneccessary calls


edavey (approved):
👍

PR #1845, opened 2024-05-02
Add Bookings@Version to support OptimisticLocking

davidatkinsuk:
Whilst we have logical protections in our code to stop entities being modified in illegal ways (e.g. reallocating an assessment that’s already been allocated to someone else), this doens’t protect us when we receive two requests to modify an entity concurrently. Assuming the second transactions starts before the first finishes, any changes made in the first transaction will be lost when the second transaction completes. This creates many confusing and hard to trace errors.

By using the JPA @Version attribute and adding a versioning column, JPA/Hibernate will automatically detect such situtations and throw an error on transcation commit. Whilst this lead to a bad user experience in these scenarios, it will at least protect the integrity of our entities state, and we can then look at making UI changes to help avoid the errors from happening

A numeric column was chosen instead of timestamp as it’s generally deemed to be the more reliable solution, although a timestamp is incredibly unlikely to be the same when the lock is gained/released.

This commit only modifies the Booking entity. Once we’re happy this doesn’t cause any issues, we’ll look at adding locking to the other aggregate roots


edavey (approved):
This makes good sense, thanks for this initiative.

PR #1846, opened 2024-05-02
APS-735 - Withdrawals improvements

davidatkinsuk:
This commit makes several changes to the Withdrawable Code in preparation for checking if an arrival has been recorded in delius, blocking the withdrawal. Improvements include:

* Improving the WithdrawalTree to capture the reason why a withdrawal is blocked (e.g. arrival recorded in CAS1).
* General refactor or withdrawal tests, making them more readable and removing duplication
* Add a new /applications/${application.id}/withdrawablesWithNotes endpoint that provides some high level notes when application withdrawal is blocked, alongside the list of withdrawable elements
* Add a WithdrawableTree class to better represent a tree's root node, and provide a convenient way to print the tree including notes to the logs

The tree when printed now appears as follows:

```
Application(c2b), withdrawable:Y, mayDirectlyWithdraw:Y, BLOCKED
---> PlacementRequest(20c), withdrawable:Y, mayDirectlyWithdraw:Y, BLOCKED
------> Booking(dc3), withdrawable:N, mayDirectlyWithdraw:Y, BLOCKING - ArrivalRecordedInCas1
---> PlacementApplication(254), withdrawable:Y, mayDirectlyWithdraw:Y
------> PlacementRequest(bbf), withdrawable:Y, mayDirectlyWithdraw:N
---------> Booking(76b), withdrawable:Y, mayDirectlyWithdraw:Y
------> PlacementRequest(0b3), withdrawable:Y, mayDirectlyWithdraw:N
---> PlacementApplication(835), withdrawable:Y, mayDirectlyWithdraw:Y
---> Booking(bde), withdrawable:Y, mayDirectlyWithdraw:Y

Notes: [1 or more placements cannot be withdrawn as they have an arrival]
```


davidatkinsuk:
> just an initial question about terminology: when you say:
> 
> ```
> ------> Booking(dc3), withdrawable:N, mayDirectlyWithdraw:Y, BLOCKING - ArrivalRecordedInCas1
> ```
> 
> do we mean that the arrival was recording in the digital CAS1 service or do we mean that we've detected that arrival was recorded in the legacy Delius AP Diary? If it's the latter, then I think we could use clearer language such as `ArrivalRecordedInDelius`

This is referring to digital CAS1 service (integration with delius is added in a subsequent PR that i'm working on now). I do agree CAS1 is ambiguous in this context. The next PR will introduce an enum for the delius equivalent

How about: ArrivalRecordedInAPI vs ArrivalRecordedInDelius? Any suggestions welcome!

edavey:
> This is referring to digital CAS1 service (integration with delius is added in a subsequent PR that i'm working on now). I do agree CAS1 is ambiguous in this context. The next PR will introduce an enum for the delius equivalent
> 
> How about: ArrivalRecordedInAPI vs ArrivalRecordedInDelius? Any suggestions welcome!

👍 yes I realised (looking properly at this PR) that the new `BlockingReason` enum has only one scenario the "arrival recorded in the new digital service" and that you will add the "arrival recorded in Delius" blocking reason in due course. 

Actually I do think that "CAS1" is the right term for the new digital service. As far as I've seen "CAS" and "Community Accommodation Service" are the latest digital terms so I would favour something like `ArrivalRecordedInCas1` versus `ArrivalRecordedInDelius`...

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/WithdrawalTest.kt
```diff
@@ -1313,9 +1300,12 @@ class WithdrawalTest : IntegrationTestBase() {
     return application
   }
 
+  data class DatePeriod(val start: LocalDate, val duration: Int)
```
edavey, line 1349: could it be confusing that we already have an API class of `DatePeriod` with a different construction `(start_date, end_date)`? Perhaps it's not an issue as this is just a local class for testing? But if you think it could be confusing, then a different class name like `DateSpan` might work?

edavey (approved):
👍 this is going to be a big help in usability (as well as testing and debugging). Thanks for all the improvements along the way.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/WithdrawalTest.kt
```diff
@@ -1469,5 +1404,33 @@ class WithdrawalTest : IntegrationTestBase() {
...
+      WithdrawableType.placementApplication,
+      placementApplication.placementDates.map { toDatePeriod(it.expectedArrival, it.duration) },
+    )
+
+  fun toWithdrawable(booking: BookingEntity) =
+    Withdrawable(
+      booking.id,
+      WithdrawableType.booking,
+      listOf(DatePeriod(booking.arrivalDate, booking.departureDate)),
+    )
```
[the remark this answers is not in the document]
davidatkinsuk: These functions are is really just for the tests only because i didn't want to depend on the WithdrawableTransformer, which is how we do this transform in the 'actual' code.

There is a related question as to whether we should introduce an interface to each Entities service for the 'getWithdrawableState' function, which they all implement. Whilst i think we could do that to make this common structure clear in code, because of the nature of how Cas1WithdrawableTreeBuilder uses these functions i don't think there's any syntactic advantage to doing this

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/WithdrawalTest.kt
davidatkinsuk: I agree, it can be confusing. I'll change it DateSpan in my next PR and also make it private

PR #1847, opened 2024-05-02
APS-765: Add a missing 'national team to AP Area' mapping

edavey:
We map this to the North East AP Area, as per
the other `N41*` team codes.


davidatkinsuk (approved):
LGTM 👍

PR #1434, opened 2024-02-01
CAS1: Improve seeding in local/dev/test environments

edavey:
This PR introduces the 'AutoScript' seeding which is in use in CAS2. On each boot of the Spring server, the `ApprovedPremisesAutoScript` utility runs, to create realistic data entities. At present it creates applications in the following states:

- **in progress**: unsubmitted
- **submitted**: with an `Assessment` but not assessed
- **info_required**: with an associated `AssessmentClarificationNote`
- **accepted**: with a `PlacementRequest`
- **applied_for_placement**: with a `PlacementApplication` (parole) but no decision recorded

More planned!

NOTE: before running localling, clear out existing data, eg with:

```sql

TRUNCATE TABLE applications CASCADE;
TRUNCATE table assessments CASCADE;

```

PR #1849, opened 2024-05-03
APS-768 Seed CAS1 Application using Auto Script

davidatkinsuk:
We previously had some default applications seeded via a ‘local only’ flyway script. Unfortunatley these became stale over time and generally cause issues using the application.

This commit removes those applications and instead bootstraps a single application to the point of submission using an auto script.

Unlike CAS2, we only want the seed job to run locally are it relies on locally managed test data (i.e. CRNs). With improvements this may be usable in other environments. For this reason this commit introduces separate auto script enabled configuration for CAS1 and CAS2


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/Cas1AutoScript.kt
```diff
@@ -0,0 +1,98 @@
...
+
```
edavey, line 29: Did you consider holding  the `data` and `document` JSON structures separately? 

This is what we were doing in CAS2 with separate files for the application data and application document blobs.  When there were changes to the questionnaire it was then quite easy to compare the seed fixture in the API with the UI's integration tests fixture (e.g. [applicationData.json](https://github.com/ministryofjustice/hmpps-community-accommodation-tier-2-ui/blob/b191ff80659ca31a79e9d62a87ea00868aa311c7/integration_tests/fixtures/applicationData.json))

e.g. https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1318

It might make for a more readable and maintainable approach?

Here's [CAS1 work in progress](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1434/files) in which I suggest separate fixture files for:

- application data
- application document
- assessment data
- assessment document
- placement application data
- placement application document

edavey (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/Cas1AutoScript.kt
davidatkinsuk: That's a great idea, i'll do that before merging. Currently this only pushes the application to pre-submission so we'll only need one blob for now.

PR #1850, opened 2024-05-03
CAS-372 Create Api endpoint for types of prison release

muhammad-elabdulla:
This [PR CAS-372](https://dsdmoj.atlassian.net/browse/CAS-372) is to add new endpoint to referral data for prison release types. The type for temporary accommodation will be:

- Conditional release date (CRD) licence
- End of custody supervised licence (ECSL)
- Licence, following fixed-term recall
- Licence, following standard recall
- Parole
- Post sentence supervision (PSS)

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ReferenceDataController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ReferenceDataController.kt
@@ -13,6 +13,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.LocalAuthority
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.LostBedReason
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.MoveOnCategory
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.NonArrivalReason
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.PrisonReleaseType
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ProbationDeliveryUnit
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ProbationRegion
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ReferralRejectionReason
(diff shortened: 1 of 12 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ReferenceDataController.kt
```diff
@@ -189,4 +194,15 @@ class ReferenceDataController(
 
     return ResponseEntity.ok(referralRejectionReasons.map(referralRejectionReasonTransformer::transformJpaToApi))
   }
+
+  override fun referenceDataPrisonReleaseTypesGet(
+    xServiceName: ServiceName?,
+  ): ResponseEntity<List<PrisonReleaseType>> {
+    val prisonReleaseTypes = when (xServiceName != null) {
+      true -> prisonReleaseTypeRepository.findAllByServiceScope(xServiceName.value)
+      else -> prisonReleaseTypeRepository.findAll(Sort.by(Sort.Direction.ASC, "sortOrder"))
```
davidatkinsuk, line 203: I think we should deliberately only support CAS3 if we're not sure if this endpoint will be used by other CASs in the future e.g.

```
    if (xServiceName != ServiceName.temporaryAccommodation) {
      throw ForbiddenProblem()
    }
```

If we do this CAS3 can happily change this endpoint without having to worry about whether CAS1/CAS2 is using it, which can sometimes be hard to determine without looking at UI code. I guess we can also remove the default findAll function too if we do this?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PrisonReleaseTypeEntity.kt
```diff
@@ -0,0 +1,34 @@
...
+import javax.persistence.Id
+import javax.persistence.Table
+
+@Repository
+interface PrisonReleaseTypeRepository : JpaRepository<PrisonReleaseTypeEntity, UUID> {
+  @Query("SELECT m FROM PrisonReleaseTypeEntity m WHERE m.serviceScope = :serviceName OR m.serviceScope = '*' ORDER BY m.sortOrder")
+  fun findAllByServiceScope(serviceName: String): List<PrisonReleaseTypeEntity>
+
+  @Query("SELECT m FROM PrisonReleaseTypeEntity m WHERE m.serviceScope = :serviceName OR m.serviceScope = '*' AND m.isActive = true ORDER BY m.sortOrder")
+  fun findActiveByServiceScope(serviceName: String): List<PrisonReleaseTypeEntity>
```
davidatkinsuk, line 17: I don't think this function is used?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PrisonReleaseTypeEntity.kt
```diff
@@ -0,0 +1,34 @@
...
+
+  @Query("SELECT m FROM PrisonReleaseTypeEntity m WHERE m.isActive = true ORDER BY m.sortOrder")
+  fun findActive(): List<PrisonReleaseTypeEntity>
```
davidatkinsuk, line 20: I don't think this function is used?

davidatkinsuk (approved):
Looks good, just a few suggestions from me on how we could maybe simplify?

PR #1851, opened 2024-05-03
APS-735 Check Placement Arrival Status in Delius on Withdrawal

davidatkinsuk:
When building the Cas1 Withdrawable Tree for a placement we now check the status of the placement in delius wrt. arrivals. If the placement has been marked as arrived in delius we ‘block’ the withdrawal as we would if the placement was marked as arrived in CAS1. We also include a specific note in the withdrawal tree indicating that 1 or placements have arrivals marked in delius and as such can’t be withdrawn.

Use of this functionality is controlled by the “cas1-check-delius-withdrawal-status” feature flag, which will return false by default

There is no caching on the delius referrals endpoint that we use to retrieve this data. Caching wouldn’t make sense in this context because we always want the latest state wrt. arrivals. If we find that the endpoint is being invoked too often we could look at either

1) caching a response that indicates there is an arrival (assuming this is easy with redis integration)
2) write the delius arrival on the booking entity itself (or via a new table)

Both of these approaches assume that once marked as a arrived in delius this can’t be undone (e.g. if the person was marked as arrived by accident)

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/BookingServiceTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/BookingServiceTest.kt
@@ -126,7 +126,9 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.ApplicationServi
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.AssessmentService
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.BookingService
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.CruService
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.DeliusService
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.DomainEventService
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.FeatureFlagService
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.GetBookingForPremisesResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.OffenderService
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.PlacementRequestService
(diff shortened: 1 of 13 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/BookingServiceTest.kt
```diff
@@ -6868,6 +6874,7 @@ class BookingServiceTest {
       )
 
       every { mockUserAccessService.userMayCancelBooking(user, booking) } returns true
+      every { mockFeatureFlagService.getBooleanFlag("cas1-check-delius-withdrawal-status", false) } returns true
```
edavey, line 6877: a tiny details, but did you name this `cas1-check-delius-withdrawal-status` because you anticipate the potential need for other Delius tests which affect withdraw-ability? If not, would `cas1-check-delius-arrival-status` be more descriptive?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas1/Cas1WithdrawableTreeBuilderTest.kt
```diff
@@ -176,22 +177,125 @@ Notes: []
...
+    val result = service.treeForApp(application, user)
+
+    assertThat(result.render(includeIds = false).trim()).isEqualTo(
+      """
+Application(), withdrawable:Y, mayDirectlyWithdraw:Y, BLOCKED
+---> PlacementApplication(), withdrawable:N, mayDirectlyWithdraw:N, BLOCKED
+------> PlacementRequest(), withdrawable:N, mayDirectlyWithdraw:N, BLOCKED
+---------> Booking(), withdrawable:Y, mayDirectlyWithdraw:N, BLOCKING - ArrivalRecordedInDelius
+
+Notes: [1 or more placements cannot be withdrawn as they have an arrival recorded in Delius]
```
edavey, line 293: 👍

edavey (approved):
very clear, this should reduce our support issues around withdrawals and the AP Diary. I note your question about whether Arrivals in AP Diary can be deleted/removed/undone...

PR #1852, opened 2024-05-07
CAS-369 Remove non PDU values from the PDU list

muhammad-elabdulla:
This [PR CAS-369](https://dsdmoj.atlassian.net/browse/CAS-369) is to remove non PDU values from the PDU list in temporary accommodation


marcocolluramoj (approved):
LGTM

PR #1856, opened 2024-05-07
CAS1 E2Es: Use NPX playwright install over docker image

richpjames:
If we use the docker image we have to manually ensure it is kept up to date with the version in the CAS1 UI repo's playwright/test version but if we use NPX then this will be automated


davidatkinsuk (approved):
LGTM 👍

PR #1857, opened 2024-05-07
CAS1 E2ES: Reinstate Playwright docker image

richpjames:
https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1856
attempted to remove the dependency on the Playwright Docker image but
ultimately failed. I couldn't workout how to fully install all the
dependencies needed without using the docker image. This means we will
have to avoid updating the version of Playwright/test in the UI repo
unless we also update the docker image here


davidatkinsuk (approved):
lgtm 👍

PR #1859, opened 2024-05-08
APS-774 filter out docs with null name

davidatkinsuk:
We have observed rare cases where the community API has returned a document without a name specified.

This commit filters out such documents as we cannot automatically generate a filename without knowing the file extension/type, which is not provided.


edavey (approved):
This looks like the right way to handle what seems to be a very rare and invalid element

PR #1860, opened 2024-05-08
CAS2-437 move SQL logic to database VIEW

danhumphreys-moj:
Ref: https://dsdmoj.atlassian.net/browse/CAS2-437

TLDR: this PR removes some of the complex and common SQL logic out of the application layer into a database VIEW which can then be used with [JPA query methods](https://docs.spring.io/spring-data/jpa/reference/jpa/query-methods.html).

The /cas2/applications endpoint is used by the frontend to provision the referrer dashboards.

Recently those dashboards have been extended with new logic such as fetching results by page, prison code and submitted/not submitted. Our approach was to add new query methods with mostly duplicate SQL each time. We knew this wasn’t going to be too sustainable but with national roll out planned we pushed through and compounded the problem.

[We now want to add in another piece of logic to automatically filter out old applications from one of those pages.](https://dsdmoj.atlassian.net/browse/CAS2-380?atlOrigin=eyJpIjoiMDc3NWQ2NjlkMmU5NGMxZmI0MzQ3NTk5NTliZTk3NmYiLCJwIjoiaiJ9) 

At this point we want to pause adding new features and refactor how this data is queried. This should make the codebase easier to read, harder for bugs to appear and easier to extend with new features.

Database views and JPA query methods have been investigated and deemed an alternative way to read sets of applications in differing states.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas2ApplicationEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas2ApplicationEntity.kt
@@ -2,14 +2,11 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity
 
 import org.hibernate.annotations.OrderBy
 import org.hibernate.annotations.Type
-import org.springframework.data.domain.Page
-import org.springframework.data.domain.Pageable
 import org.springframework.data.domain.Slice
 import org.springframework.data.jpa.repository.JpaRepository
 import org.springframework.data.jpa.repository.Lock
 import org.springframework.data.jpa.repository.Query
 import org.springframework.stereotype.Repository
-import java.sql.Timestamp
 import java.time.LocalDate
 import java.time.OffsetDateTime
 import java.util.UUID
@@ -22,198 +19,9 @@ import javax.persistence.OneToMany
 import javax.persistence.OneToOne
 import javax.persistence.Table
 
-const val SUBMITTED_APPLICATION_SUMMARY_FIELDS =
-  """
-    ,
-    a.hdc_eligibility_date as hdcEligibilityDate,
-    asu.label as latestStatusUpdateLabel,
-    CAST(asu.status_id AS TEXT) as latestStatusUpdateStatusId
-  """
-
-const val DEFAULT_APPLICATION_SUMMARY_FIELDS =
-  """
-    CAST(a.id AS TEXT) as id,
-    a.crn,
-    a.noms_number as nomsNumber,
-    CAST(a.created_by_user_id AS TEXT) as createdByUserId,
-    nu.name as createdByUserName,
-    a.created_at as createdAt,
-    a.submitted_at as submittedAt
-  """
-
-const val ALL_APPLICATION_SUMMARY_FIELDS =
-  DEFAULT_APPLICATION_SUMMARY_FIELDS + SUBMITTED_APPLICATION_SUMMARY_FIELDS
-
 @Suppress("TooManyFunctions")
 @Repository
 interface Cas2ApplicationRepository : JpaRepository<Cas2ApplicationEntity, UUID> {
-
-  @Query(
-    """
-SELECT 
-    $ALL_APPLICATION_SUMMARY_FIELDS
-FROM cas_2_applications a
-LEFT JOIN
-    (SELECT DISTINCT ON (application_id) su.application_id, 
-      su.label, su.status_id
-    FROM cas_2_status_updates su
-    ORDER BY su.application_id, su.created_at DESC) as asu
-ON a.id = asu.application_id
-JOIN nomis_users nu ON nu.id = a.created_by_user_id
-WHERE a.created_by_user_id = :userId
-AND (a.conditional_release_date IS NULL OR a.conditional_release_date >= current_date)
-ORDER BY createdAt DESC
-""",
-    countQuery =
-    """
-    SELECT COUNT(*)
-      FROM cas_2_applications a
-    WHERE a.created_by_user_id = :userId
-    """,
-    nativeQuery = true,
-  )
-  fun findAllCas2ApplicationSummariesCreatedByUser(userId: UUID, pageable: Pageable?):
-    Page<Cas2ApplicationSummary>
-
-  @Query(
-    """
-SELECT
-    $ALL_APPLICATION_SUMMARY_FIELDS
-FROM cas_2_applications a
-    LEFT JOIN
-        (SELECT DISTINCT ON (application_id) su.application_id, 
-          su.label, su.status_id
-        FROM cas_2_status_updates su
-        ORDER BY su.application_id, su.created_at DESC) as asu
-ON a.id = asu.application_id
-JOIN nomis_users nu ON nu.id = a.created_by_user_id
-WHERE a.referring_prison_code = :prisonCode
-AND (a.conditional_release_date IS NULL OR a.conditional_release_date >= current_date)
-ORDER BY createdAt DESC
-""",
-    countQuery =
-    """
-    SELECT COUNT(*)
-      FROM cas_2_applications a
-    WHERE a.referring_prison_code = :prisonCode
-    """,
-    nativeQuery = true,
-  )
-  fun findAllCas2ApplicationSummariesByPrison(prisonCode: String, pageable: Pageable?):
-    Page<Cas2ApplicationSummary>
-
-  @Query(
-    """
-SELECT
-    $ALL_APPLICATION_SUMMARY_FIELDS
-FROM cas_2_applications a
-LEFT JOIN
-    (SELECT DISTINCT ON (application_id) su.application_id, 
-      su.label, su.status_id
-    FROM cas_2_status_updates su
-    ORDER BY su.application_id, su.created_at DESC) as asu
-ON a.id = asu.application_id
-JOIN nomis_users nu ON nu.id = a.created_by_user_id
-WHERE a.created_by_user_id = :userId
-AND a.submitted_at IS NOT NULL
-AND a.conditional_release_date >= current_date
-ORDER BY createdAt DESC
-""",
-    countQuery =
-    """
-    SELECT COUNT(*)
-      FROM cas_2_applications a
-    WHERE a.created_by_user_id = :userId
-    AND a.submitted_at IS NOT NULL
-    """,
-    nativeQuery = true,
-  )
-  fun findSubmittedCas2ApplicationSummariesCreatedByUser(userId: UUID, pageable: Pageable?):
-    Page<Cas2ApplicationSummary>
-
-  @Query(
-    """
-SELECT
-    $ALL_APPLICATION_SUMMARY_FIELDS
-FROM cas_2_applications a
-LEFT JOIN
-    (SELECT DISTINCT ON (application_id) su.application_id, 
-      su.label, su.status_id
-    FROM cas_2_status_updates su
-    ORDER BY su.application_id, su.created_at DESC) as asu
-ON a.id = asu.application_id
-JOIN nomis_users nu ON nu.id = a.created_by_user_id
-WHERE a.referring_prison_code = :prisonCode
-AND a.submitted_at IS NOT NULL
-AND a.conditional_release_date >= current_date
-ORDER BY createdAt DESC
-""",
-    countQuery =
-    """
-    SELECT COUNT(*)
-      FROM cas_2_applications a
-    WHERE a.referring_prison_code = :prisonCode
-    AND a.submitted_at IS NOT NULL
-    """,
-    nativeQuery = true,
-  )
-  fun findSubmittedCas2ApplicationSummariesByPrison(prisonCode: String, pageable: Pageable?):
-    Page<Cas2ApplicationSummary>
-
-  @Query(
-    """
-SELECT
-    $DEFAULT_APPLICATION_SUMMARY_FIELDS
-FROM cas_2_applications a
-JOIN nomis_users nu ON nu.id = a.created_by_user_id
-WHERE a.created_by_user_id = :userId
-AND a.submitted_at IS NULL
-ORDER BY createdAt DESC
-""",
-    nativeQuery = true,
-  )
-  fun findUnsubmittedCas2ApplicationSummariesCreatedByUser(userId: UUID, pageable: Pageable?):
-    Page<Cas2ApplicationSummary>
-
-  @Query(
-    """
-SELECT
-    $DEFAULT_APPLICATION_SUMMARY_FIELDS
-FROM cas_2_applications a
-JOIN nomis_users nu ON nu.id = a.created_by_user_id
-WHERE a.referring_prison_code = :prisonCode
-AND a.submitted_at IS NULL
-ORDER BY createdAt DESC
-""",
-    nativeQuery = true,
-  )
-  fun findUnsubmittedCas2ApplicationSummariesByPrison(prisonCode: String, pageable: Pageable?):
-    Page<Cas2ApplicationSummary>
-
-  @Query(
-    """
-SELECT
-    $ALL_APPLICATION_SUMMARY_FIELDS
-FROM cas_2_applications a
-LEFT JOIN
-    (SELECT DISTINCT ON (application_id) su.application_id, 
-      su.label, su.status_id
-    FROM cas_2_status_updates su
-    ORDER BY su.application_id, su.created_at DESC) as asu
-ON a.id = asu.application_id
-JOIN nomis_users nu ON nu.id = a.created_by_user_id
-WHERE a.submitted_at IS NOT NULL
-""",
-    countQuery =
-    """
-    SELECT COUNT(*)
-      FROM cas_2_applications a
-      WHERE a.submitted_at IS NOT NULL
-    """,
-    nativeQuery = true,
-  )
-  fun findAllSubmittedCas2ApplicationSummaries(pageable: Pageable?): Page<Cas2ApplicationSummary>
-
   @Query(
     "SELECT a FROM Cas2ApplicationEntity a WHERE a.id = :id AND " +
       "a.submittedAt IS NOT NULL",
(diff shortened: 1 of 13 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas2ApplicationSummaryEntity.kt
```diff
@@ -0,0 +1,115 @@
...
+//    """
+//    SELECT COUNT(*)
+//      FROM cas_2_applications a
+//    WHERE a.created_by_user_id = :userId
+//    """,
+//    nativeQuery = true,
+//  )
+  fun findByUserId(userId: String, pageable: Pageable?): Page<Cas2ApplicationSummaryEntity>
+
+//  @Query(countQuery =
```
danhumphreys-moj, line 26: Left in count queries in case these are needed to fix failing integration tests.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas2ApplicationSummaryEntity.kt
```diff
@@ -0,0 +1,115 @@
...
+//    nativeQuery = true,
+//  )
+  fun findBySubmittedAtIsNotNull(pageable: Pageable?): Page<Cas2ApplicationSummaryEntity>
+}
+
+@Entity
+@Table(name = "cas_2_application_summary")
+data class Cas2ApplicationSummaryEntity(
+  @Id
+  val id: String,
```
danhumphreys-moj, line 95: Could change to UUID type.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/ApplicationService.kt
```diff
@@ -67,46 +69,46 @@ class ApplicationService(
     }
   }
 
-  fun getAllApplicationsForUser(user: NomisUserEntity, pageCriteria: PageCriteria<String>): Pair<MutableList<Cas2ApplicationSummary>, PaginationMetadata?> {
-    val response = applicationRepository.findAllCas2ApplicationSummariesCreatedByUser(user.id, getPageable(pageCriteria))
+  fun getAllApplicationsForUser(user: NomisUserEntity, pageCriteria: PageCriteria<String>): Pair<MutableList<Cas2ApplicationSummaryEntity>, PaginationMetadata?> {
```
danhumphreys-moj, line 72: These methods can be refactored further, possibly using `when`.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas2/ApplicationsTransformer.kt
```diff
@@ -40,22 +41,22 @@ class ApplicationsTransformer(
...
     personName: String,
   ): uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Cas2ApplicationSummary {
     return uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model
       .Cas2ApplicationSummary(
-        id = jpaSummary.getId(),
-        createdByUserId = jpaSummary.getCreatedByUserId(),
-        createdByUserName = jpaSummary.getCreatedByUserName(),
-        createdAt = jpaSummary.getCreatedAt().toInstant(),
-        submittedAt = jpaSummary.getSubmittedAt()?.toInstant(),
+        id = UUID.fromString(jpaSummary.id),
```
danhumphreys-moj, line 49: If go back to using interface, these changes would not be required.

src/main/resources/db/migration/all/20240502000000__create_cas_2_summary_view.sql
```diff
@@ -0,0 +1,22 @@
+-- create application summary view
+DROP VIEW IF EXISTS cas_2_application_summary;
+CREATE OR REPLACE VIEW cas_2_application_summary AS SELECT
+    CAST(a.id AS TEXT),
```
danhumphreys-moj, line 4: Is this CAST required?  String or UUID type?

src/main/resources/db/migration/all/20240502000000__create_cas_2_summary_view.sql
```diff
@@ -0,0 +1,22 @@
...
+    a.crn,
+    a.noms_number,
+    CAST(a.created_by_user_id AS TEXT),
```
danhumphreys-moj, line 7: Similar to above.

src/main/resources/db/migration/all/20240502000000__create_cas_2_summary_view.sql
```diff
@@ -0,0 +1,22 @@
...
+    CAST(asu.status_id AS TEXT),
+    a.referring_prison_code
+FROM cas_2_applications a
+LEFT JOIN (SELECT DISTINCT ON (application_id) su.application_id, su.label, su.status_id
+    FROM cas_2_status_updates su
+    ORDER BY su.application_id, su.created_at DESC) as asu
+    ON a.id = asu.application_id
+JOIN nomis_users nu ON nu.id = a.created_by_user_id
+WHERE (a.conditional_release_date IS NULL OR a.conditional_release_date >= current_date)
+ORDER BY a.created_at DESC
```
danhumphreys-moj, line 22: In theory this `ORDER BY` does not need to be there but can't seem to get sorting to work when relying on Pageable to do it.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas2/ApplicationServiceTest.kt
```diff
@@ -79,21 +80,22 @@ class ApplicationServiceTest {
...
-        override fun getNomsNumber() = randomStringMultiCaseWithNumbers(6)
-        override fun getCreatedByUserId() = UUID.fromString("836a9460-b177-433a-a0d9-262509092c9f")
-        override fun getCreatedByUserName() = "first last"
-        override fun getCreatedAt() = Timestamp(Instant.parse("2023-04-19T13:25:00+01:00").toEpochMilli())
-        override fun getSubmittedAt() = Timestamp(Instant.parse("2023-04-19T13:25:30+01:00").toEpochMilli())
-        override fun getHdcEligibilityDate() = LocalDate.parse("2023-04-29")
-        override fun getLatestStatusUpdateLabel(): String? = null
-        override fun getLatestStatusUpdateStatusId(): UUID? = null
-      }
+      val applicationSummary = Cas2ApplicationSummaryEntity(
```
danhumphreys-moj, line 83: Would not be required if interface used - can go back to original.

src/main/resources/db/migration/all/20240502000000__create_cas_2_summary_view.sql
```diff
@@ -0,0 +1,22 @@
+-- create application summary view
+DROP VIEW IF EXISTS cas_2_application_summary;
```
danhumphreys-moj, line 2: Using CREATE OR REPLACE requires that the new view includes the existing columns in the view to be replaced (but can add new ones).  Therefore a DROP is required to refresh the view.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas2ApplicationSummaryEntity.kt
danhumphreys-moj: Issue with integration tests resolved and all are now passing.

src/main/resources/db/migration/all/20240502000000__create_cas_2_summary_view.sql
danhumphreys-moj: Pagination and sorting issue resolved by now using `getPageableOrAll` since `getPageable` returns null if page is not supplied as query parameter, which means results aren't sorted.

danhumphreys-moj:
If we can remove the dependence on [ApplicationsTransformer.kt](https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/d5c2aa1dc2746a6123286a1085642624f7b720df/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/ApplicationsTransformer.kt#L180) the `AppSummary` interface can also go from Cas2ApplicationEntity.kt.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas2ApplicationSummaryEntity.kt
Harriethw: feels like it should be a UUID right? 
It looks like we do a CAST on sql queries elsewhere in the repo `CAST(a.id AS TEXT) as id` but I don't really know why? Maybe something to do with the `ApplicationSummary` interface, or the need to convert it between psql land and jpa land?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas2ApplicationEntity.kt
Harriethw, line 1: Can the `AppSummary` and ` Cas2ApplicationSummary` interface be deleted now? It looks like its only use is in the CAS2 ApplicationsTransformer in a `getStatusFromSummary` that is never used?

Harriethw:
this is magic! looks like it makes things so much simpler.
What would happen if we had to add new fields? Would that be added in another migration file?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/ApplicationService.kt
danhumphreys-moj: Done in [this](d45028763b419ebd20009d447c21515d90f723a5) commit.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas2ApplicationEntity.kt
danhumphreys-moj: Yes, done this.

src/main/resources/db/migration/all/20240502000000__create_cas_2_summary_view.sql
danhumphreys-moj: Regarding your question on adding new fields @Harriethw.  Running CREATE OR REPLACE will add additional columns, but any existing ones need to be there.  I've removed the DROP since I don't think it's necessary.

danhumphreys-moj:
> this is magic! looks like it makes things so much simpler. What would happen if we had to add new fields? Would that be added in another migration file?

danhumphreys-moj:
> > this is magic! looks like it makes things so much simpler. What would happen if we had to add new fields? Would that be added in another migration file?

Added comment above.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/ApplicationService.kt
```diff
@@ -49,58 +49,25 @@ class ApplicationService(
...
-
-  fun getSubmittedApplicationsByPrison(prisonCode: String, pageCriteria: PageCriteria<String>): Pair<MutableList<Cas2ApplicationSummaryEntity>, PaginationMetadata?> {
-    val response = applicationSummaryRepository.findByPrisonCodeAndSubmittedAtIsNotNull(prisonCode, getPageableOrAllPages(pageCriteria))
-    val metadata = getMetadata(response, pageCriteria)
-    return Pair(response.content, metadata)
-  }
-
-  fun getUnsubmittedApplicationsByPrison(prisonCode: String, pageCriteria: PageCriteria<String>): Pair<MutableList<Cas2ApplicationSummaryEntity>, PaginationMetadata?> {
-    val response = applicationSummaryRepository.findByPrisonCodeAndSubmittedAtIsNull(prisonCode, getPageableOrAllPages(pageCriteria))
+    val response = repositoryFunctionMap.get(Pair(prisonCode == null, isSubmitted))!!(
```
tahb, line 67: Clever!

src/main/resources/db/migration/all/20240502000000__create_cas_2_summary_view.sql
```diff
@@ -0,0 +1,20 @@
+-- create application summary view
```
danhumphreys-moj, line 1: Is the date and time of this filename used to determine the order in which these migration jobs are run?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas2ApplicationSummaryEntity.kt
danhumphreys-moj: I've left these as is.  Changing these three CASTed strings back to UUID types gives org.springframework.dao.InvalidDataAccessApiUsageException, the cause of which I haven't been able to resolve.  For the same of getting this PR over the line I think it's acceptable as it stands.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas2/ApplicationsTransformer.kt
danhumphreys-moj: No need to use an interface as before.  Dataclass works as expected.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas2/ApplicationServiceTest.kt
danhumphreys-moj: See [comment](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1860#discussion_r1596450939) above.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/ApplicationsTransformer.kt
```diff
@@ -175,13 +174,6 @@ class ApplicationsTransformer(
     }
   }
 
-  private fun getStatusFromSummary(entity: DomainCas2ApplicationSummary): ApplicationStatus {
```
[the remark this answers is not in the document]
danhumphreys-moj: Yes you're right.  I've added some more info to the commit message.

Since the Cas2ApplicationSummary interface is no longer required it has been removed, along with a legacy function getStatusFromSummary() in ApplicationsTransformer which isn't used (even prior to this work).

src/main/resources/db/migration/all/20240502000000__create_cas_2_summary_view.sql
danhumphreys-moj: @Harriethw @tom I got some checksum errors when I ran ap-tools and built the docker images but these were resolved when I cleared out the volumes.  Just wanted to make sure this should all work ok when we deploy to prod?

src/main/resources/db/migration/all/20240502000000__create_cas_2_summary_view.sql
```diff
@@ -0,0 +1,20 @@
...
+    asu.label,
+    CAST(asu.status_id AS TEXT),
+    a.referring_prison_code
+FROM cas_2_applications a
+LEFT JOIN (SELECT DISTINCT ON (application_id) su.application_id, su.label, su.status_id
+    FROM cas_2_status_updates su
+    ORDER BY su.application_id, su.created_at DESC) as asu
+    ON a.id = asu.application_id
+JOIN nomis_users nu ON nu.id = a.created_by_user_id
+WHERE (a.conditional_release_date IS NULL OR a.conditional_release_date >= current_date)
```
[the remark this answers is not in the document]
danhumphreys-moj: Also regarding whether `conditional_release_date IS NULL` is required, it is because this query is also used to return all (in progress and submitted applications) for the assessor dashboard.  In progress applications where CRD will be null need to be included here.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas2ApplicationSummaryEntity.kt
Harriethw: I think that it would be less messy to have this as a UUID - I believe that the Kotlin side is clever enough to know that this is a UUID, even though we are doing that cast in the psql query, and I've just tried locally and it seems to work. 
(You'll see in the old code that the ApplicationSummary interface had a `getID` function used on the transformer that returned a UUID even though we were CASTing to string there too.)

src/main/resources/db/migration/all/20240513133000__create_cas_2_summary_view.sql
```diff
@@ -0,0 +1,21 @@
+-- create application summary view
+DROP VIEW IF EXISTS cas_2_application_summary;
```
Harriethw, line 2: could the live table creating be put into the same sql file? They seems to depend on each other, and as they haven't yet been merged into main we are free to move them about

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas2/ApplicationServiceTest.kt
```diff
@@ -125,22 +127,23 @@ class ApplicationServiceTest {
 
   @Nested
   inner class GetApplicationsWithPrisonCode {
```
Harriethw, line 129: are we missing some unit test coverage of `getApplications` here?
It looks like this block tests with prison code, but is there a test for when prison code is null?
I wonder if we tightened up the unit tests around this method if we could look at paring back the integration tests in the future...

danhumphreys-moj: Added a more recent commit 68396fd0028cd9bf36471e836afbcd3ad1b612c2 with these tests in.

src/main/resources/db/migration/all/20240513133000__create_cas_2_summary_view.sql
danhumphreys-moj: Yes good idea!  Haver merged the two sql files into one.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas2ApplicationSummaryEntity.kt
danhumphreys-moj: Have changed `id` here from String to UUID.

danhumphreys-moj:
@Harriethw ready for approval?

Harriethw (approved):
thanks for persevering Dan! And introducing some great code that will simplify things immensely 👍

PR #1867, opened 2024-05-10
APS-786 request for placement api fixes

davidatkinsuk:
Prior to this commit the new request for placements API was returning reallocated and unsubmitted requests for placements. The API this is replacing did not return these placement_applications as they are effectively of no use to the user:

* a reallocated placement_application is effectively a soft-deleted placement_application
* there is currently no way for user to complete an unsubmitted placement_application once they navigate away from creating it


edavey (approved):
thanks for these very clearly documented fixes

PR #1869, opened 2024-05-10
APS-787 Request For Placement Status Fixes

davidatkinsuk:
Ensures request for placement states returned by the RequestForPlacementService align with the state transition diagram https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/4872569106/CAS1+Request+for+Placement+State+Transitions

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/RequestForPlacementTransformer.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/RequestForPlacementTransformer.kt
@@ -68,13 +68,16 @@ class RequestForPlacementTransformer(
 
   private fun PlacementApplicationEntity.deriveStatus(): RequestForPlacementStatus = when {
     this.isWithdrawn() -> RequestForPlacementStatus.requestWithdrawn
+
     this.application.getLatestBooking() != null -> RequestForPlacementStatus.placementBooked
 
     this.decision == PlacementApplicationDecision.REJECTED -> RequestForPlacementStatus.requestRejected
 
+    this.decision == PlacementApplicationDecision.ACCEPTED -> RequestForPlacementStatus.awaitingMatch
+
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/RequestForPlacementTransformer.kt
```diff
@@ -68,13 +68,16 @@ class RequestForPlacementTransformer(
 
   private fun PlacementApplicationEntity.deriveStatus(): RequestForPlacementStatus = when {
     this.isWithdrawn() -> RequestForPlacementStatus.requestWithdrawn
+
     this.application.getLatestBooking() != null -> RequestForPlacementStatus.placementBooked
 
     this.decision == PlacementApplicationDecision.REJECTED -> RequestForPlacementStatus.requestRejected
 
+    this.decision == PlacementApplicationDecision.ACCEPTED -> RequestForPlacementStatus.awaitingMatch
```
davidatkinsuk, line 82: Do we need state for 'not matched'? Is that a temporary state that will eventually lead to matched or rejected (in which case maybe we don't need a specific state for it?

edavey: there is "unable to match" which means we've: 
- tried to find a match 
- created a `booking.not-made` domain event to represent and communicate this attempt and failure

we will then re-attempt the match "later" -- but this "waiting list" workflow has not been thought through properly yet

davidatkinsuk: OK, in that case maybe that's one to pick up during match and manage. I've added a note on the state transition to capture this potential path

edavey (approved):
👍

PR #1870, opened 2024-05-10
APS-768 Fix Cas1AutoScript

davidatkinsuk:
This previously failed as the incorrect path was used to load the resource. This commit also uses the Spring Resource loaded for consistency with Cas2Autoscript, and adds an integration test catch Autoscript regression


Harriethw (approved):
👍

PR #1873, opened 2024-05-13
APS-716 CAS1 Email Tidy in preparation for using CRU email address for Reply-To

davidatkinsuk:
This PR ensures that all code sending emails for CAS1 do so via a specialised email service (e.g. Cas1AssessmentEmailService). This simplifies the subsequent changes required to ensure a CRU-specific reply-to email address is used for all CAS1 emails.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/TasksTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/TasksTest.kt
@@ -2427,16 +2427,16 @@ class TasksTest {
     }
 
     @Test
-    fun `Reallocate assessment to different assessor returns 201, creates new assessment, deallocates old one`() {
+    fun `Reallocate assessment to different assessor returns 201, creates new assessment, deallocates old one, sends emails`() {
       `Given a User`(roles = listOf(UserRole.CAS1_WORKFLOW_MANAGER)) { _, jwt ->
-        `Given a User`(roles = listOf(UserRole.CAS1_ASSESSOR)) { user, _ ->
+        `Given a User`(roles = listOf(UserRole.CAS1_ASSESSOR)) { currentlyAllocatedUser, _ ->
           `Given a User`(
             roles = listOf(UserRole.CAS1_ASSESSOR),
           ) { assigneeUser, _ ->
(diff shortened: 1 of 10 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/assessmentservice/AcceptAssessmentTest.kt
```diff
@@ -92,7 +92,7 @@ class AcceptAssessmentTest {
   private val userAllocator = mockk<UserAllocator>()
   private val objectMapperMock = mockk<ObjectMapper>()
   private val taskDeadlineServiceMock = mockk<TaskDeadlineService>()
-  private val assessmentEmailServiceMock = mockk<Cas1AssessmentEmailService>()
+  private val csa1AssessmentEmailServiceMock = mockk<Cas1AssessmentEmailService>()
```
edavey, line 95: tiny typo here in `csa1AssessmentEmailServiceMock` I think

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/TasksTest.kt
```diff
@@ -2427,16 +2427,16 @@ class TasksTest {
     }
 
     @Test
-    fun `Reallocate assessment to different assessor returns 201, creates new assessment, deallocates old one`() {
+    fun `Reallocate assessment to different assessor returns 201, creates new assessment, deallocates old one, sends emails`() {
```
edavey, line 2430: neatly done!

edavey (approved):
Very clear, thanks for structuring the commits so carefully. 

If I followed correctly, the `Cas1EmailNotificationService` defined and unit tested in the first commit isn't used anywhere yet. Would it be better to remove that commit from this PR and instead introduce it in a future PR, where you first use it?

davidatkinsuk:
> If I followed correctly, the `Cas1EmailNotificationService` defined and unit tested in the first commit isn't used anywhere yet. Would it be better to remove that commit from this PR and instead introduce it in a future PR, where you first use it?

Yes, this is a fair point. It was useful to aid my thinking but isn't actually used currently so i've removed that commit from teh PR

PR #1875, opened 2024-05-13
APS-794 Fix CAS1 Request for Placement PA Status Transform

davidatkinsuk:
Previously we used placements on the application as a whole to determine if a placement_application’s status is ‘placementBooked’. This commit fixes that and instead uses the placements associated with the specific placement_application to determine this.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PlacementApplicationEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PlacementApplicationEntity.kt
@@ -121,11 +121,15 @@ data class PlacementApplicationEntity(
 
   /**
    * Supporting multiple dates in a single Placement Application is legacy behaviour. Any new
-   * placement application will only ever one and only one date in this collection
+   * placement application will have one and only one date in this collection
    */
   @OneToMany(mappedBy = "placementApplication")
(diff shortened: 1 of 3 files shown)


richpjames (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PlacementApplicationEntity.kt
```diff
@@ -126,6 +126,10 @@ data class PlacementApplicationEntity(
   @OneToMany(mappedBy = "placementApplication")
   var placementDates: MutableList<PlacementDateEntity>,
 
+  /**
+   * Supporting multiple placements requests in a single Placement Application is legacy behaviour. Any new
+   * placement application will only ever one and only one date in this collection
```
edavey, line 131: did you mean to say something along the lines of "will only have one date in this"?

edavey (approved):
thks, looks good to me - just a tiny typo in the comment perhaps?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PlacementApplicationEntity.kt
davidatkinsuk: Good spot, fixed

PR #1877, opened 2024-05-13
CAS2-465:  Restrict offender search to within the user's active prison

Harriethw:
Can be merged after this Tools update https://github.com/ministryofjustice/hmpps-approved-premises-tools/pull/67 ✅ 
And this UI update https://github.com/ministryofjustice/hmpps-community-accommodation-tier-2-ui/pull/700 ✅  
And this API update https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1879 ✅ 

_As a POM 
When I try to create an application
With a prison number for a person who is not within my active prison
Then I am not allowed to continue_

This PR checks the inmate detail's `agencyId` against the user's `activeAgencyId`. If they are not within the same prison, then we return an unauthorised error.

More info about when this check is done and where data is from here in Confluence https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/4887511138/Integration+APIs+for+offender+data

More info about where a Nomis users caseload is set here:
https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/4881088539/How+to+set+a+user+s+active+caseload+aka+prison+in+dev+test

AFTER THIS HAD MERGED:

Our test referrers in dev (LCA and POM) and any of our dev/test users will need to have their prison set to Moorland, [following instructions here](https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/4881088539/How+to+set+a+user+s+active+caseload+aka+prison+in+dev+test)

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/OffenderService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/OffenderService.kt
@@ -216,8 +216,15 @@ class OffenderService(
     val inmateDetail = when (inmateDetailResponse) {
       is ClientResult.Success -> inmateDetailResponse.body
       is ClientResult.Failure.StatusCode -> when (inmateDetailResponse.status) {
-        HttpStatus.NOT_FOUND -> return AuthorisableActionResult.NotFound()
-        HttpStatus.FORBIDDEN -> return AuthorisableActionResult.Unauthorised()
+        HttpStatus.NOT_FOUND -> {
+          logFailedResponse(inmateDetailResponse)
+          return AuthorisableActionResult.NotFound()
+        }
+
+        HttpStatus.FORBIDDEN -> {
+          logFailedResponse(inmateDetailResponse)
+          return AuthorisableActionResult.Unauthorised()
+        }
         else -> {
           logFailedResponse(inmateDetailResponse)
           null
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/OffenderService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/OffenderService.kt
@@ -69,10 +69,7 @@ class OffenderService(
       val probationOffenderDetail = probationOffenderDetailList[0]
 
       // check for restrictions or exclusions
-      val isLao = probationOffenderDetail.currentExclusion == true || probationOffenderDetail.currentRestriction == true
-      if (isLao) {
-        return ProbationOffenderSearchResult.Forbidden(nomsNumber)
-      }
+      if (hasRestrictionOrExclusion(probationOffenderDetail)) return ProbationOffenderSearchResult.Forbidden(nomsNumber)
 
       // check inmate details from Prison API
       val inmateDetails = getInmateDetailsForProbationOffender(probationOffenderDetail)
@@ -85,6 +82,10 @@ class OffenderService(
         ProbationOffenderSearchResult.Forbidden(nomsNumber)
       }
     }
+  }
+
+  private fun hasRestrictionOrExclusion(probationOffenderDetail: ProbationOffenderDetail): Boolean {
+    return probationOffenderDetail.currentExclusion == true || probationOffenderDetail.currentRestriction == true
   }
 
   private fun isOffenderSamePrisonAsUser(inmateDetail: InmateDetail?, currentUser: NomisUserEntity): Boolean {
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/OffenderService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/OffenderService.kt
@@ -10,6 +10,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.ClientResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.PrisonsApiClient
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.ProbationOffenderSearchApiClient
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.datasource.OffenderDetailsDataSource
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.NomisUserEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.PersonInfoResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.PersonRisks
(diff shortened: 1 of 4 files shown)


Harriethw:
The forbidden error this will throw I believe is indistinguishable from the forbidden error from an LAO - I thought this would be fine as we had planned to remove the LAO restriction, but now it looks like we might not. Do we need to send some message to the UI to show why it was forbidden? If so, I think this should be a separate PR/ticket and done in tandem with UI?

danhumphreys-moj:
> The forbidden error this will throw I believe is indistinguishable from the forbidden error from an LAO - I thought this would be fine as we had planned to remove the LAO restriction, but now it looks like we might not. Do we need to send some message to the UI to show why it was forbidden? If so, I think this should be a separate PR/ticket and done in tandem with UI?
> 
> UPDATE:
> 
> I don’t think we should distinguish between the two errors for security reasons, because by adding [that context we give a bit of information to a bad actor](https://owasp.org/www-project-top-10-insider-threats/docs/2023/INT08_2023-Information_Leakage#:~:text=Description,security%20measures%20or%20personnel%20negligence.). (e.g. If I had hacked an account for a POM at HMPPS Leeds, and I got the error saying this person is not in my prison, I would now know that an offender is NOT in Leeds prison, or would know they are an LAO).

Was wondering whether we needed a ticket in the backlog to investigate this but I think better not to add another ticket on the off chance if might be required in the future and wait for a firm requirement to come through from a user.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/OffenderService.kt
```diff
@@ -66,17 +67,30 @@ class OffenderService(
...
     } else {
       val probationOffenderDetail = probationOffenderDetailList[0]
+
+      // check for restrictions or exclusions
       val isLao = probationOffenderDetail.currentExclusion == true || probationOffenderDetail.currentRestriction == true
       if (isLao) {
         return ProbationOffenderSearchResult.Forbidden(nomsNumber)
       }
 
+      // check inmate details from Prison API
```
danhumphreys-moj, line 74: As it stands we check for restrictions or exclusions on the offender then check to see if they are in the same prison.  Might be worth switching this order so that we aren't in a position to return a forbidden error for an offender with a restriction, that potentially is from a prison different to where the POM is based?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas2/OffenderServiceTest.kt
```diff
@@ -273,17 +296,29 @@ class OffenderServiceTest {
 
       val inmateDetail = InmateDetailFactory()
         .withOffenderNo(nomsNumber)
+        .withAssignedLivingUnit(
```
danhumphreys-moj, line 299: Just curious what has changed that we need to add an AssignedLivingUnit?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/OffenderService.kt
```diff
@@ -39,9 +40,9 @@ class OffenderService(
 
   private val log = LoggerFactory.getLogger(this::class.java)
 
-  fun getPersonByNomsNumber(nomsNumber: String): ProbationOffenderSearchResult {
-    fun logFailedResponse(probationResponse: ClientResult.Failure<List<ProbationOffenderDetail>>) =
-      log.warn("Could not get inmate details for $nomsNumber", probationResponse.toException())
+  fun getPersonByNomsNumber(nomsNumber: String, currentUser: NomisUserEntity): ProbationOffenderSearchResult {
+    fun logFailedResponse(probationResponse: ClientResult.Failure<List<ProbationOffenderDetail>>?) =
```
danhumphreys-moj, line 44: This could be moved to the "log" commit.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/OffenderService.kt
```diff
@@ -87,6 +84,10 @@ class OffenderService(
     }
   }
 
+  private fun hasRestrictionOrExclusion(probationOffenderDetail: ProbationOffenderDetail): Boolean {
+    return probationOffenderDetail.currentExclusion == true || probationOffenderDetail.currentRestriction == true
```
danhumphreys-moj, line 88: Minor..
`return probationOffenderDetail.currentExclusion || probationOffenderDetail.currentRestriction`

danhumphreys-moj (approved):
Looks good Harriet.  Just one thing regarding the order of exclusion/restrictio and same prison checks.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas2/OffenderServiceTest.kt
Harriethw: we are checking `isOffenderSamePrisonAsUser`  which compares `assignedLivingUnit.agencyId` of the InmateDetails to the user's `activeCaseloadId`. so we need to set the inmate details up with the same prison as the user  for the happy path

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/OffenderService.kt
Harriethw: `currentExclusion` and `currentRestriciton` could be null, so Kotlin doesn't like doing this, the linter suggests adding the `== true`

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/OffenderService.kt
Harriethw: good spot!

Harriethw: turns out I didn't need it at all!

Harriethw:
> Was wondering whether we needed a ticket in the backlog to investigate this but I think better not to add another ticket on the off chance if might be required in the future and wait for a firm requirement to come through from a user.

As I've said above, I don't think even if there were a user request for this that it should be done, as it would be leaking information.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/OffenderService.kt
Harriethw: hmm having looked at it again I'm wondering if LAO checks should go first actually - we already have the info from the Delius context on whether they are an LAO, so if they are we don't need to go and do the Prison API call and we can return early. AFAIK the controls on LAOs should apply even if the offender is within the same prison.

PR #1878, opened 2024-05-14
CAS-370 Add prison release type to temporary accommodation application submit

muhammad-elabdulla:
This [PR CAS-370](https://dsdmoj.atlassian.net/browse/CAS-370) is to add prison release types when submit temporary accommodation application

### DIFF
--- a/src/main/resources/db/migration/all/20240513120304__add_application_prison_release_types_table.sql
+++ b/src/main/resources/db/migration/all/20240513120304__add_application_prison_release_types_table.sql
@@ -1,4 +1,4 @@
-CREATE TABLE application_prison_release_types (
+CREATE TABLE temporary_accommodation_application_prison_release_types (
     application_id UUID NOT NULL,
     prison_release_type_id UUID NOT NULL,
     PRIMARY KEY (application_id, prison_release_type_id),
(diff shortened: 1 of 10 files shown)


src/main/resources/db/migration/all/20240513120304__add_application_prison_release_types_table.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE application_prison_release_types (
```
davidatkinsuk, line 1: I think this should maybe be called 'temporary_accommodation_application_prison_release_types' as it's only applicable to cas3 application?

davidatkinsuk (approved):
Just one comment from me

src/main/resources/db/migration/all/20240513120304__add_application_prison_release_types_table.sql
muhammad-elabdulla: Changed. I named it as application_prison_release_types if another CAS need to use it in the future

davidatkinsuk: I was considering that but noticed we had a simple table for CAS1 that was prefixed with approved_premises_applicaitons. we can always rename (and move relationship to the common application entity in code) if needed by other CASs

PR #1879, opened 2024-05-14
CAS2-465: Add correct caseloads to seeded data

Harriethw:
Now that we are introducing a restriction to [creating applications based on caseload](https://dsdmoj.atlassian.net/jira/software/c/projects/CAS2/boards/1347?assignee=712020%3Ad443cab3-dd14-40a2-b99d-f6c119c3259b&selectedIssue=CAS2-465), we need to ensure that all our local and dev example and seeded data will work. 

The prison code is relevant in three main places
1. The offender - the `InmateDetails` returned from the Prison API returns `assignedLivingUnit.agencyId` which is equivalent to the prison code.
2. The user - the Nomis User Roles API returns `activeCaseloadId` which is equivalent to the prison code.
3. The Application - Our submitted CAS2 Applications have the `referringPrisonCode` which is taken from the InmateDetails referred to above.

This AP Tools PR setup handled 1 by setting the local users to have the same prison code as our local example offenders. https://github.com/ministryofjustice/hmpps-approved-premises-tools/pull/67

This PR now handles 3 for seeded data locally and in dev/test.
The local InmateDetails are seeded in the Prison API here - https://github.com/ministryofjustice/prison-api/blob/37b7392bc69960f159de27e72aee19a9ae7e6ad5/src/main/resources/db/migration/data/R__1_15__OFFENDERS.sql#L26
They have the code `LEI`, so  for all local applications we are setting the prison code to be LEI.
The dev/test InmateDetails for the users listed here https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/4467949648/CRNs+for+Testing+in+Development+Environment 
however all have the code `MDI`, so for all dev/test applications we are setting the prison code to be LEI.

I've also replaced the seeded `nomsNumber` with an offender who has data on the Prison API in dev/test, as the current one does not.


danhumphreys-moj (approved):
Thanks for the extra information Harriet!

tahb (approved):

PR #1880, opened 2024-05-15
APS-768 Cas1AutoScript shouldn’t block startup on failure

davidatkinsuk:
This commit ensures that if the Cas1AutoScript fails to run startup of the service is not blocked, and any changes made to the database are rolled back


edavey (approved):
nicely done, thanks

PR #1882, opened 2024-05-15
APS-716 Use CRU Email Address for CAS1 Emails

davidatkinsuk:
This PR introduces a Cas1EmailNotificationService that should be used for sending any CAS1 related emails. Both email functions on this service require a corresponidng ApprovedPremisesApplicationEntity to be provided, from which the reply-to email address will be determined (from the corresponding ap_area)


edavey (approved):
Thanks, sorry for the delay in reviewing.

PR #1883, opened 2024-05-15
CAS-370 Capture the type of prison release for referral report

muhammad-elabdulla:
This [PR CAS-370](https://dsdmoj.atlassian.net/browse/CAS-370) is: 

- To capture the type of prison release when temporary accommodation application is submitted
- To show the type of prison release in referral report
- Remove the types of prison release from backend as they will be created in UI


davidatkinsuk (changes requested):
Could you please rebase/squash the commits to get rid of the 'fix' commit and the 'merge' commit

Could you please update the commit to explain what change is being made and why? I think it is:

* Removing the Prison Release Type Reference Data from the API as this will now be managed in the UI
* When submitting an application instead of receiving the Identifiers for reference data we now instead receive a simple list of a Print Release Type name/descriptions

Technically, all looks good 👍

muhammad-elabdulla:
Changes are made

davidatkinsuk:
> Changes are made

Thanks @muhammad-elabdulla . I still see a merge commit, can this be removed/squashed. Thanks

muhammad-elabdulla:
I agree that if there are a list of commits if will be difficult to read but with the main commit and merge commit should be Ok

davidatkinsuk:
> I agree that if there are a list of commits if will be difficult to read but with the main commit and merge commit should be Ok

We have an established principle of only the merges being those created by github merging the PR.

Is there a particular reason you can't remove this merge commit? Should be a matter of squashing and maybe a rebase if main has changed. I don't think we should be including such commits unless absolutely unavoidable?

davidatkinsuk:
> > I agree that if there are a list of commits if will be difficult to read but with the main commit and merge commit should be Ok
> 
> We have an established principle of only the merges being those created by github merging the PR.
> 
> Is there a particular reason you can't remove this merge commit? Should be a matter of squashing and maybe a rebase if main has changed. I don't think we should be including such commits unless absolutely unavoidable? I could be missing something.

Maybe a case of resetting to the commit before the merge commit, stash changes hanging around from the merge commit and then doing a rebase from there

muhammad-elabdulla:
I don't see that have merge commit when there are a conflict to be included. We cannot keep rebasing the PRs if it take same time until its approved

davidatkinsuk:
> I don't see that have merge commit when there are a conflict to be included. We cannot keep rebasing the PRs if it take same time until its approved. We need also to be flexible in some way

I think there are two issues here:

* regarding rebasing until approval ... normally i'd approve PRs even if i have comments, but in the past people have gone ahead and merged their PR without event responding to comments. i'll happily approve this PR with the understanding that comments will be addressed or responded to to the point we're both happy. Also in this case i absolutely don't think we should be merging PRs with spurious merge commits as we rebase on this project, not merge . That's why i didn't instantly approve. I've approved now to avoid any delays after rebase
* we absolutely should not have merge commits in the project (excluding the github PR merge), i don't think it's unreasonable for the PR to be fixed up to remove these.

davidatkinsuk (approved):
Approved with the understanding that the merge commit will be removed

PR #1885, opened 2024-05-15
APS-791 Add missing indexes for FK columns

davidatkinsuk:
There are several FKs used in reports or hibernate relationships that are not indexed. This commit adds those missing indexes


src/main/resources/db/migration/all/20240515143555__add_missing_indexes_to_fks.sql
```diff
@@ -0,0 +1,3 @@
+CREATE INDEX ON domain_event(assessment_id);
```
edavey, line 1: should this be `domain_events(..)`?

davidatkinsuk: that would explain all the test failures!

edavey (approved):
👍

PR #1887, opened 2024-05-15
ADR-023: Begin extracting CAS1 code to its own namespace

edavey:
See new ADR 23 for details.


davidatkinsuk (approved):
LGTM and i think this will really help with code maintenance and comprehension going forward 👍

PR #1889, opened 2024-05-17
CAS-322 Migrate CAS3 users PDU from nDelius

muhammad-elabdulla:
This [PR CAS-322](https://dsdmoj.atlassian.net/browse/CAS-322) is to get the last user ndelius borough to update users table for cas3 users

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
@@ -15,7 +15,6 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.UserSortField
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.ClientResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.CommunityApiClient
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ProbationAreaProbationRegionMappingRepository
-import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ProbationDeliveryUnitEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ProbationDeliveryUnitRepository
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ProbationRegionEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ProbationRegionRepository
@@ -219,10 +218,11 @@ class UserService(
       is ClientResult.Failure -> staffUserDetailsResponse.throwException()
     }
 
-    if (deliusUser.teams?.firstOrNull()?.borough?.code != null && user.probationDeliveryUnit?.deliusCode !== deliusUser.teams.firstOrNull()?.borough?.code) {
-      when (val probationDeliveryUnit = findProbationDeliveryUnit(deliusUser.teams)) {
+    val userLastBoroughCode = orderDeliusUserTeams(deliusUser)?.firstOrNull()?.borough?.code
+    if (userLastBoroughCode != null && user.probationDeliveryUnit?.deliusCode != userLastBoroughCode) {
+      when (val probationDeliveryUnit = probationDeliveryUnitRepository.findByDeliusCode(userLastBoroughCode)) {
         null -> {
-          throw Exception("Unable to find community API borough code ${deliusUser.teams.first().borough.code} in CAS")
+          throw Exception("Unable to find community API borough code $userLastBoroughCode in CAS")
         }
 
         else -> {
@@ -252,7 +252,8 @@ class UserService(
       }
     }
 
-    deliusUser.teams?.firstOrNull()?.let { team ->
+    val userTeams = orderDeliusUserTeams(deliusUser)
+    userTeams?.firstOrNull()?.let { team ->
       probationDeliveryUnitRepository.findByDeliusCode(team.borough.code)?.let { probationDeliveryUnit ->
         user.probationDeliveryUnit = probationDeliveryUnit
       }
@@ -319,7 +320,8 @@ class UserService(
       }
     }
 
-    val staffProbationDeliveryUnit = findProbationDeliveryUnit(staffUserDetails.teams)
+    val userLastBoroughCode = orderDeliusUserTeams(staffUserDetails)?.firstOrNull()?.borough?.code
+    val staffProbationDeliveryUnit = userLastBoroughCode?.let { probationDeliveryUnitRepository.findByDeliusCode(it) }
 
     val apArea = cas1UserMappingService.determineApArea(staffProbationRegion, staffUserDetails)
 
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
```diff
@@ -355,6 +357,10 @@ class UserService(
     return teams?.firstOrNull()?.borough?.let { probationDeliveryUnitRepository.findByDeliusCode(it.code) }
   }
 
+  private fun orderUserDeliusTeams(deliusUser: StaffUserDetails): List<StaffUserTeamMembership>? {
```
marcocolluramoj, line 360: It might be slightly easier to read if we rename this to `orderedDeliusUserTeams`

marcocolluramoj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
```diff
@@ -219,10 +219,11 @@ class UserService(
       is ClientResult.Failure -> staffUserDetailsResponse.throwException()
     }
 
-    if (deliusUser.teams?.firstOrNull()?.borough?.code != null && user.probationDeliveryUnit?.deliusCode !== deliusUser.teams.firstOrNull()?.borough?.code) {
-      when (val probationDeliveryUnit = findProbationDeliveryUnit(deliusUser.teams)) {
+    val userTeams = orderUserDeliusTeams(deliusUser)
+    if (userTeams?.firstOrNull()?.borough?.code != null && user.probationDeliveryUnit?.deliusCode !== userTeams.firstOrNull()?.borough?.code) {
```
davidatkinsuk, line 223: i don't think we want referential integrity checks provided by !==.

similar to java i suspect comparing strings by reference could leave to false negatives. I think != should be used instead.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
```diff
@@ -219,10 +219,11 @@ class UserService(
...
+      when (val probationDeliveryUnit = findProbationDeliveryUnit(userTeams)) {
         null -> {
-          throw Exception("Unable to find community API borough code ${deliusUser.teams.first().borough.code} in CAS")
+          throw Exception("Unable to find community API borough code ${userTeams.first().borough.code} in CAS")
```
davidatkinsuk, line 226: userTeams.first().borough.code or variations on it are used in 3 places in this function. Maybe line 221 could be changed to :

`val firstBoroughCode = orderUserDeliusTeams(deliusUser)?.firstOrNull()?.borough?.code`

And then firstBoroughCode could be used through the if block?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
muhammad-elabdulla: function renamed

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
```diff
@@ -252,7 +253,8 @@ class UserService(
       }
     }
 
-    deliusUser.teams?.firstOrNull()?.let { team ->
+    val userTeams = orderUserDeliusTeams(deliusUser)
+    userTeams?.firstOrNull()?.let { team ->
```
davidatkinsuk, line 257: we call orderUserDetailsTeam(deliusUser) twice in this function. Could call it once and capture the firstOrNull in a variable e.g.

val firstUserTeam = orderUserDeliusTeams(deliusUser)?.firstOrNull()

This of course slightly changes the approach suggested in my previous comment!

davidatkinsuk (approved):
Looks good although I think the referential integrity check may cause issues. Also a few comments from me re: duplication

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
muhammad-elabdulla: I know that !== is for referential comparing IO used it because i find that in some parts of the project they using it to compare nullable strings

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
muhammad-elabdulla: I need userTeams as I use it in line 224

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
muhammad-elabdulla: We calling it once only in this function

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
davidatkinsuk: In that case, i'm pretty sure that code is wrong too as it could lead to false negatives but in that context the side affect is minor as all it will do is unnecessarily update the user - https://www.baeldung.com/kotlin/string-comparison.

In kotlin using == to compare things that may be null is safe (unlike java) https://kotlinlang.org/docs/equality.html#structural-equality

(on a wider note, that userHasChanged function is broken for other reasons see APS-579)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
davidatkinsuk: i can see orderUserDeliusTeams being used on 256 and 222 but this is just minor duplication so happy to merge if you're happy

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
davidatkinsuk: my suggestion was to capture in a variable for first usage on onwards but this is just minor duplication so happy to merge if you're happy

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
muhammad-elabdulla: I changed the code by removing the function findProbationDeliveryUnit and get the last Borough code for the user as you suggested

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
muhammad-elabdulla: I changed the code by removing the function findProbationDeliveryUnit and get the last Borough code for the user as you suggested

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
muhammad-elabdulla: I agree that using == to compare strings is better approach but I still not get why it used === in the project

davidatkinsuk: I suspect it was a just a mistake

davidatkinsuk: possibly by someone who mostly codes javascript, i think === is more common there possibly

PR #1891, opened 2024-05-20
CAS2-440 soft delete by status

danhumphreys-moj:
Ref: https://dsdmoj.atlassian.net/browse/CAS2-440

**User Story**
As a referrer
I want old applications to disappear from my dashboards
So that they do not clutter up the space

**Description**
This feature will automatically hide applications from dashboards once certain criteria are met. 

**Acceptance Criteria**
API: Soft delete submitted applications once a period of 14 days has elapsed from when the following statuses were applied to them:

- Referral withdrawn
- Referral cancelled
- Awaiting arrival

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ApplicationTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ApplicationTest.kt
@@ -5,6 +5,9 @@ import com.fasterxml.jackson.databind.JsonNode
 import com.fasterxml.jackson.databind.node.NullNode
 import com.ninjasquad.springmockk.SpykBean
 import io.mockk.clearMocks
+import io.mockk.every
+import io.mockk.mockkStatic
+import io.mockk.unmockkAll
 import org.assertj.core.api.Assertions
 import org.junit.jupiter.api.AfterEach
 import org.junit.jupiter.api.BeforeEach
@@ -34,6 +37,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApplicationRe
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas2ApplicationEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ExternalUserEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.NomisUserEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.community.OffenderDetailSummary
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.randomDateAfter
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.randomDateBefore
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.randomDateTimeBefore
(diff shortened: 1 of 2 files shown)


danhumphreys-moj:
Temporarily changed to Draft since wanted to check with @Harriethw whether there might be some changes required to the seeded data to not hide applications in dev/test environments.

src/main/resources/db/migration/all/20240521110000__create_live_view.sql
```diff
@@ -0,0 +1,48 @@
...
+    a.created_at,
+    a.submitted_at,
+    a.hdc_eligibility_date,
+    a.label,
+    a.status_id,
+    a.referring_prison_code
+FROM cas_2_application_summary a
+WHERE (a.conditional_release_date IS NULL OR a.conditional_release_date >= current_date)
+AND (a.label IS NULL
+    OR (a.label = 'Referral withdrawn' AND a.status_created_at > (current_date - INTERVAL '14 DAY'))
```
Harriethw, line 38: I'm a bit worried that this is creating a dependency on the label for a couple of reasons - 
1. If we were to add a new type of status update would it be incorrectly hidden in this case? It wouldn't be clear to me that I would need to update this view query whenever a new status update is added. Is there a way of having the query only affect those three that we know have the 14 day cut off?
2. The label is the human readable label for the status update which could change, I think it should be based on the `status_id` which is the static ID for the type of status update (see `Cas2ApplicationStatusSeeding`). This will make it unreadable though so might need some comments 😬

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ApplicationTest.kt
```diff
@@ -179,6 +183,107 @@ class Cas2ApplicationTest : IntegrationTestBase() {
   @Nested
   inner class GetToIndex {
 
+    @Test
+    fun `return unexpired applications when applications GET is requested`() {
+      val unexpiredSubset = setOf(
```
Harriethw, line 188: does this cover applications without a status update as well?

danhumphreys-moj: No it didn't, good point.  I've added another couple of applications which don't have status updates.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ApplicationTest.kt
```diff
@@ -179,6 +183,112 @@ class Cas2ApplicationTest : IntegrationTestBase() {
...
+      val unexpiredDateTimesUnexpiredSubset = List(unexpiredSubset.size) { unexpiredDateTime() }
+      val expiredDateTimesExpiredSubset = List(expiredSubset.size) { expiredDateTime() }
+
+      val unexpiredApplicationIds = mutableSetOf<UUID>()
+      val expiredApplicationIds = mutableSetOf<UUID>()
+
+      `Given a CAS2 POM User` { userEntity, jwt ->
+        `Given an Offender` { offenderDetails, _ ->
+
+          repeat(2) {
```
danhumphreys-moj, line 242: @Harriethw ref: https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1891#discussion_r1611763264

src/main/resources/db/migration/all/20240521110000__create_live_view.sql
danhumphreys-moj: Good points.  Thanks!

Ref 1. I have changed the logic from IN LIST () to NOT IN LIST () so this addresses this point.

Ref 2. I was conscious it wasn't great comparing strings in the original SQL so comparing by UUID is better, I've made this change and have added a comment to say which UUID corresponds to which status.  Maybe at some point in the future these statuses will go into a dedicated Postgres table.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ApplicationTest.kt
```diff
@@ -179,6 +183,112 @@ class Cas2ApplicationTest : IntegrationTestBase() {
...
+          withCrn(offenderDetails.otherIds.crn)
+          withCreatedAt(OffsetDateTime.now().minusDays(28))
+          withConditionalReleaseDate(LocalDate.now().plusDays(1))
+        }
+      }
+
+      fun createStatusUpdate(status: Pair<String, UUID>, application: Cas2ApplicationEntity) {
+        val status = cas2StatusUpdateEntityFactory.produceAndPersist {
+          withLabel(status.first)
+          withStatusId(status.second)
```
danhumphreys-moj, line 223: I had this originally down as `withId()` and stumped myself for ages :o((

danhumphreys-moj:
I have left the seeded dates as they are since the status createdAt dates isn't hard coded.  Hopefully ok as it is.

danhumphreys-moj:
Just looking at integration tests.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ApplicationTest.kt
```diff
@@ -179,6 +183,112 @@ class Cas2ApplicationTest : IntegrationTestBase() {
...
+          unexpiredSubset.forEachIndexed { index, status ->
+            val application = createApplication(userEntity, offenderDetails)
+            every { OffsetDateTime.now() } returns unexpiredDateTimesUnexpiredSubset[index]
+            createStatusUpdate(status, application)
+            unexpiredApplicationIds.add(application.id)
+          }
+
+          expiredSubset.forEachIndexed { index, status ->
+            val application = createApplication(userEntity, offenderDetails)
+            every { OffsetDateTime.now() } returns expiredDateTimesExpiredSubset[index]
```
Harriethw, line 264: sorry, I'm finding all the date time stuff and the unions quite hard to parse...
are you mocking `OffsetDateTime.now()` here because it is called in the status update factory? and you want to set the status update's `createdAt` to a date that is 14 days or more in the past?

Would it be possible to ditch all these lists of date times and this mocking and just say what date you want status update to be have as the `createdAt` ?

```


      fun createStatusUpdate(status: Pair<String, UUID>, application: Cas2ApplicationEntity, createdAt: OffsetDateTime) {
        val status = cas2StatusUpdateEntityFactory.produceAndPersist {
          withLabel(status.first)
          withStatusId(status.second)
          withApplication(application)
          withAssessor(externalUserEntityFactory.produceAndPersist())
          withCreatedAt(createdAt)
        }
      }
 ```
 
 I think we generally only mock a return when its being used within the code under test, not to setup the testing scenario

File: src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ApplicationTest.kt [hunk 6]
```diff
@@ -179,6 +183,112 @@ class Cas2ApplicationTest : IntegrationTestBase() {
   @Nested
   inner class GetToIndex {
 
+    @Test
```
Harriethw, line 186: could we delete the `Get all applications returns 200 with correct body` now as a I guess the same functionality will be tested by this test? aware that this file is getting really long now!

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ApplicationTest.kt
```diff
@@ -179,6 +183,112 @@ class Cas2ApplicationTest : IntegrationTestBase() {
...
+        Pair("Awaiting decision", UUID.fromString("ba4d8432-250b-4ab9-81ec-7eb4b16e5dd1")),
+        Pair("On waiting list", UUID.fromString("a919097d-b324-471c-9834-756f255e87ea")),
+        Pair("Place offered", UUID.fromString("176bbda0-0766-4d77-8d56-18ed8f9a4ef2")),
+        Pair("Offer accepted", UUID.fromString("fe254d88-ce1d-4cd8-8bd6-88de88f39019")),
+        Pair("Could not be placed", UUID.fromString("758eee61-2a6d-46b9-8bdd-869536d77f1b")),
+        Pair("Incomplete", UUID.fromString("4ad9bbfa-e5b0-456f-b746-146f7fd511dd")),
+        Pair("Offer declined or withdrawn", UUID.fromString("9a381bc6-22d3-41d6-804d-4e49f428c1de")),
+      )
+
+      val expiredSubset = setOf(
```
Harriethw, line 199: maybe this could be named as `expiredStatuses` and `unexpiredStatuses`, to make it a bit more specific?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ApplicationTest.kt
Harriethw: this can be ignored - should have read the commit message 🙃

Harriethw (approved):
approving as just have some minor suggestions to naming/ combining the tests.
I would also suggest squashing these commits together, so that the tests for the change you are making in the migration is in the same commit. 🧪

PR #1892, opened 2024-05-20
APS-809 Add metadata to domain event model

davidatkinsuk:


doc/architecture/decisions/0023-use-domain-event-metadata-for-reports.md
```diff
@@ -0,0 +1,29 @@
+# 8. Use Domain Event Metadat for Reports
```
edavey, line 1: tiny typo on "Metadata"

doc/architecture/decisions/0023-use-domain-event-metadata-for-reports.md
```diff
@@ -0,0 +1,29 @@
...
+
+The Domain Event model will be updated to include a Map<String,String> of metadata, with the intention of using this data when generating reports instead of directly accessing data/document json. APIs that create Domain Events will be updated to capture any fields required in metadata that aren't already provided in first class fields manner.
+
+This is a good fit for capturing report-specific data because domain events already provide a 'snapshot' of the application's state during key reporting events (e.g. application submitted, application withdrawn).
+
+Furthermore, in places where reports access domain event data json (something which is less of an issue because the API controls the domain event schema), it would be preferred if the equivalent data was instead captured in the metadata, simplifying reporting definition and providing a small performance improvement.
+
+## Consequences
+
+This change will ensure that the API code no longer needs to understand the structure of json in the data/document columns, as was the original intention for the data/document JSON.
```
edavey, line 29: Is it also worth including the "consequence" that this will remove the brittle coupling between reports and the JSON structures which are liable to change?

edavey (approved):
Thanks. This is a really important initiative for making the reporting less fragile and easier to maintain.

doc/architecture/decisions/0023-use-domain-event-metadata-for-reports.md
edavey: i.e. that we are increasing reliability. Not only did the report have to understand the structure, it had to pray that the structure did not change in a breaking way...

davidatkinsuk: That's a really good point, i've updated the ADR to reflect this.

PR #1893, opened 2024-05-20
APS-799 Remove feature flags

davidatkinsuk:
This PR removes two feature flags for features that have been live for over a week without issues:

* cas1-remove-document-conviction-id-filters
* cas1-check-delius-arrival-status-on-withdrawal


edavey (approved):
Thanks for the tidy-up.

PR #1897, opened 2024-05-20
[APS-798] Show events associated with offline applications in the personal timeline

danlivings-dxw:
> See [APS-798 on the Approved Premises Jira board](https://dsdmoj.atlassian.net/browse/APS-798).

This PR updates the behaviour of the `GET /people/{crn}/timeline` endpoint so that events for both regular applications and offline applications can be retrieved for a given CRN. The `cas1-provide-offline-application-events-in-personal-timeline` [feature flag](doc/architecture/decisions/0022-handling-feature-flags-with-flipt.md) can be used to configure this behaviour: when `false` (by default), the current behaviour of only showing events for regular applications will be applied.

To support this, the `ApplicationTimeline` API schema has been modified so that the `status` and `createdBy` properties are no longer required (as offline applications do not contain this information), and a new `isOfflineApplication` boolean property has been added.

The `PersonalTimelineTransformer` class has been modified to support offline applications. The data model used by the transformer has been changed to the newly-introduced `ApplicationTimelineModel` class, which consists of a `BoxedApplication` and a list of `TimelineEvent`s.

A `BoxedApplication` provides a typed wrapper around either an `ApprovedPremisesApplicationEntity` or an `OfflineApplicationEntity`, which can be used as follows:

```kotlin
val regularApplicationEntity = ApprovedPremisesApplicationEntityFactory().produce()
val boxedApplication1 = BoxedApplication.of(regularApplicationEntity)

val offlineApplicationEntity = OfflineApplicationEntityFactory().produce()
val boxedApplication2 = BoxedApplication.of(offlineApplicationEntity)

// Exposes regular application through `value`
println(boxedApplication1.value.status)
// Exposes offline application through `value`
println(boxedApplication2.value.service)

// Extract a value from either variant using `map`
val boxedApplication = randomChoice(boxedApplication1, boxedApplication2)
val applicationId = boxedApplication.map(
  ApprovedPremisesApplicationEntity::id,
  OfflineApplicationEntity::id,
)
```

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/PersonalTimelineTransformer.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/PersonalTimelineTransformer.kt
@@ -6,6 +6,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.PersonalTimeli
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ServiceName
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.TimelineEvent
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApprovedPremisesApplicationEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.OfflineApplicationEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.PersonInfoResult
 
 @Component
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/PersonalTimelineTransformer.kt
```diff
@@ -32,4 +41,41 @@ class PersonalTimelineTransformer(
...
+    timelineEvents = timelineEvents,
+  )
+}
+
+data class ApplicationTimelineModel(
+  val application: BoxedApplication,
+  val timelineEvents: List<TimelineEvent>,
+)
+
+sealed interface BoxedApplication {
```
davidatkinsuk, line 63: It's a shame that OfflineApplicationEntity a subtype of the ApplicationEntity, i'm sure there were reasons why and this is a nice way to reintroduce some polymorphism

davidatkinsuk (approved):
LGTM 👍

PR #1898, opened 2024-05-21
APS-826 - Write reports directly to servlet outputstream

davidatkinsuk:
Before this commit, reports were being written to a ByteArrayOutputStream in memory which was then being converted into an input stream to provide back to the spring framework via a ResponseEntity. This was not desirable for a few reasons:

1. This approach lead to the report being loaded into memory twice, due to the call to byteArrayOutputstream.toByteArray()
2. We couldn’t leverage any efficiencies afforded by streaming to an output stream in the underlying report generation framework (although as we currently create a full XLSX in memory, this isn’t being taken advantage of)

To enable writing to the outputstream in the report controller, we need to return a `ResponseEntity<StreamingResponseBody>` instead of `ResponseEntity<Resource>`. Because the controller function return types are inherited from code generated by the open api generator, a custom open api generator template has been added to use `ResponseEntity<StreamingResponseBody>` when the API indicates a binary file is being returned


openapi/returnTypes.mustache
```diff
@@ -0,0 +1 @@
+{{#isResponseBinary}}org.springframework.web.servlet.mvc.method.annotation.StreamingResponseBody{{/isResponseBinary}}{{^isResponseBinary}}{{#isMap}}Map<String, {{{returnType}}}>{{/isMap}}{{#isArray}}{{#reactive}}Flow{{/reactive}}{{^reactive}}{{{returnContainer}}}{{/reactive}}<{{{returnType}}}>{{/isArray}}{{^returnContainer}}{{{returnType}}}{{/returnContainer}}{{/isResponseBinary}}
```
edavey, line 1: nice 🕵️

edavey (approved):
Bravo, well done figuring that out. Very neat.

danhumphreys-moj:
Looks good from CAS 2 side of things - thanks!

- Checked that the CAS2 reporting e2e test runs successfully against this branch.
- Manually checked that the 3 x CAS2 reports can still be downloaded successfully

davidatkinsuk:
> Looks good from CAS 2 side of things - thanks!
> 
> * Checked that the CAS2 reporting e2e test runs successfully against this branch.
> * Manually checked that the 3 x CAS2 reports can still be downloaded successfully

brill, thanks Dan

davidatkinsuk:
> Looks good from CAS 2 side of things - thanks!
> 
> * Checked that the CAS2 reporting e2e test runs successfully against this branch.
> * Manually checked that the 3 x CAS2 reports can still be downloaded successfully

FYI @danhumphreys-moj we found an issue when calling these updated APIs from the CAS3 UI. Root cause was that we were no longer sending the Content-Type header. That's now fixed. When you get a moment would you mind retesting against CAS2 UI (although CAS1 still works so i'm not expecting any issues). If you're busy let me know and i'll take a look at running CAS2 UI locally myself 👍

danhumphreys-moj:
Ran the same tests again @davidatkinsuk - all good!

PR #1900, opened 2024-05-22
APS-826 remove intermediate report models

davidatkinsuk:
Add framework to stream report data to output and use it for the CAS1 Placement Matching Outcomes as a working example of usage

Currently, when constructing a report we load the entire report content into memory at multiple points:

1. In a List<A>, where A is the repository row definition
2. Then mapped into a List<B>, where B is the service layer row definition
3. Then written to an XLSX document

This commit adds a framework to ‘stream’ report output directly to its intended target. In this first version there is only the option to write to XSLX, which still results in the report being loaded into memory once (see step 3 above)

If CSV was used instead of XSLX, we could directly stream the CSV document to the servlet outputstream, ensuring we never load the full report into memory.

This change has been made in a way that it can easily be applied to all existing reports. SQL would need to be modified as follows:

1. Modify the output column names to use lower_cased_underscore_naming instead of camel case. This is required because postgres will always lowercase column names in the output. The new reporting framework will convert these back to camel case when outputing to XLSX
2. Because all values are written as String or Int types into XSLX, if specific date formatting is required this should be achieved by formatting dates to varchar/text types in the SQL itself


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/Cas1PlacementMatchingOutcomesReportTest.kt
```diff
@@ -22,16 +23,18 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.PlacementRequ
...
+import java.time.format.DateTimeFormatter
 import java.util.UUID
 
 class Cas1PlacementMatchingOutcomesReportTest : IntegrationTestBase() {
 
   object Constants {
     const val REPORT_MONTH = 1
     const val REPORT_YEAR = 2020
+
+    val DATE_FORMAT: DateTimeFormatter = DateTimeFormatter.ofPattern("dd/MM/uuuu")
```
edavey, line 37: interesting to learn about `u` versus `y`: https://stackoverflow.com/a/41178418 Is it worth a note or can we assume that Java/Kotlin devs agree on this as best practice?

edavey (approved):
very interesting, thanks for making an isolated example here with the "placement matching outcomes" report.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/Cas1PlacementMatchingOutcomesReportTest.kt
davidatkinsuk: I've got into the habit of always using uuuu as it helps spot if we accidentally use a negative year (although the likelyhood of that is very slim!). I agree that yyyy is far more understandable, but given we typically always show uuuu next to dd/MM i think it's understandable enough.

PR #1901, opened 2024-05-22
HAAR-2669: 🔧 Move to new prod nomis user roles api in cloud platform

SimonMitchellMOJ:


edavey (approved):
thanks for the update

PR #1904, opened 2024-05-23
Fix/aps 326 user name in timeline

RobBoothMOJ:
Resolves a bug in CAS1 where automatic, system generated actions are attributed to the user who was logged in and triggered them, in the timeline.

A new, nullable column, triggerSource on domainEvent will provide the ability to record whether the event was triggered by 'USER' or 'SYSTEM'.

When saving an AssessmentAllocated event, CAS1 will record the triggerSource as 'USER' if allocatingUser has a value otherwise a triggerSource of 'SYSTEM' will be recorded.

A corresponding change to the UI has been made so that the timeline will better reflect when an event was triggered by the system.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas1/Cas1AssessmentDomainEventServiceTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas1/Cas1AssessmentDomainEventServiceTest.kt
@@ -143,7 +143,7 @@ class Cas1AssessmentDomainEventServiceTest {
     }
 
     @Test
-    fun `assessmentAllocated allocating user is optional`() {
+    fun `assessmentAllocated allocating user is system`() {
       val assigneeUserStaffDetails = StaffUserDetailsFactory().produce()
       every { communityApiClient.getStaffUserDetails(assigneeUser.deliusUsername) } returns ClientResult.Success(
         HttpStatus.OK,
@@ -159,7 +159,7 @@ class Cas1AssessmentDomainEventServiceTest {
           withArg {
             Assertions.assertThat(it.data.eventDetails.allocatedBy).isNull()
           },
-          any(),
+          TriggerSourceType.SYSTEM,
         )
       }
     }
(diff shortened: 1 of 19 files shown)


RobBoothMOJ:
Resolves a bug in CAS1 where automatic, system generated actions are attributed to the user who was logged in and triggered them, in the timeline. 

A new, nullable column, triggerSource on domainEvent will provide the ability to record whether the event was triggered by 'USER' or 'SYSTEM'.

When saving an AssessmentAllocated event, CAS1 will record the triggerSource as 'USER' if allocatingUser has a value otherwise a triggerSource of 'SYSTEM' will be recorded.

A corresponding change to the UI has been made so that the timeline will better reflect when an event was triggered by the system.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas1/Cas1AssessmentDomainEventServiceTest.kt
```diff
@@ -154,6 +159,7 @@ class Cas1AssessmentDomainEventServiceTest {
           withArg {
             Assertions.assertThat(it.data.eventDetails.allocatedBy).isNull()
           },
+          any(),
```
davidatkinsuk, line 162: i think this should be checking that user is TriggerSourceType.SYSTEM, given the test name? It might be better to rename the test to 'assessmentAllocated allocating user is system'?

davidatkinsuk (approved):
Thanks Rob, code wise everything is looking really good, barring one minor comment.

Regarding the commits - i think some commits may be depending on code in subsequent commits to compile. Ideally each commit should be self-contained.

For example the 'added triggerSource to api definition for timelineEvent' introduces the triggerSource API property but that has already been referenced in a prior commit 'added triggerSource to api definition for timelineEvent' when writing to [ApplicationTimelineNoteTransformer.kt](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1904/commits/8d6d6c922c01741b7a331f0f8bc6c0b2b7f23e9b#diff-f4cba18facd43760fe06e31a9235c7b3f430aae0fbfed84f24b8af55d15ad07a).

Also i think some commits may be a little too fine grained. For example 'added triggerSource param to mock' appears to be fixing a compilation issue introduced by a priory commit. That should be included in whatever commit changed the function signature that required the change

I think you could probably squash the commits down as follows:

* First three commits into a 'Added triggerSource to DomainEventEntity' commit
* Remaining two commits into a 'Added triggerSource to api definition for TimelineEvent' commit

For context, on the first commit i'd probably mention that it defaults to null and it's currently only being set for the APPROVED_PREMISES_ASSESSMENT_ALLOCATED

Hope that makes sense 👍

PR #1905, opened 2024-05-23
Add a script to help devs directly connect to the local postgres container

tahb:
We have some documentation on how to use an IDE to help with this[1] however sometimes it is useful to have a raw SQL prompt, in the same way we do on real environments via `script/remote_db`.

This script gets us straight onto the locally running container with a PSQL REPL.

[1] https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/4750868858/API+database+reader+Connecting+to+Remote+DBs

### DIFF
--- a/script/local_db
+++ b/script/local_db
@@ -1,0 +1,15 @@
+#!/bin/sh
+
+# Check if Docker is running
+docker info > /dev/null 2>&1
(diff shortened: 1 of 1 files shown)


davidatkinsuk (approved):
This is a really nice idea, i'll definitely be using it 👍

PR #1908, opened 2024-05-24
Added 'awatingSentence' to possible values for SituationOption enum

RobBoothMOJ:
Adding a new value for 'Bail Placement' Sentence Type  - 'awaitingSentence'.
Added `awaitingSentence` as an Enum value for `SituationOption` in the API definition.

New value already covered in existing test (`ApplicationServiceTest`) which enumerates all possible values for SituationOption.


davidatkinsuk (approved):
LGTM 👍

PR #1909, opened 2024-05-24
[APS-801] Introduce `/cas1/` namespace for lost beds endpoints

danlivings-dxw:
> See [APS-801 on the Approved Premises Jira board](https://dsdmoj.atlassian.net/browse/APS-801).

This PR introduces the CAS1-specific API specification (cas1-api.yml), starting with the 'lost beds' API endpoints. These new endpoints replicate the behaviour of the existing lost beds endpoints in the `PremisesController` as applied specifically to CAS1.

The configuration for the CAS1-specific spec has the `useTags` property set to `true` to allow for controllers to be decomposed more easily (for example, by default all endpoints starting with `/premises/` would be put into the `PremisesController`).

This PR is the first of a two-stage process to implement APS-801, as removing the CAS1-specific logic from the existing endpoints requires the frontend to make the switch in all environments first. The second PR can be found here: #1911

### DIFF
--- a/src/main/resources/static/cas1-api.yml
+++ b/src/main/resources/static/cas1-api.yml
@@ -1,0 +1,224 @@
+openapi: 3.0.1
+info:
+  title: 'Community Accommodation Services: Approved Premises (CAS1)'
+  version: 1.0.0
+servers:
+  - url: /cas1
+paths:
+  /premises/{premisesId}/lost-beds:
+    post:
+      tags:
+        - lost beds
+      summary: Posts a lost bed to a specified approved premises
+      parameters:
+        - name: premisesId
+          in: path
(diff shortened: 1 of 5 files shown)


src/main/resources/static/cas1-api.yml
```diff
@@ -0,0 +1,224 @@
+openapi: 3.0.1
+info:
+  title: 'Community Accommodation Services: Tier 2 (CAS2)'
```
edavey, line 3: I think we want a CAS1 title, something like
> Community Accommodation Services: Approved Premises (CAS1)

edavey (approved):
This works really well, thanks for figuring out how to use the `tags` property in the OpenAPI tools generator.

PR #1910, opened 2024-05-24
CAS-385 Bookings have duplicated confirmation records

muhammad-elabdulla:
This [PR CAS-385](https://dsdmoj.atlassian.net/browse/CAS-385) is to fix a bug in production when a booking have two confirmation records.
- Will add Transactional around the create confirmation to make sure that it will rollback if its failed
- Remove the duplicated record for the affected booking


davidatkinsuk (approved):
LGTM 👍 

Whilst i think the general direction of travel is to use 'seed jobs' to 'fix' data in prod instead of migrations, i appreciate in this instance that is a lot of work to fix something that shouldn't happen again, so happy with this pragmatic approach.

PR #1911, opened 2024-05-24
[APS-801] Remove CAS1-specific logic from shared 'lost beds' endpoints

danlivings-dxw:
> See [APS-801 on the Approved Premises Jira board](https://dsdmoj.atlassian.net/browse/APS-801).

This PR removes the CAS1-specific logic from the lost beds endpoints found in the shared `PremisesController` and introduces an error response to alert CAS1 API consumers to the new endpoint.

This PR is the second of a two-stage process to implement APS-801, as removing the CAS1-specific logic from the existing endpoints requires the frontend to make the switch in all environments first. The first PR (#1909) should be merged into `main` before this one.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
@@ -53,7 +53,6 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.UpdateRoom
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApprovedPremisesEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.TemporaryAccommodationPremisesEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserQualification
-import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserRole
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.PersonInfoResult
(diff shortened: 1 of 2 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/LostBedsTest.kt
```diff
@@ -29,7 +29,7 @@ class LostBedsTest : InitialiseDatabasePerClassTestBase() {
   @Nested
   inner class GetAllLostBeds {
     @Test
-    fun `Get All Lost Beds without JWT returns 401`() {
+    fun `List Lost Beds without JWT returns 401`() {
```
edavey, line 32: Should these tweaks to the spec titles actually be part of the previous PR 1900 https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1909/files#diff-0a762b45881962842eff1ea864b80ee2b8c8b9923d96a176061d01c36d48b9ab

danlivings-dxw: They were an artifact of rebasing the branch for #1909 without updating this branch as well

edavey:
I was just reviewing our overall plans with @richpjames and we noted that we need to hold off merging this until we have released the UI work which switches our existing lost beds API calls over to these new `/cas1/*` endpoints.

davidatkinsuk:
@richpjames do we know if this PR is unblocked (E.g. is the UI using /cas1/* API calls for lost beds/oosb). Or is this something that can't be done until OOSB is rolled out to all regions?

davidatkinsuk:
Given that cas-1 lost-beds functionality was moved to specific cas1 endpoints (see APS-801), and those endpoints should no longer be in use, I think this is fine to merge

davidatkinsuk (approved):
LGTM. Note that the only change i made was rebasing.

PR #1912, opened 2024-05-24
Change how tests allocate wiremock ports

davidatkinsuk:
This PR changes how we assign the port to wiremock to resolve port-clashes we’re seeing when running tests concurrently in the pipeline

1. The port range is no longer limited. Instead we use ServerSocket(0) to offer up any available open port
2. We don’t ‘lock’ the port to the java process once determined (by capturing it in the WiremockPortHolder as an object variable). This avoids issues when there are multiple spring boot contexts within one java process, something which is possible when running multiple spring boot integration tests with differing spring configuration
3. Instead of stopping wiremock after each test, we instead leave it running and instead reset it’s state. Whilst not strictly required because the port is ‘locked’ by the shared file, this ensures the port is also always in use so can’t be taken/considered. Leave wiremock running may also have a minor performance improvement


edavey (approved):
Thanks for tackling this and breaking out the port configuration into a separate class.

PR #1913, opened 2024-05-25
CAS-327 Repoint notes to assessments - part 1

danhumphreys-moj:
This PR is the work required to carry out the first part of the task to repoint notes to assessments as described in:
https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/4897341539/Associate+Application+Notes+with+Assessments+CAS2-327

**Database**

Since we cannot achieve our desired status in one step since we need to ensure that we don’t break the front end, we need to introduce an interim database schema which we can use to ensure continuity across both the front and back ends.  The interim schema will be:

image

This means temporarily having both the existing `application_id` field and the new `assessment_id` field in the database until the front end is changed to use a new endpoint.  Swapping the front end functionality so that the `cas_2_assessments.id` to `cas_2_application_notes.assessment_id` relationship can be created for new applications and then removing the `cas_2_applications` to cas_2_application_notes relationship.

_Tasks_

The specific database related tasks required to achieve this interim schema are:

- Create a nullable assessment_id UUID field in cas_2_application_notes table.
- Create PK-FK relationship and between cas_2_assessments.id and cas_2_application_notes.assessment_id.
- Create an index on cas_2_application_notes.assessment_id.

**Data Model**

There is a requirement for an interim model where the `Cas2ApplicationNoteEntity` will both `application` and `assessment` fields.  Once the front end has been updated, the application field can be removed from this data class.  The interim `Cas2ApplicationNoteEntity` that will contain both fields is:

image

The assessment field will be of type `Cas2AssessmentEntity`.

_Tasks_

Specific tasks related to the data model are:

- Add a new nullable field assessment to `Cas2ApplicationNoteEntity` of type `Cas2AssessmentEntity`; join the column to `assessment_id` and annotate `@ManyToOne`.

**Endpoints**

Similarly to the database and data model, we will temporarily need an interim solution where both endpoints will be specified in the OpenAPI specification:

`/submissions/{applicationId}/notes `- existing endpoint

`/assessments/{assessmentId}/notes` - new endpoint

OAuth roles that can access this endpoint are defined in `OAuth2ResourceServerSecurityConfiguration`, the interim solution will require that roles have access to both endpoints:

`authorize(HttpMethod.POST, "/cas2/submissions/*/notes", hasAnyRole("LICENCE_CA", "POM", "CAS2_ASSESSOR"))`

`authorize(HttpMethod.POST, "/cas2/assessments/*/notes", hasAnyRole("LICENCE_CA", "POM", "CAS2_ASSESSOR"))
`

_Tasks_

- Edit `OAuth2ResourceServerSecurityConfiguration` to give new permissions to `/cas2/assessments/*/notes` endpoints, permissions should be as per the existing endpoint.
- Create new `/assessments/{assessmentId}/notes` POST endpoint in `cas2-api.yml`.  The parameters, request body and response will be as per the existing POST endpoint (`/submissions/{applicationId}/notes`).

**Controllers**

As per other aspects of this work the will be an ‘interim’ controller state which is required.  A new `AssessmentsController.assessmentsAssessmentIdNotesPost()` method will be required, this method will override the method auto generated in the delegate interface `AssessmentsCas2Delegate`.  There will be a controller method to post notes in the `SubmissionsController` (existing) and a new controller method in the `AssessmentsController`, and a new method in the `ApplicationsNoteService`.  Later the legacy methods can be deleted once the front end has been migrated to use the assessments endpoint.

The `assessmentsAssessmentIdNotesPost` method will call a new `createAssessmentNote()` method in the `ApplicationNoteService`.  The method `createAssessmentNote()` is essentially as per the existing `createApplicationNote()`, with the difference being it is passed an `assessment_id` rather than an `application_id`. 

This new controller and service method will essentially exist at the same time as the `submissionsApplicationIdNotesPost()` and `createApplicationNote()` until the front end can be updated to use the new endpoints.  These methods will be annotated `@Deprecated`.

New integration tests will need to be created to test the new controller method based on the existing coverage for of `submissionsApplicationIdNotesPost()` and the new service method will require unit tests.

This work will enable us to check that the `cas_2_application_notes.assessment_id` is being populated correctly and ensure that the `AssessmentsController` and `Cas2ApplicationNoteEntity` are working as expected, and will then allow us to do the data migration before changing the front end.

This new controller functionality state will be therefore be as follows:

image

_Tasks_

The specific tasks related to this work are:

- Create new service method `createAssessmentNote()` which is as per `createApplicationNote()` but will take the `assessment_id` as a method argument.  This method will look up the application based on the `assessment_id` supplied.
- Mark `createApplicationNote()` as deprecated.
- Write unit tests for new `createAssessmentNote()` coverage based on existing coverage.  New inner test class will be `CreateAssessmentNote` and test methods will be as per `CreateApplicationNote` methods.
- Create `createAssessmentNote()` unit tests to test when assessment is `AuthorisableActionResult.NotFound()`.
- Rename (refactor) `ApplicationNoteService` class to `AssessmentNoteService`.
- Create body of `AssessmentsController.assessmentsAssessmentIdNotesPost()` which is as per `SubmissionsController.submissionsApplicationIdNotesPost()` but will call `createAssessmentNote()`.
- Mark now legacy controller method `submissionsApplicationIdNotesPost()`  as deprecated.
- Create new integration tests to test new controller assessment notes functionality.  Create new class `Cas2AssessmentNotesTest` which will be as per existing tests in `Cas2ApplicationNotesTest` but calling the new assessments endpoint.
- Refactor `Cas2ApplicationsNoteTest.kt `as `Cas2AssessmentsNoteTest.kt`.
- Pass `assessment` to `saveNote()` for both `createApplicationNote()` and `createAssessmentNote()` and update associated tests.

**Data Migration**

_Tasks_

- Add new migration job `Cas2NoteMigrationJob.kt` which will be similar to `Cas2StatusUpdateMigrationJob` and will populate assessment ids for status notes that only have application ids.  A join will be required via the `cas_2_applications` table to determine the `assessment_id` for each particular `application_id`.  At this point all historical status updates will all have a populated `assessment_id`. 
- Add the new job to `MigrationJobService.kt`.
- Add `Cas2NoteEntityFactory` to `produce()` example `Cas2NoteEntity` objects which are required for the migration integration test.  Also needs to be added to `IntegrationTestBase`.
- Add integration test` Cas2MigrateNotesTest.kt` to test migration job functionality.
- Add additional checks to test that the right number of records has been affected.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/migration/Cas2NoteMigrationJob.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/migration/Cas2NoteMigrationJob.kt
@@ -1,0 +1,50 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.migration
+
+import org.slf4j.LoggerFactory
+import org.springframework.data.domain.PageRequest
+import org.springframework.data.domain.Slice
+import org.springframework.transaction.support.TransactionTemplate
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas2ApplicationNoteEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas2ApplicationNoteRepository
+
+class Cas2NoteMigrationJob(
+  private val noteRepository: Cas2ApplicationNoteRepository,
+  private val transactionTemplate: TransactionTemplate,
+  private val pageSize: Int,
+) : MigrationJob() {
+  private val log = LoggerFactory.getLogger(this::class.java)
+  override val shouldRunInTransaction = true
+
+  override fun process() {
+    log.info("Starting CAS2 note migration process...")
+
+    var hasNext = true
+    var slice: Slice<Cas2ApplicationNoteEntity>
+    var page = 1
+    var notesWithoutAssessment = noteRepository.countAllNotesWithoutAssessment()
+    val notesWithAssessment = noteRepository.countAllNotesWithAssessment()
+
+    while (hasNext) {
+      log.info("Getting page $page for max page size $pageSize")
+      slice = noteRepository.findAllNotesWithoutAssessment(PageRequest.of(0, pageSize))
+      slice.content.forEach { note ->
+        transactionTemplate.executeWithoutResult {
+          log.info("Adding assessment to note ${note.id}")
+          note.assessment = note.application.assessment
+          noteRepository.save(
+            note,
+          )
+        }
+      }
+      hasNext = slice.hasNext()
+      page += 1
+    }
+    log.info("Before the migration:")
+    log.info("notes without assessments: $notesWithoutAssessment")
+    log.info("notes with assessments: $notesWithAssessment")
+    log.info("After the migration:")
+    log.info("notes without assessments: ${noteRepository.countAllNotesWithoutAssessment()}")
+    log.info("notes with assessments: ${noteRepository.countAllNotesWithAssessment()}")
+    log.info("CAS2 note migration process complete!")
+  }
+}
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/AssessmentNoteService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/AssessmentNoteService.kt
@@ -28,7 +28,7 @@ import java.time.OffsetDateTime
 import java.util.UUID
 
 @Service("Cas2ApplicationNoteService")
-class ApplicationNoteService(
+class AssessmentNoteService(
   private val applicationRepository: Cas2ApplicationRepository,
   private val assessmentRepository: Cas2AssessmentRepository,
   private val applicationNoteRepository: Cas2ApplicationNoteRepository,
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/AssessmentNoteService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/AssessmentNoteService.kt
@@ -45,8 +45,12 @@ class AssessmentNoteService(
   private val log = LoggerFactory.getLogger(this::class.java)
 
   @Deprecated("Superseded by createAssessmentNote().")
+  @SuppressWarnings("ReturnCount")
   fun createApplicationNote(applicationId: UUID, note: NewCas2ApplicationNote):
     AuthorisableActionResult<ValidatableActionResult<Cas2ApplicationNoteEntity>> {
+    val assessment = assessmentRepository.findFirstByApplicationId(applicationId)
+      ?: return AuthorisableActionResult.NotFound()
+
     val application = applicationRepository.findByIdOrNull(applicationId)
       ?: return AuthorisableActionResult.NotFound()
 
@@ -63,7 +67,7 @@ class AssessmentNoteService(
       return AuthorisableActionResult.Unauthorised()
     }
 
-    val savedNote = saveNote(application, note.note, user)
+    val savedNote = saveNote(application, assessment, note.note, user)
 
     sendEmail(isExternalUser, application, savedNote)
 
@@ -96,7 +100,7 @@ class AssessmentNoteService(
       return AuthorisableActionResult.Unauthorised()
     }
 
-    val savedNote = saveNote(application, note.note, user)
+    val savedNote = saveNote(application, assessment, note.note, user)
 
     sendEmail(isExternalUser, application, savedNote)
 
@@ -202,13 +206,14 @@ class AssessmentNoteService(
     }
   }
 
-  private fun saveNote(application: Cas2ApplicationEntity, body: String, user: Cas2User): Cas2ApplicationNoteEntity {
+  private fun saveNote(application: Cas2ApplicationEntity, assessment: Cas2AssessmentEntity, body: String, user: Cas2User): Cas2ApplicationNoteEntity {
     val newNote = Cas2ApplicationNoteEntity(
       id = UUID.randomUUID(),
       application = application,
       body = body,
       createdAt = OffsetDateTime.now(),
       createdByUser = user,
+      assessment = assessment,
     )
 
     return applicationNoteRepository.save(newNote)
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ApplicationNotesTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ApplicationNotesTest.kt
@@ -553,3 +553,542 @@ class Cas2ApplicationNotesTest(
     }
   }
 }
+
+class Cas2AssessmentNotesTest(
+  @Value("\${url-templates.frontend.cas2.application-overview}") private val applicationUrlTemplate: String,
+  @Value("\${url-templates.frontend.cas2.submitted-application-overview}") private val assessmentUrlTemplate: String,
+) : IntegrationTestBase() {
+
+  @SpykBean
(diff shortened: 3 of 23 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/migration/Cas2NoteMigrationJob.kt
```diff
@@ -0,0 +1,50 @@
...
+          noteRepository.save(
+            note,
+          )
+        }
+      }
+      hasNext = slice.hasNext()
+      page += 1
+    }
+    log.info("Before the migration:")
+    log.info("notes without assessments: $notesWithoutAssessment")
```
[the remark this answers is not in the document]
davidatkinsuk: > Are we confident that log.info will work on jobs? I remember the cache warming jobs didn't work when running on a real environment, though it might have been fixed since @davidatkinsuk was asking about this recently.

I haven't changed anything in this area but i'm pretty sure info logging is only an issue when the thread wasn't spawned by an HTTP/API request. So in this case i think logging should work fine

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ApplicationNotesTest.kt
```diff
@@ -553,3 +553,542 @@ class Cas2ApplicationNotesTest(
     }
   }
 }
+
+class Cas2AssessmentNotesTest(
```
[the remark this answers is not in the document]
danhumphreys-moj: I think it's ok in this file for now, since when we get to removing the legacy code after the front end has been updated `Cas2AssessmentNotesTest` will be the only tests in this file we can rename it to `Cas2AssessmentNotesTest.kt`.  We will be removing the `Cas2AssessmentNotesTest` class which test the (to be) legacy endpoint `submissions/$applicationId/notes`.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas2/ApplicationNoteServiceTest.kt
```diff
@@ -39,6 +40,7 @@ import java.util.UUID
 
 class ApplicationNoteServiceTest {
   private val mockApplicationRepository = mockk<Cas2ApplicationRepository>()
+  private val mockAssessmentRepository = mockk<Cas2AssessmentRepository>()
```
Harriethw, line 43: are there unit tests for the new service method? Quite minor, but could they be moved to the same commit that adds the method, so this is a 'complete'/atomic commit?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas2/ApplicationNoteServiceTest.kt
```diff
@@ -463,6 +463,21 @@ class ApplicationNoteServiceTest {
       }
     }
 
+    @Nested
+    inner class WhenAssessmentIsNotFound {
+      @Test
+      fun `returns Not Found`() {
```
Harriethw, line 469: ah yep I see the unit tests are here :)
It's a minor bit of feedback, but might help in future when you're organising work/git history to think about what makes an atomic commit - I think the general rule is that each commit should be able to build and run on its own, which means the tests should be in the same commit imo.  (not saying that I'm always perfect at this 😆 )

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/AssessmentNoteService.kt
```diff
@@ -70,6 +74,39 @@ class ApplicationNoteService(
...
+    }
+
+    val isExternalUser = httpAuthService.getCas2AuthenticatedPrincipalOrThrow().isExternalUser()
+    val user = getCas2User(isExternalUser)
+
+    if (!isExternalUser && !nomisUserCanAddNote(application, user as NomisUserEntity)) {
+      return AuthorisableActionResult.Unauthorised()
+    }
+
+    val savedNote = saveNote(application, note.note, user)
```
Harriethw, line 99: I might have missed something in the docs, but do we not need to start saving the assessment on a note from this point on? So that any new Notes created have an Assessment ID.

Harriethw: and we need to start doing it for both `createAssessmentNote` and `createApplicationNote` so that the deprecated end point currently in use starts saving Assessments on a new Note too, and we can safely do the migration to catch the legacy ones

tahb: Good catch! We duplicate `createAssessmentNote()` but I didn't notice the `saveNote()` actually does the same thing in both cases. Though the commit points to this being a deliberate step I'm not sure why, this does seem to be the critical part.

danhumphreys-moj: Yes - without this it doesn't work.  I added a note manually yesterday afternoon and expected to see the assessment id being populated but it wasn't.  Then saw assessment was the missing from saveNote() as a parameter.

1. Made me think there is a missing integration test somewhere?  i.e. create a note, then check that the assessment id is returned.  Couldn't see where this test exists or whether we need to create it?

3. Also checking that passing `assessment` to `saveNote()` in requires the following in `AssessmentNoteService.createApplicationNote()`?:
```
val assessment = assessmentRepository.findFirstByApplicationId(applicationId)
      ?: return AuthorisableActionResult.NotFound()
```

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas2/ApplicationNoteServiceTest.kt
danhumphreys-moj: Have squashed these commits into one now.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas2/ApplicationNoteServiceTest.kt
danhumphreys-moj: Squashed into one now.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/AssessmentNoteService.kt
danhumphreys-moj: I've done this in a recent commit and updated the necessary tests.  Hopefully the right way to do it, and will only be used temporarily until the front end is changed (which will be shortly after).

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/AssessmentNoteService.kt
```diff
@@ -45,8 +45,12 @@ class AssessmentNoteService(
   private val log = LoggerFactory.getLogger(this::class.java)
 
   @Deprecated("Superseded by createAssessmentNote().")
+  @SuppressWarnings("ReturnCount")
   fun createApplicationNote(applicationId: UUID, note: NewCas2ApplicationNote):
     AuthorisableActionResult<ValidatableActionResult<Cas2ApplicationNoteEntity>> {
+    val assessment = assessmentRepository.findFirstByApplicationId(applicationId)
+      ?: return AuthorisableActionResult.NotFound()
```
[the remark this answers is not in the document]
Harriethw: In this case, would it be a bit safer to get the assessment via the application that you get below? 
`application.assessment`

Harriethw (approved):
thanks for such a comprehensively documented PR!

PR #1917, opened 2024-05-28
Cas2-216: Abandon an in-progress application

Harriethw:
Given the referrer has an in progress application
When they withdraw the application
Then it will no longer appear on their in progress dashboard

When they try to request, submit or update an abandoned application
Then the service will not permit it

## Changes in this PR
https://dsdmoj.atlassian.net/browse/CAS2-216 

UI PR here: https://github.com/ministryofjustice/hmpps-community-accommodation-tier-2-ui/pull/727

We want to allow users to flag in progress applications as abandoned. In order to do this we will add an `abandonedAt` value to CAS2 Applications - similar to the `submitted_at` value, this timestamp will determine whether an application has been abandoned or not.

We chose to use the terminology of 'abandoned' because 'withdrawal' or 'cancel' could apply to applications post-submission rather than in progress applications. 

Once it has an abandoned timestamp, users should not be able to submit, update or view the application.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/ApplicationsController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/ApplicationsController.kt
@@ -136,6 +136,26 @@ class ApplicationsController(
     return ResponseEntity.ok(getPersonDetailAndTransform(updatedApplication))
   }
 
+  @Transactional
+  override fun applicationsApplicationIdAbandonPut(applicationId: UUID): ResponseEntity<Unit> {
+    val user = userService.getUserForRequest()
+
+    val validationResult = when (val applicationResult = applicationService.abandonApplication(applicationId, user)) {
+      is AuthorisableActionResult.NotFound -> throw NotFoundProblem(applicationId, "Application")
+      is AuthorisableActionResult.Unauthorised -> throw ForbiddenProblem()
+      is AuthorisableActionResult.Success -> applicationResult.entity
+    }
+
+    when (validationResult) {
+      is ValidatableActionResult.GeneralValidationError -> throw BadRequestProblem(errorDetail = validationResult.message)
+      is ValidatableActionResult.FieldValidationError -> throw BadRequestProblem(invalidParams = validationResult.validationMessages)
+      is ValidatableActionResult.ConflictError -> throw ConflictProblem(id = validationResult.conflictingEntityId, conflictReason = validationResult.message)
+      is ValidatableActionResult.Success -> validationResult.entity
+    }
+
+    return ResponseEntity.ok(Unit)
+  }
+
   private fun getPersonNamesAndTransformToSummaries(applicationSummaries: List<Cas2ApplicationSummaryEntity>):
     List<uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Cas2ApplicationSummary> {
     val crns = applicationSummaries.map { it.crn }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/ApplicationService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/ApplicationService.kt
@@ -204,6 +204,12 @@ class ApplicationService(
       return AuthorisableActionResult.Unauthorised()
     }
 
+    if (application.abandonedAt != null) {
+      return AuthorisableActionResult.Success(
+        ValidatableActionResult.GeneralValidationError("This application has already been abandoned"),
+      )
+    }
+
     if (application.submittedAt != null) {
       return AuthorisableActionResult.Success(
         ValidatableActionResult.GeneralValidationError("This application has already been submitted"),
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/ApplicationService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/ApplicationService.kt
@@ -100,6 +100,10 @@ class ApplicationService(
     val applicationEntity = applicationRepository.findByIdOrNull(applicationId)
       ?: return AuthorisableActionResult.NotFound()
 
+    if (applicationEntity.abandonedAt != null) {
+      return AuthorisableActionResult.NotFound()
+    }
+
     val canAccess = userAccessService.userCanViewApplication(user, applicationEntity)
 
     return if (canAccess) {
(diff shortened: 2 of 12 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ApplicationAbandonTest.kt
```diff
@@ -0,0 +1,156 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.integration.cas2
+
+import com.ninjasquad.springmockk.SpykBean
```
Harriethw, line 3: I've split these tests out from `Cas2ApplicationTest` as that file is getting so big its unreadable imo...I think there's a bit of precedence with this in the other CAS tea,s with the `Person` integration tests.

src/main/resources/db/migration/all/20240528100000__create_cas_2_summary_views_add_abandoned.sql
```diff
@@ -0,0 +1,47 @@
...
+    a.created_at,
+    a.submitted_at,
+    a.hdc_eligibility_date,
+    a.label,
+    a.status_id,
+    a.referring_prison_code,
+    a.abandoned_at
+FROM cas_2_application_summary a
+WHERE (a.conditional_release_date IS NULL OR a.conditional_release_date >= current_date)
+AND a.abandoned_at IS NULL
```
Harriethw, line 40: just realised I don't have integration tests to check that abandoned applications are not returned

Harriethw: added check in `Cas2ApplicationTest` which I think is sufficient to check the logic has been implemented?

danhumphreys-moj (approved):
Looks good to me!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/ApplicationService.kt
```diff
@@ -192,6 +192,40 @@ class ApplicationService(
...
+
+    if (application.abandonedAt != null) {
+      return AuthorisableActionResult.Success(
+        ValidatableActionResult.Success(application),
+      )
+    }
+
+    application.apply {
+      this.abandonedAt = OffsetDateTime.now()
+      this.data = null
```
tahb, line 219: Just checking that this won't break the unsubmitted application report? If I remember rightly we should have pulled everything we need out into first class columns, avoiding any json digging?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/ApplicationsController.kt
```diff
@@ -136,6 +136,26 @@ class ApplicationsController(
     return ResponseEntity.ok(getPersonDetailAndTransform(updatedApplication))
   }
 
+  @Transactional
```
tahb, line 139: On this commits message:

> it's not a hard
DELETE request because the Entity will remain in
our db, but its also not really an UPDATE request
because the entity is being deleted from the UI.

Agree on it not being a DELETE but I do think it is an UPDATE request at least as far as the API is concerned. If I recall we hit problems when trying to reuse the existing update endpoint, it was set up as a PUT instead of a PATCH despite it modifying only the `data` field. That meant we couldn't easily extend it to handle our new type of update. 

Instead of initiating _another_ migration for that endpoint given our remaining time, we decided to bring up a new action and leave the old for another day. If we'd had more time we could have converted that endpoint into a PATCH to keep it more conventional, changed the frontends request and then extended the action with our new behaviour.

This context may be worth throwing into the commit if I've remembered it right.

tahb: Mind you, having now read the contents of the action below it has enough logic to likely merit extracting into it's own action anyway, for simplicity, readability etc.

tahb (approved):
Great work noticing and covering all of the corner cases here! In particular the idea to wipe the `data` from an application.

LGTM aside from some minor comments :)

src/main/resources/db/migration/all/20240528100000__create_cas_2_summary_views_add_abandoned.sql
```diff
@@ -0,0 +1,47 @@
+-- replace application summary views with new abandoned_at column
```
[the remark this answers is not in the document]
Harriethw: yeah :(

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/ApplicationService.kt
Harriethw: hmm it doesn't break the report but it doesn't filter out the abandoned applications from the report...that would [have to happen here](https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/b7d5dec5f0b7ef046d14ec9cc5205c3d6fc10cf4/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas2UnsubmittedApplicationsReport.kt).

To avoid holding this up, shall I make a new subtask on this ticket and handle any MI logic in a new PR?
Do we want them removed from the MI report, or do we want them included but with a column for whether they were withdrawn?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/ApplicationService.kt
```diff
@@ -204,6 +204,12 @@ class ApplicationService(
       return AuthorisableActionResult.Unauthorised()
     }
 
+    if (application.abandonedAt != null) {
+      return AuthorisableActionResult.Success(
+        ValidatableActionResult.GeneralValidationError("This application has already been abandoned"),
```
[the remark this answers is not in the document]
Harriethw: I think its the other way round - this one is the one where we are trying to abandon it, the one at top is where we are trying to submit

src/main/resources/db/migration/all/20240528100000__create_cas_2_summary_views_add_abandoned.sql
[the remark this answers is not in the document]
Harriethw: I think the saving grace is that our integration tests would fail if a column were left out, but yep it isn't ideal

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/ApplicationService.kt
tahb: [As mentioned on Slack](https://mojdt.slack.com/archives/C04T6SC31HA/p1717064275466769?thread_ts=1717063601.939889&cid=C04T6SC31HA) my thoughts are in agreement that we keep them in the submitted report and create a follow-up task to add a column which shares the abandoned state with Suzie 👍

PR #1919, opened 2024-05-29
APS-835 Add Seed Job to Update a CAS1 Application Event Number

davidatkinsuk:
This commit adds a seed job that can be used to update the event number and associated offence id/conviction id for a given application.

It will also update the event number in any Application Assessed, Application Submitted or Booking Made domain events (these are the only domain events for which probation-integration needs a valid deliusEventNumber value)

This job has been created to allow us to recover from situations where the event number associated with an application do not exist in delius (i.e. they have been deleted in delius). In this case domain event processing in probation-integration will fail. Once the event number has been updated for an application and its associated domain event's, those domain events can be replayed using the Cas1DomainEventReplaySeedJob, and they should then be processed by probation-integration without issue.

Because the seed job reads the complete domain event json into objects, updates the event number and then rewrites the json from these objects, care needs to be taken to ensure that there is no data loss caused by changes to the Domain Event's schema since the JSON was originally written to the Database (although historically, the Domain Event JSON Schema is very stable)

Test evidence:

Given the seed CSV:

```
application_id,event_number,offence_id,conviction_id
"0f771648-bf12-4e4d-8111-14e03085849b","5454","2323","4545"
```

After applying seed job:

Screenshot 2024-05-29 at 15 46 43

```
approved_premises_localdev=# select event_number, offence_id, conviction_id from approved_premises_applications where id = '0f771648-bf12-4e4d-8111-14e03085849b';
 event_number | offence_id | conviction_id 
--------------+------------+---------------
 5454         | 2323       |          4545
```

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApplicationEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApplicationEntity.kt
@@ -242,6 +242,18 @@ WHERE taa.probation_region_id = :probationRegionId AND a.submitted_at IS NOT NUL
     type: Class<T>,
     pageable: Pageable?,
   ): Slice<TemporaryAccommodationApplicationEntity>
+
+  @Modifying
+  @Query(
+    """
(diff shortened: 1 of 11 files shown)


edavey (approved):
very thorough, thanks for all the testing of domain event changes. 👍

PR #1920, opened 2024-05-29
[APS-805] Introduce database model for CAS1 out-of-service beds

danlivings-dxw:
> See [APS-805 on the Approved Premises Jira board](https://dsdmoj.atlassian.net/browse/APS-805).

This PR introduces the `cas1_out_of_service_beds` and `cas1_out_of_service_bed_reasons` tables and associated JPA entities as an intended future replacement for the concept of lost beds within CAS1.

Currently, the data models for these are very close to the existing lost beds model, although we expect these that they will diverge significantly as we implement Match and Manage functionality.

A migration job is provided that copies all CAS1-appropriate `lost_bed_reasons` to the `cas1_out_of_service_bed_reasons` table, which will simplify the process of migrating the existing lost beds when it becomes necessary.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas1OutOfServiceBedEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas1OutOfServiceBedEntity.kt
@@ -1,0 +1,36 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity
+
+import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.stereotype.Repository
+import java.time.LocalDate
+import java.time.OffsetDateTime
+import java.util.UUID
+import javax.persistence.Entity
+import javax.persistence.Id
+import javax.persistence.JoinColumn
+import javax.persistence.ManyToOne
+import javax.persistence.Table
+
+@Repository
+interface Cas1OutOfServiceBedRepository : JpaRepository<Cas1OutOfServiceBedEntity, UUID>
+
+@Entity
+@Table(name = "cas1_out_of_service_beds")
+data class Cas1OutOfServiceBedEntity(
+  @Id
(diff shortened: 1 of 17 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas1OutOfServiceBedEntity.kt
```diff
@@ -0,0 +1,36 @@
...
+  @ManyToOne
+  @JoinColumn(name = "premises_id")
+  val premises: ApprovedPremisesEntity,
+  @ManyToOne
+  @JoinColumn(name = "out_of_service_bed_reason_id")
+  val reason: Cas1OutOfServiceBedReasonEntity,
+  @ManyToOne
+  @JoinColumn(name = "bed_id")
+  val bed: BedEntity,
+  val createdAt: OffsetDateTime,
```
edavey, line 31: how about using the ` @CreationTimestamp` annotation here to set this automatically? 

We're using this on several entities already. I tend to think that having automatic `created_at` and `updated_at` timestamps should be standard hygiene on all tables and would be v happy to see that here.

https://www.baeldung.com/hibernate-creationtimestamp-updatetimestamp

edavey (approved):
🚀 thanks, this looks ideal Thanks for the data migration job to be triggered on demand. I've suggested using automatic timestamps on the `cas1_out_of_service_beds`.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas1OutOfServiceBedEntity.kt
danlivings-dxw: I think we'd still have to pass in an `OffsetDateTime` when constructing an instance anyway, so I'm not sure what we gain to make up for making the origin of the timestamp less explicit.

PR #1921, opened 2024-05-29
Added metadata for application withdrawn domain event

RobBoothMOJ:
When a APPROVED_PREMISES_PLACEMENT_APPLICATION_WITHDRAWN domain event is created, capture the corresponding placement application’s id (UUID) in domain event metadata as CAS1_PLACEMENT_APPLICATION_ID

This is required for an efficient JOIN in updated reports

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1PlacementApplicationDomainEventService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1PlacementApplicationDomainEventService.kt
@@ -14,6 +14,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.Request
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.RequestForPlacementType
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.ClientResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.CommunityApiClient
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.MetaDataName
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.PlacementApplicationEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.PlacementType
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserEntity
@@ -129,6 +130,9 @@ class Cas1PlacementApplicationDomainEventService(
           timestamp = eventOccurredAt,
           eventType = EventType.placementApplicationWithdrawn,
           eventDetails = eventDetails,
+        ),
+        metadata = mapOf(
+          MetaDataName.CAS1_PLACEMENT_APPLICATION_ID to placementApplication.id.toString(),
         ),
       ),
     )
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1ApplicationDomainEventService.kt
```diff
@@ -139,6 +139,9 @@ class Cas1ApplicationDomainEventService(
           eventType = EventType.applicationWithdrawn,
           eventDetails = getApplicationWithdrawn(application, withdrawingUser, eventOccurredAt),
         ),
+        metadata = mapOf(
+          MetaDataName.CAS1_PLACEMENT_APPLICATION_ID to application.id.toString(),
```
davidatkinsuk, line 143: It's confusing (and we want to change names at some point), but Application and PlacementApplication are different things, so this metadata needs to be captured in the Cas1PlacementApplicationDomainEventService and not here

davidatkinsuk (changes requested):
Hi @RobBoothMOJ , i've added a comment as i think we're capturing this in a wrong place.

Thanks for the really helpful description in the PR. I think this would be really helpful to also capture in the commit so we have this context available in the commit history.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1ApplicationDomainEventService.kt
RobBoothMOJ: Moved the change to the correct Service.

davidatkinsuk (approved):
Just a few minor comments from me to address, then happy for this to be merged 👍

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas1/Cas1PlacementApplicationDomainEventServiceTest.kt
```diff
@@ -193,6 +194,8 @@ class Cas1PlacementApplicationDomainEventServiceTest {
             assertThat(it.crn).isEqualTo(CRN)
             assertThat(it.nomsNumber).isEqualTo(application.nomsNumber)
             assertThat(it.occurredAt).isWithinTheLastMinute()
+            assertThat(it.metadata.containsKey(MetaDataName.CAS1_PLACEMENT_APPLICATION_ID))
+            assertThat(it.metadata[MetaDataName.CAS1_PLACEMENT_APPLICATION_ID].toString() == it.applicationId.toString())
```
davidatkinsuk, line 198: we'll need a second function call to apply an assertion e.g. containsEntry(key,value)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1PlacementApplicationDomainEventService.kt
```diff
@@ -130,6 +131,9 @@ class Cas1PlacementApplicationDomainEventService(
           eventType = EventType.placementApplicationWithdrawn,
           eventDetails = eventDetails,
         ),
+        metadata = mapOf(
+          MetaDataName.CAS1_PLACEMENT_APPLICATION_ID to application.id.toString(),
+        ),
```
davidatkinsuk, line 136: I think this should be placementApplication.id.String()

PR #1924, opened 2024-05-30
Improve CAS2 Status Update Test Stability

davidatkinsuk:
Prior to this commit we were checking for a specific time value in email personalisation, but if the test started and finished in different minutes of the hour this test could fail.

A better solution would be to manage a mutable/fixed clock during integration tests so we can ensure the system clock used in code reports a known date-time, but that is a non-trivial change

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2StatusUpdateTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2StatusUpdateTest.kt
@@ -2,7 +2,10 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.integration.cas2
 
 import com.ninjasquad.springmockk.SpykBean
 import io.mockk.clearMocks
-import org.assertj.core.api.Assertions
+import io.mockk.every
+import io.mockk.mockkStatic
+import io.mockk.unmockkAll
+import org.assertj.core.api.Assertions.assertThat
 import org.junit.jupiter.api.AfterEach
 import org.junit.jupiter.api.Nested
 import org.junit.jupiter.api.Test
@@ -92,8 +95,8 @@ class Cas2StatusUpdateTest(
             withId(assessmentId)
           }
 
-          Assertions.assertThat(realStatusUpdateRepository.count()).isEqualTo(0)
-          Assertions.assertThat(realStatusUpdateDetailRepository.count()).isEqualTo(0)
+          assertThat(realStatusUpdateRepository.count()).isEqualTo(0)
+          assertThat(realStatusUpdateDetailRepository.count()).isEqualTo(0)
 
           webTestClient.post()
             .uri("/cas2/assessments/$assessmentId/status-updates")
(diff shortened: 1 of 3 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2StatusUpdateTest.kt
```diff
@@ -212,41 +214,45 @@ class Cas2StatusUpdateTest(
...
+            assertThat(appliedStatus.statusDetails?.find { detail -> detail.id == persistedStatusDetailUpdate?.statusDetailId })
               .isNotNull()
 
             emailAsserter.assertEmailsRequestedCount(1)
-            emailAsserter.assertEmailRequested(
+            val email = emailAsserter.assertEmailRequested(
               toEmailAddress = applicant.email!!,
               templateId = "ef4dc5e3-b1f1-4448-a545-7a936c50fc3a",
               personalisation = mapOf(
                 "applicationStatus" to "Offer declined or withdrawn",
```
Harriethw, line 242: looks like I left a stray comment in here that can be deleted - not sure what it meant by `This needs to be something else`, maybe I had meant to come back and look at the datetime stuff!

davidatkinsuk: > looks like I left a stray comment in here that can be deleted - not sure what it meant by `This needs to be something else`, maybe I had meant to come back and look at the datetime stuff!

Thanks for flagging it, i've removed it 👍

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2StatusUpdateTest.kt
```diff
@@ -212,41 +214,45 @@ class Cas2StatusUpdateTest(
...
+                "dateStatusChanged" to submittedAt.toLocalDate().toCas2UiFormat(),
                 "nomsNumber" to "123NOMS",
                 "applicationType" to "Home Detention Curfew (HDC)",
                 "applicationUrl" to applicationOverviewUrlTemplate.replace("#id", application.id.toString()),
               ),
               replyToEmailId = notifyConfig.emailAddresses.cas2ReplyToId,
             )
 
+            assertThat(email.personalisation["timeStatusChanged"] as String).isIn(
+              requestMadeAt.toCas2UiFormattedHourOfDay(),
```
Harriethw, line 252: Is this checking the time on the email against two timestamps - one made at the start of the test, and one at the point that this assertion takes place? So if the test was really lagging it might have happened minutes away from  the second `.now()` as well? 

It looks like its quite easy to mock OffsetDateTime.now(), if I understand what Dan has done here, so I wonder if it would be worth doing that?  https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/d4640a3f832030380bc42307871fc79d2a7806e9/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ApplicationTest.kt#L246

davidatkinsuk: That's correct, when it failed for me I suspect the test started on the 59th second of a minute and ended in the 0th second of the next minute. Whilst i'd be surprised if the test took > 60 seconds to run, I agree that what i've done here isn't water tight and not particularly readable either.

I agree mocking statics is a better and more readable solution so i've updated the commit to do that instead. Ideally we'd use a 'clock' in the code that we can then mutate during integration tests but that's a bit more effort to setup.

I had to introduce a try/finally block to make sure we tidy the static mock up, sorry if the diff is a bit hard to read!

Harriethw (approved):
thanks for this improvement!

PR #1926, opened 2024-05-30
APS-837 Refactor ReportJdbcTemplate to consume ResultSets

davidatkinsuk:
In preparation for adding the option to write reports to CSVs, we need to add the ability for the component consuming the report SQL output to deal directly with the ResultSet. This allows us to leverage open CSVs ability to write a CSV document direct from the result set.

This commit refactors the existing Excel Jdbc Consumer to deal with result sets directly instead of implementing the now removed JdbcReportConsumer. This change actually reduces the number of callbacks and interfaces so hopefully results in code that’s easier to read.


RobBoothMOJ (approved):
Approved

PR #1927, opened 2024-05-31
APS-837 Use CSV for reports part 2 - set report filename header in API and remove unused reports

davidatkinsuk:
* Add filename via content-disposition header for all CAS1 reports
* Move all active CAS1 report code from ReportService into Cas1ReportService
* Remove all unused CAS1 reports including API endpoints


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ReportsController.kt
```diff
@@ -222,11 +234,17 @@ class ReportsController(
...
+    return generateStreamingResponse(
+      contentType = ContentType.XLSX,
+      fileName = createCas1ReportName("placement-matching-outcomes", year, month, ContentType.XLSX),
+    ) { outputStream ->
       cas1ReportService.createPlacementMatchingOutcomesReport(properties, outputStream)
     }
   }
 
+  private fun createCas1ReportName(name: String, year: Int, month: Int, contentType: ContentType) =
+    "$name-$year-${month.toString().padStart(2, '0')}.${contentType.extension}"
```
edavey, line 180: just an idea but I think users can find it helpful if downloads are timestamped. They often run the same report multiple times and can get in a muddle identifying the "right" version. e.g.

`placement-matching-outcomes-2024-june-20240604_0942.xlsx`

edavey (approved):
👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ReportsController.kt
davidatkinsuk: Good idea, i'll do that before merging

PR #1932, opened 2024-06-03
APS-825 - Use jsonb for domain_events.data

davidatkinsuk:
The domain_events.data column stores JSON data. We use data stored in this column in a lot of reports, and it has been noted that these reads are causing notable performance hits (e.g. it can add seconds to query execution times).

This commit updates domain_events.data to use the ‘jsonb’ type instead of the ‘json’ type. This should significantly improve read speed, in exchange for a slight performance penalty on inserts. As we typically only insert a handful of domain events per thread of execution, if any at all, this is an acceptable trade off.

See https://dsdmoj.atlassian.net/browse/APS-825 for test evidence. I've ensured that domain events are still processed by probation-integration, and timelines can still be rendered

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas2ApplicationStatusUpdatesReport.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas2ApplicationStatusUpdatesReport.kt
@@ -23,7 +23,7 @@ interface Cas2ApplicationStatusUpdatesReportRepository : JpaRepository<DomainEve
          ) AS updatedAt
 
       FROM domain_events events
-      LEFT JOIN LATERAL json_array_elements(events.data -> 'eventDetails' -> 'newStatus' -> 'statusDetails') as details ON true
+      LEFT JOIN LATERAL jsonb_array_elements(events.data -> 'eventDetails' -> 'newStatus' -> 'statusDetails') as details ON true
       WHERE events.type = 'CAS2_APPLICATION_STATUS_UPDATED'
         AND events.occurred_at  > CURRENT_DATE - 365
       GROUP BY events.id  
--- a/src/main/resources/db/migration/all/20240614123435__use_jsonb_for_domain_events_data.sql
+++ b/src/main/resources/db/migration/all/20240614123435__use_jsonb_for_domain_events_data.sql
@@ -1,0 +1,1 @@
+ALTER TABLE domain_events ALTER COLUMN "data" SET DATA TYPE jsonb USING "data"::jsonb;


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas2ApplicationStatusUpdatesReport.kt
```diff
@@ -23,7 +23,7 @@ interface Cas2ApplicationStatusUpdatesReportRepository : JpaRepository<DomainEve
          ) AS updatedAt
 
       FROM domain_events events
-      LEFT JOIN LATERAL json_array_elements(events.data -> 'eventDetails' -> 'newStatus' -> 'statusDetails') as details ON true
+      LEFT JOIN LATERAL jsonb_array_elements(events.data -> 'eventDetails' -> 'newStatus' -> 'statusDetails') as details ON true
```
danhumphreys-moj, line 26: Just did a quick manual check to see that this report still works and all looks ok.

src/main/resources/db/migration/all/20240603102707__use_jsonb_for_domain_events_data.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE domain_events ALTER COLUMN "data" SET DATA TYPE jsonb USING "data"::jsonb;
```
danhumphreys-moj, line 1: Just wondering if there is such a thing as calculating a checksum on the JSON text field before, running the alter table, then doing a checksum after to make sure it hasn't changed, and wrapping in a SQL commit?

Just so we can make sure data before = data after?

danhumphreys-moj: Just thinking out loud..

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas2ApplicationStatusUpdatesReport.kt
davidatkinsuk: Thanks for checking 👍

src/main/resources/db/migration/all/20240603102707__use_jsonb_for_domain_events_data.sql
davidatkinsuk: FYI i've checked that domain events that existed in the database before converting data from json to jsonb can be replayed (i.e. still processed by probation-integration) and these worked without issue. Details are on the ticket. I'm going to merge this now.

danhumphreys-moj (approved):

PR #1934, opened 2024-06-04
APS-854 - Revise CAS1 Application Raw Data Report

davidatkinsuk:
See https://dsdmoj.atlassian.net/browse/APS-854 for a link to a spreadsheet that defines the updated report

This commit provides a rewrite of the CAS1 RAW Applications Data Report, adding and removing several fields and outputting the report as a CSV.

The commit also adds the Cas1SimpleApiClient in an attempt to make tests interacting with the API to setup test data more readable.

Instead of replacing the original report, this report is added as a new version of the original report. This allows us to feature flag the availability of the report on the UI as we’re not currently in a position to backfill required data in production.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -69,6 +69,8 @@ dependencies {
     exclude("org.apache.httpcomponents", "httpclient")
   }
 
+  implementation("com.opencsv:opencsv:5.9")
+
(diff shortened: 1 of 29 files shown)


RobBoothMOJ (approved):
Approves

PR #1935, opened 2024-06-04
APS-863 Don’t use JPA Callbacks to Update App Status

davidatkinsuk:
Before this PR we used JPA Listeners to manage updating the application status when a miscellaneous entities were persisted.

Not only does this make comprehension of the code difficult, it technically isn’t allowed according to the JPA Spec, and has caused us issues when attempting to update to Spring 3 / Hibernate 6:

“In general, the lifecycle method of a portable application should not invoke EntityManager or Query operations, access other entity instances, or modify relationships within the same persistence context. A lifecycle callback method may modify the non-relationship state of the entity on which it is invoked.”

This commit calls the Listener components directly as an intermediate step towards using a state machine to manage application state.

As part of this change the order that the ‘listeners’ were invoked in has changed, leading to the AssessmentListener overwriting the Withdrawn Application Status. AssessmentListener has been updated to correct this, and the WithdrawalTest updated to assert on application state after withdrawal. This is a good example of why using listeners for managing status is opaque, confusing and as such, should be discouraged!


edavey (approved):
Thanks for laying out these fiddly changes (and the fix on withdrawn state)  so clearly

PR #1938, opened 2024-06-05
CAS-322 Migrate CAS3 users PDU from nDelius

muhammad-elabdulla:
This [PR CAS-322 ](https://dsdmoj.atlassian.net/browse/CAS-322) is to improve the migration process to assign to the user a PDU that match the latest active team with a valid PDU in CAS3. In same cases the user have many teams in nDelius not all of them have a valid PDU in CAS3.


davidatkinsuk (approved):
LGTM 👍 

Going forward could the first line of the commit please be <= 50 characters? This avoids truncation in github's commit listing.

PR #1939, opened 2024-06-05
CAS-391 Fix an issue with referral rejected in referrals report

muhammad-elabdulla:
This [PR CAS-391](https://dsdmoj.atlassian.net/browse/CAS-391) is to fix the value shown in the referralRejected column in the report:

- Show **Yes** if the HPT rejected the referral and selected one of the available options
- Show **No** in all other cases


davidatkinsuk (approved):
LGTM

PR #1940, opened 2024-06-05
[APS-845][APS-846] Introduce `GET /cas1/out-of-service-beds` endpoint

danlivings-dxw:
> See [APS-845](https://dsdmoj.atlassian.net/browse/APS-845) and [APS-846](https://dsdmoj.atlassian.net/browse/APS-846) on the Approved Premises Jira board.

This PR introduces a `GET /cas1/out-of-service-beds` endpoint to the CAS1 API. This endpoint works similarly to the already existing `GET /cas1/premises/{premisesId}/out-of-service-beds`, except that:
- it is not scoped to a particular premises
- it includes both active and cancelled out-of-service beds
- it only includes current or future out-of-service beds.

To facilitate the needs of the UI, the `OutOfServiceBed` schema in the API specification has been updated and now looks like:
```jsonc
{
  "id": "00000000-0000-0000-0000-000000000000",
  "createdAt": "2024-01-01T12:34:56.789Z",
  "outOfServiceFrom": "2024-01-02",
  "outOfServiceTo": "2024-01-03",
  "bed": { // NamedId
    "id": "00000000-0000-0000-0000-000000000000",
    "name": "Bed Name"
  },
  "room": { // NamedId
    "id": "00000000-0000-0000-0000-000000000000",
    "name": "Room Name"
  },
  "premises": { // NamedId
    "id": "00000000-0000-0000-0000-000000000000",
    "name": "Premises Name"
  },
  "apArea": { // NamedId
    "id": "00000000-0000-0000-0000-000000000000",
    "name": "AP Area"
  },
  "reason": { // Cas1OutOfServiceBedReason
    "id": "00000000-0000-0000-0000-000000000000",
    "name": "Example Reason",
    "isActive": true
  },
  "referenceNumber": "AB123456",
  "notes": "Some notes",
  "daysLostCount": 2,
  "temporality": "past", // Temporality
  "status": "cancelled", // Cas1OutOfServiceBedStatus
  "cancellation": { // Cas1OutOfServiceBedCancellation
    "id": "00000000-0000-0000-0000-000000000000",
    "createdAt": "2024-01-01T12:34:56.789Z",
    "notes": "Some cancellation notes"
  }
}
```

[APS-845]: https://dsdmoj.atlassian.net/browse/APS-845?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[APS-846]: https://dsdmoj.atlassian.net/browse/APS-846?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/main/resources/static/_shared.yml
```diff
@@ -4860,3 +4868,21 @@ components:
         - id
         - name
         - isActive
+    NamedId:
```
edavey, line 4871: nice!

src/main/resources/static/codegen/built-api-spec.yml
```diff
@@ -9416,25 +9416,30 @@ components:
         createdAt:
           type: string
           format: date-time
-        startDate:
+        outOfServiceFrom:
           type: string
           format: date
-        endDate:
+        outOfServiceTo:
           type: string
```
edavey, line 9423: thanks for making these names clearer

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/transformer/cas1/Cas1OutOfServiceBedTransformerTest.kt
```diff
@@ -69,8 +99,26 @@ class Cas1OutOfServiceBedTransformerTest {
...
+      "-10,-5,6,past",
+      "-5,5,11,current",
+      "-1,0,2,current",
+      "0,0,1,current",
+      "0,1,2,current",
+      "5,10,6,future",
+    ],
+  )
+  @ParameterizedTest
+  fun `transformJpaToApi transforms correctly when cancelled`(
```
edavey, line 114: could we slim this spec right down maybe? Does it perhaps only need to assert:

```
assertThat(result.status).isEqualTo(Cas1OutOfServiceBedStatus.cancelled)
assertThat(result.cancellation).isEqualTo(cancellation)
```

if so, then maybe making it much smaller will help readers understand the change in behaviour which is being described?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserAccessService.kt
```diff
@@ -99,6 +99,12 @@ class UserAccessService(
   fun userCanManagePremisesOutOfServiceBed(user: UserEntity, premises: PremisesEntity) =
     userCanManagePremisesLostBeds(user, premises)
 
+  fun currentUserCanViewOutOfServiceBeds() =
+    userCanViewOutOfServiceBeds(userService.getUserForRequest())
+
+  fun userCanViewOutOfServiceBeds(user: UserEntity) =
+    user.hasRole(UserRole.CAS1_WORKFLOW_MANAGER)
```
edavey, line 106: 👍 this is correct (though is going to change to something clearer soon, e.g. `CRU_TEAM`)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas1OutOfServiceBedEntity.kt
```diff
@@ -15,6 +15,9 @@ import javax.persistence.Table
 
 @Repository
 interface Cas1OutOfServiceBedRepository : JpaRepository<Cas1OutOfServiceBedEntity, UUID> {
+  @Query("SELECT oosb FROM Cas1OutOfServiceBedEntity oosb WHERE oosb.endDate >= CURRENT_DATE ORDER BY oosb.startDate")
```
edavey, line 18: I tend to think that filtering by temporality ought perhaps to come later? And that initially we should return all out of service beds?

Also, is it worth being explicit about the ordering direction e.g `ORDER BY oosb.startDate DESC`.

I for one couldn't say for sure what the default ordering on a date field would be!

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/OutOfServiceBedTest.kt
```diff
@@ -60,7 +132,7 @@ class OutOfServiceBedTest : InitialiseDatabasePerClassTestBase() {
 
     @ParameterizedTest
     @EnumSource(value = UserRole::class, names = [ "CAS1_FUTURE_MANAGER", "CAS1_MANAGER", "CAS1_MATCHER" ])
-    fun `Get All Out-Of-Service Beds returns OK with correct body when user has one of roles FUTURE_MANAGER, MANAGER, MATCHER`(role: UserRole) {
+    fun `Get All Out-Of-Service Beds On Premises returns OK with correct body when user has one of roles FUTURE_MANAGER, MANAGER, MATCHER`(role: UserRole) {
```
edavey, line 135: thanks for this clarification

edavey (approved):
Nice work, thanks

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas1OutOfServiceBedEntity.kt
danlivings-dxw: I've added the `WHERE` clause as it's the default filter described in [APS-847](https://dsdmoj.atlassian.net/browse/APS-847), so no changes to the default behaviour should occur when filtering is fully implemented.

The default sort direction when not specified is `ASC` but I can make this explicit.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/transformer/cas1/Cas1OutOfServiceBedTransformerTest.kt
danlivings-dxw: I've updated this.

PR #1943, opened 2024-06-06
APS-872 Capture internal server errors in sentry

davidatkinsuk:
Before this change we would return an unhandled exception as a 500 ‘internal server error’ back to the UI, but not raise an alert in Sentry. This meant we were relying on the UI to raise a Sentry Alert. Not only is this not the correct ownership of responsibility, this meant that we were getting unhelpful alerting information being raised as the UI doesn’t know what caused the error.

This commit ensures all unhandled exceptions are logged in sentry.


edavey (approved):
nicely done, thks

PR #1944, opened 2024-06-07
APS-864 Generate unique id on all placement requests

RobBoothMOJ:
For reporting purposes we need to generate a unique UUID for any request for placement that is part of the original application (i.e. where approved_premises_applications.arrival_date is defined)

This will be captured in a new table placement_applications_automatic and backfilled for existing applications which have an arrival_date defined.

### DIFF
--- a/src/main/resources/db/migration/all/20240607130747__create_placement_applications_automatic_table.sql
+++ b/src/main/resources/db/migration/all/20240607130747__create_placement_applications_automatic_table.sql
@@ -1,0 +1,14 @@
+CREATE TABLE placement_applications_automatic (
+    id UUID NOT NULL,
+    application_id UUID NOT NULL,
+    submitted_at timestamp NOT NULL,
+    expected_arrival_date timestamp NOT NULL,
+
+    PRIMARY KEY (id),
+
+    CONSTRAINT fk_application_id
+        FOREIGN KEY(application_id)
+            REFERENCES applications(id)
+);
+
+CREATE INDEX ON placement_applications_automatic(application_id);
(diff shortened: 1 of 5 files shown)


src/main/resources/db/migration/all/20240607134906__create_legacy_placement_applications_automatic_records.sql
```diff
@@ -0,0 +1,12 @@
+-- create placementApplicationsAutomatic records for existing applications with arrival date
```
davidatkinsuk, line 1: very nice!

can we please replace the the word 'create' in the filename with 'backfill' as that's the common term we use to describe these operations. I think we can then remove the comment in the file too.

src/main/resources/db/migration/all/20240607130747__create_placement_applications_automatic_table.sql
```diff
@@ -0,0 +1,14 @@
...
+                                   application_id               UUID                        NOT NULL,
+
+                                   PRIMARY KEY (id),
+
+                                   CONSTRAINT fk_application_id
+                                       FOREIGN KEY(application_id)
+                                           REFERENCES applications(id)
+);
+
+-- index for foreign key to applications table
```
davidatkinsuk, line 13: i don't think these comments are needed?

src/main/resources/db/migration/all/20240607130747__create_placement_applications_automatic_table.sql
```diff
@@ -0,0 +1,14 @@
+-- create new table for Placement Applications Automatic
```
davidatkinsuk, line 1: given the entity is defined in the first commit, this migration should really be in the commit too (otherwise the prior commit fails to run). It's probably easiest to just merge the first two commits.

davidatkinsuk (approved):
This is great. I did spot a typo on the first commit. I'd suggest copying the summary you've put into the PR into the first commit as it gives good context for when reviewing the commit history

PR #1949, opened 2024-06-11
APS-839 Swapped out obtaining Probation Area Name from cruService, super-seeded by user.apArea.

RobBoothMOJ:
Swapped out obtaining Probation Area Name from CruService.cruNameFromProbationAreaCode to user.apArea.name as this new method super-seeds the Cru service mapping which looks to be out of date.


davidatkinsuk:
Hi @RobBoothMOJ could you please truncate the first line of the commit to 60 characters and also put the useful context in the PR description into the commit message?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AssessmentService.kt
```diff
@@ -74,7 +74,6 @@ class AssessmentService(
   private val domainEventService: DomainEventService,
   private val offenderService: OffenderService,
   private val communityApiClient: CommunityApiClient,
-  private val cruService: CruService,
```
davidatkinsuk, line 77: I think CruService can be deleted now? 🤞 that will make a net-negative line count PR 😄

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AssessmentService.kt
```diff
@@ -571,7 +570,7 @@ class AssessmentService(
                   name = staffDetails.probationArea.description,
                 ),
                 cru = Cru(
-                  name = cruService.cruNameFromProbationAreaCode(staffDetails.probationArea.code),
+                  name = user.apArea?.name.toString(),
```
davidatkinsuk, line 573: I _think_ toString() is redundant here? (and everywhere else name is used in this PR)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1AssessmentDomainEventService.kt
```diff
@@ -185,7 +185,7 @@ class Cas1AssessmentDomainEventService(
                 name = staffDetails.probationArea.description,
               ),
               cru = Cru(
-                name = cruService.cruNameFromProbationAreaCode(staffDetails.probationArea.code),
+                name = userService.getUserForRequest().apArea?.name.toString(),
```
davidatkinsuk, line 188: Although we do it alot, i'm not a fan of using userService.getUserForRequest() outside of the controller layer. Instead of doing this, could we please pass the UserEntity in from the calling AssessmentService.

Furthermore, i've noticed that the StaffUserDetails being resolved in AssessmentService.acceptAssessment are only being used by this domain event function, so instead we should stop loading the StaffUserDetails in  AssessmentService.acceptAssessment and instead load them in this function instead (which makes this function more like the other domain event functions you've modified)

davidatkinsuk (approved):
Thanks @RobBoothMOJ , just a few minor comments from me 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AssessmentService.kt
RobBoothMOJ: There's a type issue between 
apArea.name  (nullable)
cru.name (non-nullable)
toString() mitigates this but does mask the possibility of null getting through as 'null'
What's the desired behaviour?   
Would !! be more appropriate?

davidatkinsuk: Ah, good point!

This works because toString is an extension function that can operate on nulls. In this case it would return the string "null". See https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/to-string.html

user.apArea should _really_ be non-nullable as we always populate it, so the null scenario should never happen here. Regardless, i think it's probably more readable if we state what is returned when apArea is null. For consistency with what CruService was doing, maybe we could use this in all cases...

`user.apArea?.name ?: "Unknown CRU"`

davidatkinsuk (approved):
LGTM 👍 

one tiny suggestion, up to you if you want to do it

PR #1951, opened 2024-06-11
APS-823 - Use internal enum for CAS1 assess aptype metadata

davidatkinsuk:
We were previously using the value of the API type to populate the MetaDataName.CAS1_REQUESTED_AP_TYPE metadata. This is inconsistant with how we capture this value on application submission.


RobBoothMOJ (approved):
approved

edavey (approved):

PR #1953, opened 2024-06-12
APS-904 Block date changes on cancelled bookings

davidatkinsuk:
Whilst unlikely, it is technically possible to change dates on the UI by bookmarking the date change page, or constructing the URL manually, as was observed in production. This commit ensures that such changes will be blocked, ensuring the database isn’t in a logically inconsistant state.


RobBoothMOJ (approved):
Approved

PR #1957, opened 2024-06-12
APS-906 - Provide ‘other reason’ on Booking API Model

davidatkinsuk:
When a booking is cancelled with the reason ‘other’, we capture a free text description from the user, but don’t show this anywhere on the UI.

This commit adds the free text reason to the Booking model, making it available when retrieiving  a booking


RobBoothMOJ (approved):
Approved

PR #1959, opened 2024-06-13
APS-365 Removed unnecessary casting in Task SQL

davidatkinsuk:
This casting causes issues after upgrading to spring boot 3 / hibernate 6


RobBoothMOJ (approved):
Approved

PR #1963, opened 2024-06-13
Set local allocation rules to match dev

richpjames:
In dev (and other enviornments) the environment variables are set in values-[env].yml files in the helm_deploy folder. All of these environments (except the deprecated test env) have legacy-allocator-enabled to be false. In order to make local development (via ap-tools) more like other environments I think legacy-allocator-enabled should be false.
Here I also set 'priority' to match other envs.

### DIFF
--- a/src/main/resources/application-local.yml
+++ b/src/main/resources/application-local.yml
@@ -62,7 +62,7 @@ user-allocations:
         north-west: JIMSNOWLDAP
         midlands: JIMSNOWLDAP
         south-east-eastern: JIMSNOWLDAP
-        south-west-south central: JIMSNOWLDAP
+        south-west-south-central: JIMSNOWLDAP
         wales: JIMSNOWLDAP
         nat: JIMSNOWLDAP
     female-assessments:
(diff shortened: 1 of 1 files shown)


src/main/resources/application-local.yml
```diff
@@ -66,6 +67,6 @@ user-allocations:
         wales: JIMSNOWLDAP
         nat: JIMSNOWLDAP
     reallocated-assessments:
-      enabled: true
+      enabled: false
     female-assessments:
-      enabled: true
+      enabled: false
```
davidatkinsuk, line 72: This appears to be 'true' in the helm configs for dev/prod, so maybe should leave as true?

src/main/resources/application-local.yml
```diff
@@ -66,6 +67,6 @@ user-allocations:
         wales: JIMSNOWLDAP
         nat: JIMSNOWLDAP
     reallocated-assessments:
```
davidatkinsuk, line 69: i can't find this referenced anywhere, so i think it can be removed (reallocated-assessments)?

davidatkinsuk (approved):
Just a few small comments but I definitely agree with this change 👍

PR #1965, opened 2024-06-14
APS-920 - Remove unused OAsys Client Function

davidatkinsuk:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/client/ApOASysContextApiClient.kt
```diff
@@ -38,10 +37,6 @@ class ApOASysContextApiClient(
     path = "/needs-details/$crn"
   }
 
-  fun getHealthDetails(crn: String) = getRequest<HealthDetails> {
-    path = "/health-details/$crn"
```
edavey, line 42: should we: 

- include a reference to our wiremock example: https://github.com/ministryofjustice/hmpps-approved-premises-tools/blob/main/wiremock/mappings/ApAndOasys_GetHeath.json   ?
- create an accompanying PR to remove the wiremock example stub from API-Tools?

edavey (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/client/ApOASysContextApiClient.kt
davidatkinsuk: Hi Ed, good catch, I hadn't considered wiremock in ap-tools. I'm not sure that we need to explicitly ref it in the commit, but i'll create a PR to remove it. I've also found a wiremock config in this project, so i've removed that in this commit too.

PR #1966, opened 2024-06-14
CAS-367 show rejection reason and is withdrawn in the HPT timeline view

muhammad-elabdulla:
This [PR CAS-367](https://dsdmoj.atlassian.net/browse/CAS-367) is to add the:

- Rejection reason
- Rejection reason details
- is withdrawn

to the last rejection note in the notes for the timeline history

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/AssessmentReferralHistoryNoteTransformer.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/AssessmentReferralHistoryNoteTransformer.kt
@@ -2,23 +2,19 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer
 
 import org.springframework.stereotype.Component
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ReferralHistoryNote
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ReferralHistoryNoteMessageDetails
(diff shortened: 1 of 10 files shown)


danhumphreys-moj (approved):
Looks ok to me Muhammad.  Just some observations but nothing that prevents me from approving I think.

- I'm not sure the acceptance criteria reflects the tests in this ticket.  This PR looks like it returns all notes for the assessment.  I'm going to presume that the work in this PR is correct and what is required by the front end and that the acceptance test in the ticket is wrong (since it makes sense to return all notes for the timeline history).
- The integration tests `anyMatch` rather than testing to make sure the returned notes are in the right (date?) order
- I'm not sure why we have a `withCreatedAt` on line 1947 of `AssessmentTest` and none of the other notes do.
- It also seems that this specific integration test tests different scenarios: an assessment is created, with a rejection reason; withdrawn is set to true; a note is added; the status changed to submitted; the status changed to rejected 90 days ago; the status changed to unallocated; then in review; then ready to take place; then rejected again 4 days ago.  Not sure whether this an assessment test, notes test or timeline test?  I still have some work to do so hopefully this will become clear.
- Also the description of the PR is consistent with the ticket details but does not reflect the work done in the PR.

muhammad-elabdulla:
- This tickets is to add the details for the last rejection note is not altering the order or the notes. I updated the acceptance criteria to include that it will return all the all notes for the timeline history
- I don't think the test is created to test the time line of the notes is only to make sure that is returning what expected in each status. But in this case we need to make sure that the rejection reason and details is added only to the last rejection note only, this why I included dates.

PR #1967, opened 2024-06-14
APS-853 - Backfill CAS1_REQUESTED_AP_TYPE metadata

davidatkinsuk:
Now this is being via the API, we need to backfill any previously submitted and assessed application metadata

Note - this should only be merged once the corresponding UI functionality (APS-822) has gone live


RobBoothMOJ (approved):
Approved

RobBoothMOJ (approved):
Approved

PR #1968, opened 2024-06-14
APS-892 - Added isWithdrawn property to placementApplication entity

RobBoothMOJ:
This allows current 'decision' value to be kept when an application is withdrawn rather than being overwritten with a withdrawn decision. 
Backfill script will set the isWithdrawn property on existing records where a withdrawn decision already exists. 
WITHDRAW and WITHDRAWN_BY_PP PlacementApplicationDecision values are marked deprecated in the entity.

**NOTE** : This cannot be merged until APS-893 has gone live.  
APS-893 covers the removal of the two deprecated decisions ('WITHDRAW', 'WITHDRAWN_BY_PP') in the UI. 
Merging this change in prior to the corresponding UI change will result in additional records to be backfilled.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/WithdrawalTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/WithdrawalTest.kt
@@ -236,8 +236,7 @@ class WithdrawalTest : IntegrationTestBase() {
      * | -> Request for placement 4 (Reallocated)     | -            |
      * | -> Request for placement 5 (With Decision)   | Yes          |
      * | -> Request for placement 6 (Withdrawn)       | -            |
-     * | -> Request for placement 7 (Withdrawn by PP) | -            |
-     * | -> Request for placement 7 (Rejected)        | Yes            |
+     * | -> Request for placement 7 (Rejected)        | Yes          |
      * ```
      */
     @Test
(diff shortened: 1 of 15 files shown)


davidatkinsuk:
Could you please put the ticket number in the PR title. This will then appear in Slack when releasing.

Also, could you please add a note saying that this can't be merged until APS-893 has gone live (this is the UI ticket that removes the 'WITHDRAW' and 'WITHDRAWN_BY_PP' options from the UI). Otherwise we'll have more records to backfill.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/WithdrawalTest.kt
```diff
@@ -279,18 +279,6 @@ class WithdrawalTest : IntegrationTestBase() {
...
-              application,
-              DateSpan(now(), duration = 2),
-              decision = PlacementApplicationDecision.WITHDRAW,
-            )
-
-            createPlacementApplication(
-              application,
-              DateSpan(now(), duration = 2),
-              decision = PlacementApplicationDecision.WITHDRAWN_BY_PP,
-            )
```
davidatkinsuk, line 292: I think for this test we need at least one placement application that is legitimately withdrawn (i.e. where the new isWithdrawn boolean is set to true). This shouldn't exist in the response from the API, so it's just a matter of creating a placement request here (essentially folding down the original two withdrawn placements applications with different decisions types into a single placement application with the new boolean set to true).

Also, could you please review the javadoc on the test function and update it to reflect this

davidatkinsuk (approved):
Looks good, only minor comments from me.

Please note my comments re: not merging until the corresponding UI change has gone live

PR #1969, opened 2024-06-14
APS-853 migration job to backfill app reason for short notice

davidatkinsuk:
This PR adds a migration job to backfill app reason for short notice and short notice description into metadata.

Whilst we could have used a SQL based migration for this, testing in pre-prod showed this took over 40 seconds to run for each metadata value which would lead to a non-trivial prod outage on deploy (assuming it runs without issue over that length of time).

This PR also moves all the cas1 migration jobs into a cas1 specific package


edavey (approved):
Looks very clear. Thanks for including the detailed logging which I think will allow you to verify the expected behaviour and outcome in preprod.

PR #1972, opened 2024-06-17
APS-853 - Fix short notice migration job transaction issue

davidatkinsuk:
The migration job could not lazy load metadata associated with a domain event because the domain was being loaded outside of the transaction currently in scope. This commit fixes that by extending the scope of the transaction


RobBoothMOJ (approved):
Approved

PR #1973, opened 2024-06-17
APS-909 : Handle null failureDescription when booking not made.

RobBoothMOJ:
Resolves issue with 'null' appearing as a reason, on the timeline, when a booking is not made and failureDescription has not been populated (is null).

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/domainevents/DomainEventDescriber.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/domainevents/DomainEventDescriber.kt
@@ -17,7 +17,6 @@ import java.time.LocalDate
 import java.time.ZoneOffset
 import java.time.temporal.ChronoUnit
 import java.util.UUID
-import javax.xml.datatype.DatatypeConstants.DAYS
 
 @SuppressWarnings("TooManyFunctions")
 @Component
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/domainevents/DomainEventDescriber.kt
```diff
@@ -113,7 +112,11 @@ class DomainEventDescriber(
 
   private fun buildBookingNotMadeDescription(domainEventSummary: DomainEventSummary): String? {
     val event = domainEventService.getBookingNotMadeEvent(domainEventSummary.id())
-    return event.describe { "A placement was not made for the placement request. The reason was: ${it.eventDetails.failureDescription}" }
+    var failureReason = ""
+    if (event?.data?.eventDetails?.failureDescription != null) {
+      failureReason = " The reason was: ${event?.data?.eventDetails?.failureDescription}"
+    }
```
davidatkinsuk, line 118: In kotlin if statements can return values. I think we can use that along with the 'let' scope function and the elvis operator to remove the variable mutation whilst (hopefully) keeping it readible:

```
val failureReason = event?.data?.eventDetails?.failureDescription?.let { 
   " The reason was: $it" 
} ?: ""
```

davidatkinsuk (approved):
Looks good, just one minor style suggestion 👍

PR #1975, opened 2024-06-18
APS-855 - CAS1 Report PII shouldn’t be returned by default

davidatkinsuk:
Previously we relied on the open-api default value of ‘false’ for includePii, but testing has shown that this is resolving to true in code.

This commit manages the default values in controllers instead, and updates the reports tests that check PII to ensure the default behaviour is observed


RobBoothMOJ (approved):
Approved

PR #1976, opened 2024-06-19
Cas 365/remove id cast and mockkstatic

danhumphreys-moj:
Without the fixes contained in this branch upgrading spring boot and Hibernate results in:

- a JPA class cast exception (since UUID id is cast to TEXT in SQL but then has type UUID in the entity)
- mockkstatic no longer working in the associated CAS 2 integration tests

Adding these features to main now since:
- it will help reduce the number of changes in the forthcoming merge of the upgrade branch
- there is no reason for them not to be merged in beforehand

Note that the new view creation SQL is as per [20240528100000__create_cas_2_summary_views_add_abandoned.sql](https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/main/src/main/resources/db/migration/all/20240528100000__create_cas_2_summary_views_add_abandoned.sql) except `cas_2_application_summary` and it's dependent view `cas_2_application_live_summary` are dropped and then re-created without the `id` cast to `TEXT` on line 3.

It is not sufficient to just `REPLACE` the view since the type of the field is changing and hence the requirement to `DROP` and then `CREATE`.

### DIFF
--- a/src/main/resources/db/migration/all/20240618100000__create_cas_2_summary_views_change_id_type.sql
+++ b/src/main/resources/db/migration/all/20240618100000__create_cas_2_summary_views_change_id_type.sql
@@ -1,0 +1,48 @@
+-- remove id cast to TEXT
+DROP VIEW IF EXISTS cas_2_application_summary CASCADE;
+CREATE OR REPLACE VIEW cas_2_application_summary AS SELECT
+    a.id,
+    a.crn,
+    a.noms_number,
+    CAST(a.created_by_user_id AS TEXT),
+    nu.name,
+    a.created_at,
+    a.submitted_at,
+    a.hdc_eligibility_date,
+    asu.label,
+    CAST(asu.status_id AS TEXT),
+    a.referring_prison_code,
+    a.conditional_release_date,
+    asu.created_at AS status_created_at,
+    a.abandoned_at
+FROM cas_2_applications a
+LEFT JOIN (SELECT DISTINCT ON (application_id) su.application_id, su.label, su.status_id, su.created_at
+    FROM cas_2_status_updates su
+    ORDER BY su.application_id, su.created_at DESC) as asu
+    ON a.id = asu.application_id
+JOIN nomis_users nu ON nu.id = a.created_by_user_id;
+
+-- filter applications that do not have abandoned_at
+CREATE OR REPLACE VIEW cas_2_application_live_summary AS SELECT
+    a.id,
+    a.crn,
+    a.noms_number,
+    a.created_by_user_id,
+    a.name,
+    a.created_at,
+    a.submitted_at,
+    a.hdc_eligibility_date,
+    a.label,
+    a.status_id,
+    a.referring_prison_code,
+    a.abandoned_at
+FROM cas_2_application_summary a
+WHERE (a.conditional_release_date IS NULL OR a.conditional_release_date >= current_date)
+AND a.abandoned_at IS NULL
+AND a.status_id IS NULL
+   OR (a.status_id = '004e2419-9614-4c1e-a207-a8418009f23d' AND a.status_created_at > (current_date - INTERVAL '14 DAY')) -- Referral withdrawn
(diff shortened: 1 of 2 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ApplicationTest.kt
```diff
@@ -243,26 +237,27 @@ class Cas2ApplicationTest : IntegrationTestBase() {
             unexpiredApplicationIds.add(createApplication(userEntity, offenderDetails).id)
           }
 
-          mockkStatic(OffsetDateTime::class)
-
-          unexpiredSubset.union(expiredSubset).forEachIndexed { index, status ->
+          unexpiredSubset.union(expiredSubset).forEachIndexed { _, status ->
```
davidatkinsuk, line 240: I think this can be simplified to forEach now instead of forEachIndexed? (there's a few other places this can be done)

src/main/resources/db/migration/all/20240618100000__create_cas_2_summary_views_change_id_type.sql
```diff
@@ -0,0 +1,48 @@
...
+    a.abandoned_at
+FROM cas_2_applications a
+LEFT JOIN (SELECT DISTINCT ON (application_id) su.application_id, su.label, su.status_id, su.created_at
+    FROM cas_2_status_updates su
+    ORDER BY su.application_id, su.created_at DESC) as asu
+    ON a.id = asu.application_id
+JOIN nomis_users nu ON nu.id = a.created_by_user_id;
+
+-- filter applications that do not have abandoned_at
+CREATE OR REPLACE VIEW cas_2_application_live_summary AS SELECT
```
davidatkinsuk, line 26: does this view need updating? I've looked at the current version and I can't see any cast of UUIDs? I suspect i'm misunderstanding something about views.

davidatkinsuk (approved):
Thanks @danhumphreys-moj . I think the code setting createdAt is far more readable now too, so that's a bonus.

Just a few minor comments from me

src/main/resources/db/migration/all/20240618100000__create_cas_2_summary_views_change_id_type.sql
danhumphreys-moj: This view `cas_2_application_live_summary` is filtered data from the first view `cas_2_application_summary`.

cas_2_application_summary.id was previously TEXT (see [here](https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/main/src/main/resources/db/migration/all/20240528100000__create_cas_2_summary_views_add_abandoned.sql)).

But this cast is removed in this PR so the query will return id as UUID as per the table column type.

This is therefore consistent with the UUID type in [Cas2ApplicationSummaryEntity](https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/9751be6832dfa0823ba70af677c0b6662eb912a9/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas2ApplicationSummaryEntity.kt#L34).

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ApplicationTest.kt
danhumphreys-moj: Yes - changed now.

src/main/resources/db/migration/all/20240618100000__create_cas_2_summary_views_change_id_type.sql
danhumphreys-moj: Hopefully this makes sense @davidatkinsuk?

davidatkinsuk: Yep all good 👍

PR #1978, opened 2024-06-20
CAS1: Seed Delius user in dev/local envs for testing 'CRU Member' persona/role

edavey:
See:

- UI: Addition of `CAS1_E2E_CRU_MEMBER_*` https://github.com/ministryofjustice/hmpps-approved-premises-ui/pull/1902/commits/1c96a5affd5849ea0929336520ad7d74817085f2 for "Restrict 'View all OOSBs to CRU Member role'"
- AP-Tools: seeding/stubbing of Delius user (`SHEILAHANCOCKNPS`) used for this persona https://github.com/ministryofjustice/hmpps-approved-premises-tools/pull/71

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/Cas1AutoScript.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/Cas1AutoScript.kt
@@ -53,14 +53,37 @@ class Cas1AutoScript(
         seedUser.roles.forEach { role ->
           userService.addRoleToUser(user = user, role = role)
         }
-        seedLogger.info("  -> User '${user.name}' (${user.deliusUsername}) seeded with roles ${user.roles}")
+        val roles = user.roles.map { it.role }.joinToString(", ")
+        seedLogger.info("  -> User '${user.name}' (${user.deliusUsername}) seeded with roles $roles")
       }
     }
   }
 
   private fun usersToSeed(): List<SeedUser> {
     return listOf(
-      SeedUser(username = "SHEILAHANCOCKNPS", roles = listOf(UserRole.CAS1_CRU_MEMBER)),
+      SeedUser(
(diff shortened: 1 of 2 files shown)


davidatkinsuk (approved):
LGTM. 

One thought - is it worth adding these changes as new migrations to avoid disruption for devs having to reset the database? I appreciate that then results in test data that's harder to reason on as it's split across multiple files.

edavey:
> LGTM.
> 
> One thought - is it worth adding these changes as new migrations to avoid disruption for devs having to reset the database? I appreciate that then results in test data that's harder to reason on as it's split across multiple files.

This is a good point. I think having seeding in migrations is pretty horrid which I why I've previously tried to push the "auto script" method, so that seeding is represented in one place in a meaningful way. So yes I will move this... Thanks for the input.

edavey:
> LGTM.
> 
> One thought - is it worth adding these changes as new migrations to avoid disruption for devs having to reset the database? I appreciate that then results in test data that's harder to reason on as it's split across multiple files.

I've had another go at this and added the Delius user via `Cas1AutoScript`. Do you think that the `seedUsers()` function here will have legs? That we can extend it cover more complex use-cases? I could have a go at moving all the user seeding from `db/migration/local+dev+test/R__1_create_dev_users.sql` I guess?

davidatkinsuk:
> > LGTM.
> > One thought - is it worth adding these changes as new migrations to avoid disruption for devs having to reset the database? I appreciate that then results in test data that's harder to reason on as it's split across multiple files.
> 
> I've had another go at this and added the Delius user via `Cas1AutoScript`. Do you think that the `seedUsers()` function here will have legs? That we can extend it cover more complex use-cases? I could have a go at moving all the user seeding from `db/migration/local+dev+test/R__1_create_dev_users.sql` I guess?

I much prefer this approach, so if you have the bandwidth i think we should migrate them all over from SQL into here. I think it'll be much easier to maintain. I wonder if we want to clear any existing roles before updating, or maybe that's overkill?

PR #1981, opened 2024-06-21
APS-943 Backfill CAS1 Application Timeliness

davidatkinsuk:
This commit provides a SQL migration to populate the approved_premises_application.timeliness value for any application where it isn’t currently set. This is ensure the value is always available in reports.

This value was only populated in the database as-of 2024-03-20. This migration will only update applications where the value is currently null (i.e. from before 2024-03-20).

The logic used to determine the timeliness values matches that used in the UI (noticeTypeFromApplication.ts), which is applied as part of the application submission logic.

### DIFF
--- a/src/main/resources/db/migration/all/20240621101458__cas1_backfill_application_timeliness.sql
+++ b/src/main/resources/db/migration/all/20240621101458__cas1_backfill_application_timeliness.sql
@@ -1,0 +1,23 @@
+-- This logic mimicks the logic used in the UI code (noticeTypeFromApplication.ts)
+with apps_to_update as (
+    SELECT
+        a.id,
+        a.submitted_at,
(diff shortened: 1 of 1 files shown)


edavey (approved):
very clear, thanks. 

Is this backfilling required for reporting? I'm guessing there's no operational need here (the most recent applications missing the `notice_type` seem to be from March 19th so will already have been assessed long since?)

davidatkinsuk:
> very clear, thanks.
> 
> Is this backfilling required for reporting? I'm guessing there's no operational need here (the most recent applications missing the `notice_type` seem to be from March 19th so will already have been assessed long since?)

Thanks Ed. Yes, this is for reporting only. I was hoping we wouldn't need to do it but because we include applications _withdrawn_ in the 'reporting month' we end up with quite a few applications from pre March 19th showing up with no timeliness

PR #1982, opened 2024-06-21
APS-914 update profile response for non staff members

RobBoothMOJ:
Adds a new /profile/v2 endpoint which provides a more meaningful response in the event that a Staff record, for the logged in user, can't be found in Delius. 

Instead of a 500 response, which the current profile endpoint returns when the staff record can't be found, the new endpoint will return a new response which provides:

- deliusUserName (username / record searched for in Delius)
- loadError (present when the staff record cannot be found)
- user (present when the corresponding record has been found in Delius)

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
@@ -28,12 +28,12 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserRole
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserRoleAssignmentEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserRoleAssignmentRepository
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.specification.hasQualificationsAndRoles
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.GetUserResponse
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.PaginationMetadata
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.UserWorkload
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.community.StaffProbationArea
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.community.StaffUserDetails
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.BadRequestProblem
-import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.NotFoundProblem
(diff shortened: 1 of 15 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
```diff
@@ -342,6 +342,12 @@ class UserService(
         updatedAt = null,
       ),
     )
+    return GetUserResponse(savedUser, true)
+  }
+
+  fun getExistingUserOrCreate(username: String): UserEntity {
+    val getUserResponse = getExistingUserOrCreate(username, dontThrowExceptionOnStaffRecordNotFound = false)
+    return getUserResponse.user!!
   }
```
davidatkinsuk, line 361: i propose we move this function definition to be above its overloaded equivalent. it may just be me but i find overloaded functions easier to understand that way?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
```diff
@@ -342,6 +342,12 @@ class UserService(
         updatedAt = null,
       ),
     )
+    return GetUserResponse(savedUser, true)
+  }
+
+  fun getExistingUserOrCreate(username: String): UserEntity {
+    val getUserResponse = getExistingUserOrCreate(username, dontThrowExceptionOnStaffRecordNotFound = false)
+    return getUserResponse.user!!
```
davidatkinsuk, line 360: i think this can be flattened into a single line, removing the need for the curly braces and return type definition e.g.

```
fun getExistingUserOrCreate(username: String) = getExistingUserOrCreate(username, dontThrowExceptionOnStaffRecordNotFound = false).user!!
```

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/UserServiceTest.kt
```diff
@@ -79,6 +81,63 @@ class UserServiceTest {
...
+        .produce()
+
+      every { mockUserRepository.findByDeliusUsername(username) } returns user
+
+      assertThat(userService.getExistingUserOrCreate(username)).isEqualTo(user)
+      verify(exactly = 1) { userService.getExistingUserOrCreate(username, false) }
+    }
+
+    @Test
+    fun `getExistingUserOrCreate with invalid username does not throw error when dontThrowExceptionOnStaffRecordNotFound set true`() {
```
davidatkinsuk: nice! i think the test could be renamed to reflect that it's about staff record not existing, for the code to get to this point the username would be valid. (other test names may need tweaking too)

davidatkinsuk (approved):
Looks great, just some minor comments from me. I put some comments on the first commit without realising some things were changed on the second commit, i think i've removed all the ones that you'd already made on the second commit :)

Can we please hold fire on merging until Monday. Because it's a change to such a fundamental part of the system I want to make sure we're around when it goes live 👍

PR #1985, opened 2024-06-24
Remove unused ‘forceApDeliusContextApi’ variable

davidatkinsuk:
This is no longer used, as we no longer have the option of using Community API instead of the AP Delius Context when get calling getOffenderSummariesByCrns


RobBoothMOJ (approved):
Approved

PR #1986, opened 2024-06-24
APS-867: migrate CAS1 LostBeds to Cas1OutOfServiceBeds

edavey:
Run the on-demand migration with:

```
script/run_migration_job update_cas1_lost_beds_to_out_of_service_beds
```

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/migration/Cas1LostBedsToOutOfServiceBedsMigrationJob.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/migration/Cas1LostBedsToOutOfServiceBedsMigrationJob.kt
@@ -1,0 +1,152 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.migration
+
+import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.data.jpa.repository.Query
+import org.springframework.data.repository.findByIdOrNull
+import org.springframework.stereotype.Repository
+import org.springframework.transaction.support.TransactionTemplate
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApprovedPremisesEntity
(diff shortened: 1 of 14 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/migration/Cas1LostBedsToOutOfServiceBedsMigrationJob.kt
```diff
@@ -0,0 +1,118 @@
...
+      Cas1OutOfServiceBedEntity(
+        id = lostBed.id,
+        premises = lostBed.premises as ApprovedPremisesEntity,
+        bed = lostBed.bed,
+        createdAt = OffsetDateTime.now(),
+        cancellation = null,
+        revisionHistory = mutableListOf(),
+      ),
+    )
+
```
edavey, line 68: do you think, that for full information in case we need to unpick any problem in the future, we ought to record the fact that the `Cas1OutOfServiceBedEntity` with the given (LostBed) id has indeed been created. e.g. with another line of logging such as:

```
"Created CAS1 out-of-service bed {id: ${outOfServiceBed.id}, premises: ${outOfServiceBed.premises.id}, bed: ${l outOfServiceBed.bed.id}}"
```

and similarly to log the successful creation of:

- `Cas1OutOfServiceBedCancellationEntity` where appropariate
- `Cas1OutOfServiceBedRevisionEntity` in every case

edavey (commented):
I can't actually approve this as Github thinks I'm the author! 

Very clear and thorough work, thank you. 

I do think we could add a bit more logging of the created entities for more complete visibility on the outcome of the migration in case any remedial or additional work is needed.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/migration/Cas1LostBedsToOutOfServiceBedsMigrationJob.kt
danlivings-dxw: I've added some additional logging to capture this.

danlivings-dxw (approved):
Approving on @edavey's behalf

PR #1987, opened 2024-06-24
CAS-365 cast updatedData as jsonb and change to SQL.

danhumphreys-moj:
Without this change upgrading spring and hibernate to later versions causes a SQL exception since we are attempting to update a jsonb field with string data.  Casting to jsonb will resolve this error.  Casting also requires a change to use SQL rather than JPQL.


davidatkinsuk (approved):
LGTM 👍

PR #1988, opened 2024-06-24
Replace Spring HttpMethod with enum.

danhumphreys-moj:
Without this change and when Spring and Hibernate are upgraded Jackson is not able to unmarshall JSON into HttpMethod objects and the following error occurs:

`com.fasterxml.jackson.databind.exc.MismatchedInputException: Cannot construct instance of org.springframework.http.HttpMethod (although at least one Creator exists): cannot deserialize from Object value (no delegate- or property-based Creator)`

HttpMethod was previously an enum class (Spring v 4.14) which was serializable by Jackson but following the upgrade it is now neither an enum class nor serializable.

This PR adds in a serializable enum HTTP method class and uses it to construct a temporary `MarshallablePreemptiveCacheMetadata` from which the `PreemptiveCacheMetadata` object can be created.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/client/WebClientCache.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/client/WebClientCache.kt
@@ -214,4 +214,36 @@ class WebClientCache(
     val hasResponseBody: Boolean,
     val attempt: Int?,
   )
+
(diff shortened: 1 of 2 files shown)


davidatkinsuk:
This has been superseded by:

https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/2089
https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/2085

PR #1989, opened 2024-06-25
APS-950 - Add Seeded Applications and users for testing LAO

davidatkinsuk:
This commit adds configuration for two new users to test LAO scenarios in CAS1

* LAOFULLACCESS
* LAORESTRICTED

It also adds pre-created applications ready for submission for the following CRNs, both created by LAOFULLACCESS:

* X400000 exclusion for LAOFULLACCESS (whitelist)
* X400001 restricted from LAORESTRICTED (blacklist)

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/Cas1AutoScript.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/Cas1AutoScript.kt
@@ -30,21 +30,32 @@ class Cas1AutoScript(
 
   @SuppressWarnings("TooGenericExceptionCaught")
   @Transactional
-  fun script(
-    deliusUserName: String = "JIMSNOWLDAP",
-    crn: String = "X320741",
-  ) {
+  fun script() {
     seedLogger.info("Auto-Scripting for CAS1")
+
+    seedUsers()
+
+    createApplication(deliusUserName = "JIMSNOWLDAP", crn = "X320741")
+    createApplication(deliusUserName = "LAOFULLACCESS", crn = "X400000")
+    createApplication(deliusUserName = "LAOFULLACCESS", crn = "X400001")
+  }
+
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/Cas1AutoScript.kt
```diff
@@ -56,6 +67,8 @@ class Cas1AutoScript(
         val roles = user.roles.map { it.role }.joinToString(", ")
         seedLogger.info("  -> User '${user.name}' (${user.deliusUsername}) seeded with roles $roles")
```
edavey, line 68: I was thinking it might be helpful if we also logged `SeedUser.documentation` here to give a message like:

> -> User Fred Bloggs (Fred.Bloggs) seeded with roles UserRole.CAS1_MANAGER - For local use in development and testing

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/Cas1AutoScript.kt
```diff
@@ -84,16 +97,42 @@ class Cas1AutoScript(
...
+      SeedUser(
+        username = "LAOFULLACCESS",
+        roles = listOf(
+          UserRole.CAS1_ASSESSOR,
+          UserRole.CAS1_MATCHER,
+          UserRole.CAS1_MANAGER,
+          UserRole.CAS1_WORKFLOW_MANAGER,
+          UserRole.CAS1_ADMIN,
+          UserRole.CAS1_REPORT_VIEWER,
+          UserRole.CAS1_APPEALS_MANAGER,
```
edavey, line 109: I curious as to why you grant **all** the roles at this point? Are there scenarios where ADMIN and REPORT_VIEWER experience different behaviour around LAOs?

edavey (approved):
This looks excellent, thanks for pushing ahead with clarifying the handling of LAO scenarios.

PR #1990, opened 2024-06-25
APS-875 - Add seed job to link adhoc booking to placement request

davidatkinsuk:
Whilst users should always create bookings linked to a corresponding placement_request (when a placement request exists), it’s logically possible in the system to create an adhoc booking for the same requested dates for the same CRN. Whilst this is discouraged (as are use of adhoc bookings in general), it’s still possible. In this scenario, the placement_request is still showing as ‘not matched’, despite the adhoc booking existing. This is because the placement_request isn’t linked to the adhoc booking.

This seed job has been created to link an adhoc booking to the placement_request that it should have been created from, to help users recover from the scenario described above (stop showing the placement_request as ‘not matched)

This will only work for adhoc bookings where the target placement_request isn’t already associated with a booking, and it exists on the same application as the booking (something that won’t always be the case for adhoc bookings). These constraints are deliberately restrictive to minimimse the amount of manual testing required for this manual data ‘correction’


RobBoothMOJ (approved):
Approved

PR #1992, opened 2024-06-26
CAS-434 Fix an issue when calling person departure update domain event

muhammad-elabdulla:
This [PR CAS-434](https://dsdmoj.atlassian.net/browse/CAS-434) is to fix an issue for the person departure update domain event to make sure that the event type raised is personDepartureUpdated

Also add checks for the another tests in the Domain Event Builder to assert the type of the domain events


davidatkinsuk (approved):
LGTM 👍 

I'm assuming that we're not going to try and 'fix' any existing domain entries that had the incorrect type and then replay? (i suspect there may no way of differentiating them from personDeparted events)

muhammad-elabdulla:
The domain entries have the right type personDepartureUpdated

PR #1994, opened 2024-06-27
APS-966 - Remove unnecessary table locks on app submission

davidatkinsuk:
When an application is submitted we acquire a pessimistic lock. Prior to this commit the lock would apply to any eagerly loaded relationship on the application, unneccessarily locking tables that would not be updated (e.g. reference data).

In the current version of the code this results in a minor performance issue, but we’ve found that when upgrading to the latest version of hibernate locking across these relationships is no longer possible, and the locks are not being acquired, leading to tests failing.

This commit reduces the scope of locking by introducing a new simplified version of the application entity that defines no relationships. Locking on this entity instead results in the required ‘select for update’ type lock, but only on the applications table itself e.g.

`Hibernate: select simpleappl0_.id as id2_3_ from applications simpleappl0_ where simpleappl0_.id=? for update of simpleappl0_`

Attempts were made to do this directly via the EntityManager, but in all cases the lock was never being acquired.


danhumphreys-moj (approved):
Have tested these changes on the spring boot upgrade branch and the locking issues no longer occur.

PR #1995, opened 2024-06-27
APS-964 - Remove unused person types

davidatkinsuk:
The FullPersonInfo and RestrictedPersonInfo types are not used in the API or any UI code.

These are being removed to avoid confusion with new types that we’re going to add to differentiate Person Summary responses


Rosie-Brigham (approved):
👍

PR #1996, opened 2024-06-27
Improve readability of task allocation logic

davidatkinsuk:


src/main/resources/static/codegen/built-cas3-api-spec.yml
```diff
@@ -327,6 +327,7 @@ components:
           $ref: '#/components/schemas/Task'
         users:
           type: array
+          description: Users to which this task can be allocated
```
edavey, line 330: I wouldn't normally nick pick like this but as this commit is all about polishing language then I think "Users to _whom_ this task can be allocated" may be a bit clearer

edavey (approved):

src/main/resources/static/codegen/built-cas3-api-spec.yml
davidatkinsuk: Yes, agreed. Fixed :)

PR #1997, opened 2024-06-27
Cas 328 store updated release and required from dates

daveawc:
A PR for [ CAS-328 ](https://dsdmoj.atlassian.net/browse/CAS-328) to store and updated release date or accommodation required from date for a temporary accommodation assessment. Added new columns to the temporary_accommodation_assessments table, and the new values have been added as fields of the response and get GET request.

I think we need more discussion about the 'arrival date' etc fields, currently this information is stored in multiple areas (in the application json itself, in the temporary_accommodation_application table, the booking table, and now the temporary_accommodation_assessment table). There's a requirement for the new fields for reporting, however there is better ways that this can be achieved, as this will fast become unmanigable.

### DIFF
--- a/src/main/resources/static/api.yml
+++ b/src/main/resources/static/api.yml
@@ -3719,7 +3719,7 @@ paths:
         - name: X-Service-Name
           in: header
           required: false
-          description: If given, only premises for this service will be returned
+          description: Required for operations specific to each CAS
           schema:
             $ref: '_shared.yml#/components/schemas/ServiceName'
         - in: path


src/main/resources/db/migration/all/20240627100000__update_temporary_accommodation_assessments.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE temporary_accommodation_assessments
+    ADD COLUMN release_date DATE NULL;
+ALTER TABLE temporary_accommodation_assessments
+    ADD COLUMN accommodation_required_from_date DATE NULL;
```
Rosie-Brigham, line 4: @froddd just tagging you in here as this change may affect your work on editing release dates

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/factory/TemporaryAccommodationAssessmentEntityFactory.kt
```diff
@@ -12,6 +12,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ReferralRejec
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.TemporaryAccommodationAssessmentEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.randomDateTimeBefore
+import java.time.LocalDate
```
Rosie-Brigham, line 15: As they are more precise, shall we default all dates from now on to dateTimes?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas3/Cas3AssessmentService.kt
```diff
@@ -0,0 +1,72 @@
...
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.findAssessmentById
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.results.AuthorisableActionResult
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.results.ValidatableActionResult
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.UserAccessService
+import java.time.LocalDate
+import java.util.UUID
+
+@Service
+class Cas3AssessmentService(
```
davidatkinsuk, line 18: good to see these being split into CAS specific services instead of adding more if/elses :)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas3/Cas3AssessmentService.kt
```diff
@@ -0,0 +1,72 @@
...
+    val currentReleaseDate = getCurrentReleaseDate(assessment)
+    val currentAccommodationDate = getCurrentAccommodationDate(assessment)
+
+    updateAssessment.releaseDate?.apply {
+      if (this.isAfter(currentAccommodationDate)) return notAfterValidationResult(currentAccommodationDate)
+      assessment.releaseDate = this
+    } ?: updateAssessment.accommodationRequiredFromDate?.apply {
+      if (this.isBefore(currentReleaseDate)) return notBeforeValidationResult(currentReleaseDate)
+      assessment.accommodationRequiredFromDate = this
+    }
```
davidatkinsuk, line 52: i'll be honest, i'm slightly confused by the 'if releaseDate.apply returns null we instead update requiredFromDate' construct. Given the check on line 37 checks that at least one is null i think it's fine to just put the two updates on completely separate lines? (E.g. no need for the ?:). I could be misreading this though!

davidatkinsuk (approved):
I learnt some new things from this PR so that's always a bonus :) 

No functional issues that i can see, just some proposed improvements. Approved but would appreciate checking my comments and taking action as appropriate before merging

One thing to consider for future PRs (it'll be painful to do it now). We have a general principle of keeping commits as small as reasonable to make it easier to review the code. This isn't always possible or practical of course but i think the case of this PR it _could_ have been split into the following commits:

* introduce cas3assessment service
* general cas3assessment formatting tidy up
* the new functionality you've introduced

I use a tool called Tower which makes splitting and rejoining commits reasonably straight forward.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas3/Cas3AssessmentService.kt
```diff
@@ -0,0 +1,72 @@
...
+
+  private fun getCurrentAccommodationDate(assessment: TemporaryAccommodationAssessmentEntity): LocalDate =
+    assessment.accommodationRequiredFromDate
+      ?: assessment.typedApplication<TemporaryAccommodationApplicationEntity>().arrivalDate?.toLocalDate()!!
+
+  private fun notBeforeValidationResult(existingDate: LocalDate): AuthorisableActionResult.Success<ValidatableActionResult<AssessmentEntity>> =
+    AuthorisableActionResult.Success(ValidatableActionResult.GeneralValidationError("Accommodation required from date cannot be before release date: $existingDate"))
+
+  private fun notAfterValidationResult(existingDate: LocalDate): AuthorisableActionResult.Success<ValidatableActionResult<AssessmentEntity>> =
+    AuthorisableActionResult.Success(ValidatableActionResult.GeneralValidationError("Release date cannot be after accommodation required from date: $existingDate"))
```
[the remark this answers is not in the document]
davidatkinsuk: The nice thing about CasResult is you don't have this weird construct of 

```Success(Error)```

Which always confuses me!

PR #1999, opened 2024-06-27
APS-922 : Capture domain event for Placement Application Decisions

RobBoothMOJ:
Definition of the domain event and endpoint for retrieval added to API. 
Functions added to both save and get domain events for placement application decisions. 
Save function called in PlacementApplicationService- recordDecision DomainEventDescriber added.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/domainevents/DomainEventDescriber.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/domainevents/DomainEventDescriber.kt
@@ -2,6 +2,7 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.domainevents
 
 import org.springframework.data.repository.findByIdOrNull
 import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.RequestForPlacementAssessed
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.RequestForPlacementType
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.StaffMember
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.AssessmentClarificationNoteRepository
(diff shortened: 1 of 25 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/domainevents/DomainEventDescriber.kt
```diff
@@ -198,6 +200,21 @@ class DomainEventDescriber(
...
+
+    return event.describe { data ->
+      val details = data.eventDetails
+
+      val description = when (details.decision) {
+        RequestForPlacementAssessed.Decision.accepted -> "A request for placement assessment was accepted."
+        RequestForPlacementAssessed.Decision.rejected -> "A request for placement assessment was rejected."
+      }
+      val summary = details.decisionSummary?.let { "The reason was: $it" } ?: ""
+      "$description $summary"
```
davidatkinsuk, line 214: nice 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/DomainEventEntity.kt
```diff
@@ -266,6 +269,11 @@ enum class DomainEventType(val typeName: String, val typeDescription: String, va
     "An Approved Premises Request for Placement has been created",
     TimelineEventType.approvedPremisesRequestForPlacementCreated,
   ),
+  APPROVED_PREMISES_REQUEST_FOR_PLACEMENT_ASSESSED(
+    Cas1EventType.requestForPlacementAssessed.value,
+    "An assessment has been made on an Approved Premises Request",
```
davidatkinsuk, line 274: I think this should probably be 'A request for placement has been assessed'?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1PlacementApplicationDomainEventService.kt
```diff
@@ -181,4 +185,51 @@ class Cas1PlacementApplicationDomainEventService(
...
+    val eventOccurredAt = Instant.now()
+    val application = placementApplication.application
+
+    val assessor = assessedByUser.let { domainEventTransformer.toStaffMember(it) }
+
+    val assessmentDecision = when (placementApplication.decision) {
+      PlacementApplicationDecision.ACCEPTED -> RequestForPlacementAssessed.Decision.accepted
+      PlacementApplicationDecision.REJECTED -> RequestForPlacementAssessed.Decision.rejected
+      null -> throw IllegalArgumentException("PlacementApplicationDecision was null")
+      else -> throw IllegalArgumentException("PlacementApplicationDecision '${placementApplication.decision}' has been deprecated")
```
davidatkinsuk, line 204: I think it may be a tiny bit more resilient to future changes if we explicitly include the two remaining cases instead of using an 'else'. Outcome for both would be identical of course. You can specify both cases on one line using a comma to separate them to reduce code duplication, see https://kotlinlang.org/docs/control-flow.html#when-expression

davidatkinsuk (approved):
LGTM, thanks for the thorough automated testing. I've had a few minor comments.

As discussed it would be good to get some screenshots from the UI showing the E2E working (mainly because this is liable to go live as soon as merged, so we want to make sure it doesn't cause any issues on the timeline)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1PlacementApplicationDomainEventService.kt
RobBoothMOJ: Good call.   I had them separately but ran into a code complexity issue.

PR #2007, opened 2024-07-02
APS-978 - Add Update NOMS Number Seed Job

davidatkinsuk:
This commit adds a seed job that allows us to update the NOMS number associated with a given CRN on applicationa and bookings. This is required when the NOMS number is updated in delius, which is rare, but can happen.

This helps reduce the number of 404 errors we see when the inmate cache is refreshed because we’re removing usage of invalid NOMS numbers.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/SeedUpdateNomsNumberSeedJobTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/SeedUpdateNomsNumberSeedJobTest.kt
@@ -1,0 +1,127 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.integration
+
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.TestInstance
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.data.repository.findByIdOrNull
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.SeedFileType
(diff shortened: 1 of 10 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/SeedUpdateNomsNumberSeedJobTest.kt
```diff
@@ -0,0 +1,93 @@
...
+      ),
+    )
+
+    seedService.seedData(SeedFileType.updateNomsNumber, "valid-csv")
+
+    val updatedApplication = approvedPremisesApplicationRepository.findByIdOrNull(application.id)
+    assertThat(updatedApplication!!.nomsNumber).isEqualTo(NEW_NOMS)
+
+    val updatedBooking = bookingRepository.findByIdOrNull(booking.id)
+    assertThat(updatedBooking!!.nomsNumber).isEqualTo(NEW_NOMS)
```
edavey, line 73: would it be worth having a couple more "belt and braces" assertions here to verify that we aren't updating other (or all!) applications / bookings e.g.

```
assertThat(reloadedOtherApplication!!.nomsNumber).isEqualTo(OTHER_NOMS)
assertThat(reloadedOtherBooking!!.nomsNumber).isEqualTo(OTHER_NOMS)
```

edavey (approved):
Just a minor suggestion on perhaps including some further assertions on what we're updating...

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/SeedUpdateNomsNumberSeedJobTest.kt
davidatkinsuk: Good idea, done!

PR #2008, opened 2024-07-02
CAS-434 Fix issue with existing person departure updated domain event

muhammad-elabdulla:
This [PR CAS-434](https://dsdmoj.atlassian.net/browse/CAS-434) is to create a job to fix the current person departure domain events with incorrect event type. 

This job will update the event type in the data field for the current domain events from  **accommodation.cas3.person.departed** to **accommodation.cas3.person.departed.updated**

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/DomainEventEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/DomainEventEntity.kt
@@ -79,6 +79,8 @@ interface DomainEventRepository : JpaRepository<DomainEventEntity, UUID> {
 
   fun findByApplicationId(applicationId: UUID): List<DomainEventEntity>
 
+  fun findByType(type: DomainEventType): List<DomainEventEntity>
+
   fun getByApplicationIdAndType(applicationId: UUID, type: DomainEventType): DomainEventEntity
 
   @Modifying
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/DomainEventEntity.kt
```diff
@@ -108,7 +110,7 @@ data class DomainEventEntity(
   val occurredAt: OffsetDateTime,
   val createdAt: OffsetDateTime,
   @Type(type = "com.vladmihalcea.hibernate.type.json.JsonType")
-  val data: String,
+  var data: String,
```
davidatkinsuk, line 113: you can use the existing DomainEventRepository.updateData() function, so no need to make this mutable

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/Cas3UpdateDomainEventTypeForPersonDepartureUpdatedJobTest.kt
```diff
@@ -0,0 +1,69 @@
...
+            CAS3PersonDepartureUpdatedEvent(
+              id = UUID.randomUUID(),
+              timestamp = Instant.now().randomDateTimeBefore(),
+              eventType = EventType.personDepartureUpdated,
+              eventDetails = CAS3PersonDepartedEventDetailsFactory().produce(),
+            ),
+          ),
+        )
+      }
+    }.take(3).toList()
```
davidatkinsuk, line 49: can we please add a domain event of some other type and check it's unmodified after executing the seed job?

davidatkinsuk (approved):
Looks good, just a few small comments from me to address

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/Cas3UpdateDomainEventTypeForPersonDepartureUpdatedJobTest.kt
muhammad-elabdulla: Added another domain events

PR #2009, opened 2024-07-02
APS-987 - Add config to exclude NOMS from cache refresh

davidatkinsuk:
When a NOMS number can’t be found in the prisons-api, we receive a 404. If this occurs multiple times when refreshing inmate details in the cache (via InmateDetailsCacheRefreshWorker), we start to receive Sentry errors that this has occurred.

It’s possible that the resolution to the issue (typically Delius or NOMIS having the wrong NOMS number for the CRN) may take days or weeks. In the meantime we no longer need to see the cache refresh warnings, and we should stop attempting to refresh the inmate details as it will fail.

This commit provides a mechanism to exclude NOMS from inmate details cache refresh. Entries can be added to the cache_refresh_exclusions_inmate_details table using SQL

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cache/preemptive/InmateDetailsCacheRefreshWorker.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cache/preemptive/InmateDetailsCacheRefreshWorker.kt
@@ -6,10 +6,13 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.PreemptiveCacheEn
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.PrisonsApiClient
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApplicationRepository
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.BookingRepository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.CacheRefreshExclusionsInmateDetailsRepository
 
+@SuppressWarnings("LongParameterList")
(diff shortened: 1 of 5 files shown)


edavey (approved):
will you be setting these excluded NOMs numbers by a migration job?

davidatkinsuk:
> will you be setting these excluded NOMs numbers by a migration job?

My plan was to just use INSERT SQL against prod given it's a 'low risk' table, but if you'd prefer i can add a seed job that adds and removes from the list (I guess we can't use SQL migration scripts because then real CRNs/NOMs would be in the public github)

edavey:
> > will you be setting these excluded NOMs numbers by a migration job?
> 
> My plan was to just use INSERT SQL against prod given it's a 'low risk' table, but if you'd prefer i can add a seed job that adds and removes from the list (I guess we can't use SQL migration scripts because then real CRNs/NOMs would be in the public github)

yes makes sense - I think you're right that in this case it's OK to use the SQL console. 

One advantage of having a script which runs against an uploaded CSV might be that the script could output a list of all the current exclusions -- which ought to allow us to spot if anything has gone haywire when setting exclusions via the console...

davidatkinsuk:
> > > will you be setting these excluded NOMs numbers by a migration job?
> > 
> > 
> > My plan was to just use INSERT SQL against prod given it's a 'low risk' table, but if you'd prefer i can add a seed job that adds and removes from the list (I guess we can't use SQL migration scripts because then real CRNs/NOMs would be in the public github)
> 
> yes makes sense - I think you're right that in this case it's OK to use the SQL console.
> 
> One advantage of having a script which runs against an uploaded CSV might be that the script could output a list of all the current exclusions -- which ought to allow us to spot if anything has gone haywire when setting exclusions via the console...

It's a fair point, we can revisit if we find ourselves using these exclusions alot (i think this is the first time i've seen an issue where we've not being able to 'fix' the application's noms number)

PR #2012, opened 2024-07-03
APS-964 : Task API - add PersonSummary

RobBoothMOJ:
PersonSummary includes specific information on the type of person - regular, lao, not found or unknown. 
It supersedes 'personName' which was used to derive this information and is now deprecated.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/TasksController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/TasksController.kt
@@ -17,7 +17,6 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.TaskSortField
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.TaskType
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.TaskWrapper
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.convert.EnumConverterFactory
-import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApplicationEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.AssessmentEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.PlacementApplicationEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.PlacementRequestEntity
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/TaskTransformer.kt
```diff
@@ -96,6 +105,35 @@ class TaskTransformer(
...
+      is PersonSummaryInfoResult.NotFound, is PersonSummaryInfoResult.Unknown -> PersonSummaryDiscriminator.unknownPersonSummary
+    }
+
+    val name: String? = if (personType == PersonSummaryDiscriminator.fullPersonSummary) {
+      getNameFromPersonSummaryInfoResult(offenderSummary)
+    } else {
+      null
+    }
+
+    return FullPersonSummary(crn = application.crn, personType = personType, name = name)
```
davidatkinsuk, line 134: I think we should be returning one of three types here:

* FullPersonSummary
* RestrictedPersonSummary
* UnknownPersonSummary

I think the other 2 will need to be defined in the yml, but neither will have any additional fields other than what is inherited from #/components/schemas/PersonSummary.

This can probably be shifted into the when block on line 122, and what ever that returns can be immediately returned as the function result

Take a look at PersonTransformer.transformModelToPersonApi for how we do this for the non-summary types

src/main/resources/static/_shared.yml
```diff
@@ -255,8 +255,11 @@ components:
           type: string
           format: uuid
           example: 6abb5fa3-e93f-4445-887b-30d081688f44
+        personSummary:
+          $ref: '#/components/schemas/PersonSummary'
         personName:
           type: string
+          deprecated: true
```
davidatkinsuk, line 262: could we please add a 'description' property describing what should be used instead of this deprecated field (personSummary). If you search for other deprecate fields in this file it'll provide an example

src/main/resources/static/_shared.yml
```diff
@@ -782,6 +786,22 @@ components:
             - dateOfBirth
             - sex
             - status
+    PersonSummary:
+      type: object
+      properties:
+        crn:
+          type: string
+        name:
```
davidatkinsuk, line 794: I think 'name' should only be a property on the FullPersonSummary, so should be moved into the FullPersonSummary definition. This should be a 'required' field

src/main/resources/static/_shared.yml
```diff
@@ -4568,6 +4588,15 @@ components:
...
         - UnknownPerson
+    FullPersonSummary:
+      allOf:
+        - $ref: '#/components/schemas/PersonSummary'
+    PersonSummaryDiscriminator:
+      type: string
+      enum:
+        - FullPersonSummary
+        - RestrictedPersonSummary
+        - UnknownPersonSummary
```
davidatkinsuk, line 4599: Like FullPersonSummary: I think we also need a RestrictedPersonSummary: and UnknownPersonSummary:. Neither of these need to provide additional properties e.g.
```
    RestrictedPersonSummary:
      allOf:
        - $ref: '#/components/schemas/PersonSummary'
    UnknownPersonSummary:
      allOf:
        - $ref: '#/components/schemas/PersonSummary'         
```

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/TasksController.kt
```diff
@@ -305,7 +303,7 @@ class TasksController(
   ): PlacementRequestTask {
     return taskTransformer.transformPlacementRequestToTask(
       placementRequest = placementRequest,
-      personName = getPersonNameFromApplication(placementRequest.application, offenderSummaries),
+      offenderSummaries = offenderSummaries,
```
davidatkinsuk, line 306: nice!

davidatkinsuk (changes requested):
A few comments from me to address.

davidatkinsuk (approved):
Looks great! 👍

PR #2015, opened 2024-07-03
[APS-936] Introduce `CAS1_JANITOR` user role as prerequisite for out-of-service bed cancellation

danlivings-dxw:
> See [APS-936 on the Approved Premises Jira board](https://dsdmoj.atlassian.net/browse/APS-936).

This PR introduces the `CAS1_JANITOR` user role and applies it as a prerequisite for a user to access the `POST /premises/{premisesId}/out-of-service-beds/{outOfServiceBedId}/cancellations` endpoint.

This new role is intended to be expanded in future to allow a small number of users various permissions to modify or remove data beyond the capabilities of regular users.


edavey (approved):
👍 thanks for implementing the new style `UserRole` enum class

PR #2016, opened 2024-07-03
APS-953: Enforce presence of OutOfServiceBedRevision.notes

edavey:
We will be migrating existing LostBed records to the new OutOfServiceBed models. As `LostBed.notes` are nullable we won't enforce the presence of notes on the `OutOfServiceBedRevision` entity yet. Instead we use a validation step within the `OutOfServiceBedService`.

Once the new OOSB system is released and the `LostBed` data has been migrated we can make `OutOfServiceBedRevision.notes` non-nullable


danlivings-dxw (commented):
I think that the unit test is failing because the new check has only been added to the `createOutOfServiceBed` method and the tests have only been changed for `updateOutOfServiceBed`.

edavey:
> I think that the unit test is failing because the new check has only been added to the `createOutOfServiceBed` method and the tests have only been changed for `updateOutOfServiceBed`.

Thanks @danlivings-dxw, I'm not sure how I missed that!

edavey:
@danlivings-dxw Could I get your eyes on this again please?

danlivings-dxw (approved):

PR #2020, opened 2024-07-03
CAS-408 set existing move on category to false, added new category

daveawc:
deactivated move on category, and created a new one. [https://dsdmoj.atlassian.net/browse/CAS-408](url)


froddd (approved):
LGTM

PR #2021, opened 2024-07-03
APS-1001: make CAS1 out-of-service date names consistent

edavey:
We're currently using:

- `startDate` and `endDate ` when creating new OOSB records and making revisions (POST and PUT)

- `outOfServiceFrom` and `outOfServiceTo` when representing stored records and revisions (GET)

In this commit we harmonise the naming to use `startDate ` and `endDate ` throughout.

This consistency makes work on the UI clearer.


danlivings-dxw (approved):
Looks good to me, thanks for cleaning this up.

PR #2022, opened 2024-07-04
APS-922 : API - Domain Event for Placement Application Decisions

RobBoothMOJ:
Added arrival date and duration to the domain event to clarify which request the decision is for.
Existing Request For Placement decision (assessed) domain events are cleared from the DB.
Recording this domain event isn't actually live yet so clearing them out is not a problem.

### DIFF
--- a/src/main/resources/db/migration/all/20240704135216__delete_assessed_request_for_placements_domain_events.sql
+++ b/src/main/resources/db/migration/all/20240704135216__delete_assessed_request_for_placements_domain_events.sql
@@ -1,0 +1,10 @@
+DELETE FROM domain_events_metadata
+WHERE
+    domain_event_id IN (
+        SELECT ID FROM domain_events
+        WHERE domain_events.type = 'APPROVED_PREMISES_REQUEST_FOR_PLACEMENT_ASSESSED'
+     );
+
+DELETE FROM domain_events
+WHERE
+    domain_events.type = 'APPROVED_PREMISES_REQUEST_FOR_PLACEMENT_ASSESSED';
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/domainevents/DomainEventDescriberTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/domainevents/DomainEventDescriberTest.kt
@@ -528,6 +528,7 @@ class DomainEventDescriberTest {
   @EnumSource(value = RequestForPlacementAssessed.Decision::class)
   fun `Returns expected description for request for placement assessed event with summary`(decision: RequestForPlacementAssessed.Decision) {
     val domainEventSummary = DomainEventSummaryImpl.ofType(DomainEventType.APPROVED_PREMISES_REQUEST_FOR_PLACEMENT_ASSESSED)
+    val expectedTermUsed = if (decision == RequestForPlacementAssessed.Decision.rejected) "was" else "is"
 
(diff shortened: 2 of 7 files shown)


src/main/resources/db/migration/all/20240704135216__delete_assessed_request_for_placements_domain_events.sql
```diff
@@ -0,0 +1,10 @@
+DELETE FROM domain_events_metadata
+WHERE
+    domain_event_id IN (
+        SELECT ID FROM domain_events
+        WHERE domain_events.type = 'APPROVED_PREMISES_REQUEST_FOR_PLACEMENT_ASSESSED'
+     );
```
davidatkinsuk, line 6: i hadn't considered clearing the metadata too, good catch

davidatkinsuk (changes requested):
As discussed i think we need the describer updating to show this newly captured data, similar to the other placement request domain event. I think the buildRequestForPlacementDescription function in DomainEventDescriber can be reused

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/domainevents/DomainEventDescriberTest.kt
```diff
@@ -537,21 +538,24 @@ class DomainEventDescriberTest {
...
+          .withDuration(7)
           .produce(),
       )
     }
 
     val result = domainEventDescriber.getDescription(domainEventSummary)
 
     assertThat(result).isEqualTo(
-      "A request for placement assessment was $decision. The reason was: $decision",
+      "A request for placement assessment was $decision. The reason was: $decision The placement request $expectedTermUsed for Friday 3 May 2024 to Friday 10 May 2024 (1 week)",
```
davidatkinsuk, line 550: i think there may be a full stop missing here after decision?

davidatkinsuk (approved):
just one minor thing from me 👍

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/domainevents/DomainEventDescriberTest.kt
RobBoothMOJ: There is, it's relying (assuming) that the free-text entered is a complete sentence ending with a full stop. 

A few options - 

1. Leave it and hope the text entered is good (possible missing full stop). 
2. Add a full stop (might mean multiple full stops are present in the rendered string).
3. Parse the string and ensure that there is always only 1 full stop present.

RobBoothMOJ: I'll change the test to include a narrative.

davidatkinsuk: Yeah that's tricky. I guess another alternative is to re-order the message so the reason always appears last e.g.

```A request for placement assessment was $decision. The placement request $expectedTermUsed for Friday 3 May 2024 to Friday 10 May 2024 (1 week). The reason was: $decision```

My second preference is to add a fullstop and risk having two.

RobBoothMOJ: Hopefully covered those both off now.

davidatkinsuk (approved):
Looks great 👍

PR #2023, opened 2024-07-04
Fix performance issue when generate CAS3 utilisation report for all regions

muhammad-elabdulla:
This [PR CAS-414](https://dsdmoj.atlassian.net/browse/CAS-414) is to fix a performance issue when generate utilisation report in CAS3 for all regions. Which includes:
- Reduce the amount of data retrieved in the report queries by changing it to sql queries 
- Get all the information needed for bedspace, bookings and lost bed space in one query instead of calling the database for each row in the report
- Return only the data related to temporary accommodation which is needed in the report 
- Add integration tests to check the new report queries


davidatkinsuk:
Hi @muhammad-elabdulla , could you please summarise how this differs from the PR I reviewed previously but was then reverted due to performance issues?

muhammad-elabdulla:
Hi @davidatkinsuk 
There was some issues in the sql queries with joins between tables and I added integration tests to cover them in CAS3ReportsTest

davidatkinsuk (approved):
LGTM given this is almost identical to https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/2000/files (which i've already reviewed), but with some fixes to joins in the SQL 👍

PR #2025, opened 2024-07-05
APS-1010 - Don't present withdrawn requests for placement for assessment

davidatkinsuk:
Commit https://github.com/ministryofjustice/hmpps-approved-premises-api/commit/c4b088c11b8fd0269dc3a85f74962a1f1130016a removed the usage of placement_application.decision to store if a request for placement was withdrawn. Due a lack of regression we didn’t capture that this modified the behaviour of the /tasks endpoint as as such it started to return withdrawn requests for placements (because the SQL relied on the withdrawn state being set in the decision).

This commit fixes that logic and adds regression tests to ensure withdrawn requests for placements and assessments are not returned by the endpoint.


RobBoothMOJ (approved):
Approved

PR #2027, opened 2024-07-05
CAS2-506 Add new sub status under Referral Cancelled.

danhumphreys-moj:
Ref: https://dsdmoj.atlassian.net/browse/CAS2-506

New sub status required to cover scenario when prisoners are transferred to another prison and the referral has to be cancelled titled: "Person transferred to another prison".


muhammad-elabdulla (approved):

muhammad-elabdulla:
LGTM

PR #2029, opened 2024-07-05
Revert removal of generated open api YML files

davidatkinsuk:
The YML files in the folder are generated by the openApiGenerate gradle task.

Whilst it's generally bad practice to check in code that can be generated on the fly, we currently required these files to available in the git hub repository to allow the user interface projects to bull the latest 'compiled' open api specs for the project.

A previous commit removed these generated files as we hadn’t appreciated how they were being used


richpjames:
In case it's helpful in future, this is where the CAS1 repo uses these autogenerated specs: https://github.com/ministryofjustice/hmpps-approved-premises-ui/blob/123783bccc1173532b51f6f1f4d4dcd989e7e42c/server/testutils/jest.setup.ts#L44

richpjames (approved):

PR #2031, opened 2024-07-05
[APS-954] Introduce out-of-service beds report

danlivings-dxw:
> See [APS-954 on the Approved Premises Jira board](https://dsdmoj.atlassian.net/browse/APS-954).

This PR introduces a new report for out-of-service beds, accessible using the `GET /cas1/reports/outOfServiceBeds` endpoint. This report is functionally almost identical to the existing lost beds report and is intended as a drop-in replacement for the CAS1 use case.


edavey (approved):
looks goods to me, thks

PR #2035, opened 2024-07-08
CAS-328 - update logic and add test coverage

daveawc:
corrected an issue where the dates were reading from the wrong field, updated tests to cover it. Also included generated files (🤮). Added more tests and improved existing coverage.


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/AssessmentTest.kt
```diff
@@ -2697,7 +2698,7 @@ class AssessmentTest : IntegrationTestBase() {
   }
 
   @Test
-  fun `GET temporary accommodation assessment contains release date and accommodation required from date`() {
+  fun `GET temporary accommodation assessment contains updated release date and accommodation required from date`() {
```
froddd, line 2701: To make this test more explicit, would it be worth creating the application with `withPersonReleaseDate(originalDate)` and `withArrivalDate(originalDate)`, then creating the assessment with `withReleaseDate(updatedDate)` and `withAccommodationRequiredFromDate(updatedDate)` then asserting that the dates are equal to the `updatedDate`?

froddd (approved):
Looks good, one suggestion to differentiate both tests more explicitly

froddd (approved):
Looks good!

PR #2036, opened 2024-07-09
Fix performance issue when generate CAS3 utilisation report for all regions

muhammad-elabdulla:
This [PR CAS-414](https://dsdmoj.atlassian.net/browse/CAS-414) is to fix a performance issue when generate utilisation report in CAS3 for all regions. Which includes:
- Reduce the amount of data retrieved in the report queries by changing it to sql queries 
- Get all the information needed for bedspace, bookings and lost bed space in one query instead of calling the database for each row in the report
- Return only the data related to temporary accommodation which is needed in the report 
- Add integration tests to check the new report queries

This PR is different from previous one:
- Move the Booking cancellations and turnaround in separate sql queries.
- Add new integration tests to cover the scenarios when there are  multiple booking cancellations and turnarounds for the same bedspace during the report period


davidatkinsuk (approved):
Approved 👍

PR #2038, opened 2024-07-09
APS-1034 : Capture Cancellation ID in Domain Event metadata 

RobBoothMOJ:
Added CAS1_CANCELLATION_ID metadataName and record cancellation ID in domain event metadata.  
Migration to backfill the domain event metadata with the cancellation id to be able access 'other' narrative text.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
```diff
@@ -1253,6 +1254,9 @@ class BookingService(
             cancellationReason = reason.name,
           ),
         ),
+        metadata = mapOf(
+          MetaDataName.CAS1_CANCELLATION_ID to booking.cancellations.first().id.toString(),
```
davidatkinsuk, line 1258: we should have direct access to the cancellation entity in this function, can we derive the id from that instead please for clarity?

src/main/resources/db/migration/all/20240709103500__backfill_cas1_cancelled_bookings_domain_events_metadata.sql
```diff
@@ -0,0 +1,22 @@
...
+AND
+    cancellations.id::text
+	NOT IN
+		(
+		SELECT DISTINCT value
+		FROM
+		    domain_events_metadata
+        WHERE
+            name = 'CAS1_CANCELLATION_ID'
+        )
```
davidatkinsuk, line 22: nice!

davidatkinsuk (approved):
Just one minor comment from me 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
davidatkinsuk: Yep i think pass the CancellationEntity straight into createCas1CancellationDomainEvent and then get the id from that. Sorry, i should have checked, i forgot this was in it's own private function

PR #2040, opened 2024-07-09
CAS 365, = null changed to IS null, required for Spring 3.2.

danhumphreys-moj:
CAS 365, = null changed to IS null, required for Spring 3.2.


davidatkinsuk (approved):
LGTM 👍 

(i suspect this migration job is also a good candidate for removal, we should review that one day :) )

PR #2045, opened 2024-07-10
Remove generic /reports endpoint

davidatkinsuk:
Each cas now has a specific /casx/reports endpoint. This endpoint hasn’t been used in production since 24/6/24.

See ticket for screenshot showing that this endpoint hasn't been used since 24/6/24 https://dsdmoj.atlassian.net/browse/APS-948


davidatkinsuk:
closing this, will just remove the CAS1 endpoints

PR #2048, opened 2024-07-10
APS-948 Repoint all CAS1 report tests to use /cas1/reports

davidatkinsuk:
The /reports endpoint is deprecated and should no longer be used.

The UI is using the /cas1/reports so this update ensures the tests better reflect usage of the API in prod


RobBoothMOJ (approved):
Approved

PR #2049, opened 2024-07-10
APS-1000: Provide cancelled booking information on Timeline

RobBoothMOJ:
Added original booking information to the description which includes detail about the booked premises and booking start and end dates. 
Added user inputted information text to timeline when reason for cancelling is 'Other' and a text narrative has been added.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/domainevents/DomainEventDescriber.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/domainevents/DomainEventDescriber.kt
@@ -6,6 +6,8 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.Request
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.RequestForPlacementType
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.StaffMember
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.AssessmentClarificationNoteRepository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.BookingRepository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.CancellationReasonRepository
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.DomainEventType
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.DomainEvent
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.DomainEventSummary
@@ -24,6 +26,7 @@ import java.util.UUID
 class DomainEventDescriber(
   private val domainEventService: DomainEventService,
   private val assessmentClarificationNoteRepository: AssessmentClarificationNoteRepository,
+  private val bookingRepository: BookingRepository,
 ) {
 
   @SuppressWarnings("CyclomaticComplexMethod")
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/domainevents/DomainEventDescriber.kt
```diff
@@ -121,9 +123,22 @@ class DomainEventDescriber(
...
   private fun buildBookingCancelledDescription(domainEventSummary: DomainEventSummary): String? {
     val event = domainEventService.getBookingCancelledEvent(domainEventSummary.id())
-    return event.describe { "The placement was cancelled. The reason was: '${it.eventDetails.cancellationReason}'" }
+    return event.describe { data ->
+      val booking = bookingRepository.findByIdOrNull(data.eventDetails.bookingId)
+        ?: throw RuntimeException("Booking ID ${data.eventDetails.bookingId} with cancellation not found")
+      if (booking.cancellations.count() != 1) {
+        throw RuntimeException("Booking ID ${data.eventDetails.bookingId} does not have one cancellation")
+      }
+      val cancellation = booking.cancellations.first()
```
davidatkinsuk, line 135: we should be using the cancellation id that added as metadata via https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/2038

But - we'll need to start adding the metadata to the DomainEventSummary object to achieve this :( I think it may just be a case of adding `d.metadata` to the query in DomainEventRepository.findAllTimelineEventsByApplicationId and adding the map definition to the DomainEventSummary interface?

We can then traverse cancellation.booking to get hold of the booking. This should at least reduce the amount of potential errors we deal with here?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/domainevents/DomainEventDescriber.kt
```diff
@@ -121,9 +123,22 @@ class DomainEventDescriber(
...
+      val otherReasonText = if (cancellation.reason.name == "Other" && !cancellation.otherReason.isNullOrEmpty()) ": ${cancellation.otherReason}." else ""
```
davidatkinsuk, line 136: I _think_ it may be better to add the UUID for the 'other' reason to CancellationReasonRepository.Constants and use that to identify the 'other' reason. Neither approach is ideal but we've used that approach elsewhere.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/domainevents/DomainEventDescriber.kt
```diff
@@ -121,9 +123,22 @@ class DomainEventDescriber(
...
+
+      "A placement at ${booking.premises.name} booked for " +
+        "${booking.arrivalDate.toUiFormat()} to ${booking.departureDate.toUiFormat()} " +
+        "was cancelled. The reason was: '${data.eventDetails.cancellationReason}'$otherReasonText"
```
davidatkinsuk, line 140: nice

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/domainevents/DomainEventDescriberTest.kt
```diff
@@ -235,25 +248,157 @@ class DomainEventDescriberTest {
...
+        id = it,
+        timestamp = Instant.now(),
+        eventType = EventType.bookingCancelled,
+        eventDetails = BookingCancelledFactory()
+          .withBookingId(bookingEntity.id)
+          .withCancellationReason("Other")
+          .produce(),
+      )
+    }
+    every { mockBookingRepository.findByIdOrNull(bookingEntity.id) } returns bookingEntity
```
davidatkinsuk, line 397: thanks for the very thorough testing

davidatkinsuk (approved):
a few comments from me on this, hopefully won't be too much work

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/domainevents/DomainEventDescriber.kt
davidatkinsuk: Discussed with Rob. Getting metadata via the JPQL is proving tricky as this doesn't appear to be supported (loading in map).

We think the correct solution is to actually use the DomainEventEntity directly here, and add explicit relationships to linked entities (e.g. ApplicationEntity, BookingEntity etc.). We need to be careful these aren't loaded by default (eagerly) but in this query (or whatever we replace it with) we _do_ load these eagerly

PR #2051, opened 2024-07-10
Improve error message when cache refresh fails

davidatkinsuk:
This commit updates the error message to clarify the issue is related to refreshing the cache, and includes the URL being requested in the error message to make the requested value clearer


edavey (approved):
👍 thks for this improvement

PR #2057, opened 2024-07-11
Improve database refresh docs

davidatkinsuk:
Indicate how to check the database state in prod/preprod is the same before running the script


RobBoothMOJ (approved):
Approved

PR #2060, opened 2024-07-11
Return 500 if can’t determine user region, not 400

davidatkinsuk:
Before this commit we would return a BadRequestProblem if we could not determine the user’s region. This was incorrect becuase the fault was not in the request, but instead the issue was don’t know which region to assign the user to. Furthermore, by returning a 400 we were not raising any sentry alerts when this issue happened.

This commit changes such errors to return a RuntimeException (as this should never occur), which leads to a 500 being returned to the caller


RobBoothMOJ (approved):
Approved

PR #2064, opened 2024-07-12
CA-365 remove inadvertent use of AWS NotFoundException.

danhumphreys-moj:
Change captureException() to captureMessage() results in NotFoundException no longer needed.  See: https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1961#discussion_r1662114917


davidatkinsuk (approved):
Approving this whilst fresh in my mind. LGTM 👍 

Note that because we're now using captureMessage instead of captureException we're loosing the stack trace from the alert (i think that's why previously the code was adding a NotFoundException. But - given the alert message is very clear and unique to CAS2, i don't think this is an issue.

PR #2065, opened 2024-07-12
Add port forward script for db connections

davidatkinsuk:
This script replaces the existing remote_db script (when using the ‘-c’ argument to start the JAQY SQL CLI)


script/remote_db_tunnel.sh
```diff
@@ -0,0 +1,215 @@
...
+    echo "Forwarding port. Use ctl-c to close port forwarding"
+    kubectl -n "$NAMESPACE" port-forward $PORT_FORWARD_CONTAINER_NAME $PORT:$PORT
+else
+    # This is ran in the background so we can start the cli
+    # we use trap at the top of this script to kill it on exit
+    kubectl -n "$NAMESPACE" port-forward $PORT_FORWARD_CONTAINER_NAME $PORT:$PORT &
+    port_forward_pid=$!
+    echo "Port forward pid is $port_forward_pid"
+
+    sleep 1
```
williamfalconeruk, line 169: QQ - what's the purpose of the sleep?

davidatkinsuk: Good question - i think it was returning before the container was ready (presumably by milliseconds) but let me try removing it and seeing if it's stable as it shouldn't really be required. If not stable i'll look at adding a check to make sure the containers ready instead.

davidatkinsuk: i've removed the sleep and it seems to work without issue. you already have a 'pod ready' check earlier on in the script, so i'm not sure why i added this!

williamfalconeruk (approved):
Looking Good!

PR #2066, opened 2024-07-12
CAS-472 move on categories returning inactive ref data

daveawc:
PR to fix an issue with the order of precedence in the move on category query, as it was returning categories that had the is_active field set to false. Also improved test coverage in the area, including tests for categories that have a '*' for the service scope.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ReferenceDataTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ReferenceDataTest.kt
@@ -1,9 +1,13 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.integration
 
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Assertions.assertEquals
 import org.junit.jupiter.api.Test
 import org.springframework.beans.factory.annotation.Autowired
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.CancellationReason
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.MoveOnCategory
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ServiceName
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.MoveOnCategoryEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ProbationDeliveryUnitEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.ApAreaTransformer
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.CancellationReasonTransformer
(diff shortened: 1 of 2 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ReferenceDataTest.kt
```diff
@@ -365,9 +368,63 @@ class ReferenceDataTest : IntegrationTestBase() {
...
+    moveOnCategoryEntityFactory.produceAndPersistMultiple(10)
+    inactivateExistingMoveOnCategories()
+    moveOnCategoryEntityFactory.produceAndPersistMultiple(10)
+
+    val jwt = jwtAuthHelper.createValidAuthorizationCodeJwt()
+
+    val result =
+      webTestClient
+        .get()
+        .uri("/reference-data/move-on-categories?includeInactive=true")
```
danhumphreys-moj, line 381: Is there a test for `includeInactive=false`?  includeInactive is nullable by the looks of it; is there a test for `includeInactive=null `too?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ReferenceDataTest.kt
```diff
@@ -365,9 +368,63 @@ class ReferenceDataTest : IntegrationTestBase() {
...
+        .header("Authorization", "Bearer $jwt")
+        .exchange()
+        .expectBodyList(MoveOnCategory::class.java)
+        .returnResult()
+        .responseBody
+
+    assertThat(result!!.size).isEqualTo(moveOnCategoryRepository.count())
+    assertThat(result.filter { !it.isActive }).isNotEmpty
```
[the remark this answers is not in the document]
danhumphreys-moj: Records are cleared out from the database before each integration test so I think we should be able to confirm the number?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ReferenceDataTest.kt
danhumphreys-moj: Should `includeInactive` have a default value in the spec then?  I can't see it. 

It is not required so null, true or false could be supplied.

danhumphreys-moj: Ahh ok, I see how this works [here](https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/705b73b7c5ca9d46dcd142b150b4644b7b2d72ce/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ReferenceDataController.kt#L101).

I think the tests for not suppling it are covered off already in your other tests.

Probably just a test to check that supplying it equal to false gives the same as supplying a null value?

Would be good I think since we're covering off all bases of this signature:

```
  override fun referenceDataDepartureReasonsGet(
    xServiceName: ServiceName?,
    includeInactive: Boolean?,
  )
```

Will approve in any case.

danhumphreys-moj (approved):

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ReferenceDataTest.kt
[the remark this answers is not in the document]
danhumphreys-moj: Super - thanks for explaining.

PR #2069, opened 2024-07-15
APS-1027 Implement user manager role for user management.

RobBoothMOJ:
Add a new user manager role.
This role and the existing janitor role are given access to user management end points.


davidatkinsuk (approved):
LGTM, thanks for the thorough testing

PR #2072, opened 2024-07-16
APS-1073 - Alert on InternalServer & NotImplemented Problems

davidatkinsuk:
Whilst we already alert on unhandled exceptions, we are not alerting when the code explicitly throws an InternalServerErrorProblem or NotImplementedProblem. This commit resolves that.

As part of this change we now throw a new UnhandledExceptionProblem if we can’t handle an exception. This better describes the problem, and avoids the issue of the InternalServerErrorProblem that we used to return being alerted on again by the new code in logInternalServerErrorProblem.


edavey (approved):
👍 thanks

PR #2073, opened 2024-07-16
APS-1067 - Allow  assessor users to list assigned tasks

davidatkinsuk:
As a pre-requisite to allowing assesors to be assigned tasks, we need to ensure they can view any assigned tasks on the UI without a permissions error ocurring. This commit enables that functionality.


RobBoothMOJ (approved):
Approved

PR #2080, opened 2024-07-17
APS-1022 - Add CAS1 ‘Gender for AP’ API property

davidatkinsuk:
This being added in preparation for supporting Female APs, allowing us to add it in suitable places on the UI.

This is currently hardcoded to ‘Male’ as only applicants for Male APs are supported. As support for female APs is added this will be derived dynamically.


RobBoothMOJ (approved):
Approved

PR #2081, opened 2024-07-18
APS-1087 - Simplify Open API Task Config in Gradle

davidatkinsuk:
Introduce a function to reduce duplication of Open API Task Config


edavey (approved):
thanks for DRYing this up

PR #2082, opened 2024-07-18
[FIX][APS-1065] Ensure out-of-service beds always sort correctly

danlivings-dxw:
> See [APS-1065 on the Approved Premises Jira board](https://dsdmoj.atlassian.net/browse/APS-1065).

Currently the `GET /cas1/out-of-service-beds` does not always sort as expected. This issue was not caught by unit or integration testing, which tests all sort fields in both ascending and descending order, and in the tests sorting works correctly.

The underlying issue is that the `findAllById` method provided on the `CrudRepository` interface (and by extension, our JPA repositories) explicitly does not guarantee the order of the returned results. The order is therefore dependent entirely on how the results are returned from the database, which may be non-deterministic.

This PR therefore introduces a new utility method to wrap the `findAllById` method while guaranteeing the order of the returned elements, which solves the issue. To do this, this PR adds the `findAllByIdOrdered` method as an extension to the `JpaRepository` interface, which guarantees that the returned entities will be ordered according to the provided list of
IDs, i.e. that the first entity will have as its ID the first element of the input list, the second entity will have the second element of the input list as its ID, and so on.

This is achieved by constructing a map from the results returned by the original `findAllById` method, using the ID as the key, and then transforming the input list of IDs to the desired entities using the map.

To do this in a generic way requires using reflection. This can be written manually, however it is possible to get the ID using a `JpaEntityInformation` instance. To do this requires an active `EntityManager`, which is difficult to get an instance of outside of a Spring context. Therefore, the `RepositoryUtilHelper` class acts as a bridge to the static context - as a Spring component, the `EntityManagerFactory` is autowired on construction, and the `@PostConstruct` method registers itself to the static `lateinit` instance - allowing it to be accessed by the utility functions.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1OutOfServiceBedService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1OutOfServiceBedService.kt
@@ -22,6 +22,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.results.AuthorisableActi
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.results.ValidatableActionResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.UserService
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.PageCriteria
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.findAllByIdOrdered
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.getMetadata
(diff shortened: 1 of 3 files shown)


edavey (approved):
Thanks for figuring this out. I'm surprised that it isn't possible to use one of the ordering features of JPA e.g. (https://www.baeldung.com/spring-data-sorting#1-sorting-with-the-orderby-method-keyword), but I'm sure you've explored what's possible thoroughly!

danlivings-dxw:
@edavey the problem is that the list of IDs is ordered already, and that the `findAllById` method translates to something like `SELECT * FROM cas1_out_of_service_beds WHERE id IN (:ids)`. I don't think it's possible to do something equivalent to `ORDER BY (:ids)` in SQL, or at least not cleanly.

PR #2083, opened 2024-07-18
Upper case CRN when call Approved-premises-and-delius endpoints

muhammad-elabdulla:
This PR is to make sure that when a call is made to Approved premises and delius endpoints the CRN is provided in uppercase as these endpoints are case sensitive


Rosie-Brigham (approved):
👍

PR #2084, opened 2024-07-18
APS-1087 - Upgrade open api generator

davidatkinsuk:


edavey (approved):
well done figuring out a pragmatic solution!

PR #2085, opened 2024-07-18
APS-365 Use custom HttpMethod for caching metadata

davidatkinsuk:
Before this commit we used Spring’s HttpMethod in our PreemptiveCacheMetadata class. Instances of this class are marshalled into JSON (via jackson) and then stored in Redis. Metadata is then subsequently retrieved and unmarhsalled back into a Kotlin instance.

As of Spring Boot 3, Spring’s HttpMethod is no longer unmarshallable into a Kotlin instance as it is not an enum and doesn’t provide a default constructor. Therefore, in preparation for the spring boot upgrade this commit adds a new HttpMethod implementation that can be marshalled and unmarshalled, whilst remaining structurely identical to the Spring HttpMethod previous used (meaning existing cache entries can be used with this class).


RobBoothMOJ (approved):
Approved

danhumphreys-moj:
Excellent.  This is much simpler.  Glad we managed to get a better solution than littering the code with `MarshallableHttpMethod` as it was before.

PR #2087, opened 2024-07-19
APS-1090 Add placement_application -> application_id index

davidatkinsuk:
We have been observing version poor performance from the /applications/all endpoint with high page numbers. This fundamental index was missing. Adding it resolves these performance issues


muhammad-elabdulla (approved):

muhammad-elabdulla:
LGTM

PR #2092, opened 2024-07-19
APS-1028: Implement Assessment Appeals Manager Role

RobBoothMOJ:
Add granular permissions to CAS1 roles to allow roles to be selected based on a permission requirement. 

'CAS1_ASSESS_APPEALED_APPLICATION' permission is used to get active users (with the matching role. - CAS1_APPEALS_MANAGER, CAS1_ASSESSOR) who can  (re)assess an appealed assessment.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/UserTransformer.kt
```diff
@@ -102,4 +105,19 @@ class UserTransformer(
...
+    userRole.role.permissions.forEach {
+      listOfPermissions += when (it) {
+        UserPermission.CAS1_VIEW_ASSESSMENTS -> ApiUserPermission.viewAssessments
+        UserPermission.CAS1_PROCESS_AN_APPEAL -> ApiUserPermission.processAnAppeal
+        UserPermission.CAS1_ASSESS_PLACEMENT_APPLICATION -> ApiUserPermission.assessPlacementApplication
+        UserPermission.CAS1_ASSESS_PLACEMENT_REQUEST -> ApiUserPermission.assessPlacementRequest
+        UserPermission.CAS1_ASSESS_APPLICATION -> ApiUserPermission.assessApplication
+        UserPermission.CAS1_ASSESS_APPEALED_APPLICATION -> ApiUserPermission.assessAppealedApplication
+      }
+    }
```
davidatkinsuk, line 120: instead of using a mutable map you could maybe do something like the following:

```
return userRole.role.permissions.map {
   when(it) {
     ... existing code here ...
   }
}
```

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/UserServiceTest.kt
```diff
@@ -861,4 +863,57 @@ class UserServiceTest {
       }
     }
   }
+
+  @Nested
+  inner class getAllocatableUsersForAllocationType {
```
davidatkinsuk, line 868: thanks for improving testing of existing functionality here

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/UserEntity.kt
```diff
@@ -316,8 +316,12 @@ enum class UserRole(val service: ServiceName, val cas1ApiValue: ApprovedPremises
   CAS1_EXCLUDED_FROM_ASSESS_ALLOCATION(ServiceName.approvedPremises, ApprovedPremisesUserRole.excludedFromAssessAllocation),
   CAS1_EXCLUDED_FROM_MATCH_ALLOCATION(ServiceName.approvedPremises, ApprovedPremisesUserRole.excludedFromMatchAllocation),
   CAS1_EXCLUDED_FROM_PLACEMENT_APPLICATION_ALLOCATION(ServiceName.approvedPremises, ApprovedPremisesUserRole.excludedFromPlacementApplicationAllocation),
-  CAS1_APPEALS_MANAGER(ServiceName.approvedPremises, ApprovedPremisesUserRole.appealsManager),
-  CAS1_JANITOR(ServiceName.approvedPremises, ApprovedPremisesUserRole.janitor),
+  CAS1_APPEALS_MANAGER(
+    ServiceName.approvedPremises,
+    ApprovedPremisesUserRole.appealsManager,
+    listOf(UserPermission.CAS1_PROCESS_AN_APPEAL, UserPermission.CAS1_VIEW_ASSESSMENTS, UserPermission.CAS1_ASSESS_APPEALED_APPLICATION),
+  ),
```
davidatkinsuk, line 323: I _think_ we may have a UI/API deployment race condition issue

If this API was deployed before UI changes we would allow appealed assessments to be allocated to to users with the role 'CAS1_APPEALS_MANAGER', but the UI wouldn't let them view assessments assigned to them

If the UI was deployed before the API there would be lots of errors because the API wouldn't be returning permissions (and users with existing roles won't be able to view the assessments anymore because they'll have no permissions!)

I think the answer is to use feature flags to determine whether the CAS1_APPEALS_MANAGER can be assigned appealed applications (i.e. they have the CAS1_ASSESS_APPEALED_APPLICATION permission). This will be disabled in production but enabled in all other environments (allowing us to do UAT).

Once we've starting using permissions across the UI this will no longer be an issue, but it'll just be some pain we'll need to deal with whilst migrating (And to be fair, this is the sort of issue we're introducing permissions to avoid!)

Let's have a chat about this when you're ready because it's awkward to inject a feature flag into enum construction so we'll need to add some indirection to do this. Also i have a PR to 'fix' feature flags so we'll need that merging first - https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/2094

davidatkinsuk, line 323: Have you tried running through the whole process on the UI with a user that only has the CAS1_APPEALS_MANAGER role? I've made similar changes before and been caught out because there have been permission checks when submitted assessments that lead to permission errors

davidatkinsuk (changes requested):
Thanks @RobBoothMOJ . I've marked as Request changes mainly because i think there may be an issue if we merge in it's current state - see my comment about deployment race conditions

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/UserEntity.kt
```diff
@@ -302,9 +302,9 @@ data class UserRoleAssignmentEntity(
   override fun hashCode() = Objects.hash(id, role)
 }
 
-enum class UserRole(val service: ServiceName, val cas1ApiValue: ApprovedPremisesUserRole?) {
-  CAS1_ASSESSOR(ServiceName.approvedPremises, ApprovedPremisesUserRole.assessor),
-  CAS1_MATCHER(ServiceName.approvedPremises, ApprovedPremisesUserRole.matcher),
+enum class UserRole(val service: ServiceName, val cas1ApiValue: ApprovedPremisesUserRole?, val permissions: List<UserPermission> = emptyList()) {
+  CAS1_ASSESSOR(ServiceName.approvedPremises, ApprovedPremisesUserRole.assessor, listOf(UserPermission.CAS1_ASSESS_APPLICATION, UserPermission.CAS1_ASSESS_APPEALED_APPLICATION)),
+  CAS1_MATCHER(ServiceName.approvedPremises, ApprovedPremisesUserRole.matcher, listOf(UserPermission.CAS1_ASSESS_PLACEMENT_APPLICATION, UserPermission.CAS1_ASSESS_PLACEMENT_REQUEST)),
```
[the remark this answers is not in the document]
RobBoothMOJ: CAS1_VIEW_ASSESSMENTS is currently the permission being leveraged when displaying the "Assess Approved Premises Applications" tile in the UI. 
The tile is displayed (when the permission exists) for CAS1_APPEALS_MANAGER and CAS1_JANITOR and also for users with the CAS1_ASSESSOR role.   

Adding it to the CAS1_ASSESSOR role makes sense.  I'm not sure about the matcher role though?   Adding it to the matcher role means we need a new permission to determine whether the tile should be displayed; the matcher role should not see the tile.  It also makes the permission redundant (currently) as it won't be leveraged anywhere.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/UserEntity.kt
RobBoothMOJ: Resolved with introduction of `cas1-appeal-manager-can-assess-applications` feature flag, in UserService.

davidatkinsuk (approved):
LGTM 👍

PR #2093, opened 2024-07-21
CAS-365 - Remove implicit JPA Timestamp to Date conversions

davidatkinsuk:
There are several JPA queries that require an implicit version from a timestamptz field in the database to a LocalDate value in the Kotlin Model. This relies on JPA/Hibernate to determine how to convert between these two types (and understand in what timezone this should be applied). As of Hibernate 6 this conversion is no longer supported.

This commit removes this implicit conversion and instead manages the conversion ‘in code’ using the JDK timezone (which is Europe/London on deployment). Regardless of HIbernate compatibility, this results in code that better identifies the type of the database column we are using to provide this data.


Rosie-Brigham (approved):
Looks good, thank you for the clarification of the use of date and dateTime over slack

PR #2094, opened 2024-07-21
Remove Flipt Integration for Feature Flags

davidatkinsuk:
The instance of flipt we are using is being deprecated. Whilst we determine how we are managing feature flags going forward, this commit reverts to previous approach of using environment variables, albeit via a more structured ‘Feature Flag Service’ to encourage consistency in flag definition and usage.


Rosie-Brigham (approved):
👍

PR #2096, opened 2024-07-22
Revert Upper case CRN when call Approved-premises-and-delius endpoints

muhammad-elabdulla:
This PR is to revert the change to pass upper case CRN when calling Approved-premises-and-delius endpoints. 

The current code is expecting that if the user pass a CRN in lower case it will not return any data. Changing it to pass a CRN in uppercase will associate the data with the lower CRN which may causes some issues.


davidatkinsuk (approved):
LGTM 👍

PR #2098, opened 2024-07-22
Fix an issue in CAS3 Booking report to calculate Actual Nights Stayed

muhammad-elabdulla:
This [PR CAS-483](https://dsdmoj.atlassian.net/browse/CAS-483) is to fix in issue in calculate the Actual Nights Stayed to include start date in CAS3 booking report


davidatkinsuk (approved):
LGTM 👍

PR #2099, opened 2024-07-22
CAS-445 Get current user details from delius and update

daveawc:
This commit adds and endpoint to return the current users details as stored in delius, as well as updating the user table so it's up to date. 

It uses the existing community api rather than the integration services, as the code is existing and other things are using the user table, this will require extra caution and testing, so will transition to use the integration service when the rest of the app changes over.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/UsersController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/UsersController.kt
@@ -109,6 +109,16 @@ class UsersController(
     )
   }
 
+  override fun getCurrentUserDetails(xServiceName: ServiceName): ResponseEntity<User> {
+    if (!userAccessService.userCanViewOwnUserDetails()) {
+      throw ForbiddenProblem()
+    }
+
+    val userFromDelius = userService.getUpdatedCurrentUserDetails(xServiceName)
+    val userTransformed = userTransformer.transformJpaToApi(userFromDelius, xServiceName)
+    return ResponseEntity.ok(userTransformed)
+  }
+
   override fun usersDeliusGet(name: String, xServiceName: ServiceName): ResponseEntity<User> {
     if (!userAccessService.currentUserCanManageUsers(xServiceName)) {
       throw ForbiddenProblem()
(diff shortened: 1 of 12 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas3/DomainEventService.kt
```diff
@@ -203,7 +204,7 @@ class DomainEventService(
         occurredAt = domainEvent.occurredAt.atOffset(ZoneOffset.UTC),
         createdAt = OffsetDateTime.now(),
         data = objectMapper.writeValueAsString(domainEvent.data),
-        service = "CAS3",
+        service = ServiceName.temporaryAccommodation.name,
```
davidatkinsuk, line 207: I'm pretty sure "CAS3" is correct in this case, looking at the discriminator in the yml

src/main/resources/static/_shared.yml
```diff
@@ -3197,7 +3197,7 @@ components:
         propertyName: service
         mapping:
           CAS1: '#/components/schemas/ApprovedPremisesUser'
-          CAS3: '#/components/schemas/TemporaryAccommodationUser'
+          temporary-accommodation: '#/components/schemas/TemporaryAccommodationUser'
```
davidatkinsuk, line 3200: is there a particular reason we've done this? we should be using the CAS naming where possible

src/main/resources/static/api.yml
```diff
@@ -4238,6 +4238,38 @@ paths:
                 $ref: '_shared.yml#/components/schemas/Problem'
         500:
           $ref: '_shared.yml#/components/responses/500Response'
+  /users/me:
```
davidatkinsuk, line 4241: we already have an endpoint to get current user's details on /profile (or /profile/v2 preferably).

we could add an optional param to that to indicate the user's details should be refreshed before being retrieved? I'm concerned that we're introducing an additional path to do something similar.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/UsersController.kt
```diff
@@ -109,6 +109,16 @@ class UsersController(
     )
   }
 
+  override fun getCurrentUserDetails(xServiceName: ServiceName): ResponseEntity<User> {
+    if (!userAccessService.userCanViewOwnUserDetails()) {
```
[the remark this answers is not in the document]
Rosie-Brigham: I can't think of a reason why we need this - will double check with Mahesh but my gut is to 🔥 it.

src/main/resources/static/api.yml
davidatkinsuk: The wrapper on /profile/v2 is so we can provide context back to the caller if the user doesn't have staff information so the UI can display something useful to the user when this occurs. The response still contains the same User type (which in your case will always be a TemporaryAccommodationUser)

If the UI calls /profile and there is no staff information the API will throw a 500 so the user will see 'something went wrong' which then leads to a manual support request when it could be handled with a useful message.

I believe CAS3 UI is currently calling /profile. So ideally, if you can, we should make this enhancement to /profile/v2 and also update the CAS3 UI to use this new version _and_ also handle what happens when no staff member exists e.g. where loadError=staff_record_not_found. An example of what CAS1 does is here - https://dsdmoj.atlassian.net/browse/APS-597. Rosie may know if there's an equivalent CAS3 ticket.

PR #2112, opened 2024-07-24
APS-1085 Improve PersonTransformer readability

davidatkinsuk:
This commit removes repeated use of the same property traversal in PersonTransformer. This helps clarifies the key entities in various result data classes, in preparation for some rationalisation and simplification work, with the ultimate aim of removing data classes bound to the community API


RobBoothMOJ (approved):
Approved

PR #2116, opened 2024-07-24
CAS-445 update user details from delius

daveawc:
Added an optional parameter to the /profile/v2 endpoint so that user details can be pulled from delius and updated before being returned to the front end.

[https://dsdmoj.atlassian.net/browse/CAS-445](url)


davidatkinsuk (approved):
LGTM with some proposal re removing a little but of duplication

PR #2121, opened 2024-07-25
Change to use bulk retrieval of person information when there a group of offenders

muhammad-elabdulla:
This [PR CAS-465](https://dsdmoj.atlassian.net/browse/CAS-465) is to use bulk retrieval of person information when there are a group of offenders that need to retrieve their person information. This will improve the performance of the code and reduce the number of calls to `/probation-cases/summaries` Api end point
- Create a new function `getInfoForPersons` to get a bulk of person information and use it instead of the current `getInfoForPerson` function in case there are a group of offenders. This function will call `getOffenderSummariesByCrns` to unify the point to get offenders information.
- Change the current `getInfoForPerson` to use the new function
- Replace the code when currently call the old function to use the new one when there are a group of offenders that need to retrieve their personal information


davidatkinsuk (approved):
LGTM 👍

PR #2122, opened 2024-07-25
Change call to get person info async and add logs for generating CAS3 booking report

muhammad-elabdulla:
This [PR CAS-451](https://dsdmoj.atlassian.net/browse/CAS-451) is to fix an issue in logs and add more logs for generating CAS3 booking report


davidatkinsuk (approved):
approved to allow testing against bulk data

PR #2120, opened 2024-07-25
APS-1068: Send essential and desirable characteristics when booking CAS1 space

edavey:
On reflection we've realised that sending:

- need characteristics 
    - `wheelchair` 
    - `enSuite` 
    - `single` 
    - `limitedMobility` 
    - `catered`

- risk characteristics 
    - `arson` 
    - `posesSexualRiskToAdults` 
    - `posesSexualRiskToChildren` 
    - `hateBasedOffences`
    - `atRiskOfExploitation` 
    - `posesNonSexualRiskToChildren`

does not make good sense. This is because:

1. we already have a list of space characteristics (combination of premises and room characteristics) in the form of the placement requirements associated with the placement request.

The list of possible space characteristics is:

- `acceptsChildSexOffenders`
- `acceptsHateCrimeOffenders`
- `acceptsNonSexualChildOffenders`
- `acceptsSexOffenders`
- `hasArsonInsuranceConditions`
- `hasBrailleSignage`
- `hasCallForAssistance`
- `hasCrib7Bedding`
- `hasEnSuite`
- `hasFixedMobilityAids`
- `hasHearingLoop`
- `hasLift`
- `hasNearbySprinkler`
- `hasSmokeDetector`
- `hasStepFreeAccess`
- `hasStepFreeAccessToCommunalAreas`
- `hasTactileFlooring`
- `hasTurningSpace`
- `hasWheelChairAccessibleBathrooms`
- `hasWideAccessToCommunalAreas`
- `hasWideDoor`
- `hasWideStepFreeAccess`
- `isArsonDesignated`
- `isArsonSuitable`
- `isCatered`
- `isFullyFm`
- `isGroundFloor`
- `isGroundFloorNrOffice`
- `isIAP`
- `isSingle
- `isSingle`
- `isStepFreeDesignated`
- `isSuitedForSexOffenders`
- `isTopFloorVulnerable`
- `isWheelchairAccessible`
- `isWheelchairDesignated`

2. recording the characteristics on our space bookings will simplify the process of searching for suitable available spaces:

- "space bookings" have essential characteristics
- "spaces" have characteristics
- "space searches" have essential characteristics (this is to come)

PR #2123, opened 2024-07-25
APS1068 Search for spaces using space characteristics

edavey:
**NB: This follows on from https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/2120**

---

Now that we have modelled "space characteristics" (a combination
of premises characteristics and room characteristics) and are
using these to make a space booking with lists of:

- `essentialCharacteristics`, and
- `desirableCharacteristics`

It's clear that the more abstract concepts of `needCharacteristics` and
`riskCharacteristics` should be UI concerns only. We are
already intimately involved with these raw space characteristics on
the UI, e.g. in https://github.com/ministryofjustice/approved-premises-ui/blob/7d7041a6a99934b89afc6834928ad4f7eb0fa360/server/utils/placementCriteriaUtils.ts

characteristics_requirements_categories

### DIFF
--- a/src/main/resources/static/_shared.yml
+++ b/src/main/resources/static/_shared.yml
@@ -640,14 +640,10 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/ApType'
-        needCharacteristics:
+        spaceCharacteristics:
           type: array
           items:
-            $ref: '#/components/schemas/Cas1SpaceNeedCharacteristic'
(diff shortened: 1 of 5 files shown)


richpjames (approved):

danlivings-dxw (approved):
I think this makes implementing the search filters much easier from an API perspective

PR #2134, opened 2024-08-01
Drop implicit ‘.csv’ for seed job execution

davidatkinsuk:
Before this commit when running a seed job the API required that the ‘.csv’ extension was not included when specifying the filename. This has proven to be counter-intuative and leads to user error.


RobBoothMOJ (approved):
Approved

PR #2136, opened 2024-08-01
APS-1132 - Remove ‘characteristics’ col from AP seed job

davidatkinsuk:
This column is no longer used and has been replaced by boolean columns for each characteristic


RobBoothMOJ (approved):
Approved

PR #2137, opened 2024-08-01
CAS-445 update user details from profile

daveawc:
PR to update the user details from delius when calling the /v2 profile endpoint. The call to delius is skipped if the user is created as part of the get request.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ProfileController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ProfileController.kt
@@ -7,6 +7,8 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.ProfileApiDelegate
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ProfileResponse
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ServiceName
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.User
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.GetUserResponse
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.results.AuthorisableActionResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.UserService
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.UserTransformer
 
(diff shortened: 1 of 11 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ProfileController.kt
```diff
@@ -23,7 +25,11 @@ class ProfileController(
...
   override fun profileV2Get(xServiceName: ServiceName): ResponseEntity<ProfileResponse> {
     val username = userService.getDeliusUserNameForRequest()
-    val userResponse = userService.getUserForProfile(username)
-    return ResponseEntity(userTransformer.transformProfileResponseToApi(username, userResponse, xServiceName), HttpStatus.OK)
+    var getUserResponse = userService.getUserForProfile(username)
+    if (getUserResponse.user != null && !getUserResponse.createdOnGet) {
+      val updateResponse = userService.updateUserFromCommunityApiById(getUserResponse.user!!.id, xServiceName)
+      getUserResponse = (updateResponse as AuthorisableActionResult.Success<GetUserResponse>).entity
+    }
+    return ResponseEntity(userTransformer.transformProfileResponseToApi(username, getUserResponse, xServiceName), HttpStatus.OK)
```
davidatkinsuk, line 33: nice, i like how this isolates the 'update on get' logic into the controller so we're not inadvertently updating elsewhere

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/TaskService.kt
```diff
@@ -156,7 +156,7 @@ class TaskService(
     }
 
     val assigneeUser = when (val assigneeUserResult = userService.updateUserFromCommunityApiById(userToAllocateToId, ServiceName.approvedPremises)) {
-      is AuthorisableActionResult.Success -> assigneeUserResult.entity
+      is AuthorisableActionResult.Success -> assigneeUserResult.entity.user!!
```
davidatkinsuk, line 159: i know this will never happen currently because the updateUserFromCommunityApiById will throw an exception, but i think it's probably best for consistency with how we handle GetUserResponse to make line 159 something like the following (psuedo code, i'm not sure if you can do things like this in a single when line?)

```is AuthorisationActionResult.Success && assigneeUserResult.staffRecordFound -> assigneeUserResult.entity.user!!```

davidatkinsuk (approved):
This looks great. I've got one minor suggestion to make things a bit more future proof, in case a future dev isn't paying attention when they make changes to updateUserFromCommunityApiById.

Also as you mentioned on your comment, please squash 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/TaskService.kt
davidatkinsuk: Yep, that makes sense to me. I guess it is a bit odd how we return 'Success' despite staff not found but that's a pattern we're committed to now for this profile stuff :)

PR #2138, opened 2024-08-01
APS-1134 - Add fields to CAS1 booking cancelled domain event

davidatkinsuk:
Before this commit, the CAS1 booking cancelled domain event details did not provide:

1. The date (and only the date) of when the cancellation occurred
2. The datetime of when the cancellation was recorded

The domain event did provide a datetime for when the cancellation occurred, but the time was always set to midnight UTC. This is currently being used by probation-integration services to set the time the contact occurred. This has resulted in two issues:

1. The cancellation appears to happen at either midnight or 1am, depending on the current daylight savings
2. If multiple cancellations happen for the same CRN in the same day, due to limitations in contact recorded only one contact is seen (despite referrals being correctly deleted in Delius).

This commit introduces the values mentoned at the top of this description. These fields can then be used by probation-integration to resolve the aforementioned issues.

This change is not backwards compatible, because once probation-integration have changed their code to expect these new fields, we can no longer replay existing booking cancelled domain events in their current state. To support these scenarios (if they occur), we are setting domain_events.schema_version to ‘2’ for any new booking cancelled domain event. If replay of a ‘legacy’ domain event is required, the replay logic will need to be updated to detect usage of the legacy detail schema (i.e. where domain_events.schema_version is null), and the json will need updating in the database to add these two additional fields.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
@@ -1196,7 +1196,7 @@ class BookingService(
     cancellation: CancellationEntity,
     reason: CancellationReasonEntity,
   ) {
-    val dateTime = OffsetDateTime.now()
+    val now = OffsetDateTime.now()
 
     val domainEventId = UUID.randomUUID()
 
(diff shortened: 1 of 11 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BookingService.kt
```diff
@@ -1228,11 +1228,11 @@ class BookingService(
         applicationId = applicationId,
         crn = booking.crn,
         nomsNumber = offenderDetails.otherIds.nomsNumber,
-        occurredAt = dateTime.toInstant(),
+        occurredAt = now.toInstant(),
         bookingId = booking.id,
         data = BookingCancelledEnvelope(
           id = domainEventId,
-          timestamp = dateTime.toInstant(),
```
davidatkinsuk, line 1235: introduce version on domain events table, simple number. Set it to 2 in this PR

RobBoothMOJ (approved):
Approved

PR #2139, opened 2024-08-02
Reduce connection pool timeout

davidatkinsuk:
When we have experienced extreme load issues, we’ve found that the default connection pool timeout of 30s is exacerbating the issue as by the time the connection becomes available we’ve had a client timeout/give up, so the request is no longer required.

This commit reduces that timeout to 15 seconds to try to allow clients to fail fast in these circumstances

### DIFF
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -22,6 +22,7 @@ spring:
   datasource:
     hikari:
       maximum-pool-size: 20
+      connection-timeout: 15000
 
   flyway:
     repeatable-sql-migration-prefix: "R"


danhumphreys-moj:
Presumably the time taken for requests to time out is longer than 15s?

danhumphreys-moj (approved):

davidatkinsuk:
> Presumably the time taken for requests to time out is longer than 15s?

Good question. I've observed the client waiting longer than 30 seconds so i don't think it's automatically timing out in less than 15 seconds but it's worth considering:

1. A request isn't just SQL, API requests triggered by a UI request are already taking quite a long time to complete (adding to the 15 seconds)
2. Under load we're seeing queries taking a long time to run, and that's after waiting up to 30 seconds for a connection
3. Once we disable open session in view (see APS-1127), it's likely that a single request may be requesting multiple connections in one thread of execution (stacking the 'waits' it's having to do), stacking up these delays
4. Users will probably manually refresh the page before a 15 second wait

Basically, if we can't service a SQL request in 15 seconds we've agreed the system is already unsustainable so it's best to try and just fail fast instead of queueing up more load. APS-1120 will also help reduce the amount of load in these circumstances

PR #2153, opened 2024-08-06
APS-1151 - Increase pre-emptive cache lock duration

davidatkinsuk:
We’re observing that the cache lock duration is currently insufficient to refresh the entire cache.

This commit increases it from 2 minutes to 10 minutes to help us determine how long will be required once the cache state has normalised (i.e. fully refreshed).

The downside of a longer lock is that no other instance will be able to pick up the cache refresh work if the worker unexpectedly dies. Given that the cache being 100% up to date is not mission critical, a maximum delay of 10 minutes before the refresh worker starts again should not be an issue.


Rosie-Brigham (approved):
👍

PR #2157, opened 2024-08-07
APS-1089 - RemoveRole should remove duplicates

davidatkinsuk:
We have observed that either due to a bug or a backup/restore issue in pre-prod, users can have the same role assigned multiple times. Whilst this issue is mostly prevelant in pre-prod, there is a single occurance of this in prod too. This can be checked by running

select u.id, u.name, r.role, count(r.role)
from user_role_assignments r
inner join users u on u.id = r.user_id
group by u.id, u.name, r.role
having count(r.role) > 1;

This commit updates the removeRoleFromUser to remove all ocurrances of the roles recorded against a user.


RobBoothMOJ (approved):
Approved

PR #2158, opened 2024-08-07
APS-1155 - Remove OffenderDetailsCache Part 1

davidatkinsuk:
Remove the Offender Details Cache Refresh Worker and any now unused functions.

Once this is deployed to prod we can clear the cache, and then remove any existing configuration in the code for this cache.


RobBoothMOJ (approved):
Approved

PR #2159, opened 2024-08-07
APS-117: Normalise user roles list before calculating user version

RobBoothMOJ:
Ensure that duplicates are removed from the user's role list and the list is sorted before calculating user version to ensure consistent responses. 
Uppercase Delius username when executing query to retrieve roles by username.


davidatkinsuk (approved):
LGTM and thanks for the thorough tests 👍

PR #2161, opened 2024-08-07
Remove comparison checks from update user.

daveawc:
update to always save the user details that are returned from delius. Hibernate won't perform the update if the values are not changed.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ProbationRegionEntity.kt
```diff
@@ -19,6 +19,7 @@ interface ProbationRegionRepository : JpaRepository<ProbationRegionEntity, UUID>
 data class ProbationRegionEntity(
   @Id
   val id: UUID,
+  @Deprecated(message = "There can be multiple codes to a region, so use probation_area_probation_region_mappings to get the correct code, ")
```
davidatkinsuk, line 22: can we remove the trailing ", " please

davidatkinsuk (approved):
LGTM 👍

PR #2077, opened 2024-07-17
APS-1068: Proposed OpenAPI spec for "find and book CAS1 space"

edavey:
In this PR I propose an interface between the UI and API for searching for and booking CAS1 spaces.

## Search for spaces
The UI will post a set of `Cas1SpaceSearchParameters` to an CAS1 api endpoint at:

`/cas1/spaces/search`

The `Cas1SpaceSearchParameters` have this shape:

```json
{
  "startDate": "2024-07-17",
  "durationInDays": 84,
  "targetPostcodeDistrict": "SE5",
  "requirements": {
    "apType": "normal",
    "needCharacteristics": [
      "single",
      "enSuite",
      "wheelchair"
    ],
    "riskCharacteristics": [
      "atRiskOfExploitation",
      "arson"
    ],
    "gender": "male"
  }
}
```

where `needCharacteristics` is a list which may include the enum values:

```
- single
- enSuite
- wheelchair
- limitedMobility
- catered
```

and `riskCharacteristics` is a list which can include the values:

```
- atRiskOfExploitation
- arson
- posesSexualRiskToAdults
- posesSexualRiskToChildren
- posesNonSexualRiskToChildren
```

Search results are returned in the following form:

- a recap of the search criteria which generated the results
- a count of results
- a list of results. Availability is grouped by premises. Within each
  premises-result we list the available spaces. Against each available
  space we list its "need" characteristics and its "risk" characteristics:

```json
{
  "searchCriteria": {
    "startDate": "2024-07-17",
    "durationInDays": 84,
    "targetPostcodeDistrict": "SE5",
    "requirements": {
      "apType": "normal",
      "needCharacteristics": [
      "single",
      "enSuite",
      "wheelchair"
    ],
    "riskCharacteristics": [
      "atRiskOfExploitation",
      "arson"
    ],
      "gender": "male"
    }
  },
  "resultsCount": 4,
  "results": [
    {
      "premises": {
        "id": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
        "apCode": "NEHOPE1",
        "deliusQCode": "Q005",
        "apType": "normal",
        "name": "Hope House",
        "addressLine1": "1 The Street",
        "addressLine2": "Blackmore End",
        "town": "Braintree",
        "postcode": "LS1 3AD",
        "apArea": {
          "id": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
          "name": "string"
        },
        "totalSpaceCount": 22,
        "premisesCharacteristics": [
          {
[...]

PR #2162, opened 2024-08-07
[APS-1094] Implement API endpoint for creating a new CAS1 space booking

danlivings-dxw:
> See [APS-1094 on the Approved Premises Jira board](https://dsdmoj.atlassian.net/browse/APS-1094).

This PR implements the `POST /cas1/placement-requests/{placementRequestId}/space-bookings` API endpoint introduced in #2077, which allows a user to create a new booking for a space in an Approved Premises.

A new booking is created by `POST`ing to the endpoint with a request body like:
```jsonc
{ // NewCas1SpaceBooking
  "arrivalDate": "2024-09-01",
  "departureDate": "2024-09-30",
  "premisesId": "00000000-0000-0000-0000-000000000000",
  "placementRequestId": "00000000-0000-0000-0000-000000000000",
  "requirements": { // Cas1SpaceBookingRequirements
    "apType": "pipe",
    "gender": "female",
    "essentialCharacteristics": [
      // Cas1SpaceCharacteristic
      // ...
    ],
    "desirableCharacteristics": [
      // Cas1SpaceCharacteristic
      // ...
    ]
  }
}
```

If the API successfully creates a new booking, it will respond with the new booking, which will look like:
```jsonc
{
  "id": "00000000-0000-0000-0000-000000000000",
  "person": { // Person
    // ...
  },
  "requirements": { // Cas1SpaceBookingRequirements
    "apType": "pipe",
    "gender": "female",
    "essentialCharacteristics": [
      // Cas1SpaceCharacteristic
      // ...
    ],
    "desirableCharacteristics": [
      // Cas1SpaceCharacteristic
      // ...
    ]
  },
  "premises": { // NamedId
    "id": "00000000-0000-0000-0000-000000000000",
    "name": "Some AP"
  },
  "apArea": { // NamedId
    "id": "00000000-0000-0000-0000-000000000000",
    "name": "Some AP Area"
  },
  "bookedBy": { // User
    // ...
  },
  "arrivalDate": "2024-09-01",
  "departureDate": "2024-09-30",
  "createdAt": "2024-08-08T12:34:56.789Z"
}
```

It is worth noting that there is currently some redundancy in what is required in the request body. Ideally the API specification would be updated to account for this but this is yet to be agreed:
- The `placementRequestId` is specified both in the URL as a path parameter and in the request body. The value specified in the request body is currently ignored.
- The `requirements` are already stored within the specified placement request. The booking will be created using those specified in the placement request.

The API validates the information provided in the request body in the following ways:
- The `premisesId` must reference an existing Approved Premises.
- The `placementRequestId` must reference an existing placement request.
- The `arrivalDate` must be before the `departureDate`.

Space availability should be validated in the future, but is not currently implemented - this will be accounted for in a future ticket that will also add availability information to the space search endpoint.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas1/SpaceBookingController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas1/SpaceBookingController.kt
@@ -1,0 +1,47 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.controller.cas1
+
+import org.springframework.http.ResponseEntity
+import org.springframework.stereotype.Service
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.cas1.SpaceBookingsCas1Delegate
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Cas1SpaceBooking
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.NewCas1SpaceBooking
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserQualification
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.OffenderService
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.UserService
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas1.Cas1SpaceBookingService
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.cas1.Cas1SpaceBookingTransformer
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.extractEntityFromValidatableActionResult
+import java.util.UUID
+
+@Service
+class SpaceBookingController(
+  private val userService: UserService,
+  private val offenderService: OffenderService,
+  private val spaceBookingService: Cas1SpaceBookingService,
+  private val spaceBookingTransformer: Cas1SpaceBookingTransformer,
+) : SpaceBookingsCas1Delegate {
+  override fun placementRequestsPlacementRequestIdSpaceBookingsPost(
+    placementRequestId: UUID,
+    body: NewCas1SpaceBooking,
+  ): ResponseEntity<Cas1SpaceBooking> {
+    val user = userService.getUserForRequest()
+
+    val booking = extractEntityFromValidatableActionResult(
+      spaceBookingService.createNewBooking(
+        body.premisesId,
+        placementRequestId,
+        body.arrivalDate,
+        body.departureDate,
+        user,
+      ),
+    )
+
+    val person = offenderService.getPersonInfoResult(
+      booking.placementRequest.application.crn,
+      user.deliusUsername,
+      user.hasQualification(UserQualification.LAO),
+    )
+
+    return ResponseEntity.ok(spaceBookingTransformer.transformJpaToApi(person, booking))
+  }
+}
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas1SpaceBookingEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas1SpaceBookingEntity.kt
@@ -1,0 +1,37 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity
+
+import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.stereotype.Repository
+import java.time.LocalDate
+import java.time.OffsetDateTime
+import java.util.UUID
+import javax.persistence.Entity
+import javax.persistence.FetchType
+import javax.persistence.Id
+import javax.persistence.JoinColumn
+import javax.persistence.ManyToOne
+import javax.persistence.Table
+
+@Repository
+interface Cas1SpaceBookingRepository : JpaRepository<Cas1SpaceBookingEntity, UUID> {
+  fun findByPremisesIdAndPlacementRequestId(premisesId: UUID, placementRequestId: UUID): Cas1SpaceBookingEntity?
+}
+
+@Entity
+@Table(name = "cas1_space_bookings")
+data class Cas1SpaceBookingEntity(
+  @Id
+  val id: UUID,
+  @ManyToOne(fetch = FetchType.LAZY)
+  @JoinColumn(name = "premises_id")
+  val premises: ApprovedPremisesEntity,
+  @ManyToOne(fetch = FetchType.LAZY)
+  @JoinColumn(name = "placement_request_id")
+  val placementRequest: PlacementRequestEntity,
+  @ManyToOne(fetch = FetchType.LAZY)
+  @JoinColumn(name = "created_by_user_id")
+  val createdBy: UserEntity,
+  val createdAt: OffsetDateTime,
+  val arrivalDate: LocalDate,
+  val departureDate: LocalDate,
+)
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1SpaceBookingService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1SpaceBookingService.kt
@@ -1,0 +1,75 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas1
+
+import kotlinx.datetime.toKotlinDatePeriod
+import org.springframework.stereotype.Service
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApprovedPremisesEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas1SpaceBookingEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas1SpaceBookingRepository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.cas1.Cas1SpaceSearchRepository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.validated
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.results.ValidatableActionResult
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.PlacementRequestService
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.PremisesService
+import java.time.LocalDate
+import java.time.OffsetDateTime
+import java.util.UUID
+
+@Service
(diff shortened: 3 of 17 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas1SpaceBookingEntity.kt
```diff
@@ -0,0 +1,36 @@
...
+import java.util.UUID
+import javax.persistence.Entity
+import javax.persistence.Id
+import javax.persistence.JoinColumn
+import javax.persistence.ManyToOne
+import javax.persistence.Table
+
+@Repository
+interface Cas1SpaceBookingRepository : JpaRepository<Cas1SpaceBookingEntity, UUID> {
+  fun findByPremisesIdAndPlacementRequestId(premisesId: UUID, placementRequestId: UUID): Cas1SpaceBookingEntity?
```
davidatkinsuk, line 16: can we please add a multicolumn index for these two values

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas1SpaceBookingEntity.kt
```diff
@@ -0,0 +1,36 @@
...
+}
+
+@Entity
+@Table(name = "cas1_space_bookings")
+data class Cas1SpaceBookingEntity(
```
davidatkinsuk, line 21: Can we please tag all @ManyToOne relationships as lazy. We've had lots of performance issues on Spring Boot 3 and the prevailing advice seems to be that 'eager' is a bad default for @ManyToOne because it ends up generating huge 'cascading' queries. We can always change to eager once we've established how we're using these entities.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1SpaceBookingService.kt
```diff
@@ -0,0 +1,75 @@
...
+      "$.premisesId" hasValidationError "doesNotExist"
+    }
+
+    val placementRequest = placementRequestService.getPlacementRequestOrNull(placementRequestId)
+    if (placementRequest == null) {
+      "$.placementRequestId" hasValidationError "doesNotExist"
+    }
+
+    if (arrivalDate >= departureDate) {
+      "$.departureDate" hasValidationError "beforeBookingArrivalDate"
```
davidatkinsuk, line 43: i'm not sure if we're constrained by existing validation error 'codes', but i think 'shouldBeAfterArrivalDate' would be a better description for this given we're doing a >= check?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/Cas1SpaceBookingServiceTest.kt
```diff
@@ -0,0 +1,222 @@
...
+  @MockK
+  private lateinit var placementRequestService: PlacementRequestService
+
+  @MockK
+  private lateinit var spaceBookingRepository: Cas1SpaceBookingRepository
+
+  @MockK
+  private lateinit var spaceSearchRepository: Cas1SpaceSearchRepository
+
+  @InjectMockKs
```
davidatkinsuk, line 39: Nice to seeI njectMockks being used

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas1/Cas1SpaceBookingRequirementsTransformer.kt
```diff
@@ -0,0 +1,23 @@
...
+    apType = jpa.apType,
+    gender = jpa.gender,
+    essentialCharacteristics = jpa.essentialCriteria.mapNotNull { it.asCas1SpaceCharacteristic() },
+    desirableCharacteristics = jpa.desirableCriteria.mapNotNull { it.asCas1SpaceCharacteristic() },
+  )
+
+  private fun CharacteristicEntity.asCas1SpaceCharacteristic() = try {
+    Cas1SpaceCharacteristic.valueOf(this.propertyName!!)
+  } catch (_: Exception) {
+    null
```
davidatkinsuk, line 21: do we expect this to happen (e.g. are there characteristics in the database that don't map to JPA)? My main concern would be a subtle naming mismatch not being mapped. Would throwing an exception here be an issue?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas1/SpaceBookingController.kt
```diff
@@ -0,0 +1,46 @@
...
+        body.arrivalDate,
+        body.departureDate,
+        user,
+      ),
+    )
+
+    val person = offenderService.getPersonInfoResult(
+      booking.placementRequest.application.crn,
+      user.deliusUsername,
+      false,
```
davidatkinsuk, line 41: I think we should be using user.hasQualification(UserQualification.LAO) to determine this boolean (this is what we do on the active 'create booking for placement request' endpoint.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1SpaceBookingService.kt
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    if (validationErrors.any()) {
+      return fieldValidationError
+    }
+
+    premises!!
+    placementRequest!!
+
+    when (val existingBooking = cas1SpaceBookingRepository.findByPremisesIdAndPlacementRequestId(premisesId, placementRequestId)) {
```
davidatkinsuk, line 53: We'll need to add permission checks here (rob is working on new roles/permissions for creating placements as we speak).

Do we have a ticket for the things that need 'filling in' once we're happy with the general functionality? I think it'll be:

1. Domain Events
2. Permission Checks
3. Emails

Am I missing anything else?

davidatkinsuk (approved):
Thanks Dan, very neat and clear as always.

Just some minor comments/clarifications from me.

Let's discuss the following on today's standup:

* The placementRequestId is specified both in the URL as a path parameter and in the request body. The value specified in the request body is currently ignored. DA Propose - remove from body.
* The requirements are already stored within the specified placement request. The booking will be created using those specified in the placement request. DA propose - We need to understand if a user will ever have the ability to override/ignore placement request constraints. I suspect this has already been dissed? If no, we should remove them from the body.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas1SpaceBookingEntity.kt
danlivings-dxw: Good idea - I've added this.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas1SpaceBookingEntity.kt
danlivings-dxw: I've updated these. Weird that the default wasn't changed given the poor performance introduced in the new version of Spring.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1SpaceBookingService.kt
danlivings-dxw: I think there's a list in the UI that these are mapped to, but I've changed it as it's more descriptive.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas1/Cas1SpaceBookingRequirementsTransformer.kt
danlivings-dxw: I've copied this logic from `PlacementRequestTransformer.characteristicToCriteria` which swallows the exception.

I think the main benefit would be in testing, where an arbitrary `CharacteristicEntity` won't cause a test to fail because of the exception being thrown.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas1/SpaceBookingController.kt
danlivings-dxw: I believe this is redundant as the AP-and-Delius API should be taking this into account when presenting offender details, but I'm not 100% sure on this. Happy to change it if that's not the case.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1SpaceBookingService.kt
danlivings-dxw: I've created some tickets for these:
- https://dsdmoj.atlassian.net/browse/APS-1160
- https://dsdmoj.atlassian.net/browse/APS-1161
- https://dsdmoj.atlassian.net/browse/APS-1162

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas1SpaceBookingEntity.kt
davidatkinsuk: Yes, it's a strange one but i guess as it's in the JPA spec it's hard to change without breaking existing deployments :(

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas1/Cas1SpaceBookingRequirementsTransformer.kt
davidatkinsuk: OK, we'll leave as is and can always return to this later

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1SpaceBookingService.kt
davidatkinsuk: Thank you!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas1/SpaceBookingController.kt
danlivings-dxw: Okay, makes sense - change has been made.

davidatkinsuk:
I think only outstanding change is removing the placementRequestId from the body of the message. This is approved so happy for this to be merged once that's done

PR #2163, opened 2024-08-07
APS-1149 - Add retries to API calls

davidatkinsuk:
If API requests fail with a connection error (other than timeout) or with 500, 501, 502 or 503, the API client will now retry the request 1 additional time.

This helps work around stale connection errors we’ve observed in production where an idle connection in the netty connection pool has (presumably) been timed out by the target server. Given how intermittent this issue is in production,  a retry should, with luck, provide us with a working connection.

The error we’ve been seeing is “Connection prematurely closed BEFORE response; nested exception is reactor.netty.http.client.PrematureCloseException: Connection prematurely closed BEFORE response”. A specific integration test has been added for this scenario.


RobBoothMOJ (approved):
Approved

PR #2165, opened 2024-08-08
APS-1155 Remove offenderDetails Cache

davidatkinsuk:
This commit completes the removal of the offender details cache now that is has been cleared in production.

As part of this work I had to update the PreemptiveCacheTest to use the prisonsApiWebClient and the communityApiWebClient no longer has a pre-emptive cache configured


RobBoothMOJ (approved):
Approved

PR #2166, opened 2024-08-08
Cas 438 persist pdu id to applications table

danhumphreys-moj:
Ref: CAS-438, https://dsdmoj.atlassian.net/browse/CAS-438

As a part of the CAS-307: Capture PP's PDU and other core information in referral from Delius work, we are now planning to record and save the ID of the probation delivery unit (PDU).

In ticket CAS-366: FE: Implement new user journey in FE with pre-populated nDelius data the form will now return the PDU id as first class field called probationDeliveryUnitId. This needs to be saved into the db alongside the name of the PDU (determinable via the existing probation_delivery_units table).

Create a new probation_delivery_unit_id field and save the value returned from the UI probationDeliveryUnitId in it.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApplicationEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApplicationEntity.kt
@@ -25,6 +25,7 @@ import javax.persistence.DiscriminatorValue
 import javax.persistence.Entity
 import javax.persistence.EnumType
 import javax.persistence.Enumerated
+import javax.persistence.FetchType
 import javax.persistence.Id
 import javax.persistence.Inheritance
 import javax.persistence.InheritanceType
@@ -510,6 +511,9 @@ class TemporaryAccommodationApplicationEntity(
   var pdu: String?,
   var name: String?,
   var prisonReleaseTypes: String?,
+  @ManyToOne(fetch = FetchType.LAZY)
+  @JoinColumn(name = "probation_delivery_unit_id")
+  var probationDeliveryUnit: ProbationDeliveryUnitEntity?,
 ) : ApplicationEntity(
   id,
   crn,
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ApplicationTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ApplicationTest.kt
@@ -2409,7 +2409,7 @@ class ApplicationTest : IntegrationTestBase() {
     }
 
     @Test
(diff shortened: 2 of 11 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ApplicationTest.kt
```diff
@@ -2673,6 +2673,14 @@ class ApplicationTest : IntegrationTestBase() {
...
             }
 
+            val region = probationRegionEntityFactory.produceAndPersist {
+              withId(submittingUser.probationRegion.id)
+            }
+
+            val pdu = probationDeliveryUnitFactory.produceAndPersist {
+              withProbationRegion(region)
+            }
+
```
danhumphreys-moj, line 2683: Presumably this is the right way to insert a PDU and region for integration testing?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApplicationEntity.kt
```diff
@@ -510,6 +510,7 @@ class TemporaryAccommodationApplicationEntity(
   var pdu: String?,
   var name: String?,
   var prisonReleaseTypes: String?,
+  var probationDeliveryUnitId: UUID?,
```
davidatkinsuk, line 513: I would expect this to be defined as a ManyToOne relationship to the ProbationDeliveryUnitEntity instead of a UUID. Is there a reason we haven't modelled it this way? As long as the ManyToOne annotation marks the relationship as fetch = lazy, there should be any performance penalty adding this (we've learnt that the hard way with spring boot 3 upgrade 😄 ) . The column name can remain as is.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ApplicationTest.kt
davidatkinsuk: Yep, i think that'll work fine 👍

davidatkinsuk (changes requested):
Thanks Dan, I think there's one comment that needs considering before approval

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApplicationEntity.kt
danhumphreys-moj: Super, thanks for this.  Should be ok now?

davidatkinsuk: Looks good!

davidatkinsuk (approved):
LGTM, thanks for making the change

PR #2173, opened 2024-08-12
CAS2-505 Change soft delete period from 14 to 32 days.

danhumphreys-moj:
Ref: https://dsdmoj.atlassian.net/browse/CAS2-505

Applications where the status is withdrawn, cancelled or awaiting arrival and created greater than 32 days in the past can be excluded from the applications list (previously 14 days).

VIEW recreated with 32 day INTERVAL period (previously 14).  This VIEW replaces the one created in [20240618100000__create_cas_2_summary_views_change_id_type.sql](https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/main/src/main/resources/db/migration/all/20240618100000__create_cas_2_summary_views_change_id_type.sql).

Tests updated so that valid/unexpired dates are within the past 32 days (previously 14) and expired times are at least 33 days in the past.

### DIFF
--- a/src/main/resources/db/migration/all/20240812103000__create_cas_2_summary_views_make_32_days.sql
+++ b/src/main/resources/db/migration/all/20240812103000__create_cas_2_summary_views_make_32_days.sql
@@ -1,0 +1,25 @@
+-- change INTERVAL to 32 days
+DROP VIEW IF EXISTS cas_2_application_live_summary CASCADE;
+CREATE OR REPLACE VIEW cas_2_application_live_summary AS SELECT
+    a.id,
+    a.crn,
+    a.noms_number,
+    a.created_by_user_id,
+    a.name,
+    a.created_at,
+    a.submitted_at,
+    a.hdc_eligibility_date,
+    a.label,
+    a.status_id,
+    a.referring_prison_code,
+    a.abandoned_at
+FROM cas_2_application_summary a
+WHERE (a.conditional_release_date IS NULL OR a.conditional_release_date >= current_date)
+AND a.abandoned_at IS NULL
+AND a.status_id IS NULL
+   OR (a.status_id = '004e2419-9614-4c1e-a207-a8418009f23d' AND a.status_created_at > (current_date - INTERVAL '32 DAY')) -- Referral withdrawn
+   OR (a.status_id = 'f13bbdd6-44f1-4362-b9d3-e6f1298b1bf9' AND a.status_created_at > (current_date - INTERVAL '32 DAY')) -- Referral cancelled
+   OR (a.status_id = '89458555-3219-44a2-9584-c4f715d6b565' AND a.status_created_at > (current_date - INTERVAL '32 DAY')) -- Awaiting arrival
+   OR (a.status_id NOT IN ('004e2419-9614-4c1e-a207-a8418009f23d',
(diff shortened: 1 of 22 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ApplicationTest.kt
```diff
@@ -224,8 +224,8 @@ class Cas2ApplicationTest : IntegrationTestBase() {
         }
       }
 
-      fun unexpiredDateTime() = OffsetDateTime.now().randomDateTimeBefore()
```
davidatkinsuk, line 227: i don't suppose you fancy removing the default 'maxDays' attribute on randomDateTimeBefore(). I was confused by this change until i looked at the function definition and realised it used 14 by default, which i don't think is intuitive

davidatkinsuk (approved):
LGTM, just one optional proposal related to code you didn't modify to improve test code readability

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ApplicationTest.kt
danhumphreys-moj: I've removed the default values from all the functions in RandomData.kt now.  Should make it more readable across the board.  Re-requested review again.

davidatkinsuk: That's fantastic, thank you!

davidatkinsuk: (pls. consider this re-reviewed 😄  )

src/main/resources/db/migration/all/20240812103000__create_cas_2_summary_views_make_32_days.sql
```diff
@@ -0,0 +1,25 @@
...
+    a.hdc_eligibility_date,
+    a.label,
+    a.status_id,
+    a.referring_prison_code,
+    a.abandoned_at
+FROM cas_2_application_summary a
+WHERE (a.conditional_release_date IS NULL OR a.conditional_release_date >= current_date)
+AND a.abandoned_at IS NULL
+AND a.status_id IS NULL
+   OR (a.status_id = '004e2419-9614-4c1e-a207-a8418009f23d' AND a.status_created_at > (current_date - INTERVAL '32 DAY')) -- Referral withdrawn
```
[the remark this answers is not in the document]
danhumphreys-moj: Yes that would be good..  The application statuses come from [this seeding](https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/main/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/reference/Cas2ApplicationStatusSeeding.kt) file which has the UUIDs hard coded in it (and are not stored in the database).  No reason why they couldn't be static variables I suppose, just the way DXW did it.

PR #2174, opened 2024-08-13
APS-1171 Acquire lock when allocating assessments

davidatkinsuk:
We have observed issues in production where a user clicks on the ‘reallocate’ button multiple times for a given task if the backend is being slow to respond. This results in multiple copies of the assessment being created in the backend.

Whilst we have double click protection on the reallocate button this has a debounce setting of 1 second, so doesn’t block users from click the button multiple times over a longer period.

Previously we’ve used Optimistic Locking to solve these issues (i.e. @Version attribute on Entities). Whilst this helps us ensure the database is consistent (i.e. we don’t end up with multiple assessments) this will result in a generic assessment being returned to the user, and an alert being raised in sentry. A pessmistic lock acquired before reallocation results in a neater error for the user (a message saying the task has already been allocated), and no alert is raised in the backend. We already have presidence for use of a pessmistic lock on application submission.

Note that there is a better solution to problem of duplicate assessments that requires a redesign of the datamodel, but doesn’t require any sort of locking (which is preferred if possible). This is discussed here - https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/4996136966/CAS1+Proposal+Introduce+Task+Entity

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,8 @@ script/development_database
 Then in the "Gradle" panel (`View->Tool Windows->Gradle` if not visible), expand `approved-premises-api`, `Tasks`, 
 `application` and right click on `bootRunLocal` and select either Run or Debug.
 
+Note that setting a breakpoint will block the handling of any subsequent web requests (don't use breakpoints to test what happens when subsequent request are received that are in contention).
+
 ### Linting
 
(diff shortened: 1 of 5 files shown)


davidatkinsuk:
Closing this - acquiring the lock was causing reads to hang so this will compound performance issues

RobBoothMOJ (approved):
Approved

PR #2178, opened 2024-08-14
CAS-549: Ensure AnyValue type is not empty object

froddd:
Prior to version 5.5, Typescript would treat any empty object as a `Record<string, any>`. From 5.5, empty objects are now considered as a `Record<string, never>`, which is causing some issues with typechecking in the UI repositories, as the API returned a type of `AnyValue = {}` for any JSON properties.

Specifying `type: object` for the `AnyValue` type ensures the type generation script correctly assigns `Record<string, any>`.


davidatkinsuk (approved):
LGTM. As discussed our definition of AnyValue is now more like a Free Form type. But given this solves an issue in UI and doesn't appear to change our generated models in the API, I think this is fine.

PR #2181, opened 2024-08-16
CAS-412 return pdu and enable sort by pdu

danhumphreys-moj:
Ref: https://dsdmoj.atlassian.net/browse/CAS-412

In order to display PDUs for referrals list, and sort those lists by PDUs, the GET /assessments API endpoint needs to:

- Return the PDU name alongside the existing data for each referral (this should be the PDU of the application)

- Enable new sort field by PDU (alphabetical sort on PDU name)

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/AssessmentEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/AssessmentEntity.kt
@@ -122,11 +122,13 @@ interface AssessmentRepository : JpaRepository<AssessmentEntity, UUID> {
                   WHEN a.decision='ACCEPTED' AND aa.completed_at is null THEN 'READY_TO_PLACE'
                   WHEN a.decision is null AND a.allocated_to_user_id is not null THEN 'IN_REVIEW'
                   WHEN a.decision is null AND a.allocated_to_user_id is null THEN 'UNALLOCATED'
-             END  as status
+             END  as status,
+             pdu.name as probationDeliveryUnitName
         from temporary_accommodation_assessments aa
              join assessments a on aa.assessment_id = a.id
              join applications ap on a.application_id = ap.id
              left outer join temporary_accommodation_applications taa on ap.id = taa.id
+             left outer join probation_delivery_units pdu on taa.probation_delivery_unit_id = pdu.id
        where taa.probation_region_id = ?1
              and (?2 is null OR ap.crn = ?2 OR lower(taa.name) LIKE CONCAT('%', lower(?2),'%'))
              and a.reallocated_at is null
@@ -148,6 +150,7 @@ interface AssessmentRepository : JpaRepository<AssessmentEntity, UUID> {
                join assessments a on aa.assessment_id = a.id
                join applications ap on a.application_id = ap.id
                left outer join temporary_accommodation_applications taa on ap.id = taa.id
+               left outer join probation_delivery_units pdu on taa.probation_delivery_unit_id = pdu.id
          where taa.probation_region_id = ?1
                and (?2 is null OR ap.crn = ?2)
                and a.reallocated_at is null
@@ -362,6 +365,7 @@ interface DomainAssessmentSummary {
   val crn: String
   val status: DomainAssessmentSummaryStatus?
   val dueAt: Timestamp?
+  val probationDeliveryUnitName: String?
 }
 
 enum class DomainAssessmentSummaryStatus {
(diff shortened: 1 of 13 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/AssessmentEntity.kt
```diff
@@ -362,6 +365,7 @@ interface DomainAssessmentSummary {
   val crn: String
   val status: DomainAssessmentSummaryStatus?
   val dueAt: Timestamp?
+  val probationDeliveryUnitName: String?
```
danhumphreys-moj, line 368: Presumably this is ok to add here?  `DomainAssessmentSummary` is used across CAS1/3.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/util/AssessmentUtils.kt
```diff
@@ -1,25 +1,13 @@
...
-      AssessmentSortField.personName -> compareValues((a.person as? FullPerson)?.name, (b.person as? FullPerson)?.name)
-      AssessmentSortField.personCrn -> compareValues(a.person.crn, b.person.crn)
-      AssessmentSortField.assessmentArrivalDate -> compareValues(a.arrivalDate, b.arrivalDate)
-      AssessmentSortField.assessmentStatus -> when {
-        a is TemporaryAccommodationAssessmentSummary && b is TemporaryAccommodationAssessmentSummary -> compareValues(a.status, b.status)
-        else -> throw RuntimeException("Cannot compare values of types ${a::class.qualifiedName} and ${b::class.qualifiedName} due to incomparable status types.")
-      }
-      AssessmentSortField.assessmentCreatedAt -> compareValues(a.createdAt, b.createdAt)
-      AssessmentSortField.assessmentDueAt -> throw BadRequestProblem(errorDetail = "Sorting by due date is not supported for CAS3")
-    }
```
danhumphreys-moj, line 22: I think this could be refactored even more.  There seemed to be lots of legacy compares here.  AssessmentUtils.kt could probably go.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/AssessmentTest.kt
```diff
@@ -1298,6 +1304,21 @@ class AssessmentTest : IntegrationTestBase() {
             AssessmentSortField.assessmentDueAt -> {
               ExpectedResponse.Error(HttpStatus.BAD_REQUEST, "Sorting by due date is not supported for CAS3")
             }
+
+            AssessmentSortField.applicationProbationDeliveryUnit -> {
+              ExpectedResponse.OK(
+                assessments
+                  .sortedWith(
+                    compareByDescending(nullsLast()) {
```
danhumphreys-moj, line 1312: When sorting a lists of strings that includes nulls, when using ORDER BY * DESC (via PageCriteria) Postgres puts nulls last, therefore `compareByDescending` and `nullsLast()` is required for this test to pass.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/AssessmentTest.kt
```diff
@@ -3537,6 +3558,7 @@ class AssessmentTest : IntegrationTestBase() {
         allocated = assessment.allocatedToUser != null,
         status = status,
         dueAt = assessment.dueAt?.toTimestamp(),
+        probationDeliveryUnitName = (assessment.application as? TemporaryAccommodationApplicationEntity)?.probationDeliveryUnit?.name,
```
danhumphreys-moj, line 3561: This this is ok, lots of tests will fail without as?.  If `assessment.application` is not `TemporaryAccommodationApplicationEntity` this returns null due to cast failing.  If `assessment.application` is not null but `probationDeliveryUnit` is null, then null is also returned, which makes sense for applications for which the PDU hasn't been specified (and would therefore need to be null).

muhammad-elabdulla (changes requested):

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/AssessmentTest.kt
muhammad-elabdulla: It still the pdu in application assessment returned with null values line 1309
Screenshot 2024-08-20 at 15 59 12

muhammad-elabdulla (approved):

PR #2184, opened 2024-08-19
CAS-318 - enabled use-ap-and-delius-to-update-users feature

daveawc:
switched the flags to use the ap-and-delius integration api for updating staff details.

I've also added a converted from StaffUserDetails to StaffDetails, and duplicated the calls from the community API and ap-and-delius API, so we should get the same responses from both endpoints. This was an issue for tests that used both endpoints in the same test due to a difference with the data that was being mocked. It should also make it easier when switching from the community-api.


muhammad-elabdulla:
LGTM

muhammad-elabdulla (approved):

PR #2185, opened 2024-08-19
CAS-544 Refactor bedspace search to improve performance

muhammad-elabdulla:
This [PR CAS-544](https://dsdmoj.atlassian.net/browse/CAS-544) is to improve the performance of bedspace search which include:

- Make the Booking entity dependencies lazy load
- Improve the query to search closest booking more efficient execution time 

This PR was tested with upgrade spring boot 3 branch to make sure that the changes are compatible

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BookingEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BookingEntity.kt
@@ -17,6 +17,7 @@ import java.util.UUID
 import javax.persistence.Entity
 import javax.persistence.EnumType
 import javax.persistence.Enumerated
+import javax.persistence.FetchType
 import javax.persistence.Id
 import javax.persistence.JoinColumn
 import javax.persistence.ManyToOne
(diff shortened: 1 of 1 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BookingEntity.kt
```diff
@@ -129,14 +130,30 @@ interface BookingRepository : JpaRepository<BookingEntity, UUID> {
...
   @Query(
-    "SELECT b FROM BookingEntity b WHERE (b.bed, b.departureDate) IN (" +
-      "  SELECT b2.bed, MAX(b2.departureDate)" +
-      "  FROM BookingEntity b2 " +
-      "  WHERE b2.departureDate <= :date " +
-      "  AND b2.bed.id IN :bedIds " +
-      "  AND SIZE(b2.cancellations) = 0 " +
-      "  GROUP BY b2.bed " +
-      ")",
+    """
```
Rosie-Brigham, line 133: As we are moving away from a standard JPQL here to optimise the query - could you add a couple of comments in on the query to highlight exactly what is doing what.
I think that may improve readability for an engineering coming fresh to this code, if they need to increase the scope of this search. 

```
... SELECT b2.bed_id,                       -- Select the bed_id
        MAX(b2.departure_date) AS departure_date -- Find the latest departure date for each bed
        FROM bookings b2 ...
```

Aware we don't use comments too much in the API code, but unsure why

muhammad-elabdulla: Comment added. I don't prefer to add comments in the code it should be readable except it its complicated to understand

muhammad-elabdulla: Thanks @daveawc that is good point

awhodgson256: @muhammad-elabdulla I believe using Exists with the subquery is more performant than count(*) > 0

muhammad-elabdulla: @awhodgson256 I will replace it with left join

muhammad-elabdulla: @daveawc  this query is missing some rows compare with the original one

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BookingEntity.kt
```diff
@@ -128,15 +129,32 @@ interface BookingRepository : JpaRepository<BookingEntity, UUID> {
...
+      WITH ClosestBooking AS
+         (SELECT b.bed_id, MAX(b.departure_date) departure_date
+          FROM bookings b
+          LEFT JOIN cancellations c ON b.id = c.booking_id
+          WHERE b.departure_date <= :date
+            AND b.bed_id IN :bedIds
+            AND c.id IS NULL
+          GROUP BY b.bed_id)
+          
+      SELECT b.*,
```
Rosie-Brigham, line 147: Do we want to be more specific with the columns pulled through on the booking?
I am wondering if we need to pull through the application, as that then has a oneToMany relationship with assessment. Maybe we could be really specific here - or maybe go down the chain of other things being pulled in from other models and lazy load other oneToMany relationships where relevant (such as assessments in the applicationEntity)

Just a thought

muhammad-elabdulla: No because it maps the query result to BookingEntity

awhodgson256 (approved):
Looks good to me for a first iteration, lets deploy and profile database cpu usage following modification

PR #2186, opened 2024-08-21
Fix gradle intellij import issue

davidatkinsuk:
A recent change broke importing the gradle config into a fresh intellij project (because in this cases there is no directly in invoked task). This commit fixes that issue.


RobBoothMOJ (approved):
Approved

PR #2192, opened 2024-08-21
APS-801 Deprecate CAS1 lost-beds endpoints

davidatkinsuk:
These have been replaced by the out-of-service-beds endpoints.


RobBoothMOJ (approved):
Approved

PR #2197, opened 2024-08-22
CAS-451 Setting memory resource to match prod configuration

muhammad-elabdulla:
This [PR CAS-451](https://dsdmoj.atlassian.net/browse/CAS-451) is to set memory resource to match prod configuration


davidatkinsuk (approved):
LGTM 👍

PR #2199, opened 2024-08-23
APS-1197: Update query for finding oos beds

RobBoothMOJ:
Ensures that the query to get out of service beds will always return the most recent record / revision for a bed and apply temporalities correctly.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/OutOfServiceBedTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/OutOfServiceBedTest.kt
@@ -1,5 +1,7 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.integration.cas1
 
+import com.fasterxml.jackson.core.type.TypeReference
+import org.assertj.core.api.Assertions.assertThat
 import org.hamcrest.Matchers.containsInAnyOrder
 import org.junit.jupiter.api.Nested
 import org.junit.jupiter.api.Test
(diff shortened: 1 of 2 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/OutOfServiceBedTest.kt
```diff
@@ -1244,6 +1250,164 @@ class OutOfServiceBedTest : InitialiseDatabasePerClassTestBase() {
     }
   }
 
+  @Nested
+  inner class GetOutOfServiceBedWithMultipleRevisionsInAllTemporalities : InitialiseDatabasePerClassTestBase() {
```
davidatkinsuk, line 1254: i appreciate trying to group these tests together, but i think they would be better sitting under the existing GetAllOutOfServiceBeds inner class? I've found that when we have multiple classes testing the same thing I tend to completely miss one of the sets of tests. I'm not sure if you can put nested classes in nested classes, which would be the easiest way to merge them

Does this not duplicate the existing 'Get All Out-Of-Service Beds filters by temporality correctly' test to some degree? Presumably it wasn't straight forward to 'fix' that test to cover the additional scenarios?

davidatkinsuk (approved):
functionally looks great, but i have some comments on tests

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/OutOfServiceBedTest.kt
RobBoothMOJ: I reworked the original test code and I think there was an issue in the way it was setting up the test data, with a separate oosb record for each revision instead of 1 oosb record with multiple revisions.  As such, it would always return a response for every temporality and when all 3 temporalities were in the specification it would return 3 separate records rather than limiting the responses to always return only the latest one.

PR #2201, opened 2024-08-23
APS-1192: Update query for finding oos beds

RobBoothMOJ:
Ensures that the query will ignore out of service bed records which have been cancelled and only return those out of service bed records which are still active.


muhammad-elabdulla (approved):
LGTM

PR #2202, opened 2024-08-27
APS-1214 - Add CAS1 API to list space bookings for premises

davidatkinsuk:
‘statuses’ are not considering in filtering, sorting or in the body of the response. These will be added in a subsequent iteration.


RobBoothMOJ (approved):
Approved

PR #2204, opened 2024-08-27
Convert eager load to lazy for performance

davidatkinsuk:
This commit updates some @ManyToOne and @OneToOne relationships to use eager loading to reduce the number of unneccessary database calls in production.

Placement Application - Typically this is viewed for a specific application, in which case application information isn’t required. Loading application information can lead to a cascade of eager loads.

Placement Date Entity - the relationships now marked as eager were only used in the hashcode/equals generation, which was unneccessary given that the ID uniquely identifies an instance

Placement Request Entity - the assessment property is not typically required but eagerly loaded. This is now lazy


muhammad-elabdulla (approved):
LGTM

davidatkinsuk:
closing this as it uncovered an issue when comparing classes on hibernate proxies. i'm going to recreate/address via separate pr

PR #2205, opened 2024-08-28
APS-1200: Update to only retrieve conviction documents for application

RobBoothMOJ:
Added cas1_only_list_conviction_documents feature flag
The feature flag determines whether only conviction documents for a specified application should be returned (cas1_only_list_conviction_documents = true) or offender documents should also be included (cas1_only_list_conviction_documents = false)


helm_deploy/values-prod.yaml
```diff
@@ -106,6 +106,7 @@ generic-service:
     USER-ALLOCATIONS_RULES_EMERGENCY-AND-SHORT-NOTICE-ASSESSMENTS_ALLOCATE-TO-USERS_WALES: KEVINPENNY1NPS
 
     FEATURE-FLAGS_CAS1-APPEAL-MANAGER-CAN-ASSESS-APPLICATIONS: true
+    FEATURE-FLAGS_CAS1_ONLY_LIST_CONVICTION_DOCUMENTS: true
```
davidatkinsuk, line 109: this should be false until testing is complete

we don't appear to have a value added for pre-prod - was this intended for there instead?

src/main/resources/application.yml
```diff
@@ -283,4 +283,5 @@ reports:
 
 feature-flags:
   cas1-appeal-manager-can-assess-applications: false
+  cas1_only_list_conviction_documents: true
```
davidatkinsuk, line 286: feature flags should generally be false by default

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/DocumentTransformer.kt
```diff
@@ -17,25 +17,14 @@ class DocumentTransformer {
   fun transformToApiFiltered(groupedDocuments: GroupedDocuments, convictionId: Long?) =
     transformToApi(groupedDocuments) { it.convictionId == convictionId.toString() }
 
-  fun transformToApiUnfiltered(groupedDocuments: GroupedDocuments) =
-    transformToApi(groupedDocuments) { true }
+  fun transformToApiUnfiltered(groupedDocuments: GroupedDocuments, onlyConvictionDocuments: Boolean) =
+    transformToApi(groupedDocuments, onlyConvictionDocuments) { true }
```
davidatkinsuk, line 21: This function is called transformToApiUnfiltered, but it's now doing filtering.

I wonder if it would be cleaner to remove transformToApiFiltered and transformToApiUnfiltered and have callers call transformToApi directly?

davidatkinsuk (changes requested):
Thanks Rob, i've marked as request changes as currently feature flag is true for prod where as should be false until testing complete

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ApplicationDocumentsTest.kt
```diff
@@ -39,16 +39,6 @@ class ApplicationDocumentsTest : InitialiseDatabasePerClassTestBase() {
         }
 
         val groupedDocuments = GroupedDocumentsFactory()
-          .withOffenderLevelDocument(
```
davidatkinsuk, line 42: can you recall why this was removed? given we're testing onlyConvictionDocuments = false i'd expect we'd want a non-conviction document in the results?

davidatkinsuk (approved):
LGTM, one minor comment to respond to

PR #2206, opened 2024-08-28
APS-1215 Add additional fields to Cas1SpaceBooking API

davidatkinsuk:
This commit adds additional fields and renames some existing fields on the Cas1SpaceBooking API Type to provide the information required for the placement details view.

The Cas1SpaceBookingTransformer has not been updated to correctly populate these new fields, adding placeholder values for any new mandatory fields. Population of these fields will be managed by a subsequent commit as new data model columns are required to support this.


RobBoothMOJ (approved):
Approved

PR #2207, opened 2024-08-28
CAS-365 - Add explicit count query for OOSB

davidatkinsuk:
When migrating to spring boot 3 hibernate 6 is struggling to generate a count query for the out of service beds query. This commit explicitly defines a count query to work around this issue


RobBoothMOJ (approved):
Approved

PR #2209, opened 2024-08-28
Add ability to define CAS-specific schemas

davidatkinsuk:
This commit reworks the buildSpecWithSharedComponentsAppended function in the gradle script to allow us to use cas-specific schema definitions. This helps better separate CAS API definitions, whilst ensuring irrelevant type definitions from one CAS API are not added to the other CAS API definitions.

This solution isn’t perfect. To keep the open api generator happy the new ‘cas1-schemas.yml’ needs a leading

```
components:
  schemas:
```

But this is stripped out when creating the ‘built’ versions of the schemas.

Furthermore, this only allows custom schemas to be defined per cas. ‘responses’ and ‘headers’ still have to be defined in _shared.yml


RobBoothMOJ (approved):
Approved

PR #2211, opened 2024-08-28
CAS-464 Extend bedspace search to include property attributes

muhammad-elabdulla:
This [PR CAS-464](https://dsdmoj.atlassian.net/browse/CAS-464) is to extend the current bedspace search to include filtering results by the following property attributes:
- Shared Property
- Single Occupancy

The change made is to create a CTE to return the premises Ids that match the filtering criteria. 

I added another condition in the where condition for Shared Property and Single Occupancy to include the properties that they don't have property attributes.


froddd:
Swagger doesn't seem to show the new search parameters, could these be added to the OpenAPI definitions to make sure they're up to date?

awhodgson256 (approved):
LGTM

muhammad-elabdulla:
@froddd this is a known issue as the BedSearchParameters for temporary accommodation inherent from BedSearchParameters and this parameters are only for temporary accommodation. These why it will not show in the swagger @Rosie-Brigham start looking at it

PR #2214, opened 2024-08-29
CAS-547 Remove crn parameter from assessment api endpoint

muhammad-elabdulla:
This [PR CAS-547](https://dsdmoj.atlassian.net/browse/CAS-547) is to remove the crn parameter from the assessment end point as it was replaced by the new parameter crnOrName


davidatkinsuk (approved):
LGTM 👍

PR #2216, opened 2024-08-29
CAS1 OOSB conflict check defect fix

davidatkinsuk:
When checking for out of service bed conflicts we were consider _all_ revision dates for related out of service bed record. This commit changes the logic to only consider the latest revision for a relation out of service bed record.

A similar fix to a different query was made via https://github.com/ministryofjustice/hmpps-approved-premises-api/commit/01b63d5cf9fe0bf0acbcf4921f5fa26c724fb943


RobBoothMOJ (approved):
Approved

PR #2217, opened 2024-08-29
APS-1202: Update to only retrieve adjudications within last 12 months for CAS1

RobBoothMOJ:
When feature flag is set AND request is for CAS1 the resulting data set will only contain adjudications which took place in the last 12 months. 

Added cas1_only_list_adjudications_up_to_12_months feature flag (true) 
Added X-Service-Header to API to determine whether request is for CAS1 or CAS3 
Feature flag determines whether adjudications older than 12 months are returned (true) if the request is also for CAS1


davidatkinsuk (approved):
LGTM 👍

PR #2219, opened 2024-08-29
refactor cas3 domain event service

daveawc:
Refactored the properties to its own config class to make testing easier
Refactore unit tests so we only use 1 service, and change the mockk response when we want a differerent answer, rather than having 2 mockk services.


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas3/DomainEventServiceTest.kt
```diff
@@ -71,23 +90,12 @@ class DomainEventServiceTest {
 
   private val detailUrl = "http://example.com/123"
 
-  private val domainEventService = buildService(emitDomainEventsEnabled = EventType.entries)
-  private val domainEventServiceEmittingDisabled = buildService(emitDomainEventsEnabled = listOf())
```
davidatkinsuk, line 75: this is a really nice pattern, i really dislike having to build multiple services to test config variations 👍

davidatkinsuk (approved):
LGTM 👍

PR #2220, opened 2024-08-30
APS-1226 - Add Job to truncate OOSB records for deleted beds

davidatkinsuk:
A bed can be deleted by setting its end date. Currently this is a manual operation (i.e. there is no UI operation to do this)

Currently it’s possible for an out of service bed record to exist past the corresponding beds end date. This causes a few issues:

* deleted beds can appear as ‘currently out of service’
* deleted beds are appearing in the out of service bed reports past their deletion date

This commit adds a migration job to ensure that all out of service bed records end before the corresponding bed’s end date. This should be ran whenever modifying a bed’s enddate.


RobBoothMOJ (approved):
Approved

PR #2221, opened 2024-08-30
Replace getPageable with extension functions

davidatkinsuk:
This makes PageCriteria to Pageable slightly more readable at the call site.

THis PR also deprecates getPageable in favour of getPageableOrAllPages


RobBoothMOJ (approved):
Approved

PR #2225, opened 2024-08-30
APS-1122 - Spring Boot 3 - Use 2nd Level Cache for Ref Data Entities

davidatkinsuk:
This commit enables an in-memory (i.e. per container) second level hibernate cache.

This cache has only been applied to read-only reference data that is managed via the flyway schema migrations. This ensures that the cache will always be correct because a change in migration would come via a new release (all containers would be restarted, flushing the caches).

Note that this cache is managed by Hibernate, and is not linked to Spring’s Redis cache that we use for various function-level caches. This was a deliberate choice to keep this initial implementation simple.

All caches share the same default settings. For reference, this is show many rows exist in production for each cached entity:

ap_areas - 7
cancellation_reasons -  22
cas1_out_of_service_bed_reasons - 7
characteristics - 66
departure_reasons - 38
destination_providers - 30
json_schemas - 7
local_authority_areas - 408
lost_bed_reasons - 14
move_on_categories - 33
non_arrival_reasons - 6
postcode_districts - 2862
probation_area_probation_region_mappings - 19
probation_delivery_units - 108
probation_regions - 13
referral_rejection_reasons - 9

Given there are 16 caches and a max cache size of 2MB, this will require 32MB of heap. Given each contains has 2GB heap and typically use just over 1GB of memory, there should be sufficient head room for this.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -40,6 +40,7 @@ dependencies {
   implementation("io.hypersistence:hypersistence-utils-hibernate-63:3.7.0")
   implementation("org.locationtech.jts:jts-core:1.19.0")
   implementation("org.hibernate:hibernate-spatial:6.4.4.Final")
+  implementation("org.hibernate.orm:hibernate-jcache")
   implementation("org.flywaydb:flyway-core")
   implementation("org.springframework.boot:spring-boot-starter-data-redis")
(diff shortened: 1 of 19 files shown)


danhumphreys-moj (approved):
LGTM

PR #2226, opened 2024-08-30
added user id to CAS3 domain events

daveawc:
- reordered the config for emitting messages so they are consistent between environments
- added user id to the CAS3 domain events
- made the userentity a non-nullable parameter, as we'll always have one (i think)
- removed test that's no longer needed

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas3/DomainEventService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas3/DomainEventService.kt
@@ -24,12 +24,14 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.DomainEventEn
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.DomainEventRepository
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.DomainEventType
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.TemporaryAccommodationApplicationEntity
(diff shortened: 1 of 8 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas3/DomainEventService.kt
```diff
@@ -209,8 +214,8 @@ class DomainEventService(
         createdAt = OffsetDateTime.now(),
         data = objectMapper.writeValueAsString(domainEvent.data),
         service = "CAS3",
-        triggerSource = null,
-        triggeredByUserId = null,
+        triggeredByUserId = user?.id,
+        triggerSource = if (user != null) TriggerSourceType.USER else TriggerSourceType.SYSTEM,
```
davidatkinsuk, line 218: in CAS1 it's possible for a domain event that was triggered via an API request for a specific user to still be classed as a SYSTEM event.

For example, when an application is submitted we sometimes automatically allocate the application's assessment to a user. This 'assessment allocated' domain event will have a TriggerSourceType of SYSTEM, despite triggeredByUserId being set to a value (i.e. the user didn't decide to allocate the assessment to someone, so it would be wrong to show their name next to the domain event)

For consistency, i'd recommend that TriggerSourceType is only ever set explicitly by the calling code (i.e. set to null unless calling code says otherwise). UI can then make the assumption that if source type is null the user should be shown (which it would have to do anyway to deal with existing domain events where this isn't set).

davidatkinsuk (commented):
One comment from me. I'm not approving as suspect probably best a CAS3 API person gives approved for this one

davidatkinsuk (approved):
LGTM 👍

PR #2231, opened 2024-09-03
APS-1268 - Add CAS1 specific premises summary endpoint

davidatkinsuk:
This is similar to the existing /premises/{premisesId}/summary, but only returns the fields currently being shown on the CAS1 premises UI (i.e. it doesn’t include bookings, which will instead be provided by /premises/{premisesId}/space-bookings)


Rosie-Brigham (approved):
Looks grand, nicely done

PR #2233, opened 2024-09-03
CAS-379 CAS3 allow turnaround to be zero

muhammad-elabdulla:
This [PR CAS-379](https://dsdmoj.atlassian.net/browse/CAS-379) is to allow the value of zero days for turnaround when:
- Create/update a property
- Change the turnaround for a booking

It will return the same validation message type isNotAPositiveInteger when the value of turnaround days is negative integer


davidatkinsuk (approved):
LGTM 👍

PR #2240, opened 2024-09-05
Add additional fields to CAS1 premises summary

davidatkinsuk:
Available Beds and Out of Service Beds Count have been added to the summary as these are required by the UI


RobBoothMOJ (approved):
Approved

PR #2241, opened 2024-09-05
APS-1244 - Create Booking Made Domain Event for CAS1 Space Bookings

davidatkinsuk:
Whilst we reuse the same APPROVED_PREMISES_BOOKING_MADE domain event for space bookings, we are populating the domain_events.cas1_space_booking_Id instead of domain_events.cas1_booking_Id to allow us to easily differentiate between the two types when rendering the timeline

This commit also adds a cas1SpaceBooking timeline url type to allow the UI to differentiate between legacy bookings and cas 1 space bookings when linking to the corresponding view placement page

The screenshot below shows how a legacy booking and a cas1 space booking appear on the timeline. Note that view placement link for CAS1 space booking is incomplete. This requires a change in the UI to recognise the new URL type (via APS-1268)

Screenshot 2024-09-05 at 12 17 51


RobBoothMOJ (approved):
Approved

PR #2242, opened 2024-09-05
APS-1205: Update to limit the type of prison case notes returned

RobBoothMOJ:
Added cas1_only_list_specific_prison_note_types feature flag
Added X-Service-Header to API so the API can determine whether the request is for CAS1 or CAS3 
When the cas1 case note feature flag is set  true (cas1_only_list_specific_prison_note_types) and the incoming request is for CAS1 then the API will only return case notes which match a specified list of types recorded in the offenderService.
The existing behaviour remains unchanged for requests for CAS3 or if the feature flag is disabled.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/OffenderService.kt
```diff
@@ -423,7 +424,11 @@ class OffenderService(
     }
   }
 
-  fun getPrisonCaseNotesByNomsNumber(nomsNumber: String): AuthorisableActionResult<List<CaseNote>> {
+  fun getFilteredPrisonCaseNotesByNomsNumber(nomsNumber: String, getCas1SpecificNoteTypes: Boolean): CasResult<List<CaseNote>> {
+    val cas1PrisonNoteTypes = listOf(
```
davidatkinsuk, line 428: for clarity could we please rename this to `cas1PrisonNoteTypesToInclude`. I don't know why but my brain assumed it was notes to exclude for a few minutes and I got confused :)

davidatkinsuk (approved):
LGTM, just one minor comment

PR #2245, opened 2024-09-05
refactored update cas3 assessment to use cas result

daveawc:
This removes the use of deprecated methods and replaces it with CasResult. This is changing as part of CAS-440 so a good point to refactor it.


davidatkinsuk (approved):
LGTM 👍

PR #2246, opened 2024-09-06
added new method to convert to CasResult

daveawc:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/AssessmentController.kt
```diff
@@ -144,31 +169,14 @@ class AssessmentController(
...
-          val updated =
-            assessmentService.updateAssessment(
-              user,
-              assessmentId,
-              objectMapper.writeValueAsString(updateAssessment.data),
-            )
-          val assessmentValidationResult =
-            when (updated) {
-              is AuthorisableActionResult.Success -> updated.entity
-              is AuthorisableActionResult.NotFound -> throw NotFoundProblem(assessmentId, "Assessment")
```
davidatkinsuk, line 156: Looking at `AssessmentService.updateAssessment()`, it's not populating `assessmentId, "Assessment"` for `AuthorisableActionResult.NotFound()`, so you're loosing this context. I think this needs adding into `updateAssessment()`

davidatkinsuk (approved):
Just one comment from me (aside the discussion we've already had about not using intellij formatting going forward :D )

PR #2251, opened 2024-09-09
CAS-480 Add wheelchair accessible to bedspace search.

danhumphreys-moj:
Ref: https://dsdmoj.atlassian.net/browse/CAS-480

Extend bedspace query with wheelchair accessible property.


muhammad-elabdulla:
LGTM
I will add a integration test to test if its filtering more than one property attribute like wheelchair accessibility and single occupancy it will be retun the expected search result

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BedSearchService.kt
```diff
@@ -148,15 +148,28 @@ class BedSearchService(
           return@validated fieldValidationError
         }
 
+        var premisesCharacteristicIds = listOf<UUID>()
```
davidatkinsuk, line 151: could we maybe refactor the code to avoid the need for a mutable premiseCharacteristicIds? e.g.

```
val premiseCharacteristicIds = premisesCharacteristicsNames?.let {
    val premisesCharacteristics = characteristicService.getCharacteristicsByNames(premisesCharacteristicsNames)
    premisesCharacteristicIds = premisesCharacteristics.filter { it.isActive && it.matches(ServiceName.temporaryAccommodation.value, "premises") }.map { it.id }.toList()
} ?: emptyList()
```

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/repository/BedSearchRepository.kt
```diff
@@ -231,34 +216,49 @@ ORDER BY distance_miles;
...
+      addValue("premises_characteristic_ids", premisesCharacteristicsIds)
+      addValue("premises_characteristic_ids_count", premisesCharacteristicsIds.size)
+    }
+
+    var optionalFilters = ""
+    if (premisesCharacteristicsIds.any()) {
+      optionalFilters += "$temporaryAccommodationPremisesCharacteristicFilter AND\n"
     }
 
+    val query = temporaryAccommodationSearchQuery.replace("#OPTIONAL_FILTERS", optionalFilters)
```
davidatkinsuk, line 258: we could maybe avoid mutation by doing something like

```
val optionalFilters = if (premisesCharacteristicsIds.any()) {
      optionalFilters += "$temporaryAccommodationPremisesCharacteristicFilter AND\n"
    } else { 
       ""
} 
```

davidatkinsuk (approved):
LGTM. Just a few minor optional proposals. Could you please rebase before merging 👍

PR #2253, opened 2024-09-09
CAS-440 - Update to save assessments updates as domain events

daveawc:
This PR is the second phase in adding assessment updates, where a change will now be saved as a domain event that will **not** be emitted to delius. The next step is to add a migration script to back track any changes up to now, and then returning the domain events will be in an additional PR

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas3/Cas3AssessmentService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas3/Cas3AssessmentService.kt
@@ -1,6 +1,7 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas3
 
 import org.springframework.stereotype.Service
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.cas3.model.CAS3AssessmentUpdatedField
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.UpdateAssessment
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.AssessmentRepository
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.TemporaryAccommodationAssessmentEntity
@@ -8,6 +9,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.findAssessmentById
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.results.CasResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.UserAccessService
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.toUiFormat
 import java.time.LocalDate
 import java.util.UUID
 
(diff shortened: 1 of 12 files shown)


src/main/resources/static/cas3-domain-events-api.yml
```diff
@@ -551,6 +551,21 @@ components:
...
+      allOf:
+        - $ref: '#/components/schemas/CAS3Event'
+        - type: object
+          properties:
+            fieldName:
+              type: string
+            updatedFrom:
+              type: string
+            updatedTo:
+              type: string
```
davidatkinsuk, line 564: would it be worth making this into an array of updated fields, even if only one is being updated at the moment? If you ever need to support multiple fields in the future it'll be awkward when deserializing the JOSN as you need to consider the 'old' schema. You'll need to deal with schema version control which we have an approach for but best avoided if possible. see https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/main/doc/how-to/modifying_domain_event_schemas.md

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas3/Cas3AssessmentService.kt
```diff
@@ -36,22 +37,52 @@ class Cas3AssessmentService(
...
+      if (newReleaseDate.isAfter(currentAccommodationReqDate)) {
+        return notAfterValidationResult(currentAccommodationReqDate)
+      }
+
+      val domainEvent =
+        domainEventBuilder.buildAssessmentUpdatedDomainEvent(
+          assessment = assessment,
+          fieldName = "releaseDate",
+          updatedFrom = assessment.currentReleaseDate().toString(),
+          updatedTo = newReleaseDate.toString(),
```
davidatkinsuk, line 52: in CAS1 we use the LocalDate.toUiFormat() extension function in Dates.kt whenever converting dates to strings. Is LocalDate.toString() producing dates that match how you present dates elsewhere on the UI? (i'm assuming this will be shown in a timeline somewhere?)

davidatkinsuk (approved):
Just a few proposal from me, feel free to ignore as may not make sense in context. I'd prefer someone from CAS3 reviews this too as they'll understand the wider context 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas3/Cas3AssessmentService.kt
```diff
@@ -36,22 +39,60 @@ class Cas3AssessmentService(
...
+      }
+
+      val domainEvent =
+        domainEventBuilder.buildAssessmentUpdatedDomainEvent(
+          assessment = assessment,
+          listOf(
+            CAS3AssessmentUpdatedField(
+              fieldName = "accommodationRequiredFromDate",
+              updatedFrom = assessment.currentAccommodationRequiredFromDate().toUiFormat(),
+              updatedTo = newAccommodationRequiredFromDate.toString(),
```
davidatkinsuk, line 78: if using toUiFormat presumably would want to do that here too

muhammad-elabdulla: May be you need to use it also .toUiFormat()

muhammad-elabdulla (approved):

PR #2255, opened 2024-09-09
APS-1264 send booking made email to pr creator

davidatkinsuk:
a change was made on https://github.com/ministryofjustice/hmpps-approved-premises-api/commit/e8c803178839a27feec829a36edc2ecd803b1c95 to sending booking emails to placement request creators, if it’s a different user to the application creator. This change did not work because it assumed the relationship between booking and placement request was populated when the code was called, which was not the case.

This commit reworks the functionality in Cas1BookingEmailService, updating the function to be explicitly provided with the placement request (PlacementApplicationEntity). It also adds an integration test for this scenario.


RobBoothMOJ (approved):
Approved

PR #2256, opened 2024-09-09
APS-1274 Harmonise OOSB Reasons across environments

davidatkinsuk:
Before this commit the cas1_out_of_service_bed_reasons reference data was populated using the update_cas1_out_of_service_bed_reasons migration job, which copied the data from the lost beds reasons table.

This is awkward when testing locally because we have to remember to run the migration job step.

This commits adds a migration script that populates the update_cas1_out_of_service_bed_reasons, where they don’t already exist. This will have no affect on environments where this data is already populated (e.g. prod, preprod, dev)


RobBoothMOJ (approved):
Approved

PR #2259, opened 2024-09-10
APS-1284 - Allow call to /swagger-ui.html

davidatkinsuk:
We often receive an AuthenticationCredentialsNotFoundException for GET /swagger-ui.html. This is a legitimate call which should redirect to swagger-ui/index.html. Therefore, this request should be allowed


RobBoothMOJ (approved):
Approved

PR #2265, opened 2024-09-11
CAS1 - Add ability to withdraw all entities to JANITOR and CRU_MEMBER

davidatkinsuk:
Prior to this commit roles were used to determine if various CAS1 entities could be withdrawn. We now use permissions

This commit also adds the ability for JANITOR and CRU_MEMBER to withdrawal all entity types.


RobBoothMOJ (approved):
Approved

PR #2267, opened 2024-09-11
APS-1286 - Add cancellation fields to space booking data model

davidatkinsuk:
This PR also excludes cancelled space bookings when listing placements for a premises

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/factory/Cas1SpaceBookingEntityFactory.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/factory/Cas1SpaceBookingEntityFactory.kt
@@ -4,6 +4,7 @@ import io.github.bluegroundltd.kfactory.Factory
 import io.github.bluegroundltd.kfactory.Yielded
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApprovedPremisesApplicationEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApprovedPremisesEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.CancellationReasonEntity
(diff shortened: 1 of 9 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/factory/Cas1SpaceBookingEntityFactory.kt
```diff
@@ -129,6 +134,22 @@ class Cas1SpaceBookingEntityFactory : Factory<Cas1SpaceBookingEntity> {
     this.keyWorkerAssignedAt = { keyWorkerAssignedAt }
   }
 
+  fun withCancellationOccurredAt(occurredAt: LocalDate?) = apply {
+    this.cancellationOccurredAt = { occurredAt }
+  }
+
+  fun withCancellationRecordedAt(recordedAt: Instant?) = apply {
```
Rosie-Brigham, line 141: Could you remind me the difference between Occured at and recorded at?
Why do these need to be two separate rows?
I wonder if there is a way to make it more obvious why the two times are required, as it may not be obvious to someone new to the codebase

Rosie-Brigham (commented):
I just wonder if there is a way to be clearer about occurred and recorded?
I know we seem to be very anti comments here, but would a comment in the model help?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/factory/Cas1SpaceBookingEntityFactory.kt
davidatkinsuk: On the UI we ask the user to define the date that the cancellation occurred, which may not necessarily be the same as when it was entered into the system. It's never been clear to me why these would be different - I suspect it was to help in the early days when 'migrating' existing bookings into the system.

I'm pro comment if the code doesn't speak for itself so i've added a comment.

When we get to the wireframes for space booking cancellation we should double check if we actually still need to capture this and maybe remove it

PR #2270, opened 2024-09-11
APS-1259 - Add AP Area to Cas1PremiseSummary API Model

davidatkinsuk:
This commit also:

* Refactors the getPremiseSummary code to use a transformer to map from the entity model into the API Model
* Makes all fields on Cas1PremiseSummary mandatory


RobBoothMOJ (approved):
Approved

PR #2271, opened 2024-09-11
CAS-342 Subject Access Request - Local Testing Fixes

williamfalconeruk:
includes:
- Fixing of null date requests
- Updating endpoint authorization
- use Map interface to expose as entries which serialises correctly to JSON  on endpoint

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/CAS3SubjectAccessRequestServiceTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/CAS3SubjectAccessRequestServiceTest.kt
@@ -42,6 +42,28 @@ class CAS3SubjectAccessRequestServiceTest : SubjectAccessRequestServiceTestBase(
   }
 
   @Test
+  fun `Get CAS3 Information - Null dates check`() {
+    val (offenderDetails, _) = `Given an Offender`()
+    val result =
+      sarService.getCAS3Result(offenderDetails.otherIds.crn, offenderDetails.otherIds.nomsNumber, null, null)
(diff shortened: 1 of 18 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/CAS3SubjectAccessRequestServiceTest.kt
```diff
@@ -41,6 +41,28 @@ class CAS3SubjectAccessRequestServiceTest : SubjectAccessRequestServiceTestBase(
...
+      """ 
+     {
+         "Applications": [ ],
+         "Assessments": [ ],
+         "AssessmentReferralHistoryNotes" : [ ],
+         "Bookings": [ ],
+         "BookingExtensions": [ ],
+         "Cancellations": [ ],
+         "DomainEvents": [ ],
+         "DomainEventsMetadata": [ ]
```
davidatkinsuk, line 59: just checking an assumption - the intention of these new checks are to make sure the code doesn't fall over when dates aren't defined, and if data was defined we'd be seeing that here (but to keep things simple you haven't added any test data for these tests)

davidatkinsuk (approved):
LGTM, with one assumption noted

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/CAS3SubjectAccessRequestServiceTest.kt
williamfalconeruk: that's correct. could have made more tests for this, just covering off the scenario i found in testing yesterday.

PR #2272, opened 2024-09-11
APS-1288 - Send booking withdrawn email to placement app creator

davidatkinsuk:
This commit ensure that booking withdrawn emails for a given request for placement are sent to the person who created that request for placement, if not the same as the application’s creator.


RobBoothMOJ (approved):
Approved

PR #2274, opened 2024-09-12
APS-1286 Refactor CAS1 booking withdrawn domain event logic in preparation for space bookings

davidatkinsuk:
The goal of this PR is to extract the create booking withdrawn domain event logic from the booking service into the Cas1BookingDomainEventService, and refactor it to allow for the logic to be re-used for space bookings


RobBoothMOJ (approved):
Approved

PR #2276, opened 2024-09-12
APS-1294 Use correct default ap-and-delius-context config

davidatkinsuk:
The default URL configurations in application.yml should match those used for local deployments.


RobBoothMOJ (approved):
Approved

PR #2279, opened 2024-09-12
APS-1223: Implement M&M V2 Record Arrival API

RobBoothMOJ:
This PR implements a record arrival API for M&M V2

Adds a new User Permission : CAS1_SPACE_BOOKING_RECORD_ARRIVAL.
The endpoint is only accessible to users with CAS1_SPACE_BOOKING_RECORD_ARRIVAL permission. 
Updates existing space booking with arrival and departure dates.
Creates a domain event to record the arrival.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1SpaceBookingManagementDomainEventService.kt
```diff
@@ -0,0 +1,104 @@
...
+            premises = Premises(
+              id = premises.id,
+              name = premises.name,
+              apCode = premises.apCode,
+              legacyApCode = premises.qCode,
+              localAuthorityAreaName = premises.localAuthorityArea!!.name,
+            ),
+            applicationSubmittedOn = application.submittedAt!!.toLocalDate(),
+            keyWorker = StaffMember(
+              staffCode = updatedCas1SpaceBooking.keyWorkerStaffCode!!,
```
davidatkinsuk, line 89: really minor comment - can we use keyWorker.staffCode here instead for consistency (i appreciate you've duplicated existing code)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1SpaceBookingManagementDomainEventService.kt
```diff
@@ -0,0 +1,104 @@
...
+              staffIdentifier = keyWorker.staffIdentifier,
+              forenames = keyWorker.staff.forenames,
+              surname = keyWorker.staff.surname,
+              username = null,
+            ),
+            arrivedAt = updatedCas1SpaceBooking.actualArrivalDateTime!!,
```
davidatkinsuk, line 95: updatedCas1SpaceBooking.actualArrivalDateTime!! is used in a few places. could you please declare that as a val at the top as actualArrivalDate and use that instad

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas1/SpaceBookingController.kt
```diff
@@ -120,7 +127,22 @@ class SpaceBookingController(
...
+      bookingId,
+      cas1NewArrival,
+    )
+
+    when (result) {
+      is CasResult.Success -> Unit
+      is CasResult.FieldValidationError -> throw BadRequestProblem(invalidParams = result.validationMessages)
+      is CasResult.ConflictError -> throw ConflictProblem(id = result.conflictingEntityId, conflictReason = result.message)
+      else -> throw InternalServerErrorProblem("An error occurred recording an arrival at premises: $premisesId for space booking: $bookingId")
+    }
```
davidatkinsuk, line 143: I think we can just use extractEntityFromCasResult() to deal with error cases, see an example in getSpaceBookingById

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/Cas1SpaceBookingTest.kt
```diff
@@ -789,4 +794,238 @@ class Cas1SpaceBookingTest {
...
+        .is4xxClientError
+        .expectBody()
+        .jsonPath("title").isEqualTo("Conflict")
+        .jsonPath("status").isEqualTo(409)
+        .jsonPath("detail")
+        .isEqualTo("An arrival is already recorded for this Space Booking: ${spaceBookingWithArrivalDate.id}")
+    }
+
+    @Test
+    fun `Returns 500 Internal Server Error if unexpected failure occurs - invalid key worker )`() {
```
davidatkinsuk, line 945: i'm going to create a ticket to work what to do when this happens as i feel like it shouldn't necessarily block the activity.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/Cas1SpaceBookingTest.kt
```diff
@@ -789,4 +794,238 @@ class Cas1SpaceBookingTest {
...
+            arrivalDateTime = LocalDateTime.now().toInstant(ZoneOffset.UTC),
+          ),
+        )
+        .exchange()
+        .expectStatus()
+        .isForbidden
+    }
+
+    @Test
+    fun `Returns 400 Bad Request if the premises id supplied has no matching premises record `() {
```
davidatkinsuk, line 857: i appreciate the coverage we're getting for the error cases in the integration tests here but i do wonder if we can make our lives easier by not repeating any validation test that's already in the service unit test going forward? (especially as these scenarios shouldn't _really_ happen if the UI is validating requests).

Of course if the validation depends on a non trivial bit of SQL/database access we should integration test it. Something to consider going forward!

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas1/Cas1SpaceBookingManagementDomainEventServiceTest.kt
```diff
@@ -0,0 +1,218 @@
...
+      assertThat(data.applicationId).isEqualTo(application.id)
+      assertThat(data.applicationSubmittedOn).isEqualTo(application.submittedAt!!.toLocalDate())
+      assertThat(data.applicationUrl).isEqualTo("http://frontend/applications/${application.id}")
+      assertThat(data.arrivedAt).isEqualTo(arrivalDate)
+      assertThat(data.deliusEventNumber).isEqualTo(application.eventNumber)
+      assertThat(data.premises.id).isEqualTo(premises.id)
+      assertThat(data.premises.name).isEqualTo(premises.name)
+      assertThat(data.premises.apCode).isEqualTo(premises.apCode)
+      assertThat(data.premises.legacyApCode).isEqualTo(premises.qCode)
+      assertThat(data.premises.localAuthorityAreaName).isEqualTo(premises.localAuthorityArea!!.name)
```
davidatkinsuk, line 179: i would be included to not repeat all these assertions and just focus on the field that is different from other unit tests

davidatkinsuk (approved):
A few comments from me, I think there's one that definitely needs addressing (case worker potentially not being set), and some thoughts on how we can streamline some testing going foward

PR #2280, opened 2024-09-12
Feature/cas 440 add notes for referral date change 4

daveawc:
This PR adds in domain events when an assessment has been updated. It takes the domain events and transforms them into the existing ReferralHistoryNote format. It's not ideal, but with fitting in with existing limitations with how the referral notes has been designed/adapted, so it is what it is, without having a major refactor.

I'll look to add some additional unit tests in a following PR.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/AssessmentTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/AssessmentTest.kt
@@ -2905,8 +2905,8 @@ class AssessmentTest : IntegrationTestBase() {
           CAS3AssessmentUpdatedEvent::class.java,
         )
 
-        assertThat(releaseDateUpdatedEvent.eventType).isEqualTo("accommodation.cas3.assessment.updated")
-        assertThat(accommodationReqFromUpdatedEvent.eventType).isEqualTo("accommodation.cas3.assessment.updated")
+        assertThat(releaseDateUpdatedEvent.eventType).isEqualTo(EventType.assessmentUpdated)
+        assertThat(accommodationReqFromUpdatedEvent.eventType).isEqualTo(EventType.assessmentUpdated)
 
         assertThat(releaseDateUpdatedEvent.updatedFields).isEqualTo(
           listOf(


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/AssessmentTransformer.kt
```diff
@@ -76,6 +85,23 @@ class AssessmentTransformer(
...
+        } else {
+          assessmentReferralHistoryNoteTransformer.transformJpaToApi(it)
+        }
+      }.toMutableList()
+
+      notes.addAll(
+        cas3Events.map {
+          val user = userService.findById(it.triggeredByUserId!!)
+            .orElseThrow { NotFoundProblem(it.triggeredByUserId!!, "UserEntity") }
+          val event = objectMapper.readValue<CAS3AssessmentUpdatedEvent>(it.data)
```
davidatkinsuk, line 100: IMO the only thing that should 'understand' the JSON in the domain events is the Domain Event Service to better encapsulate how we manage the serialization/deserialization.

Furthermore, this code is assuming that cas3Events only contains CAS3 Assessment Update Events but there's nothing at compile time validating this

I propose the following:

1. transformJpaToApi should accept List <CAS3AssessmentUpdatedEvent>
2. this list is provided by your domain event service (we do something similar in CAS1). You could maybe update domainEventService.getAssessmentUpdatedEvents to return a list of uk.gov.justice.digital.hmpps.approvedpremisesapi.model.DomainEvent instead of the entity? (you'll need to add 'triggeredByUserId' to that)

src/main/resources/static/_shared.yml
```diff
@@ -2883,6 +2884,12 @@ components:
           type: string
         isWithdrawn:
           type: boolean
+        domainEventNote:
```
davidatkinsuk, line 2887: should this not just be called domainEvent? Can we link it to the actual domain event definition?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/AssessmentReferralHistoryNoteTransformer.kt
```diff
@@ -40,6 +44,16 @@ class AssessmentReferralHistoryNoteTransformer {
     else -> throw RuntimeException("Unsupported ReferralHistoryNote type: ${jpa::class.qualifiedName}")
   }
 
+  fun transformToReferralHistoryDomainEventNote(updatedEvent: CAS3AssessmentUpdatedEvent, user: UserEntity) =
+    ReferralHistoryDomainEventNote(
+      id = updatedEvent.id,
+      createdAt = updatedEvent.timestamp,
+      message = "",
```
davidatkinsuk, line 51: is it possible to make this type nullable in the schema to better reflect this may happen? I guess in this case the UI will use the message details to build a message?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
```diff
@@ -72,6 +72,8 @@ class UserService(
     return userRepository.findByNameContainingIgnoreCase(name)
   }
 
+  fun findById(id: UUID) = userRepository.findById(id)
```
davidatkinsuk, line 75: I think Kotlin neatly removes the need for Optionals given how well it deals with nulls, so maybe here we could use findByIdOrNull instead?

davidatkinsuk (changes requested):
Sorry, a bunch of comments from me. Could you please let me know when your other PR is merged and i'll review the tests

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/AssessmentTransformer.kt
davidatkinsuk: Makes sense, and keeps the code generic to handle other domain event types here if required 👍

davidatkinsuk (approved):
LGTM 👍

PR #2283, opened 2024-09-13
APS-1279: Correct wording on validation error message

RobBoothMOJ:
Correct wording on validation error message when creating a booking from a placement request


davidatkinsuk (approved):
LGTM

As discussed, whilst it would be good to remove the 'create booking for bed' logic, we decided that as this code is due for removal once space bookings is rolled out, it's not worth the effort.

PR #2285, opened 2024-09-13
CAS-633 Upgrade dataframe library to 0.13.1

danhumphreys-moj:
https://dsdmoj.atlassian.net/browse/CAS-633

Upgrading this kotlinx-dataframe dependency resolves the commons-compress vulnerability issue associated with commons-compress when the gradle OWASP job is run.

Upgrading the version alone results in a vulnerability being raised with swagger-parser-safe-url-resolver-2.1.20.jar so this transitive dependency which is a referenced library in dataframe-openapi also needs to be excluded.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -72,7 +72,9 @@ dependencies {
 
   implementation("com.github.doyaaaaaken:kotlin-csv-jvm:1.9.3")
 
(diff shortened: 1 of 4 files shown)


davidatkinsuk (approved):
LGTM 👍 

I tested this locally in CAS1 and XLSX are still reported without issue

PR #2290, opened 2024-09-17
Add X-CAS-User-ID Header for CAS1

davidatkinsuk:
This header is included alongside the X-CAS-User-Version header to support the UI detecting and handling user configuration changes.

To get the user id we had to use a slightly more ‘clunky’ bit of SQL that results in more code needed to interpret the results, but this ensures we’ll always get the user id even if the user has no roles set.


RobBoothMOJ (approved):
Approved

PR #2294, opened 2024-09-17
CAS-440 - added feature flag for returning assessmentUpdated domain events

daveawc:
Temporarily adding this feature flag until the UI is ready, as the UI will not be able to successfully render the domain event message details in it's current state.


davidatkinsuk (approved):
LGTM 👍

PR #2296, opened 2024-09-18
wip

davidatkinsuk:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas1CruManagementAreaEntity.kt
```diff
@@ -0,0 +1,33 @@
...
+@Table(name = "cas1_cru_management_areas")
+@org.hibernate.annotations.Cache(usage = CacheConcurrencyStrategy.READ_ONLY)
+data class Cas1CruManagementAreaEntity(
+  @Id
+  val id: UUID,
+  val name: String,
+  val emailAddress: String?,
+  val notifyReplyToEmailId: String?,
+) {
+  companion object {
```
Rosie-Brigham, line 30: Oh I have not come across these before, this seems neat

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas1/Cas1ReferenceDataController.kt
```diff
@@ -0,0 +1,44 @@
...
+  override fun getCruManagementAreas(): ResponseEntity<List<Cas1CruManagementArea>> {
+    val womensEstateEnabled = featureFlagService.getBooleanFlag("cas1-womens-estate-enabled")
+    return ResponseEntity.ok(
+      cas1CruManagementAreaRepository.findAll().map {
+        Cas1CruManagementArea(
+          id = it.id,
+          name = it.name,
+        )
+      }.filter {
+        womensEstateEnabled || it.id != Cas1CruManagementAreaEntity.WOMENS_ESTATE_ID
```
Rosie-Brigham, line 40: Oh this is neat.

Rosie-Brigham (approved):
Looks grand

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas1CruManagementAreaEntity.kt
davidatkinsuk: i'm not sure if you're too familiar with java but it's the equivalent of 

`public static final UUID WOMENS_ESTATE_ID = ...;`

It's more clunky but i get why kotlin designers took this approach

davidatkinsuk:
> Looks grand

Cheers. FYI I did a force push while you were reviewing. the only change was making the table name plural for consistency with other table names

PR #2297, opened 2024-09-18
APS-1335 - Assign CAS1_PREMISES_VIEW_SUMMARY to relevant role

davidatkinsuk:
As the /cas1/premises/{premiseId} endpoint is going to be used for premise summaries, we need to ensure that all roles that access this page have the neccessary permissions


RobBoothMOJ (approved):
Approved

PR #2298, opened 2024-09-18
APS-1331 - Add gender to approved premise entity

davidatkinsuk:
This also adds a new endpoint to list approved premises with an optional gender-specific filter

It also introduces an ApprovedPremisesRepository to better split CAS1 code away from other CASs


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas1/Cas1PremisesController.kt
```diff
@@ -29,4 +32,20 @@ class Cas1PremisesController(
...
       )
   }
+
+  override fun getPremisesSummaries(gender: Cas1ApprovedPremisesGender?): ResponseEntity<List<Cas1PremisesBasicSummary>> {
+    return ResponseEntity
+      .ok()
+      .body(
+        cas1PremisesService.getPremises(
+          gender = when (gender) {
+            Cas1ApprovedPremisesGender.male -> ApprovedPremisesGender.MALE
```
Rosie-Brigham, line 42: I believe this needs to be diffrent, let me just check...

src/main/resources/static/codegen/built-cas1-api-spec.yml
```diff
@@ -6558,4 +6608,9 @@ components:
         staffCode:
           type: string
       required:
-        - staffCode
\ No newline at end of file
+        - staffCode
+    Cas1ApprovedPremisesGender:
+      type: string
+      enum:
+        - male
```
Rosie-Brigham, line 6615: See above comment

Rosie-Brigham (dismissed):

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PremisesEntity.kt
```diff
@@ -259,6 +267,11 @@ class ApprovedPremisesEntity(
   status,
 )
 
+enum class ApprovedPremisesGender {
+  MALE,
```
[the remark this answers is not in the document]
davidatkinsuk: Done. I did this on an additional commit as it was going to be too difficult to unpick the individual commit.

RobBoothMOJ (approved):
Approved

PR #2299, opened 2024-09-18
CAS-593 - Remove ‘conspicuous’ logging

davidatkinsuk:
To work around an issue where cache refresh logs were not appearing in app insights, all logs were output at an ERROR level (which did appear in app insights).

As part of the spring boot 3 upgrade there is no longer a specific slf4j appender for app insights, and as such, this should no longer be an issue.

This commit removes such ‘conspicuous’ logs and instead logs them at info level


RobBoothMOJ (approved):
Approved

PR #2300, opened 2024-09-19
APS-1315 Capture CRU Management Area on Application Submission

davidatkinsuk:
This PR adds a new 'default_cru1_management_area_id' column to approved_premises_application

Initially this will always be derived from the AP Area's default Management Area. In subsequent commits we will use the ‘Women’s Estate’ Management Area if the application is for a female.

This PR also backfills the management area for any submitted applications.


RobBoothMOJ (approved):
Approved

PR #2301, opened 2024-09-19
APS-1205: Replace underscores with hyphens in feature flag names

RobBoothMOJ:
Replaced underscores with hyphens to correct issue with feature flags not being recognised in deployed environments.


davidatkinsuk (approved):
LGTM 👍

PR #2302, opened 2024-09-20
Remove direct usage of ApAreaEntityFactory in IT

davidatkinsuk:
Instead of using factories directly in tests we should, where possible, use the ‘Given’ functions as these reduce the amount of duplicated boiler plate code.

This commit uses `Given an AP Area` function whereever an ap area is required in tests. This is in preparation for adding a new mandatory field to AP Area.


RobBoothMOJ (approved):
Approved

PR #2306, opened 2024-09-20
APS-1315 Add link from ap area to CAS1 CRU management area

davidatkinsuk:
In subsequent commits the ap_areas.default_cru1_management_area_id will be used to determine a user's Cas1CruManagementAreaEntity if no override is specified.


RobBoothMOJ (approved):
Approved

PR #2310, opened 2024-09-21
APS-1239: Implement API to record M&M v2 departure

RobBoothMOJ:
Implementation of the M&M V2 API to record departure for a Space Booking.

SQL Script to add departure_reason and deperture_move_on_category to cas1_space_bookings 
Add departure_reason and departure_move_on_category to cas1SpaceBookingEntity 
Only users with CAS1_SPACE_BOOKING_RECORD_DEPARTURE can call the API


src/main/resources/static/cas1-schemas.yml
```diff
@@ -290,6 +290,10 @@ components:
           description: actual departure date or, if not known, the expected departure date
           type: string
           format: date
+        departureReason:
+          $ref: '_shared.yml#/components/schemas/DepartureReason'
+        departureMoveOnCategory:
+          $ref: '_shared.yml#/components/schemas/MoveOnCategory'
```
davidatkinsuk, line 296: in an attempt to try and reduce how much unneccessary information we provide on the API i've used the generic 'NamedId' to provide information about premise and apArea  on Cas1SpaceBooking (see line 267). I propose we do the same for these two fields and only add more detailed types if required by the UI.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/UserEntity.kt
```diff
@@ -386,6 +386,7 @@ enum class UserRole(val service: ServiceName, val cas1ApiValue: ApprovedPremises
       UserPermission.CAS1_OUT_OF_SERVICE_BED_CREATE,
       UserPermission.CAS1_SPACE_BOOKING_LIST,
       UserPermission.CAS1_SPACE_BOOKING_RECORD_ARRIVAL,
+      UserPermission.CAS1_SPACE_BOOKING_RECORD_DEPARTURE,
```
davidatkinsuk, line 389: can we add this to cru member and janitor as well please (going forward i think janitor needs everything new adding to it). See https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/4357851312/User+roles+in+Approved+Premises

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas1/SpaceBookingController.kt
```diff
@@ -140,7 +140,16 @@ class SpaceBookingController(
     bookingId: UUID,
     cas1NewDeparture: Cas1NewDeparture,
   ): ResponseEntity<Unit> {
-    return super.premisesPremisesIdSpaceBookingsBookingIdDeparturePost(premisesId, bookingId, cas1NewDeparture)
+    userAccessService.ensureCurrentUserHasPermission(UserPermission.CAS1_SPACE_BOOKING_RECORD_DEPARTURE)
+
+    extractEntityFromCasResult(
```
davidatkinsuk, line 145: it may be more readable to use `ensureEntityFromCasResultIsSuccess` instead of `extractEntityFromCasResult` here (and also in the arrival controller above). Technically it does the exact same thing but `ensureEntityFromCasResultIsSuccess` makes it (slightly?) clearer that it will throw an exception if not a success

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1SpaceBookingService.kt
```diff
@@ -168,6 +175,72 @@ class Cas1SpaceBookingService(
...
+    if (existingCas1SpaceBooking == null) {
+      "$.bookingId" hasValidationError "doesNotExist"
+    }
+
+    val departureReason = departureReasonRepository.findByIdOrNull(cas1NewDeparture.reasonId)
+    if (departureReason == null) {
+      "$.cas1NewDeparture.reasonId" hasValidationError "doesNotExist"
+    } else if (departureReason.serviceScope != "approved-premises") {
+      "$.cas1NewDeparture.reasonId" hasValidationError "incorrectDepartureReasonServiceScope"
+    }
```
davidatkinsuk, line 199: there is a possibility of having a service scope of '*' which would also match.

In BookingService there is a function `serviceScopeMatches`. Maybe we should make that a function on the Departure Reason Entity?

Also - we can probably simplify things by just returning 'doesNotExist' for both outcomes (logically the reason doesn't exist). This will make it easier to untangle from other CASs if we ever get there!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1SpaceBookingService.kt
```diff
@@ -168,6 +175,72 @@ class Cas1SpaceBookingService(
...
+
+    val moveOnCategory = moveOnCategoryRepository.findByIdOrNull(cas1NewDeparture.moveOnCategoryId)
+    if (moveOnCategory == null) {
+      "$.cas1NewDeparture.moveOnCategoryId" hasValidationError "doesNotExist"
+    } else if (moveOnCategory.serviceScope != "approved-premises") {
+      "$.cas1NewDeparture.moveOnCategoryId" hasValidationError "incorrectMoveOnCategoryServiceScope"
```
davidatkinsuk, line 205: See comment above for departureReason. We can fix in the same way (copy the serviceScopeMatches function onto the MoveOnCategoryEntity)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1SpaceBookingService.kt
```diff
@@ -168,6 +175,72 @@ class Cas1SpaceBookingService(
...
+    }
+
+    existingCas1SpaceBooking!!
+
+    if (existingCas1SpaceBooking.actualArrivalDateTime == null) {
+      return existingCas1SpaceBooking.id hasConflictError "An arrival is not recorded for this Space Booking."
+    }
+
+    if (existingCas1SpaceBooking.actualArrivalDateTime!!.isAfter(cas1NewDeparture.departureDateTime)) {
+      return existingCas1SpaceBooking.id hasConflictError "The departure date is before the arrival date."
```
davidatkinsuk, line 219: nice!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1SpaceBookingService.kt
```diff
@@ -168,6 +175,72 @@ class Cas1SpaceBookingService(
...
+    }
+
+    if (existingCas1SpaceBooking.actualDepartureDateTime != null) {
+      return existingCas1SpaceBooking.id hasConflictError "A departure is already recorded for this Space Booking."
+    }
+
+    val updatedCas1SpaceBooking = existingCas1SpaceBooking.copy(
```
davidatkinsuk, line 226: is there a reason to copy the entity here instead of updating the fields directly?

I suspect it's because actualDepartureDateTime and canonicalDepartureDate are vals. We can just change these to vars (my fault when creating the initial entity)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1SpaceBookingManagementDomainEventService.kt
```diff
@@ -36,33 +43,8 @@ class Cas1SpaceBookingManagementDomainEventService(
...
-            forenames = keyWorker.staff.forenames,
-            surname = keyWorker.staff.surname,
-            username = null,
-          )
-        }
-        is ClientResult.Failure -> keyWorkerStaffDetailsResult.throwException()
-      }
-    } ?: null
+    val offenderDetails = getOffenderForCrn(updatedCas1SpaceBooking.crn)
+    val keyWorker = getStaffMemberDetails(updatedCas1SpaceBooking.keyWorkerStaffCode)
```
davidatkinsuk, line 47: nice!

davidatkinsuk (approved):
Looks great, just a bunch of minor comments from me

PR #2311, opened 2024-09-22
Use postgres templating to reset test database

davidatkinsuk:
Before this commit the following approach was used to reset the database state between each test

1. At the start of the test session capture the state of all tables in-memory
2. At the start of each test TRUNCATE each table and restore it to the previously captured state

This approach worked but it was cumbersome and quite slow, taking around 1 second to reset the database to a baseline

This commit introduces a new approach

1. At the start of the test session create a 'template' database
2. At the start of each test DROP and recreate the API database from the 'template' database using postgres' template functionality e.g. CREATE DATABASE $databaseName TEMPLATE $templateDatabaseName

This approach is simpler to understand and much faster, taking around 250ms

It also presents an opportunity to further speed up testing in development environment by maintaining a template database across test sessions, avoiding the need to run all flyway scripts for each test session, which takes around 4 seconds. This will be added in a future commit.

One slightly awkward aspect of using postgres' TEMPLATE functionality is that the database being cloned must not have any connections to it when being cloned. For this reason we've introduce a new 'integration_test_monitor' user that devs can use to check on the state of the test database without inadvertentlty creating a connection to the template database, blocking any subsequent clones. This works because the integration_test_monitor user doesn't have CONNECT permissions for the template database.


RobBoothMOJ (approved):
Approved

PR #2313, opened 2024-09-23
APS-1314 Add CRU Management Area to user

davidatkinsuk:
This PR adds two new fields to the users table:

* cas1_cru_management_area_id
* cas1_cru_management_area_override_id

The former will always be updated when ap area is updated, using the ap areas default management area value.

This commit also backfills the cas1_cru_management_area_id value for existing users.

The latter will be configurable via an API call to be added in a subsquent commit.

### DIFF
--- a/src/main/resources/db/migration/all/20240923113600__add_cru_management_fields_to_user.sql
+++ b/src/main/resources/db/migration/all/20240923113600__add_cru_management_fields_to_user.sql
@@ -1,0 +1,5 @@
+ALTER TABLE users ADD cas1_cru_management_area_id uuid NULL;
+ALTER TABLE users ADD CONSTRAINT users_cas1_cru_management_area_id_fk FOREIGN KEY (cas1_cru_management_area_id) REFERENCES cas1_cru_management_areas(id);
+
+ALTER TABLE users ADD cas1_cru_management_area_override_id uuid NULL;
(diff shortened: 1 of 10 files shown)


src/main/resources/db/migration/all/20240923113600__add_cru_management_fields_to_user.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE users ADD cas1_cru_management_area_id uuid NULL;
+ALTER TABLE users ADD CONSTRAINT users_cas1_cru_management_area_id_fk FOREIGN KEY (cas1_cru_management_area_id) REFERENCES cas1_cru_management_areas(id);
+
+ALTER TABLE users ADD cas1_cru_management_area_override_id uuid NULL;
```
danhumphreys-moj, line 4: What is this `cas1_cru_management_area_override_id` column?  Isn't there a 1:1 relationship between users and management areas?

danhumphreys-moj (approved):

src/main/resources/db/migration/all/20240923113600__add_cru_management_fields_to_user.sql
davidatkinsuk: We determine the value of cas1_cru_management_area_id automatically when the user logs in based upon their AP Area. We have a requirement for some users to be put in to the 'women's estate' management area which will never be automatically mapped from an AP Area. So to manage this there will be on an option on the user management page to override the automatically assigned management area, allowing users to be assigned to the 'women's estate'. There will be a subsequent commit that adds the logic for this

PR #2314, opened 2024-09-23
APS-1346 Filter by CRU management area id.

danhumphreys-moj:
Ref: https://dsdmoj.atlassian.net/browse/APS-1346

We have several endpoints that allow filtering of results on AP Area ID. These need updating to support filtering on CRU Management Area ID instead. The option to filter on AP Area ID should be deprecated in favour of CRU Management Area

This PR updates `PlacementRequestsController.placementRequestsDashboardGet` (tested by `PlacementRequestsTest.Dashboard`). In the SQL this will work by filtering on the CRU Management Area linked to the application i.e. (`approved_premises_applications.cas1_cru_management_area_id`).

This is a back end change only and does not require any change to the front end (for now).  Therefore although we are passing `apAreaId` to the API, we are doing a SQL search joining on `cruManagementAreaId`.

Note that for AP Area we perform a join - this has been removed.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PlacementRequestsController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PlacementRequestsController.kt
@@ -80,6 +80,7 @@ class PlacementRequestsController(
     arrivalDateEnd: LocalDate?,
     requestType: PlacementRequestRequestType?,
     apAreaId: UUID?,
+    cruManagementAreaId: UUID?,
     page: Int?,
     sortBy: PlacementRequestSortField?,
     sortDirection: SortDirection?,
@@ -100,6 +101,7 @@ class PlacementRequestsController(
         arrivalDateEnd = arrivalDateEnd,
         requestType = requestType,
         apAreaId = apAreaId,
+        cruManagementAreaId = cruManagementAreaId,
       ),
       PageCriteria(
         sortBy = sortBy ?: PlacementRequestSortField.createdAt,
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PlacementRequestEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PlacementRequestEntity.kt
@@ -114,6 +114,7 @@ interface PlacementRequestRepository : JpaRepository<PlacementRequestEntity, UUI
         )
       )
       AND ((CAST(:apAreaId AS pg_catalog.uuid) IS NULL) OR area.id = :apAreaId)
+      AND ((CAST(:cruManagementAreaId AS pg_catalog.uuid) IS NULL) OR apa.cas1_cru_management_area_id = :cruManagementAreaId)
   """,
     nativeQuery = true,
   )
@@ -126,6 +127,7 @@ interface PlacementRequestRepository : JpaRepository<PlacementRequestEntity, UUI
     arrivalDateTo: LocalDate? = null,
     requestType: String? = null,
     apAreaId: UUID? = null,
+    cruManagementAreaId: UUID? = null,
     pageable: Pageable? = null,
   ): Page<PlacementRequestEntity>
 
--- a/src/main/resources/static/codegen/built-api-spec.yml
+++ b/src/main/resources/static/codegen/built-api-spec.yml
@@ -3192,7 +3192,15 @@ paths:
         - name: apAreaId
           in: query
           required: false
-          description: filter by approved premises area ID
+          description: filter by approved premises area ID.  Use cruManagementAreaId instead.
+          deprecated: true
+          schema:
+            type: string
+            format: uuid
+        - name: cruManagementAreaId
+          in: query
+          required: false
+          description: filter by CRU management area ID
           schema:
             type: string
             format: uuid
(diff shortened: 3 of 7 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PlacementRequestsController.kt
```diff
@@ -79,7 +79,7 @@ class PlacementRequestsController(
     arrivalDateStart: LocalDate?,
     arrivalDateEnd: LocalDate?,
     requestType: PlacementRequestRequestType?,
-    apAreaId: UUID?,
+    cruManagementAreaId: UUID?,
```
danhumphreys-moj, line 82: @davidatkinsuk looking at this I think this is wrong - from memory the front end isn't changing so no need to change `apAreaId` to `cruManagementAreaId`.  Will wait for you to confirm before reverting this.

davidatkinsuk: Correct - we need to retain apAreaId until the front end has moved over to use cruManagementArea instead

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/PlacementRequestsTest.kt
```diff
@@ -1036,6 +1113,9 @@ class PlacementRequestsTest : IntegrationTestBase() {
             ).produce(),
         )
         withApArea(apArea)
+        apArea?.defaultCruManagementArea?.let {
+          withCruManagementArea(it)
+        }
```
davidatkinsuk, line 1118: can we please add `cruManagementArea` as a first class parameter in the `createPlacementRequest() `function and use that here instead? You can create one in the calling tests using `Given an CAS1 CRU Management Area`()

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/PlacementRequestsTest.kt
```diff
@@ -741,6 +775,49 @@ class PlacementRequestsTest : IntegrationTestBase() {
...
     }
 
+    @Test
+    fun `It searches using multiple criteria including CRU management area id where user is manager`() {
+      `Given a User`(roles = listOf(UserRole.CAS1_WORKFLOW_MANAGER)) { user, jwt ->
+        `Given an Offender` { offender1Details, inmate1Details ->
+          `Given an Offender` { offender2Details, _ ->
+
+            val apArea1 = `Given an AP Area`()
+            val apArea2 = `Given an AP Area`()
```
davidatkinsuk, line 785: i think for tests related to filtering on the CRU Management Area we should build the CRU Management Area here (instead of ap area) for clarity. You can use `Given an CAS1 CRU Management Area`()` to do this. This applies to all updated tests

src/main/resources/static/codegen/built-api-spec.yml
```diff
@@ -3202,6 +3202,13 @@ paths:
           schema:
             type: string
             format: uuid
+        - name: cruManagementAreaId
```
davidatkinsuk, line 3205: could you please add

`deprecated: true` to the apAreaId parameter above, and add a note in the description saying cruManagementAreaId should be used instead?

davidatkinsuk (approved):
LGTM. Just a few minor comments from me 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PlacementRequestEntity.kt
```diff
@@ -114,6 +114,7 @@ interface PlacementRequestRepository : JpaRepository<PlacementRequestEntity, UUI
         )
       )
       AND ((CAST(:apAreaId AS pg_catalog.uuid) IS NULL) OR area.id = :apAreaId)
+      AND ((CAST(:cruManagementAreaId AS pg_catalog.uuid) IS NULL) OR apa.cas1_cru_management_area_id = :cruManagementAreaId)
```
davidatkinsuk, line 117: it just occurred to me that we should add an index on approved_premises_applications cas1_cru_management_area_id. Can maybe do that in the next PR (which will be filtering on the same value)

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/PlacementRequestsTest.kt
danhumphreys-moj: Added `cruManagementArea` to `createPlacementRequest()`.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/PlacementRequestsTest.kt
danhumphreys-moj: Updated 2 x tests to do this now.

src/main/resources/static/codegen/built-api-spec.yml
danhumphreys-moj: Done this.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PlacementRequestEntity.kt
danhumphreys-moj: Haven't done this in this PR - wanted to get this one over the line.

PR #2315, opened 2024-09-24
APS-1314 - Return User's CRU Management Area details in the API

davidatkinsuk:
This commit adds CRU Management Area fields to the ApprovedPremisesUser API Model to support:

* Determining which CRU Management Area the user is in (when retrieved via /profile/v2)
* Logic to determine/set/get override configuration for the user management page

This commit is quite large because i had fix existing test data where entries in the users table were created for CAS1 testing and the CRU Management Area was not set, something which wouldn’t be the case in production (it’s logically mandatory for CAS1 users)


RobBoothMOJ (approved):
Approved

PR #2316, opened 2024-09-24
APS-1342: Surface exceptions when obtaining risk information

RobBoothMOJ:
Errors occurring when obtaining Risk information weren't being logged. 
Added sentry logging to record when failures do occur.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/SentryService.kt
```diff
@@ -21,6 +21,9 @@ class SentryServiceImpl : SentryService {
     Sentry.captureException(throwable)
   }
 
+  /**
+   * First line of error messages MUST be identical to ensure correct grouping in alerts.
+   */
```
davidatkinsuk, line 26: nice

davidatkinsuk (approved):
LGTM 👍

PR #2318, opened 2024-09-24
APS-1352 - Remove female assessment allocation rule and associated qualification

davidatkinsuk:
**Remove legacy FemaleAssessmentRule** 

This rule was used to automatically assign assessments for ‘womens applications’ to an assessor with the ‘WOMENS’ qualification. This is going to be replaced by the new women’s estate logic. Furthermore, there is no user in any environment with the ‘WOMENS’ qualification, and there is no option to add this qualification on the UI. Therefore, this is safe to remove.

As part of this change some integration tests were updated that relied on this assessment rule to instead use the emergency and short notice assessment rule that _is_ used in production

**Remove ‘WOMENS’ qualification**

This is going to replaced by the new women’s estate functionality.

This qualification is not in use by any users in production, and can’t be assigned via the UI (checked using `select qualification from user_qualification_assignments group by qualification;`)


RobBoothMOJ (approved):
Approved

PR #2319, opened 2024-09-24
Disable pre-emptive cache refresh during tests

davidatkinsuk:
This isn’t required for testing and creates quite a bit of noise when the database is being reset at the same point in the time the refresh job runs.


RobBoothMOJ (approved):
Approved

PR #2320, opened 2024-09-24
Fix ApStaffUsersSeedJobTest class name

davidatkinsuk:
The Class Name didn’t match the filename which meant the tests weren’t being ran in circle.

The build should have been broken as the tests weren’t using the correct error message. This has now been fixed.


RobBoothMOJ (approved):
Approved

PR #2322, opened 2024-09-24
Remove CAS1 LegacyAllocatorRule

davidatkinsuk:
The LegacyAllocatorRule assigned Assessments, Placement Applications and Placement Requests to assessor automatically whenever they were submitted. Its usage has been disabled in all environments for months other than test and as such should be removed.

Removing this rule has broken several integration tests that expected assessments, placement applications and placements requests to be automatically allocted on submission. This commit updates those tests to remove that expectation, better testing the actual functionality live in production.

This commit also depreactes the auto allocation functions for placement requests and placement applications as they will always return null.


RobBoothMOJ (approved):
Approved

PR #2324, opened 2024-09-24
CAS-637 Remove CRN from booking Api endpoint

muhammad-elabdulla:
This [PR CAS-637](https://dsdmoj.atlassian.net/browse/CAS-637) is to remove the `crn` parameter from booking api endpoint as it was replaced by `crnOrName` parameter


davidatkinsuk (approved):
LGTM 👍

PR #2325, opened 2024-09-25
Rollback change to only return conviction documents

RobBoothMOJ:
Removed feature flag CAS1-ONLY-LIST-CONVICTION-DOCUMENTS Agreed on call 24/9/24
Removed filtering of documents which was only returning conviction documents if the feature flag was set.


davidatkinsuk (approved):
LGTM 👍

PR #2328, opened 2024-09-25
APS-1314 Add CAS1 Users Endpoints

davidatkinsuk:
This PR copies existing functionality from the shared /users endpoint to /cas1/users. Specifically:

* get user by id
* delete user - only used by CAS1
* update user - only used by CAS1

This is in preparation for changes to the update user endpoint. It allows us to simplify the implementation logic, and also explicitly state the ApprovedPremiseUser type in the response

A ticket has been created to remove delete user/update user from the shared endpoint once the ui is using the new endpoints - APS-1363


RobBoothMOJ (approved):
Approved

PR #2330, opened 2024-09-26
APS-1243: Implement M&M V2 Record Key Worker API.

RobBoothMOJ:
Implemented the M&M V2 API to record Key Worker for a Space Booking 
Added new User Permission: CAS1_SPACE_BOOKING_RECORD_KEYWORKER 
Endpoint is only accessible to users with CAS1_SPACE_BOOKING_RECORD_KEYWORKER permission. 
Updates existing space booking with information for the provided key worker


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1SpaceBookingService.kt
```diff
@@ -174,6 +180,43 @@ class Cas1SpaceBookingService(
...
+    if (premises == null) {
+      "$.premisesId" hasValidationError "doesNotExist"
+    }
+
+    val existingCas1SpaceBooking = cas1SpaceBookingRepository.findByIdOrNull(bookingId)
+    if (existingCas1SpaceBooking == null) {
+      "$.bookingId" hasValidationError "doesNotExist"
+    }
+
+    val staffKeyWorker = userRepository.findByDeliusStaffCode(keyWorker.staffCode)
```
davidatkinsuk, line 199: apologies ... this is my fault for not specifying this on the ticket.

we retrieve key worker details from a service. Take a look at the logic in BookingService.createCas1Arrival

```
val staffMemberResponse = staffMemberService.getStaffMemberByCode(keyWorkerStaffCode, premises.qCode)

if (staffMemberResponse !is AuthorisableActionResult.Success) {
   return "$.keyWorkerStaffId" hasSingleValidationError "notFound"
}
```

It should be a matter of dropping that in here and updating the integration test i image. looking at tests for the aforementioned function we can do this using

```
val keyWorker = ContextStaffMemberFactory().produce()
APDeliusContext_mockSuccessfulStaffMembersCall(keyWorker, "QCODE")
```

where qcode must match the premise's qcode e.g.

```
        val premises = approvedPremisesEntityFactory.produceAndPersist {
          withQCode("QCODE")
          withYieldedLocalAuthorityArea { localAuthorityEntityFactory.produceAndPersist() }
          withYieldedProbationRegion { probationRegion }
        }
```

davidatkinsuk (changes requested):
Looks great but we need to tweak how we get key worker info (my fault for not specifying)

davidatkinsuk (approved):
Looks great.

PR #2331, opened 2024-09-26
APS-1311 - Fix premises seed

davidatkinsuk:
This commit fixes the approved premises seed CSV file, adding a missing column name.


RobBoothMOJ (approved):
Approved

PR #2334, opened 2024-09-26
CAS-636 Removed use of community api in UserService

muhammad-elabdulla:
This PR removes the feature flag for using the ap-and-delius integration - it was set to true for all environments. It also removes the use of the community-api to get staff user details in the UserService, instead getting them from the AP and Delius integration.


davidatkinsuk (approved):
LGTM 👍

PR #2335, opened 2024-09-26
Add .env file support to boot run

davidatkinsuk:
This commit adds .env file support to the gradle bootRun task. This will help us simplify configuration of upstream URLs and credentials in ap-tools by allowing the same env file to be used by gradle or docker based API resources


RobBoothMOJ (approved):
Approved

PR #2336, opened 2024-09-27
APS-1365 : Modify Space Booking residency filters

RobBoothMOJ:
Changed functionality for filters when retrieving space bookings : 

- Added 'historic' filter -  only departed bookings are returned (presence of actual departure date). 
- Removed exclusion of departed bookings, previously these were not included in the resulting data set.


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/Cas1SpaceBookingTest.kt
```diff
@@ -459,7 +459,7 @@ class Cas1SpaceBookingTest {
     }
 
     @Test
-    fun `Search with no filters excludes departed bookings`() {
+    fun `Search with no filters excludes no cancelled bookings`() {
```
davidatkinsuk, line 462: i think we have a double negative here, should it be `Search with no filters excludes cancelled bookings`

davidatkinsuk (approved):
Just one really minor comment from me, looks good! 👍

PR #2337, opened 2024-09-27
APS-1367: Modify Space Booking Search to support filtering on key worker.

RobBoothMOJ:
Added new keyWorkerStaffCode filter option to the API
Employed the additional filter in the Entity's  SQL


davidatkinsuk (approved):
LGTM 👍

PR #2339, opened 2024-09-27
Fix bed count for cas1 premise summary

davidatkinsuk:
Before this commit the code populating the ‘outOfServiceBedsCount’ when viewing a premise considered any out of service bed record linked to the premise regardless of the start and end date as ‘current’.


bobmeredith (approved):
LGTM

PR #2341, opened 2024-09-30
APS-1350 add cru filter to users list

danhumphreys-moj:
We need to enhance GET `/users` as follows:

- add an optional filter on `users.cas1_cru_management_area_id`. This will be similar to the existing `ap_area_id` which should be deprecate in favour of this filter
- ~~add an additional field in the response providing the id and name of the user’s `Cas1CruManagementAreaEntity`. This should already be available ‘for free’ as it has already been added to the ApprovedPremisesUser API type.~~ This has already been done.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa.specification/UserSpecifications.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa.specification/UserSpecifications.kt
@@ -6,6 +6,7 @@ import jakarta.persistence.criteria.Predicate
 import jakarta.persistence.criteria.Root
 import org.springframework.data.jpa.domain.Specification
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ApArea
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Cas1CruManagementArea
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ProbationRegionEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserQualification
@@ -14,11 +15,13 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserRole
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserRoleAssignmentEntity
 import java.util.UUID
 
+@SuppressWarnings("LongParameterList")
 fun hasQualificationsAndRoles(
   qualifications: List<UserQualification>?,
   roles: List<UserRole>?,
   region: UUID?,
   apArea: UUID?,
+  cruManagementArea: UUID?,
   showOnlyActive: Boolean = false,
 ): Specification<UserEntity> {
   return Specification { root: Root<UserEntity>, query: CriteriaQuery<*>, criteriaBuilder: CriteriaBuilder ->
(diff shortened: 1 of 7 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/UsersTest.kt
```diff
@@ -515,6 +515,57 @@ class UsersTest : InitialiseDatabasePerClassTestBase() {
...
+              }
+
+              val userTwo = userEntityFactory.produceAndPersist {
+                withProbationRegion(probationRegion)
+                withApArea(apArea)
+                withCruManagementArea(apArea.defaultCruManagementArea)
+              }
+
+              webTestClient.get()
+                .uri("/users?cruManagementAreaId=${apArea.defaultCruManagementArea.id}")
```
danhumphreys-moj, line 545: Does it make sense to use the `defaultCruManagementArea` here?

davidatkinsuk: I think it may be cleaner if you create some CRU Management Areas yourself and assign these directly to the user's cruManagementArea property, ignoring ap areas completely e.g. 

```
val cruManagementArea1 = `Given a CAS1 CRU Management Area`()
val cruManagementArea2 = `Given a CAS1 CRU Management Area`()

...

userEntityFactory.produceAndPersist {
   ...
   withCruManagementArea(cruManagementArea1)
}
```

(rough code ... i think it will compile!)

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/UsersTest.kt
```diff
@@ -515,6 +515,57 @@ class UsersTest : InitialiseDatabasePerClassTestBase() {
...
+                withProbationRegion(probationRegion)
+                withApArea(apArea)
+                withCruManagementArea(apArea.defaultCruManagementArea)
+              }
+
+              val userTwo = userEntityFactory.produceAndPersist {
+                withProbationRegion(probationRegion)
+                withApArea(apArea)
+                withCruManagementArea(apArea.defaultCruManagementArea)
+              }
```
davidatkinsuk, line 542: can we please have another user not in the cru management area we're filtering on?

davidatkinsuk (approved):
As always looks good, and just minor comments!

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/UsersTest.kt
danhumphreys-moj: No problem, done this.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa.specification/UserSpecifications.kt
```diff
@@ -76,6 +79,16 @@ fun hasQualificationsAndRoles(
       )
     }
 
+    if (cruManagementArea != null) {
+      val cruManagementAreaid = root.get<Cas1CruManagementArea>("cruManagementArea").get<UUID>("id")
```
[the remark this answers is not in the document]
danhumphreys-moj: Done.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
```diff
@@ -131,6 +131,7 @@ class UserService(
     page: Int?,
     region: UUID?,
     apArea: UUID?,
+    cruManagementAreaId: UUID?,
   ): Pair<List<UserEntity>, PaginationMetadata?> {
```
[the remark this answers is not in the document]
danhumphreys-moj: Renamed this function to getUsers().

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/UsersTest.kt
danhumphreys-moj: Added another user in the same apArea but different CRU management area.  Test still passes as expected.

PR #2342, opened 2024-09-30
Add simple scheduled task with distributed lock

davidatkinsuk:
This commits adds a spring-based scheduled task that makes use of a redis-based distributed lock to ensure that task only runs on one instance of the application.

This is being added as a basic check that the distributed lock acts as expected in our deployment environments.


danhumphreys-moj (approved):
Great stuff!

PR #2345, opened 2024-10-01
CAS-451 Add turnaround days to the script

muhammad-elabdulla:
This [PR CAS-451](https://dsdmoj.atlassian.net/browse/CAS-541) is to:
- Include provisional bookings
- Exclude cancelled bookings 
- Add a new column the turnaround days for bedspace booking to the report. The query is a proof of concept so we are not looking to the performance in the current stage
- Remove the gap between the departure date of the last booking and the current date


davidatkinsuk:
closing this because the change appears to have been made via a different PR (I can see turnaround days in the report on main)

PR #2346, opened 2024-10-01
APS-1345 - Use CruManagementArea to auto allocate assessments

davidatkinsuk:
Before this commit we used the AP Area name and corresponding configuration to determine who (if anyone) to automatically allocate assessments to for short notice or emergency applications.

This commit instead uses cas1_cru_management_area.assessment_auto_allocation_username.

This allows us to remove alot of configuration and simplify the allocation code


danhumphreys-moj (approved):
LGTM

PR #2347, opened 2024-10-01
APS-1234 - Seed Women’s Estate Test User in local environment

davidatkinsuk:
This commit adds default roles to the CRUWOMENSESTATE user for local deployments. It also overrides their CRU Management Area to be the women’s estate management area, and sets them as the auto allocation user for assessments


bobmeredith (approved):
LGTM

PR #2354, opened 2024-10-02
APS-1389 ensure workflow manager can view CRU Dashboard

davidatkinsuk:
Give WORKFLOW_MANAGER most CRU_MEMBER permissions

The only permissions not assigned to the WORKFLOW_MANAGER from CRU_MEMBER are related to out of service beds.

This is temporary solution to support migration from the WORKFLOW_MANAGER role to CRU_MEMBER or other roles, as not all CRU users in production have the CRU_MEMBER role and nor should they currently as we don’t want all users to have access the the national out of service beds functionality


davidatkinsuk (commented):
Just one comment that i think you may have missed, other than that looks great. would mind holding fire on merging it as we have something we need to push out to prod ASAP related to the UI rollback we had to do earlier

danhumphreys-moj (approved):

PR #2355, opened 2024-10-02
APS-1390 Simplify auto allocation config

davidatkinsuk:
Whilst each allocation rule could be enabled and have its prioirty set via configuration, in all cases the exact same configuration was used. This commit removes that configuration


danhumphreys-moj (approved):
LGTM

PR #2359, opened 2024-10-03
APS-1388 add scheduled job and domain event to update statuses for expired applications

danhumphreys-moj:
Ref: https://dsdmoj.atlassian.net/browse/APS-1388 (add scheduled job)
Ref: https://dsdmoj.atlassian.net/browse/APS-1387 (add domain event)

Creates a job to be run daily that selects any application where the assessment decision is over 12 months old; and for each selected application:

- Updates the application status to ‘EXPIRED'
- Triggers an expired application domain event


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/Cas1ExpiredApplicationsScheduledJobTest.kt
```diff
@@ -0,0 +1,81 @@
...
+              .withDecision("ACCEPTED")
+              .produce(),
+          ),
+        ),
+      )
+    }
+
+    // create a domain event for this application
+
+    Cas1ExpiredApplicationsScheduledJob(
```
davidatkinsuk, line 69: you should be able to inject this in using @Autowired as it's a spring manage bean. no need to construct yourself

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApplicationEntity.kt
```diff
@@ -291,6 +291,22 @@ WHERE taa.probation_region_id = :probationRegionId AND a.submitted_at IS NOT NUL
...
+    """
+      SELECT application_id
+      FROM domain_events de
+      JOIN approved_premises_applications apa
+        ON de.application_id = apa.id
+      WHERE
+        de.type = 'APPROVED_PREMISES_APPLICATION_ASSESSED'
+        AND de.data -> 'eventDetails' ->> 'decision' = 'ACCEPTED'
+        AND apa.status <> 'EXPIRED'
+        AND de.occurred_at < current_date - 365
```
davidatkinsuk, line 305: on first glance this looks correct. i do wonder if driving it from the applications table joining onto the domain events table would be quicker

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/Cas1ExpiredApplicationsScheduledJobTest.kt
danhumphreys-moj: Have changed this.

RobBoothMOJ (approved):
Approved

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/scheduled/Cas1ExpiredApplicationsScheduledJob.kt
```diff
@@ -17,7 +33,49 @@ class Cas1ExpiredApplicationsScheduledJob(
   @SchedulerLock(name = "cas1_expire_applications", lockAtMostFor = "5m", lockAtLeastFor = "1m")
   fun expireApplications() {
     if (featureFlagService.getBooleanFlag("cas1-expired-applications-enabled")) {
-      log.info("Checking for expired applications (Dry Run)")
+      val applicationIds = applicationRepository.findAllExpiredApplications()
```
gregkhawkins, line 36: qq - is there any danger of this returning a lot of records the first time this scheduled job is kicked off? 
Just wondering if we might retrieve too many records to expire for the list to store in memory and it then fall over? 
If we are running this job regularly that problem would probably go away after the first run tbf it's just a case of ensuring it does not fall over when it kicks off against prod.
Forgive me if you have already thought of this and is a non-issue as small dataset. If unsure it might be worth running queries to get us counts against pre-prod / prod DBs etc
Thanks!

gregkhawkins (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/scheduled/Cas1ExpiredApplicationsScheduledJob.kt
danhumphreys-moj: Running this query tells us how many records we're currently expecting to update in pre-prod, and as it stands today it will be 1338 applications:

```
SELECT distinct(application_id)
FROM domain_events de
         JOIN approved_premises_applications apa
              ON de.application_id = apa.id
WHERE
    de.type = 'APPROVED_PREMISES_APPLICATION_ASSESSED'
  AND de.data -> 'eventDetails' ->> 'decision' = 'ACCEPTED'
  AND apa.status <> 'EXPIRED'
  AND de.occurred_at < current_date - 365
```

Your comment was useful since it helped me realise we needed a distinct in the query, since it seems we have one or more applications which have multiple matching records in domain_events table.  So if we didn't include the distinct what would have happened is that we would have change the status of an application to EXPIRED, then have changed the status of the same application from EXPIRED to EXPIRED again (probably not a problem), but we would have generated multiple APPLICATION_EXPIRED domain events for the same application.

Changing the query as per @davidatkinsuk suggestion above to join `domain_events` with `approved_premises_applications` would negate this problem since application_id will be distinct in `approved_premises_applications` since it's a primary key, but the query was agreed before he went on holiday so am keen not to change it.

I've added the distinct as a separate [commit](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/2359/commits/b183cbe6ff0b1a8ee3bbbfd6cfe02ff8d56725a1).  And in the process of finalising some tests we would want to do as a [comment in the ticket](https://dsdmoj.atlassian.net/browse/APS-1387?focusedCommentId=518377).

PR #2360, opened 2024-10-03
Refactored test as part of CAS-573

daveawc:
Changed the one long boolean assertion into assertions on individual fields


davidatkinsuk (approved):
LGTM 👍

PR #2362, opened 2024-10-04
Added new functions for CAS-573

daveawc:
Added new methods preparing for removing StaffUserDetails. toStaffMember() is duplicated, so this adds it into the StaffDetail class, so the others can be removed in the following PR

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/deliuscontext/StaffDetail.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/deliuscontext/StaffDetail.kt
@@ -1,5 +1,7 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.model.deliuscontext
 
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.StaffMember
+
 data class StaffDetail(
   val email: String?,
   val telephoneNumber: String?,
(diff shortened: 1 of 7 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/deliuscontext/StaffDetail.kt
```diff
@@ -16,14 +18,23 @@ data class StaffDetail(
...
 }
 
 data class PersonName(
   val forename: String,
   val surname: String,
-  val middleName: String?,
+  val middleName: String? = null,
 ) {
   fun deliusName() = "$forename $surname"
+  fun forenames() = "$forename ${middleName?.takeIf { it.isNotEmpty() } ?: ""}".trim()
```
muhammad-elabdulla, line 37: Do we need this function? I think we use full name which also include the surname

muhammad-elabdulla (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/deliuscontext/StaffDetail.kt
muhammad-elabdulla: I mean if we have forename, middleName and surname. I think the only method that is need is the full name with include all three properties. 

I am not sure if the forenames in the StaffMember includes also the middleName

PR #2366, opened 2024-10-09
CAS-1171 - Make close assessment idempotent

daveawc:
This change makes closing an already closed asseessment return OK and log out a message, instead of throwing an error.

This will prevent unnecessary alerts in the cas-events channel.


froddd (approved):
LGTM

muhammad-elabdulla (approved):

PR #2368, opened 2024-10-14
CAS-1173 remove staff user details model

daveawc:
The PR removes the usage of StaffUserDetails Factory from the unit tests, and replaced it with StaffDetailFactory


danhumphreys-moj (approved):
LGTM good work Dave!

PR #2369, opened 2024-10-15
CAS-1173 - Remove StaffUserDetails from integration tests

daveawc:
This PR removes StaffUserDetails from the integration tests, as well as removing the factory. The StaffUserDetails class will need to be removed in a later PR, as the mocked calls to get staff by code still use it. It also updates the delius name to use all three names, that was inadvertently removed in a previous commit.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/deliuscontext/CaseDetail.kt
```diff
@@ -40,7 +40,14 @@ data class Team(
   val borough: Borough?,
   val startDate: LocalDate?,
   val endDate: LocalDate?,
-)
+) {
+  fun toTeam(): uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.model.Team {
```
davidatkinsuk, line 44: could this maybe provide more context in the function name. Maybe toDomainEventTeam() (assuming that's what it is). I would say rename the target type but that'll probably just break something else as it's generated from open api :)

davidatkinsuk (approved):
LGTM 👍

PR #2374, opened 2024-10-18
APS-1386 ensure that if a placement request is submitted with expired application, it returns an error

danhumphreys-moj:
Ref: https://dsdmoj.atlassian.net/browse/APS-1386

**Problem Description**
Whilst we will hide any way to create a placement request on the front end, we should ensure that placement requests cannot be made, and will return and error if they are made.

**Desired Outcome**

- Add logic that will return an error where a placement request is made with an application that has the status ‘Expired Application’
- Return the error: ‘Placement requests cannot be made with an expired application’
- Capture the event in sentry,


RobBoothMOJ (approved):
Approved

PR #2375, opened 2024-10-18
APS-1394 Throw error if user does not have correct permission

danhumphreys-moj:
Ref: https://dsdmoj.atlassian.net/browse/APS-1394

Changed access from users with role CAS1_WORKFLOW_MANAGER to those who have CAS1_VIEW_CRU_DASHBOARD permission.


davidatkinsuk (approved):
LGTM 👍

PR #2376, opened 2024-10-21
APS-1418: Update Cas1 Premises Summary endpoint

RobBoothMOJ:
To ensure consistency with the existing, non-namespaced endpoint the following have been added to the new, namespaced endpoint: 

- filtering by area. Added optional `apAreaId`  query parameter
- bed count is returned in the summary data.  This is a 'live' bed count for the given premises, how many beds the premises has live at the time,  and not an 'available' bed count which would incorporate booking data ( i.e. live beds - booked beds).

There is only one consumer of the new summary query.   The data returned by the query has been changed to retrieve only that needed rather than complete premises records.  This was primarily due to the need for bed count, which is calculated when the query is executed, rather than being data persisted in the records themselves.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PremisesEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PremisesEntity.kt
@@ -174,8 +174,29 @@ where
 
 @Repository
 interface ApprovedPremisesRepository : JpaRepository<ApprovedPremisesEntity, UUID> {
-  @Query("SELECT p as premises FROM ApprovedPremisesEntity p WHERE :gender IS NULL OR p.gender = :gender")
-  fun findForSummaries(gender: ApprovedPremisesGender?): List<ApprovedPremisesEntity>
+  @Query(
+    """
+        SELECT
+          new uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApprovedPremisesBasicSummary(
+              p.id,
+              p.name,
+              apArea.id,
+              apArea.name,
+              CAST(COUNT(b) as int)
+              )
+        FROM 
+          ApprovedPremisesEntity p
+          LEFT JOIN p.rooms r 
+          LEFT JOIN r.beds b on (b.endDate IS NULL OR b.endDate > CURRENT_DATE) 
+          LEFT JOIN p.probationRegion region
+          LEFT JOIN region.apArea apArea
+        WHERE 
+          (:gender IS NULL OR p.gender = :gender)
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PremisesEntity.kt
```diff
@@ -174,8 +174,29 @@ where
...
+          ApprovedPremisesEntity p
+          LEFT JOIN p.rooms r 
+          LEFT JOIN r.beds b on (b.endDate IS NULL OR b.endDate > CURRENT_DATE) 
+          LEFT JOIN p.probationRegion region
+          LEFT JOIN region.apArea apArea
+        WHERE 
+          (:gender IS NULL OR p.gender = :gender)
+          AND(cast(:apAreaId as text) IS NULL OR apArea.id = :apAreaId) 
+          GROUP BY p.id, p.name, apArea.id, apArea.name 
+      """,
```
gregkhawkins, line 197: A couple of things on this query we might want to consider (if have not already!):
1. do we need to add an index onto `beds.endDate` column as that drives the counts and might be slow?
2. do we need to add an an index to the `approved_premises.gender` column also to speed up this bit in the where clause `p.gender = :gender`?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas1/Cas1PremisesTransformer.kt
```diff
@@ -26,12 +26,12 @@ class Cas1PremisesTransformer(
...
   }
 
-  fun toPremiseBasicSummary(entity: PremisesEntity): Cas1PremisesBasicSummary {
-    val apArea = entity.probationRegion.apArea!!
+  fun toPremiseBasicSummary(entity: ApprovedPremisesBasicSummary): Cas1PremisesBasicSummary {
     return Cas1PremisesBasicSummary(
       id = entity.id,
       name = entity.name,
-      apArea = NamedId(apArea.id, apArea.name),
+      apArea = NamedId(entity.apAreaId!!, entity.apAreaName),
```
gregkhawkins, line 33: we are sure `entity.apAreaId` is never null I assume? Double bang (`!!`) always scares me :)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PremisesEntity.kt
RobBoothMOJ: Although this query is basically a cut-down copy of another which already exists, I'm going to take an action to look into the performance of the query in the two areas you highlight.

gregkhawkins (approved):

PR #2382, opened 2024-10-22
APS-1418: Update Cas1 Premises Summary endpoint

RobBoothMOJ:
Added apCode to the response so the UI has all of the summary information it needs.


danhumphreys-moj (approved):
LGTM

PR #2384, opened 2024-10-23
CAS-1179 Refactor GET document by ID endpoint to use AP and Delius API

gregkhawkins:
**PR includes:**
Refactor the `GET /documents/{crn}/{documentId}` endpoint to download the specific document using the GET documents endpoints in `AP and Delius API` (instead of using the `Community API` GET documents endpoints)

**Extras**
- This PR introduces a new feature flag `FEATURE-FLAGS_GET-DOCUMENTS-FROM-AP-DELIUS` which is switched off on all environments currently (via the helm properties) and also switched off in the spring boot yml files (for AP tools and integration tests).
- This new feature flag will also be used in follow-up ticket `CAS-1178` where we refactor the `GET /applications/{applicationId}/documents` call to also use `AP and Delius API` for doc retrieval. At this stage, we can switch the new flag on which will have the affect that all documents calls in the API goes to Delius.
- This will enable us to switch on at the `AP tools` level first to make sure all docs are coming from the `Ap and Delius API` service deployed locally. As part of `CAS-1178` we will need to add some docs into the H2 DB that backs the `Ap and Delius API` (have a plan for this - see this thread: https://mojdt.slack.com/archives/C03KVQUFSER/p1729608893989549)
- Once we have proved the above, we can switch on the new feature flag for the dev and test envs - so that we can validate it on those envs first (before pre-prod and prod). 
- `CAS-1178` and `CAS-1179` will need to be shipped together as the docId used by the FE to call the `GET /documents/{crn}/{documentId}` endpoint is fed the IDs it retrieves by first calling the `GET /applications/{applicationId}/documents` endpoint. These Ids need to come from the same underlying system (i.e. either from `Community API` or by `AP and Delius API`) for this to all work seamlessly.

**Note**
- I wanted to keep the PRs as small as possible which is why I plan to separate out the work for `CAS-1178` and `CAS-1179` into different feature branches and why this only contains the work for `CAS-1179`.
- If it is better to do on the work on single branch that is fine too. - let me know preference :)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/DocumentsController.kt
```diff
@@ -31,24 +35,51 @@ class DocumentsController(
...
     )
   }
+
+  private fun getDocuments(retrieveDocumentsFromApDelius: Boolean, crn: String) = if (retrieveDocumentsFromApDelius) {
+    offenderService.getDocumentsFromApDeliusApi(crn)
+  } else {
+    offenderService.getDocumentsFromCommunityApi(crn)
+  }
+
+  private fun getDocumentFileName(documentsMetaData: Any, documentId: UUID): String? = when (documentsMetaData) {
```
davidatkinsuk, line 68: I think use of `Any` here is acceptable as it's a short-lived change

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/DocumentsController.kt
```diff
@@ -31,24 +35,51 @@ class DocumentsController(
...
+    is GroupedDocuments -> getDocumentFileName(documentsMetaData, documentId)
+    is List<*> -> getDocumentFileName(documentsMetaData, documentId)
```
davidatkinsuk, line 70: I guess Kotlin doesn't like using generic types here due to type erasure (E.g. `List<Document>`)?

If it was a long lived change we could maybe introduce an adapter that wraps the community api or ap-and-delius responses to provide the filename given an ID. Would then avoid the need to use the `Any` type and wildcard generics.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/OffenderService.kt
```diff
@@ -640,11 +641,32 @@ class OffenderService(
...
+    }
+    return AuthorisableActionResult.Success(documents)
+  }
+
   fun getDocument(
     crn: String,
     documentId: String,
     outputStream: OutputStream,
-  ) = communityApiClient.getDocument(crn, documentId, outputStream)
+    isRetrieveDocumentFromApDelius: Boolean,
```
davidatkinsuk, line 664: I don't think this matters too much as it's a short-lived change, but for consistency with the getDocuments functions maybe there should be two functions?

* `getDocumentFromCommunityApi`
* `getDocumentFromDeliusApi`

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ApplicationDocumentsTest.kt
```diff
@@ -238,4 +243,95 @@ class ApplicationDocumentsTest : InitialiseDatabasePerClassTestBase() {
       }
     }
   }
+
+  @Test
+  fun `Download document - with get-documents-from-ap-delius feature-flag on - returns 404 when not found in documents meta data`() {
+    mockFeatureFlagService.setFlag("get-documents-from-ap-delius", true)
```
davidatkinsuk, line 249: nice

davidatkinsuk (approved):
Just a few comments from me, mostly just clarifications.

Thanks for the detailed explanation in the PR description. Don't worry about it on this PR but going forward it would be good to get that information on the commit message itself for context when looking through the history. Also if possible could the first line of the commit please be limited to 50 characters.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/DocumentsController.kt
```diff
@@ -31,24 +35,51 @@ class DocumentsController(
...
+    else -> null
+  }
+
+  private fun getDocumentFileName(documentsMetaData: List<*>, documentId: UUID): String {
+    val doc = documentsMetaData.firstOrNull { doc -> (doc as Document).id == documentId.toString() } as Document?
+      ?: throw NotFoundProblem(documentId, "Document")
+    return doc.filename
+  }
+
+  private fun getDocumentFileName(documentsMetaData: GroupedDocuments, documentId: UUID): String? {
```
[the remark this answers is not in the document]
gregkhawkins: Unfortunately this String is nullable as comes from the document data object below (no changes on this from me as it the community api document object):
Screenshot 2024-10-24 at 12 27 35 (2)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/DocumentsController.kt
gregkhawkins: yes I agonised over this for a while but concluded it wopuld be short lived and was a good way to reduce duplication for that short period

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/DocumentsController.kt
gregkhawkins: makes sense, as we are going to strip out very soon will leave as-is for now. Sounds like that is okay

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/OffenderService.kt
gregkhawkins: sound good, and this will only take 5 so will make this change. Thanks!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/DocumentsController.kt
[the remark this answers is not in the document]
davidatkinsuk: Ah! sorry i misread that as coming from the new API. In that case, ignore me :)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/DocumentsController.kt
```diff
@@ -31,24 +35,51 @@ class DocumentsController(
...
-
+    val documentFilename = getDocumentFileName(documentsMetaData, documentId)
     return ResponseEntity(
       StreamingResponseBody { outputStream ->
-        offenderService.getDocument(crn, documentId.toString(), outputStream)
+        offenderService.getDocument(
+          crn,
+          documentId.toString(),
+          outputStream,
+          isRetrieveDocumentFromApDelius = documentsFromApDeliusFlag,
```
Rosie-Brigham: Tiny thing, but maybe `isDocumentFromAPDelius`? May read slightly better

Rosie-Brigham (approved):
Looks nice, only one tiny comment

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/DocumentsController.kt
gregkhawkins: will do - thanks @Rosie-Brigham

gregkhawkins: done and pushed

PR #2389, opened 2024-10-23
Fix auto seeded CAS1 application for access needs

davidatkinsuk:
A change was recently made in the UI that split the access needs questions across two screens. This commit updates the application data for the automatically created CAS1 application to reflect this change, ensuring it’s in a submittable state after being created


froddd (approved):
👍

PR #2391, opened 2024-10-23
APS-1442 - Add CRU Management Area to UserWithWorkload

davidatkinsuk:
This is required by the CAS1 UI to support filtering and showing the CRU Management Area for users that tasks can be assigned to

### DIFF
--- a/src/main/resources/static/_shared.yml
+++ b/src/main/resources/static/_shared.yml
@@ -3072,7 +3072,13 @@ components:
               items:
                 $ref: '#/components/schemas/ApprovedPremisesUserRole'
             apArea:
-              $ref: '#/components/schemas/ApArea'
+              deprecated: true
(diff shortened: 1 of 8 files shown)


src/main/resources/static/_shared.yml
```diff
@@ -3073,6 +3073,9 @@ components:
                 $ref: '#/components/schemas/ApprovedPremisesUserRole'
             apArea:
```
danhumphreys-moj, line 3074: Do we need to set deprecated: true here and add a message to use cruManagementArea instead?

danhumphreys-moj (approved):
LGTM - one minor comment

src/main/resources/static/_shared.yml
davidatkinsuk: Great point - i've made this change as apArea doesn't make much sense in this context anymore

PR #2392, opened 2024-10-23
Don’t revert CRU Management Area on user update

davidatkinsuk:
Before this commit if a user had cruManagementAreaOverride set, the value of user.cruManagementArea was reverted to the default value whenever a user was updated from delius, essentially removing the override.


RobBoothMOJ (approved):
Approved

PR #2393, opened 2024-10-23
Add request for placement Id to Cas1SpaceBooking type.

danhumphreys-moj:
Ref: [https://dsdmoj.atlassian.net/browse/APS-1400](https://dsdmoj.atlassian.net/browse/APS-1400)

`Cas1SpaceBookingType` is returned from `/premises/{premisesId}/space-bookings/{bookingId}`; request for placement id is required by the UI.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/Cas1SpaceBookingTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/Cas1SpaceBookingTest.kt
@@ -836,6 +836,7 @@ class Cas1SpaceBookingTest {
       assertThat(response.id).isEqualTo(spaceBooking.id)
       assertThat(response.otherBookingsInPremisesForCrn).hasSize(1)
       assertThat(response.otherBookingsInPremisesForCrn[0].id).isEqualTo(otherSpaceBookingAtPremises.id)
+      assertThat(response.requestForPlacementId).isEqualTo(spaceBooking.placementRequest.id)
     }
   }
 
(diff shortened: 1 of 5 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/Cas1SpaceBookingTest.kt
```diff
@@ -836,6 +836,7 @@ class Cas1SpaceBookingTest {
       assertThat(response.id).isEqualTo(spaceBooking.id)
       assertThat(response.otherBookingsInPremisesForCrn).hasSize(1)
       assertThat(response.otherBookingsInPremisesForCrn[0].id).isEqualTo(otherSpaceBookingAtPremises.id)
+      assertThat(response.requestForPlacementId).isEqualTo(spaceBooking.placementRequest.id)
```
danhumphreys-moj, line 839: This is testing the scenario when `requestForPlacementId = spaceBooking.placementRequest.id`.  Wondering if we need an extra integration test to test the scenario when `requestForPlacementId = spaceBooking.placementRequest.placementApplication?.id`.

Do we have an integration test to cover off scenario 2 here:

image

Will look at this Friday.
@davidatkinsuk

davidatkinsuk: Hi @danhumphreys-moj I think for completeness we probably should add the test. it can be a copy of this test but it only needs to assert on that specific fields to avoid too much duplication?

Ideally we wouldn't need to do this if/else on placement request id, but we'd need to do a ot of database remodelling to fix the design flaw that has led to it :(

davidatkinsuk (approved):
LGTM, we should probably add the unit test you've proposed if that's ok?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/Cas1SpaceBookingTest.kt
danhumphreys-moj: Having spoken with @davidatkinsuk this morning.  No need to add an additional integration test, however additional unit test coverage was required to check `requestForPlacementId = spaceBooking.placementRequest.placementApplication?.id` is returned when creating placement requests via placement applications - this has been added now.

PR #2394, opened 2024-10-23
APS-1444 Add ‘read only’ option to /profile/v2

davidatkinsuk:
The CAS1 UI will update its view of the user by calling /profile/v2 whenever a change in the user’s hash is detected (via the X-CAS-User-Version header). When this update is required, there is no need to invoke delius to check for any other updates, because the change in the version header indicates that our persisted user is already up to date.

This commit adds a ‘readOnly’ option to /profile/v2 that will supress any attempts to update the user from delius. This will help us avoid getting into ‘infinite update loops’ if we consider a field updated from delius in the X-CAS-User-Version header, and that field’s value is constantly changing.


RobBoothMOJ (approved):
Approved

muhammad-elabdulla (approved):
LGTM

PR #2395, opened 2024-10-24
APS-1362 Auto Seed Test Users for test/dev environments

davidatkinsuk:
This commits adds logic to CAS1 auto seeding to ensure that the required users for e2e tests in the test/dev environments exist on startup, when running in test/dev environments. This works by using the spring profile name.

It also ensures all test users are assigned all available qualifications, reflecting the current configuration.

Whilst this change helps ensure the dev/test environments are correctly configured, it also simplifies running tests locally against upstream services in dev


RobBoothMOJ (approved):
Approved

PR #2396, opened 2024-10-24
APS-1362 - Add bootRunDebug task, removing hardcoded profile

davidatkinsuk:
We currently use bootRunLocal to start the API from ap-tools. This task hardcodes the spring profile being used to ‘local’. To support running tests against upstream services, we want to be able to run against the ‘dev’ profile. Therefore, this commit introduces a new ‘bootRunDebug’ task that is identical to ‘bootRunLocal’ but removes the hardcoded profile.

The calling ap-tools code already sets the active profile in the env var, so this will not breaking existing behaviour

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -132,12 +132,25 @@ tasks {
   compileTestScala { enabled = false }
 }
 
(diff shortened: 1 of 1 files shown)


RobBoothMOJ (approved):
Approved

PR #2401, opened 2024-10-24
Change future bookings report to use getPersonSummaryInfoResults

muhammad-elabdulla:
This [PR CAS-643](https://dsdmoj.atlassian.net/browse/CAS-643) is to use  getPersonSummaryInfoResults instead of getOffenderSummariesByCrns to get offender details

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/OffenderService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/OffenderService.kt
@@ -712,8 +712,10 @@ class OffenderService(
   }
 }
 
-fun UserEntity.casLimitedAccessStrategy() = if (this.hasQualification(UserQualification.LAO)) {
+fun UserEntity.cas1LimitedAccessStrategy() = if (this.hasQualification(UserQualification.LAO)) {
   LimitedAccessStrategy.IgnoreLimitedAccess
 } else {
   LimitedAccessStrategy.ReturnRestrictedIfLimitedAccess(this.deliusUsername)
 }
+
+fun UserEntity.cas3LimitedAccessStrategy() = LimitedAccessStrategy.ReturnRestrictedIfLimitedAccess(this.deliusUsername)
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/OffenderService.kt
```diff
@@ -712,7 +712,7 @@ class OffenderService(
   }
 }
 
-fun UserEntity.cas1LimitedAccessStrategy() = if (this.hasQualification(UserQualification.LAO)) {
+fun UserEntity.casLimitedAccessStrategy() = if (this.hasQualification(UserQualification.LAO)) {
```
davidatkinsuk, line 715: this strategy makes use of the LAO Qualification, which i believe only CAS1 uses. Therefore, i don't think this should be renamed or used by CAS3. Otherwise, we're mixing CASs and it becomes difficult to untangle/reason on in the future.

For CAS3, i believe you should instead be passing through 

`LimitedAccessStrategy.ReturnRestrictedIfLimitedAccess(user.deliusUsername)`

or, you could introduce a new function like the following and use that instead to make the code a bit more readible?

`fun UserEntity.casLimitedAccessStrategy() = LimitedAccessStrategy.ReturnRestrictedIfLimitedAccess(user.deliusUsername)`

davidatkinsuk (approved):
one comment from me, otherwise looks good

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/OffenderService.kt
muhammad-elabdulla: Change the code by create a new function but to be honest the code is a little bit confused and not easy to follow when we get offender details

davidatkinsuk: Any proposals for improvement are welcome. I thought the new function definition was reasonably straight forward (at least compared to all the deprecated ones)

```
  fun getPersonSummaryInfoResults(
    crns: Set<String>,
    limitedAccessStrategy: LimitedAccessStrategy,
  ): List<PersonSummaryInfoResult>
```

PR #2403, opened 2024-10-25
APS-1448 - Add Space Booking Day Planner Logic

davidatkinsuk:
This PR adds a SpaceBookingDayPlanner service. Given a list of rooms and space bookings, this will attempt to optimally assign a a space booking to each room, where possible.

Subsequent commits will make use of this planner to plan occupancy over a period of time for a given premise, which will then be used for occupancy calculation

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/planning/SpaceBookingDayPlanner.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/planning/SpaceBookingDayPlanner.kt
@@ -1,0 +1,153 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas1.planning
+
+import org.springframework.stereotype.Service
+
+@Service
+class SpaceBookingDayPlanner {
+  /**
+   * Given a list of available beds and required bookings for a given day, will attempt to optimally
+   * assign bookings to beds to determine a premise's capacity on a given day.
+   *
+   * We do not currently intend to present the 'booking to room', instead we are only using this planning
+   * to determine the number of free spaces after bookings are planned.
+   *
+   * If we were to present 'booking to room' planning, we may wish to discuss further optimisation
+   * with premises e.g. would they want to plan bookings into the fullest multiple occupancy room possible,
+   * Or is the goal to minimise shared rooms? What's more important, minimising characteristic surplus, or
+   * minimising single bookings into room with the least other beds?
+   *
+   * Planning goals:
+   *
+   * <ol>
+   *   <li>Priority is given to bookings with constraints, those with the most constraints being planned first</li>
+   *   <li>Bookings are placed into rooms with the least surplus characteristics</li>
+   *   <li>Single occupancy bookings are never placed in rooms with other bookings</li>
+   *   <li>If there are multiple competing rooms (after above goals are satisfied), single occupancy bookings will
+   *   be placed into rooms with the least beds (ensuring minimum 'wastage' in these scenarios)</li>
+   * </ol>
+   *
+   * @param beds All beds available on the day being planned (i.e. won't include out of service beds or beds inactive according to start/end date)
+   * @param bookings All required bookings. The code assumes that the premise level characteristics have already been met
+   */
+  fun plan(
+    beds: Set<Bed>,
+    bookings: Set<SpaceBooking>,
+  ): DayPlannerResult {
+    val bedLedger = BedLedger(beds)
+    val planned = mutableListOf<DayPlannerPlannedBooking>()
+    val unplanned = mutableListOf<DayPlannerUnplannedBooking>()
+
+    val sortedBookings = bookings.toList()
+      .sortedByDescending { b -> b.requiredCharacteristics.sumOf { c -> c.weighting } }
+      .toMutableList()
+
+    sortedBookings.forEach { booking ->
+      val requiresSingleRoom = booking.requiredCharacteristics.any { it.singleRoom }
+      val characteristicsExcludingSingleRoom = booking.requiredCharacteristics.filter { !it.singleRoom }.toSet()
+
+      when (
+        val findResult = bedLedger.findBed(
+          characteristics = characteristicsExcludingSingleRoom,
+          requiresSingleRoom = requiresSingleRoom,
+        )
+      ) {
+        is FindBedResult.BedsFound -> {
+          findResult.beds.forEach { bed ->
+            planned.add(
+              DayPlannerPlannedBooking(
+                bed = bed,
+                booking = booking,
+              ),
+            )
+            bedLedger.reserve(bed)
+          }
+        }
+
+        is FindBedResult.BedNotFound -> {
+          unplanned.add(
+            DayPlannerUnplannedBooking(
+              booking = booking,
+            ),
+          )
+        }
+      }
+    }
+
+    return DayPlannerResult(
+      planned = planned.toList(),
+      unplanned = unplanned.toSet(),
+    )
+  }
+}
+
+private class BedLedger(initialState: Set<Bed>) {
+  private val allBeds = initialState.sortedBy { it.label }.toMutableList()
+  private val availableBeds = initialState.sortedBy { it.label }.toMutableList()
+  private val reservedBeds = mutableListOf<Bed>()
+  private val bedsByRoom = allBeds.groupBy { it.room }
+
+  fun findBed(
+    characteristics: Set<Characteristic>,
+    requiresSingleRoom: Boolean,
+  ): FindBedResult {
+    val bedsWithRequiredCharacteristics = findBedsWithRequiredCharacteristics(characteristics, requiresSingleRoom)
+    val bedWithLeastSurplusCharacteristics = findBedWithLeastSurplusCharacteristics(bedsWithRequiredCharacteristics, characteristics)
+
+    if (bedWithLeastSurplusCharacteristics == null) {
+      return FindBedResult.BedNotFound
+    }
+
+    return if (requiresSingleRoom) {
+      FindBedResult.BedsFound(bedsByRoom[bedWithLeastSurplusCharacteristics.room]!!.toSet())
+    } else {
+      FindBedResult.BedsFound(setOf(bedWithLeastSurplusCharacteristics))
+    }
+  }
+
+  private fun findBedsWithRequiredCharacteristics(
+    characteristics: Set<Characteristic>,
+    requiresSingleRoom: Boolean,
+  ): List<Bed> {
+    val bedsWithMatchingCharacteristics = availableBeds
+      .filter { availableBed -> availableBed.room.characteristics.containsAll(characteristics) }
+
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/planning/SpaceBookingDayPlanner.kt
```diff
@@ -0,0 +1,129 @@
...
+      if (requiresSingleRoom) {
+        FindBedResult.BedsFound(allAvailableBedsInRoom(bedWithLeastUnusedCharacteristics.room))
+      } else {
+        FindBedResult.BedsFound(setOf(bedWithLeastUnusedCharacteristics))
+      }
+    } else {
+      FindBedResult.BedNotFound
+    }
+  }
+
```
bobmeredith, line 100: I may be wrong but as far as I can see, there's nothing to force a person into the fullest multiple occupancy room possible if the booking don't have 'single' set, or, more importantly, to put a single occupancy booking into the room with the smallest number of beds.
This `findBed` routine will be called a lot and has several all-bed loops within it so we're probably talking of a performance going with n^2. If performance is an issue, I think the number of scans in this area could be reduced.

bobmeredith (commented):
Looking really good to me. The visual rendering is really nice. Just a couple of comments

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas1/planning/SpaceBookingDayPlannerTest.kt
```diff
@@ -0,0 +1,440 @@
...
+
+    result.assertResult(
+      """
+      Planned: 5
+      
+      | Bed             | Booking         | Bed Characteristics            |
+      | --------------- | --------------- | ------------------------------ |
+      | room1 bed1      | booking1        |                                |
+      | room1 bed2      | booking1        |                                |
+      | room2 bed1      | booking1        |                                |
```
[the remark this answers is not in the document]
davidatkinsuk: Excellent spot. Your expectations were correct, I messed up the room definition:

`val room2Bed1 = bed("room2 bed1", room = room1)`

This is now fixed

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/planning/SpaceBookingDayPlanner.kt
davidatkinsuk: Some great points.

> there's nothing to force a person into the fullest multiple occupancy room possible if the booking don't have 'single' set

Because we plan bookings with characteristics first (e.g. bookings requiring single rooms), i don't think we need to explicit add this logic, although I agree that this is definitely a goal of the planning. I'm going to look at documenting the goals of the planning at the top of this class and including this so we don't lose sight of it

I'll also add a unit test to make this expectation clear

> put a single occupancy booking into the room with the smallest number of beds.

This is a really good point - I'll explicitly add this logic and add a unit test

> This findBed routine will be called a lot and has several all-bed loops within it so we're probably talking of a performance going with n^2. If performance is an issue, I think the number of scans in this area could be reduced.

I'm cautiously optimistic that we won't need to optimise given that premises won't have too many beds e.g. <100. We have a series of tickets on the occupancy epic APS-1256 with an end goal of trying to plan  a period of time for a given premise against existing bookings and see what the outcome is. We can capture timings from this. I've updated the ticket to make sure we do that.

davidatkinsuk: Done! I've rewritten the documentation on the planning class to reflect the discussion above, and highlight some things to consider/optimise if we ever want to start presenting the 'booking to room' planning to premises.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/planning/SpaceBookingDayPlanner.kt
```diff
@@ -0,0 +1,147 @@
...
+              booking = booking,
+            ),
+          )
+        }
+      }
+    }
+
+    return DayPlannerResult(
+      planned = planned.toList(),
+      unplanned = unplanned.toSet(),
```
Rosie-Brigham, line 78: Am I right thinking that the unplanned here will overbooked bookings, where a suitable bed could not be found?

davidatkinsuk: Yep that's correct. I was trying to keep the language neutral wrt. planning, but these are what we would present as 'overbooked'

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/planning/SpaceBookingDayPlanner.kt
```diff
@@ -0,0 +1,147 @@
...
+      FindBedResult.BedNotFound
+    }
+  }
+
+  fun reserve(bed: Bed) {
+    availableBeds.remove(bed)
+    reservedBeds.add(bed)
+  }
+
+  private fun allBedsInRoom(room: Room) = allBeds.filter { it.room == room }.toSet()
```
Rosie-Brigham, line 123: Would it make sense to have this as a variable that is calculated once, then referred back to through each check, instead of a function?
e.g `private val bedsByRoom = allBeds.groupBy { it.room }`

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/planning/SpaceBookingDayPlanner.kt
```diff
@@ -0,0 +1,147 @@
...
+   *   <li>Single occupancy bookings are never placed in rooms with other bookings</li>
+   *   <li>If there are multiple competing rooms (after above goals are satisfied), single occupancy bookings will
+   *   be placed into rooms with the least beds (ensuring minimum 'wastage' in these scenarios)</li>
+   * </ol>
+   *
+   * @param beds All beds available on the day being planned (i.e. won't include out of service beds or beds inactive according to start/end date)
+   * @param bookings All required bookings. The code assumes that the premise level characteristics have already been met
+   */
+  fun plan(
+    beds: Set<Bed>,
```
Rosie-Brigham, line 33: Shall we add in some error handling for poor data? Do you think there could be a circumstance in which there are two beds with the same reference, which are the same bed but have been incorrectly inputted?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/planning/SpaceBookingDayPlanner.kt
```diff
@@ -0,0 +1,147 @@
...
+    val bedsWithMatchingCharacteristics =
+      if (requiresSingleRoom) {
+        availableBeds
+          .filter { availableBed -> availableBed.room.characteristics.containsAll(characteristics) }
+          .filter { availableBed -> !isRoomOccupied(availableBed) }
+          .sortedBy { availableBeds -> allBedsInRoom(availableBeds.room).size }
+      } else {
+        availableBeds
+          .filter { availableBed -> availableBed.room.characteristics.containsAll(characteristics) }
+      }
```
Rosie-Brigham, line 101: Could this logic be refactored into smaller functions, to make it simpler to read and test maybe?

e.g something like
 ```
private fun findMatchingBeds(
        characteristics: Set<Characteristic>,
        requiresSingleRoom: Boolean
    ): Set<Bed> {
        val matchingBeds = availableBeds.filter { bed ->
            bed.room.characteristics.containsAll(characteristics)
        }

        return if (requiresSingleRoom) {
            matchingBeds.filter { bed -> !isRoomOccupied(bed) }
                .sortedBy { bed -> bedsByRoom[bed.room]?.size ?: 0 }
        } else {
            matchingBeds
        }.toSet()
    }

    private fun findOptimalBed(candidates: Set<Bed>, requirements: Set<Characteristic>): Bed? =
        candidates.minByOrNull { bed ->
            bed.room.characteristics.minus(requirements).size
        }

val candidateBeds = findCandidateBeds(characteristics, requiresSingleRoom)
val optimalBed = findOptimalBed(candidateBeds, characteristics) ?: return FindBedResult.BedNotFound
        
       
FindBedResult.BedsFound(setOf(optimalBed))


```

Rosie-Brigham: Apologies for the terrible formatting

Rosie-Brigham (approved):
Looks really great - the logic seems all there. I am trying to think of things that could break it, but haven't thought of any yet.
The only thing that may be problematic is duplicate rooms or bookings(?)
But I fear that if that happens there are bigger fish to fry.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/planning/SpaceBookingDayPlanner.kt
davidatkinsuk: Whilst i tend to favour readability over optimisation i think this is a good optimisation that doesn't impact readability at all (actually probably makes it more readable). i've made the change 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/planning/SpaceBookingDayPlanner.kt
davidatkinsuk: Whilst `bed.reference` being duplicated would produce confusing 'dev only' output from the planner (e.g. the text output we use to verify behaviour), it's actually `bed.id` that would cause an issue if duplicated.  I'm inclined not to add error handling as we trust the source (our own code), and if bed names/labels were duplicated it should be very obvious on the UI and be dealt with quickly. With that said, there's no harm in adding error handling if this is something you'd like adding

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/planning/SpaceBookingDayPlanner.kt
davidatkinsuk: This is great ... i've mostly gone with code proposed as-is. I agree breaking it out into chunks makes it easier to understand. I've stuck with `bedWithLeastSurplusCharacteristics` instead of `optimalBed` as i think it makes it clearer what the optimisation is (it's not so obvious reading the code)

PR #2405, opened 2024-10-25
CAS-617 Remove PDU property from bedsearch Api endpoint

muhammad-elabdulla:
Remove PDU property from bedsearch Api endpoint as it was replaced with PDU list


davidatkinsuk (approved):
LGTM 👍

PR #2406, opened 2024-10-25
GET Documents API to use Delius API

gregkhawkins:
- Refactor to the `GET /applications/{applicationId}/documents` endpoint to retrieve all docs using the GET documents endpoint fro the `AP and Delius API` (instead of using the `Community API`)
- After some analysis, David A and I, discovered that the 'Temporary Accommodation' flow is no longer used so the related business-logic has also been stripped out


src/main/resources/static/_shared.yml
```diff
@@ -3669,7 +3669,6 @@ components:
         - id
         - level
         - fileName
-        - createdAt
```
gregkhawkins, line 3672: This can be null from `AP and Delius API` so had no choice but to make it nullable int the open-api spec

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ApplicationDocumentsTest.kt
```diff
@@ -78,64 +77,39 @@ class ApplicationDocumentsTest : InitialiseDatabasePerClassTestBase() {
...
-              documentTransformer.transformToApiUnfiltered(groupedDocuments),
+              documentTransformer.transformToApi(groupedDocuments),
             ),
           )
       }
     }
   }
 
   @Test
-  fun `Get application documents for Temporary Accommodation returns 200`() {
```
gregkhawkins, line 89: As mentioned in PR description, David A and I, discovered that the 'Temporary Accommodation' flow is no longer used so the related business-logic has also been stripped out as well as this test to cover that case

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/transformer/DocumentTransformerTest.kt
```diff
@@ -66,7 +68,7 @@ class DocumentTransformerTest {
   }
 
   @Test
-  fun `transformToApiFiltered filters out convictions other than one specified`() {
```
gregkhawkins, line 69: no longer filtering by convictionId due to the 'Temporary Accommodation' not needed discovery (see above comment)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/DocumentTransformer.kt
```diff
@@ -53,9 +43,27 @@ class DocumentTransformer {
...
+  fun transformToApi(documents: List<APDeliusDocument>): List<Document> =
+    documents
+      .filter { it.id != null }
+      .map {
+        Document(
+          id = it.id!!,
+          level =  when (it.level) {
+            DocumentLevel.conviction.value -> DocumentLevel.conviction
+            else -> DocumentLevel.offender
+          },
```
gregkhawkins, line 58: think this when statement is safe to use `offender` as the fallback as `AP and Delius API` will always give us one or the other. See line 31;
https://github.com/ministryofjustice/hmpps-probation-integration-services/blob/main/projects/approved-premises-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/DocumentService.kt

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/DocumentTransformer.kt
```diff
@@ -14,16 +14,7 @@ class DocumentTransformer {
...
 
-  fun transformToApiFiltered(groupedDocuments: GroupedDocuments, convictionId: Long?) =
-    transformToApi(groupedDocuments) { it.convictionId == convictionId.toString() }
-
-  fun transformToApiUnfiltered(groupedDocuments: GroupedDocuments) =
-    transformToApi(groupedDocuments) { true }
-
-  fun transformToApi(
-    groupedDocuments: GroupedDocuments,
-    convictionDocFilter: (ConvictionDocuments) -> Boolean,
```
gregkhawkins, line 25: no longer need to filter by convictionId due to the discovery (as mentioned in PR description) that the 'Temporary Accommodation' flow is no longer used so the related business-logic can be stripped been stripped out.

davidatkinsuk (approved):
Looks good - just some minor comments from me

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/DocumentTransformer.kt
```diff
@@ -53,9 +43,27 @@ class DocumentTransformer {
...
+          id = it.id!!,
+          level = when (it.level) {
+            DocumentLevel.conviction.value -> DocumentLevel.conviction
+            else -> DocumentLevel.offender
+          },
+          fileName = it.filename,
+          createdAt = it.dateCreated?.toInstant(),
+          typeCode = it.typeCode,
+          typeDescription = it.typeDescription,
+          description = it.description,
```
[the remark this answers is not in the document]
gregkhawkins: exactly! Once Marcus delivers that we can have a small PR with the change `description = it.extendedDescription`

src/main/resources/application.yml
```diff
@@ -308,4 +308,4 @@ feature-flags:
   cas1-womens-estate-enabled: false
   cas1-expired-applications-enabled: false
   include-assessment-updated-domain-events: false
-  get-documents-from-ap-delius: false
+  get-documents-from-ap-delius: true
```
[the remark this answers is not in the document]
gregkhawkins: moved to `application-local.yml` as discussed

PR #2409, opened 2024-10-28
CAS1 - Enable Reduce Info on Apply Functionality

davidatkinsuk:
This enables the following functionality in production:

* only list adjudications up to 12 months
* only list specific prison case note types


RobBoothMOJ (approved):
approved

PR #2410, opened 2024-10-28
APS-1457 - Disable reduce info on apply features

davidatkinsuk:
Disabling these features whilst we investigate some issues relates to existing applications using entities no longer being returned by the API


froddd (approved):
👍

RobBoothMOJ (approved):
Approved

bobmeredith (commented):
LGTM

PR #2413, opened 2024-10-28
APS-1464 Remove premises/bookings/move endpoint

davidatkinsuk:
This is not in use by any CAS UI and can be removed


Rosie-Brigham (approved):
Burn it all!

PR #2414, opened 2024-10-28
remove redis tracking from app insights

Rosie-Brigham:
This PR forms part of [API-1357](https://dsdmoj.atlassian.net/jira/software/c/projects/APS/boards/1328?selectedIssue=APS-1375), reviewing our usage of app insights (which is larrrrrge!)

I have updated the app insight config to not track redis, as there seems little business need for this
I have also reduced the sampling for the health and info endpoint to 5%, as [per this guidance](https://docs.google.com/document/d/1uG7Ne1_B2RwgREI-iMXHVnyQfuIQykDVfMQ1k-h-xHk/edit)

### DIFF
--- a/applicationinsights.json
+++ b/applicationinsights.json
@@ -8,6 +8,9 @@
   "instrumentation": {
     "logging": {
       "level": "DEBUG"
+    },
+    "redis": {
+      "enabled": false
     }
   },
   "selfDiagnostics": {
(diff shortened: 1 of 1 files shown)


applicationinsights.json
```diff
@@ -8,6 +8,9 @@
   "instrumentation": {
     "logging": {
       "level": "DEBUG"
+    },
+    "redis": {
+      "enabled": false
```
davidatkinsuk, line 13: i noticed the guidance suggests sampling redis at 5% which i _could_ see being useful just as a sanity check that we are using caching. BUT - i don't recall ever making use of these redis logs in the last year so i think it's fine to completely disable 👍

davidatkinsuk (approved):
LGTM 👍

applicationinsights.json
Rosie-Brigham: Yeah that was my thinking - chatting to other folks no one else seems to use it either. If it is important we can just turn it on again

PR #2416, opened 2024-10-29
CAS-480 Filter bedspaces which have room wheelchair attribute

muhammad-elabdulla:
This [PR CAS-480](https://dsdmoj.atlassian.net/browse/CAS-480) is to change the way that bedspaces are filtering by wheelchair. It changes to filter bedspaces with rooms have wheelchair attribute instead of properties with wheelchair attribute

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BedSearchService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BedSearchService.kt
@@ -143,14 +143,18 @@ class BedSearchService(
           when (it) {
             BedSearchAttributes.singleOccupancy -> "Single occupancy"
             BedSearchAttributes.sharedProperty -> "Shared property"
-            BedSearchAttributes.wheelchairAccessible -> "Wheelchair accessible"
+            BedSearchAttributes.wheelchairAccessible -> ""
           }
         }
 
-        val premisesCharacteristicIds = premisesCharacteristicsNames?.let {
-          val premisesCharacteristics = characteristicService.getCharacteristicsByNames(premisesCharacteristicsNames)
-          premisesCharacteristics.filter { it.isActive && it.matches(ServiceName.temporaryAccommodation.value, "premises") }.map { it.id }.toList()
-        } ?: emptyList()
+        val premisesCharacteristicIds = getTemporaryAccommodationCharacteristicsIds(premisesCharacteristicsNames, "premises")
+
+        val roomCharacteristicsNames = when {
+          propertyBedAttributes?.contains(BedSearchAttributes.wheelchairAccessible) == true -> listOf("Wheelchair accessible")
+          else -> null
+        }
+
+        val roomCharacteristicIds = getTemporaryAccommodationCharacteristicsIds(roomCharacteristicsNames, "room")
 
         val endDate = startDate.plusDays(durationInDays.toLong() - 1)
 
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BedSearchService.kt
```diff
@@ -143,13 +143,35 @@ class BedSearchService(
...
+
+        val roomCharacteristicIds = roomCharacteristicsNames?.let {
+          val roomCharacteristics = characteristicService.getCharacteristicsByNames(roomCharacteristicsNames)
+          roomCharacteristics.filter {
+            it.isActive &&
+              (
+                it.matches(ServiceName.temporaryAccommodation.value, "room") ||
+                  it.matches(ServiceName.temporaryAccommodation.value, "*")
+                )
+          }.map { it.id }.toList()
```
davidatkinsuk, line 174: these two chunks of code to resolve characteristics ids appear to be almost identical other than matching on 'room' instead of 'premises'. could this be moved into a common private function?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BedSearchService.kt
```diff
@@ -143,13 +143,35 @@ class BedSearchService(
...
+        } ?: emptyList()
+
+        val roomCharacteristicIds = roomCharacteristicsNames?.let {
+          val roomCharacteristics = characteristicService.getCharacteristicsByNames(roomCharacteristicsNames)
+          roomCharacteristics.filter {
+            it.isActive &&
+              (
+                it.matches(ServiceName.temporaryAccommodation.value, "room") ||
+                  it.matches(ServiceName.temporaryAccommodation.value, "*")
```
davidatkinsuk, line 172: looking at the definition of `CharacteristicEntity.matches`, it already considers the wildcard '*' value, so i don't think this additional check is required?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BedSearchService.kt
```diff
@@ -143,13 +143,35 @@ class BedSearchService(
...
+            it.isActive &&
+              (
+                it.matches(ServiceName.temporaryAccommodation.value, "premises") ||
+                  it.matches(ServiceName.temporaryAccommodation.value, "*")
+                )
+          }.map { it.id }.toList()
+        } ?: emptyList()
+
+        val roomCharacteristicIds = roomCharacteristicsNames?.let {
+          val roomCharacteristics = characteristicService.getCharacteristicsByNames(roomCharacteristicsNames)
```
davidatkinsuk, line 167: something to consider for the future - i think we should be matching on `CharacteristicEntity.propertyName` instead of `CharacteristicEntity.name` as the former is less likely to ever change.

davidatkinsuk (approved):
A few comments on potential code improvements, but functionally LGTM 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BedSearchService.kt
muhammad-elabdulla: Yes it already checking it

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/BedSearchService.kt
muhammad-elabdulla: For temporary accommodation most of the property name are null

PR #2418, opened 2024-10-29
CAS-440 Change to return data domain as json instead of string

muhammad-elabdulla:
This [PR CAS-440](https://dsdmoj.atlassian.net/browse/CAS-440) is to change to return data domain as json instead of string

### DIFF
--- a/src/main/resources/static/_shared.yml
+++ b/src/main/resources/static/_shared.yml
@@ -2904,7 +2904,7 @@ components:
         isWithdrawn:
           type: boolean
         domainEvent:
-          type: object
+          $ref: '#/components/schemas/AnyValue'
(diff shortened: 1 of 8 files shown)


src/main/resources/static/_shared.yml
```diff
@@ -2904,7 +2904,7 @@ components:
         isWithdrawn:
           type: boolean
         domainEvent:
-          type: object
+          $ref: '#/components/schemas/AnyValue'
```
aliuk2012, line 2907: Could we maybe structure this a bit better? 

- id
- eventType = string
- timestamp = datetime
- updatedFields = array of objects with fieldName, updatedTo, updatedFrom

PR #2421, opened 2024-10-31
Fix an issue showing overlap people staying in a bedspace when their bookings was cancelled

muhammad-elabdulla:
This PR CAS-1207 is to fix an issue showing overlap people staying in a bedspace when their bookings was cancelled


Rosie-Brigham (approved):
Well found! ⭐

PR #2426, opened 2024-10-31
Revert document's createdAt field back to non-nullable

gregkhawkins:
- Changes also include switching on the documents feature flag on preprod


davidatkinsuk (approved):
LGTM 👍

PR #2429, opened 2024-11-01
Remove Old CAS1 Seed and Migration Jobs

davidatkinsuk:
*approvedPremisesBookingAdhocProperty*

This seed job was used to populate the booking.adhoc property when it was introduced in March 2024. This was a one time exercise, so the seed job is no longer required.

*cas1LostBedsToOutOfServiceBeds*

This job was used for a one time exercise when migrating CAS1 lost beds to Out Of Service Beds. It is no longer required

*cas1UserDetailsMigrationJob*

This job was used to populate the `Cas1ApplicationUserDetailsEntity` when it was first introduced as a one time exercised. It is no longer required

*Cas1TruncateOosbMigrationJob*

This job was used to fix some existing out of service bed records where the record end date exceeded the bed's end date. This was a one time fix to address a subsequently fixed bug. This is no longer required

*Cas1ReasonForShortNoticeMetadataMigrationJob*

This job was used to populate reason for short notice in the application submission domain event. This is captured for applciation's going forward so is no longer required.


muhammad-elabdulla (approved):
LGTM

PR #2431, opened 2024-11-01
CAS-600 Ignore case when HPTs search by CRN

muhammad-elabdulla:
This [PR CAS-600](https://dsdmoj.atlassian.net/browse/CAS-600) is to ignore case sensitivity when HPTs search by CRN in referrals and bookings


davidatkinsuk (approved):
LGTM 👍

PR #2434, opened 2024-11-01
Remove /profile endpoint

davidatkinsuk:
This has been superceded by /profile/v2, which is now used by both CAS1 and CAS3


muhammad-elabdulla (approved):
LGTM

PR #2436, opened 2024-11-04
APS-1486 - Add DeliusEventNumber to SpaceBooking

davidatkinsuk:
This commit persists a delius_event_number to the cas1_space_bookings table.

Whilst each booking logically must have a delius_event_number, there are a non-trivial number of bookings created via offline_applications where the event number is not known, and without a corresponding APPROVED_PREMISES_BOOKING_MADE domain event to derive this from. For this reason, the new column is optional.


RobBoothMOJ (approved):
Approved

PR #2437, opened 2024-11-04
APS-1486 - Use space booking event no. for domain events

davidatkinsuk:
Whilst cas1_space_bookings.delius_event_number is nullable and domain events require a non-null value, we will guarantee that this is always set to a value for any space bookings that has not already occurred.

APS-1488 will be used to manage backfilling this value when migrating from bookings to space bookings

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1SpaceBookingManagementDomainEventService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1SpaceBookingManagementDomainEventService.kt
@@ -51,6 +51,7 @@ class Cas1SpaceBookingManagementDomainEventService(
     val premises = updatedCas1SpaceBooking.premises
     val offenderDetails = getOffenderForCrn(updatedCas1SpaceBooking.crn)
     val keyWorker = getStaffMemberDetails(updatedCas1SpaceBooking.keyWorkerStaffCode)
+    val eventNumber = updatedCas1SpaceBooking.deliusEventNumber!!
 
     val actualArrivalDate = updatedCas1SpaceBooking.actualArrivalDateTime!!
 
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1SpaceBookingManagementDomainEventService.kt
```diff
@@ -51,6 +51,7 @@ class Cas1SpaceBookingManagementDomainEventService(
     val premises = updatedCas1SpaceBooking.premises
     val offenderDetails = getOffenderForCrn(updatedCas1SpaceBooking.crn)
     val keyWorker = getStaffMemberDetails(updatedCas1SpaceBooking.keyWorkerStaffCode)
+    val eventNumber = updatedCas1SpaceBooking.deliusEventNumber!!
```
gregkhawkins, line 54: assume we can safely assume that the double-bang will not come back to haunt us?

gregkhawkins (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1SpaceBookingManagementDomainEventService.kt
davidatkinsuk: regardless, here's a longer description i typed out :)

Good question.

These domain events only apply to 'active' space bookings (i.e. bookings with a departure date in the future). We can guarantee that any new space booking created in our system will have this populated. All other space bookings will be migrated from legacy bookings. Of these, there's only one with a departure date in the future (2025) where we can't resolve an event number at the point of migration. For that booking i've created a ticket (APS-1488) to manually backfill the value by looking at delius to figure out what it should be :)

PR #2439, opened 2024-11-04
Add wheelchair accessible room characteristics to premises

daveawc:
The wheelchair accessible characteristic was changed to be a **room only** attribute, where previously it was used for both room and premises. This is the only remaining premises that has it on the premises level and not the rooms.


vbala-moj (approved):

davidatkinsuk (approved):
LGTM - to confirm my understanding, you're removing the wheelchair characteristic from the premise and adding it to specific rooms in the premise?

PR #2440, opened 2024-11-05
APS-1487 - Support offline applications for space bookings

davidatkinsuk:
We have several bookings linked to `offline_applications` with space bookings arriving and/or departing in the future. Therefore, these need supporting.

This commit updates the `cas1_space_bookings` datamodel to support space bookings being linked to an `offline_application` instead of an `approved_premises_application`

This requires update to email and domain event logic, which mimics the behaviour currently used for legacy `bookings`. Note that withdrawal emails are not sent for bookings (or space bookings) linked to offline bookings. A ticket APS-1493 has been created to address that.

Note that this commit is not sufficient to fully support the migration of bookings linked to offline applications to space bookings because placement requests will also not be defined for such bookings. That will be addressed in subsequent commits, along with space booking migration tests for this scenario.


src/main/resources/db/migration/all/20241104135401__add_offline_app_to_cas1_space_booking.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE cas1_space_bookings ALTER COLUMN approved_premises_application_id DROP NOT NULL;
+ALTER TABLE cas1_space_bookings ADD offline_application_id UUID NULL;
+
+ALTER TABLE cas1_space_bookings ADD FOREIGN KEY (offline_application_id) REFERENCES offline_applications(id);
+
+alter table cas1_space_bookings
+    add constraint only_one_application_set
+    check ((approved_premises_application_id is null) <> (offline_application_id is null));
```
danhumphreys-moj, line 8: Does this fall over if both are set?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas1/Cas1SpaceBookingManagementDomainEventServiceTest.kt
```diff
@@ -428,24 +490,57 @@ class Cas1SpaceBookingManagementDomainEventServiceTest {
...
 
-      checkDomainEventBookingProperties(domainEvent, spaceBookingWithKeyWorker)
+      checkDomainEventBookingProperties(domainEvent, spaceBookingWithKeyWorker, application.id)
+      val domainEventEventDetails = domainEvent.data.eventDetails
+      assertThat(domainEventEventDetails.previousKeyWorkerName).isEqualTo(previousKeyWorkerName)
+      assertThat(domainEventEventDetails.assignedKeyWorkerName).isEqualTo(keyWorkerName)
+    }
+
+    @Test
+    fun `record keyWorker assigned and emits domain event with offline application`() {
```
danhumphreys-moj, line 500: Is this testing a historical scenario?  Since we aren't expecting any future offline applications?

danhumphreys-moj (approved):
Observation: commit four is large relative to the previous three - however, looks good to me.

src/main/resources/db/migration/all/20241104135401__add_offline_app_to_cas1_space_booking.sql
davidatkinsuk: Yep that's the idea. We make that assumption in code (one is set) so it's nice to have a database level constraint to guarantee that (i've never used checks before, they're handy!)

davidatkinsuk:
> Observation: commit four is large relative to the previous three - however, looks good to me.

I tried to pull out 'stand alone' changes into separate commit to help readibility/reviewing, that's what the first three are

src/main/resources/db/migration/all/20241104135401__add_offline_app_to_cas1_space_booking.sql
danhumphreys-moj: I was thinking this DB check doesn't work if a space booking is linked to an application id and an offline application id, this would mean:

check (FALSE <> FALSE)

Which returns FALSE so not really what we want?  Do we not want an XOR type check?

davidatkinsuk: The intention _is_ an XOR check, which i believe is what we're getting here e.g. a space booking must be linked to  an application or an offline application (one and only one)

PR #2441, opened 2024-11-05
Add stub for Cas1 Space Booking, non arrival

RobBoothMOJ:
Added Space Booking Non Arrival API definition and stub


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas1/Cas1SpaceBookingController.kt
```diff
@@ -212,6 +213,14 @@ class Cas1SpaceBookingController(
       )
   }
 
+  override fun recordNonArrival(
+    premisesId: UUID,
+    bookingId: UUID,
+    cas1NonArrival: Cas1NonArrival,
+  ): ResponseEntity<Unit> {
+    return super.recordNonArrival(premisesId, bookingId, cas1NonArrival)
```
gregkhawkins, line 221: as far as I can see the result of this is sets an example response? 
Are we doing this only so that we have Swagger that consumers can look at? Also possible I'm missing something! :)

gregkhawkins (approved):

RobBoothMOJ:
Implemented this change as part of a PR 2454 : 
https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/2454

PR #2442, opened 2024-11-05
Give AP_USER_TEST_2 specific roles in dev

davidatkinsuk:
This test user is used to test report and user management functionality. This commit ensure they only have the required roles to do this.


RobBoothMOJ (approved):
Approved

PR #2446, opened 2024-11-05
APS-1492 - Support migrating bookings to space bookings for offline applications

davidatkinsuk:
This PR includes changes that allows us to migrate a subset of legacy `bookings` where:

1. There is no corresponding linked placement request (the case for all offline applications and adhoc bookings)
2. There is no corresponding APPROVED_PREMISES_BOOKING_MADE domain event

This is a small subset of bookings, but some have arrival/departure dates in the future so they must be migrated

As a side effect of this change, the following fields are no longer mandatory on the SpaceBooking API type:

* assessmentId
* bookedBy


RobBoothMOJ (approved):
Approved

PR #2447, opened 2024-11-05
Changed systemdefault to UTC in a unit test

daveawc:
The test was generating against systemDefault(), and then testing against UTC, which were different when running locally.


davidatkinsuk (approved):
Fun!

It's worth noting that integration tests (and deployments) use Europe/London but that doesn't really matter here as it's just a unit test

PR #2449, opened 2024-11-05
APS-1496 - Populate arrival and departure dates when migrating booking to space booking

davidatkinsuk:
This PR retrieves the arrival and departure date from delius when migrating bookings to space bookings, and saves them on the space booking, if defined.


RobBoothMOJ (approved):
Approved

PR #2453, opened 2024-11-06
Generate status for space booking

gregkhawkins:
**PR includes:**
- Generate all space booking statuses based on `expectedArrivalDate` / `expectedDepartureDate` / `actualArrivalDate` / `actualDepartureDate`
- The exception is the `notArrived` status which is driven by the `nonArrivalConfirmedAtDateTime`(i.e. if this is set then `notArrived`) - Spoke with @RobBoothMOJ on this
- For reference business-logic see the related ticket: https://dsdmoj.atlassian.net/browse/APS-1370

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas1/Cas1SpaceBookingStatusTransformer.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas1/Cas1SpaceBookingStatusTransformer.kt
@@ -1,0 +1,29 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.cas1
+
+import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Cas1SpaceBookingSummaryStatus
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.SpaceBookingDates
+import java.time.LocalDate
+
+@Component
+class Cas1SpaceBookingStatusTransformer {
+
+  fun transformToSpaceBookingSummaryStatus(
+    spaceBookingDates: SpaceBookingDates,
+  ): Cas1SpaceBookingSummaryStatus? {
+    val nowDate = LocalDate.now()
+    return when {
+      spaceBookingDates.isNotArrived() -> Cas1SpaceBookingSummaryStatus.notArrived
+      spaceBookingDates.hasDeparted() -> Cas1SpaceBookingSummaryStatus.departed
+      spaceBookingDates.isOverdueDeparture(nowDate) -> Cas1SpaceBookingSummaryStatus.overdueDeparture
+      spaceBookingDates.isDepartingToday(nowDate) -> Cas1SpaceBookingSummaryStatus.departingToday
+      spaceBookingDates.isDepartingWithin2Weeks(nowDate) -> Cas1SpaceBookingSummaryStatus.departingWithin2Weeks
+      spaceBookingDates.hasArrivedAndNotDeparted() -> Cas1SpaceBookingSummaryStatus.arrived
+      spaceBookingDates.isArrivalToday(nowDate) -> Cas1SpaceBookingSummaryStatus.arrivingToday
+      spaceBookingDates.isArrivalWithin2Weeks(nowDate) -> Cas1SpaceBookingSummaryStatus.arrivingWithin2Weeks
+      spaceBookingDates.isArrivalWithin6Weeks(nowDate) -> Cas1SpaceBookingSummaryStatus.arrivingWithin6Weeks
+      spaceBookingDates.isOverdueArrival(nowDate) -> Cas1SpaceBookingSummaryStatus.overdueArrival
+      else -> null
+    }
+  }
+}
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas1/Cas1SpaceBookingTransformer.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas1/Cas1SpaceBookingTransformer.kt
@@ -14,6 +14,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas1SpaceBook
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas1SpaceBookingSearchResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.PersonInfoResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.PersonSummaryInfoResult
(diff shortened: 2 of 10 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/util/SpaceBookingSummaryStatusTestHelper.kt
```diff
@@ -0,0 +1,181 @@
...
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Cas1SpaceBookingSummaryStatus
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.unit.transformer.cas1.TestCaseForSpaceBookingSummaryStatus
+import java.time.LocalDate
+import java.time.LocalDateTime
+import java.util.stream.Stream
+
+object SpaceBookingSummaryStatusTestHelper {
+  private var testCaseIndex = 0
+
+  fun spaceBookingSummaryStatusCases(): Stream<Arguments> {
```
gregkhawkins, line 13: - These test cases were first built (using TDD) to create the unit tests to drive the implementation. 
- The test cases are extensive and test the date boundaries fully (I hope!)
- It was moved here to be shared by the unit tests and integration tests
- I was mindful that using this for the integration tests might slow down the build too much as it creates 89 test cases in total
- I was also mindful that we discussed up front that we want the integration tests to be solid encase we move the business logic (to generate the statuses) to the sql query at some point (i.e. if we introduce filtering / sorting on status at some stage - though out of scope officially for now!)
- In the end, I decided to reuse this for both unit tests and integration tests and see how the build behaves. Here are the results:

1. The integration tests in the build before mine took `1m 55s`: 
before

2. The integration tests in my build (with the extra 89 tests took `2m 09s`: 
after

**Conclusion**: adding 14 secs to the build is not the end of the world and proves it works properly. 

*Happy to be contradicted on this conclusion if this is of concern and can slim down the integration tests to be 1 test case per status

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/Cas1SpaceBookingTest.kt
```diff
@@ -657,6 +681,51 @@ class Cas1SpaceBookingTest {
       assertThat(response[4].tier).isEqualTo("A")
       assertThat(response[4].person.crn).isEqualTo("CRN_CURRENT1")
     }
+
+    @Order(9999)
```
gregkhawkins, line 685: need to ensure that this test runs last due to data setup breaking other tests otherwise. Not sure if this is ideal and can investigate tearing down data at end of each test also. Wanted to avoid increasing pipeline run time further with teardown.
Not sure if this is the right solution though so please feed in!

src/main/resources/static/cas1-schemas.yml
```diff
@@ -375,6 +375,8 @@ components:
           type: string
         keyWorkerAllocation:
           $ref: '#/components/schemas/Cas1KeyWorkerAllocation'
+        status:
+          $ref: '#/components/schemas/Cas1SpaceBookingSummaryStatus'
```
gregkhawkins, line 379: as this is a new field not a breaking change for the API in my mind. That said, I know we do some magic in the UI tht auto-generates the typescript types -will this potentially cause issues on the FE side?

davidatkinsuk (approved):
This looks really good and i appreciate the clean comments explaining the approach you've taken. i've made some suggestions but i think all are optional other than removing use of canonical times

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/Cas1SpaceBookingTest.kt
gregkhawkins: this sounds good! I'll have a crack at it, thanks

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas1/Cas1SpaceBookingTransformer.kt
```diff
@@ -129,5 +130,12 @@ class Cas1SpaceBookingTransformer(
         ),
       )
     },
+    status = spaceBookingStatusTransformer.transformToSpaceBookingSummaryStatus(
+      scheduledArrivalDate = searchResult.canonicalArrivalDate,
+      scheduledDepartureDate = searchResult.canonicalDepartureDate,
```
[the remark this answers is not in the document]
gregkhawkins: ah! good catch. Will move over to the expected dates

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas1/Cas1SpaceBookingStatusTransformer.kt
```diff
@@ -0,0 +1,104 @@
...
+
+  private fun isNotArrived(nonArrivalConfirmedAt: LocalDateTime?): Boolean = nonArrivalConfirmedAt != null
+
+  private fun isDepartingWithin2Weeks(
+    nowDate: LocalDate,
+    actualArrivalDateTime: LocalDateTime?,
+    scheduledDepartureDate: LocalDate,
+    actualDepartureDateTime: LocalDateTime?,
+  ): Boolean {
+    val twoWeeksFromNow = LocalDate.now().plusDays(TWO_WEEKS_IN_DAYS)
```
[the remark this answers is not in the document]
gregkhawkins: yep, good suggestion. I was using plus days for the boundaries when creating the test data so got into this habit. Thanks :)

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/Cas1SpaceBookingTest.kt
```diff
@@ -657,6 +681,51 @@ class Cas1SpaceBookingTest {
...
+    @ParameterizedTest
+    @MethodSource("spaceBookingSummaryStatusCases")
+    fun `Search generates the correct statuses`(
+      testCaseForSpaceBookingSummaryStatus: TestCaseForSpaceBookingSummaryStatus,
+      expectedResultStatus: Cas1SpaceBookingSummaryStatus,
+      testCaseIndex: Int,
+    ) {
+      val (_, jwt) = `Given a User`(roles = listOf(CAS1_FUTURE_MANAGER))
+
+      val crnForStatusTest = "STATUS_TEST_BOOKING__$testCaseIndex"
```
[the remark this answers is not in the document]
gregkhawkins: I think this will lead to the filtered list of results getting more than 1 result (in the cases where we test the boundaries for a specific status. For example, we test the `arrivingIn2Weeks` status 14 times to cover the 14 days in the period where we service up this status.
Let me know if I'm missing something :)

davidatkinsuk (approved):
Looks great! re-approved :D

PR #2454, opened 2024-11-06
APS-1366: Add API support for Cas1 Space Booking non arrivals 

RobBoothMOJ:
Implement API to record non arrival for CAS1 Space Bookings : 

- Created API definition for non-arrival
- Implemented non-arrival function in Cas1SpaceBookingController
- Only users with CAS1_SPACE_BOOKING_RECORD_NON_ARRIVAL can call function
- Implemented the recording of non-arrival data in the Space Booking record
- Implemented the generation of a domain event to record the non-arrival
- Updated the Space Booking search query to categorise non-arrivals as historic bookings

The final requirement of APS-1366 is to remove the original non-arrival code, if CAS 3 isn't using it. 
This will be covered in a separate ticket.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas1SpaceBookingEntity.kt
```diff
@@ -44,9 +44,12 @@ interface Cas1SpaceBookingRepository : JpaRepository<Cas1SpaceBookingEntity, UUI
...
         cast(:residency as text) IS NULL OR (
-          (:residency = 'upcoming' AND b.actual_arrival_date_time IS NULL) OR
-          (:residency = 'current' AND b.actual_arrival_date_time IS NOT NULL and b.actual_departure_date_time IS NULL) OR
-          (:residency = 'historic' AND b.actual_departure_date_time IS NOT NULL)
+          (:residency = 'upcoming' AND b.actual_arrival_date_time IS NULL AND b.non_arrival_confirmed_at IS NULL) OR
+          (:residency = 'current' AND b.actual_arrival_date_time IS NOT NULL AND 
+            b.actual_departure_date_time IS NULL AND b.non_arrival_confirmed_at IS NULL) OR
+          (:residency = 'historic' AND 
+            (b.actual_departure_date_time IS NOT NULL OR b.non_arrival_confirmed_at IS NOT NULL)
+          )
```
davidatkinsuk, line 52: nice!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1SpaceBookingManagementDomainEventService.kt
```diff
@@ -98,6 +102,61 @@ class Cas1SpaceBookingManagementDomainEventService(
...
+              noms = offenderDetails?.nomsId ?: "Unknown NOMS Id",
+            ),
+            deliusEventNumber = application!!.eventNumber,
+            premises = Premises(
+              id = premises.id,
+              name = premises.name,
+              apCode = premises.apCode,
+              legacyApCode = premises.qCode,
+              localAuthorityAreaName = premises.localAuthorityArea!!.name,
+            ),
```
davidatkinsuk, line 148: not related to this change really, but we do this `ApprovedPremisesEntity` to `Premises` mapping 3 times in this class. Could we pull that out into a private function (or private extension function on the `ApprovedPremisesEntity`?)

davidatkinsuk (approved):
Perfect! just one suggestion from me that isn't really related to this change but a good opportunity to remove some duplicate code!

PR #2458, opened 2024-11-07
APS-1460: Removed expected departure date from Cas1 Arrival endpoint

RobBoothMOJ:
Removed expected departure date from CAS1 Arrival endpoint

Removed setting of previousExpectedDepartureOn but left the field in place, on the `PersonArrived` domain event.  
Discussion with David 7/5 we thought there would likely be value in this being captured and while API definitions are in a state of flux, worth keeping for the time being.


davidatkinsuk (approved):
LGTM 👍

PR #2462, opened 2024-11-07
APS-153 - Use legacy cas1 management data for seeding space bookings

davidatkinsuk:
This PR falls back to 'legacy' arrival and departure data captured on a `booking` when seeding an equivalent `space booking`. This will only be used if a corresponding record cannot be found in Delius.


RobBoothMOJ (approved):
Approved

PR #2464, opened 2024-11-07
CAS-1213 Remove Referral end point from CAS3 reports end points

muhammad-elabdulla:
This[ PR CAS-1213](https://dsdmoj.atlassian.net/browse/CAS-1213) is to remove `/reports/referrals` Api as is no longer in use and is replaced with `/reports/{reportName}`


davidatkinsuk (approved):
LGTM 👍

PR #2465, opened 2024-11-08
APS-1453 - Use transaction per premise when migrating space bookings

davidatkinsuk:
This commit allows seed jobs to manage their own transactions, and subsequently changes the ‘booking to space booking’ logic to migrate each premise in its own transaction. There’s concern that if trying to migrate a large number of premises in one execution this wll result in a very large transaction which may lead to resource issues during processing


gregkhawkins (approved):

gregkhawkins:
Guess if you concern becomes a reality we can wipe the space booking data and refactor and try again. Not live data so should be fine + prob be okay! 🤞

PR #2467, opened 2024-11-08
Make record-arrival endpoint idempotent

gregkhawkins:
Update `cas1SpaceBookingService.recordArrivalForBooking` to return a success outcome if the required `cas1NewArrival.arrivalDateTime` matches the value already set on `existingCas1SpaceBooking.actualArrivalDateTime`. If a value is already set and it doesn’t match the required value, return a conflict error as per the current behaviour.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1SpaceBookingService.kt
```diff
@@ -172,7 +174,11 @@ class Cas1SpaceBookingService(
     existingCas1SpaceBooking!!
 
     if (existingCas1SpaceBooking.actualArrivalDateTime != null) {
-      return existingCas1SpaceBooking.id hasConflictError "An arrival is already recorded for this Space Booking"
+      return if (cas1NewArrival.arrivalDateTime.toLocalDate() == LocalDate.ofInstant(existingCas1SpaceBooking.actualArrivalDateTime, ZoneOffset.UTC)) {
```
gregkhawkins, line 177: this is a bit of an assumption by me. Do we want to be strcter on the exact second of arrival. This matches on the arrival date only (i.e. the converted `LocalDate`)

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/Cas1SpaceBookingTest.kt
```diff
@@ -971,6 +972,93 @@ class Cas1SpaceBookingTest {
         .isOk
       domainEventAsserter.assertDomainEventOfTypeStored(spaceBooking.application!!.id, DomainEventType.APPROVED_PREMISES_PERSON_ARRIVED)
     }
+
+    @Test
+    fun `Recording arrival returns OK when arrival already registered and arrival date matches`() {
```
davidatkinsuk, line 977: up to you if you want to keep it, but i think it's probably overkill to have a unit and integration test as this  change hasn't introduced any database / SQL / integration changes? In this case i'd keep the unit test for the sake of execution speed.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/asserter/DomainEventAsserter.kt
```diff
@@ -25,6 +25,18 @@ class DomainEventAsserter(
     ).hasSize(expectedCount)
   }
 
+  fun assertHasZeroDomainEventsStoredOfEventType(applicationId: UUID, eventType: DomainEventType) =
+    assertDomainEventStoreCount(applicationId, eventType, expectedCount = 0)
```
davidatkinsuk, line 29: i can see why you're not reusing the existing assertDomainEventsOfTypeStored function because that's deliberately filtering out events with a space booking id. I'm going to create a ticket to remove that - i suspect it was a 'quick fix' to stop a bunch of tests failing but the behaviour is quite opaque

davidatkinsuk: i've created APS-1517

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1SpaceBookingService.kt
davidatkinsuk: The date and time is explicitly entered on the UI, so i think this should be a 'to the second' comparison (i.e. a straight comparison of Instants)

Note line 185 and use of `ZoneId`. We're actually using Europe/London as our default zone so we should use ZoneId.systemDefault() where required (although i guess that won't be required anymore in your changed code)

FYI OS timezone is being set in the Dockerfile and JVM is inheriting the OS defaults

davidatkinsuk (approved):
Looking good, just a few comments from me 👍

PR #2469, opened 2024-11-08
remove getStaffByStaffCode

daveawc:
This is the start of migration away from community api, as the ap and delius endpoint has been created to get staff by staff code.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
```diff
@@ -329,7 +329,7 @@ class UserService(
     }
 
     if (forService == ServiceName.approvedPremises) {
-      val apArea = cas1ApAreaMappingService.determineApArea(user.probationRegion, staffDetail.teamCodes(), staffDetail.username)
+      val apArea = cas1ApAreaMappingService.determineApArea(user.probationRegion, staffDetail.teamCodes(), staffDetail.username!!)
```
davidatkinsuk, line 332: It's a shame that username is now nullable to Accommodate the get by staff code API. We obviously know this isn't null when we're using the 'get by username' endpoint - can you think of anyway to 'bake this in' to the API client response or is it going to be too awkward? Maybe worth having a chat about this

it's also worth noting that we don't appear to use the username field when retrieving staff via code?

davidatkinsuk (approved):
A few questions from me but functionally looks fine!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/deliuscontext/StaffDetail.kt
```diff
@@ -8,7 +8,7 @@ data class StaffDetail(
   val staffIdentifier: Long,
   val teams: List<Team> = emptyList(),
   val probationArea: ProbationArea,
-  val username: String,
+  val username: String?,
```
muhammad-elabdulla, line 11: does the username have a null value in any case?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/client/ApDeliusContextApiClient.kt
```diff
@@ -61,6 +61,11 @@ class ApDeliusContextApiClient(
     path = "/staff/$username"
   }
 
+  @Cacheable(value = ["staffDetailsCache"], unless = IS_NOT_SUCCESSFUL)
```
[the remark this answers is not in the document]
davidatkinsuk: I think it depends on whether we think merging the classes was a good idea (this relates to my other comment). Regardless, i think using a cache is a good idea whatever we do

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
davidatkinsuk: Following on from our discussion:

* whilst staff legitimately may not have a username defined, when retrieving staff details via `getStaffDetail(username:String)` we know that the returned user (if any) _must_ have a username
* our first idea was to remove the username? field from the API response to force the calling code to use whatever username was passed into the client function, but this field was used in API responses, so could break something in the UI
* the agreed approach was to instead avoid using staffDetail.username in calling code, and instead use the non-null username value used when calling `getStaffDetail(username:String)`. This may require adding it as a function argument whereever the result of this API call is used.

davidatkinsuk: looks good!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/client/ApDeliusContextApiClient.kt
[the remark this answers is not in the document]
davidatkinsuk: have agreed to remove this for now and monitor performance

PR #2471, opened 2024-11-11
APS_1475 upgrade gov notify client to 5.2.1

danhumphreys-moj:
Ref: https://dsdmoj.atlassian.net/browse/APS-1475

The gov notify email client is currently at version 5.2.1.


davidatkinsuk (approved):
LGTM 👍

PR #2474, opened 2024-11-11
Removed STAFF-DETAIL-BY-STAFF-CODE-ENABLED feature flag.

daveawc:
This isn't currently used, so there is no way to test in pre-prod. This will be done when the UI is implemented. It also removes the references from the community api context, and all of the now redundant and unused code


davidatkinsuk (approved):
LGTM 👍

PR #2475, opened 2024-11-11
CAS-1203 Change wheelchair accessible scope for CAS3

muhammad-elabdulla:
This [PR CAS-1203](https://dsdmoj.atlassian.net/browse/CAS-1203) is to change wheelchair accessible scope to be room level instead of premises and room


davidatkinsuk (approved):
LGTM 👍

PR #2481, opened 2024-11-12
Make record departure endpoint idempotent

gregkhawkins:
Update `cas1SpaceBookingService.recordDepartureForBooking` to return a success outcome if the same required values in `Cas1NewDeparture` the are already set on `existingCas1SpaceBooking`. If different values are already set, return a conflict error as per the current behaviour when `existingCas1SpaceBooking.actualDepartureDateTime` != null


davidatkinsuk (approved):
LGTM 👍

PR #2482, opened 2024-11-12
APS-1535 - Revise criteria considered for planning

davidatkinsuk:
This commit removes isArsonDesignated and isGroundFloor from consideration when planning space booking. isArsonDesignated is legacy criteria, and isGroundFloor may be re-added in future iterations as required.


RobBoothMOJ (approved):
Approved

PR #2483, opened 2024-11-12
APS-1300: Update Application Status when withdrawing Space Booking

RobBoothMOJ:
- Move CAS1 booking functionality, when withdrawing a booking from BookingService to `Cas1ApplicationStatusService`.
- Add functionality in `Cas1SpaceBookingService`  -`withdraw` to update the application status for the withdrawn booking.
- Implement the processing to change the application status for a withdrawn / cancelled booking in `Cas1ApplicationSatusService`.

Consideration to be given to whether a space booking being withdrawn as part of 
`Cas1WithdrawableTreeOperations` should instigate a change to the associated application’s status.

A failing test : 
`Withdrawing a match request for original app dates cascades to applicable booking entity and updates the application status`
demonstrates the withdrawal of a Space Booking when traversing the associated entity tree for a withdrawn placement request.   
The initial behaviour saw the application status change to `AWAITING_PLACEMENT`, because of the cancelled Space Booking, rather than the expected `PENDING_PLACEMENT_REQUEST` because both the placement request and the space booking had been cancelled (the space booking last).

An additional, optional parameter on the `Cas1ApplicationStatusService.spaceBookingCancelled `function, `isWithdrawableTreeOperation`, can be employed to leave the application status unchanged.  
`Cas1SpaceBookingService.withdraw`, which sets the value `isWithdrawableTreeOperation` may need refined and the name of the parameter itself changed to something more suitable.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1ApplicationStatusService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1ApplicationStatusService.kt
@@ -4,12 +4,16 @@ import org.springframework.stereotype.Service
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApplicationRepository
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApprovedPremisesApplicationEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.BookingEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.BookingRepository
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas1SpaceBookingEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas1SpaceBookingRepository
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.ApprovedPremisesApplicationStatus
 
 @Service
 class Cas1ApplicationStatusService(
   val applicationRepository: ApplicationRepository,
+  val cas1SpaceBookingRepository: Cas1SpaceBookingRepository,
+  val bookingRepository: BookingRepository,
 ) {
 
   fun bookingMade(booking: BookingEntity) {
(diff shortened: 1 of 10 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1ApplicationStatusService.kt
```diff
@@ -23,6 +27,38 @@ class Cas1ApplicationStatusService(
...
+    if (isWithdrawableTreeOperation || spaceBooking.application == null) {
+      return
+    }
+    val application = spaceBooking.application!!
+    val spaceBookingsForApplication = cas1SpaceBookingRepository.findAllByApplication(application)
+    val anyActiveBookings = spaceBookingsForApplication.any { it.isActive() }
+    if (!anyActiveBookings) {
+      bookingCancelled(spaceBooking.application!!)
+    }
+  }
```
davidatkinsuk, line 55: i _think_ we may have to return to this once we have a way to determine which premise works with space bookings and which doesn't because it's possible for an application to have a mix of bookings across regions (some supporting space bookings, some not). I'll make a note on that ticket

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1ApplicationStatusService.kt
```diff
@@ -23,6 +27,38 @@ class Cas1ApplicationStatusService(
...
+
+  private fun bookingCancelled(application: ApprovedPremisesApplicationEntity) {
```
davidatkinsuk, line 57: could maybe rename this `lastBookingCancelled`?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/WithdrawalTest.kt
```diff
@@ -914,7 +914,7 @@ class WithdrawalTest : IntegrationTestBase() {
 
             withdrawApplication(application, jwt)
 
-            assertApplicationWithdrawn(application)
+            assertApplicationStatus(application, ApprovedPremisesApplicationStatus.AWAITING_PLACEMENT)
```
davidatkinsuk, line 917: this should be as before, with a status of WITHDRAWN

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1SpaceBookingService.kt
```diff
@@ -438,6 +438,10 @@ class Cas1SpaceBookingService(
       else -> throw InternalServerErrorProblem("Withdrawal triggered automatically is not supported")
     }
     cas1BookingDomainEventService.spaceBookingCancelled(spaceBooking, user, reason)
+    cas1ApplicationStatusService.spaceBookingCancelled(
+      spaceBooking,
+      isWithdrawableTreeOperation = withdrawalContext.triggeringEntityType == WithdrawableEntityType.PlacementRequest,
```
davidatkinsuk, line 443: I don't think it's the correct answer, but doing something similar to BookingService would at least maintain the existing behaviour? e.g.

```
isUserRequestedWithdrawal = withdrawalContext.triggeringEntityType == WithdrawableEntityType.SpaceBooking,
```

And then ignore the change if isUserRequestedWithdrawal?

PR #2485, opened 2024-11-13
APS-1543: Added non-arrival information to API returning a space booking

RobBoothMOJ:
A previous PR / piece of work captured Non arrival information for a space booking. 

This PR provides the way to retrieve that captured information.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas1/Cas1SpaceBookingTransformer.kt
```diff
@@ -112,6 +114,25 @@ class Cas1SpaceBookingTransformer(
     }
   }
 
+  private fun Cas1SpaceBookingEntity.extractNonArrival(): Cas1SpaceBookingNonArrival? {
+    val confirmedAt = nonArrivalConfirmedAt
+    val reason = nonArrivalReason
+    return if (confirmedAt != null && reason != null) {
```
davidatkinsuk, line 120: We have some simple functions on the Cas1SpaceBookingEntity to help determine a booking state:

```
  fun isActive() = !isCancelled()
  fun isCancelled() = cancellationOccurredAt != null
  fun hasArrival() = actualArrivalDateTime != null
```

Could we please add one for hasNonArrival() and use that here instead? I think using confirmedAt to determine this is sufficient?

(and yes we do need to update some existing code to use these :) )

davidatkinsuk (approved):
Looks good, just one minor not from me

PR #2489, opened 2024-11-13
CAS-486 Add validation to check if applications or application summary are empty

muhammad-elabdulla:
This PR is to not call delius endpoint if the applications or summary applications are empty


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ApplicationService.kt
```diff
@@ -123,10 +123,14 @@ class ApplicationService(
...
 
-    val crns = applicationSummaries.map { it.getCrn() }.distinct()
-    val offendersAccess = offenderService.canAccessOffenders(userDistinguishedName, crns)
-    return applicationSummaries.filter {
-      offendersAccess.containsKey(it.getCrn()) && offendersAccess[it.getCrn()] == true
+    if (applicationSummaries.isEmpty()) {
+      return applicationSummaries
+    } else {
+      val crns = applicationSummaries.map { it.getCrn() }.distinct()
+      val offendersAccess = offenderService.canAccessOffenders(userDistinguishedName, crns)
```
davidatkinsuk, line 130: I would prefer that we were more defensive in the canAccessOffenders function instead e.g.

```
  fun canAccessOffenders(username: String, crns: List<String>): Map<String, Boolean> {
    if(crns.isEmpty()) { return emptyMap() }
    ...
  }
```

that way dev's don't need to remember that empty lists won't be allowed when calling this function, making bugs less likely. i think we can then remove the current changes in this PR because it doesn't really add much overhead dealing with an empty list being returned? Also it makes adding unit tests simpler

davidatkinsuk, line 130: can we add unit tests please

davidatkinsuk (changes requested):

davidatkinsuk (approved):
perfect! 👍

PR #2490, opened 2024-11-13
APS-1549 - Use booking create permission

davidatkinsuk:


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/PlacementRequestsTest.kt
```diff
@@ -1692,8 +1692,8 @@ class PlacementRequestsTest : IntegrationTestBase() {
     }
 
     @Test
-    fun `Create a Booking from a PR allocated to other user and current user is a Workflow Manager returns a 200`() {
-      givenAUser(roles = listOf(UserRole.CAS1_WORKFLOW_MANAGER)) { user, jwt ->
+    fun `Create a Booking from a PR allocated to other user and current user is a CRU Member returns a 200`() {
+      givenAUser(roles = listOf(UserRole.CAS1_CRU_MEMBER)) { user, jwt ->
```
gregkhawkins, line 1696: can see that you are refactoring the pre-existing test to test the equivalent functionality based on `CAS1_CRU_MEMBER` instead of `CAS1_WORKFLOW_MANAGER` so all good. Just wondering if we test anywhere a user without this perm which results in a 403?

gregkhawkins (approved):

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/PlacementRequestsTest.kt
davidatkinsuk: Good spot, there was one that needed updating. I've done this to be exhaustive wrt. testing all roles. I've also updated this test to check all roles which have the correct permission

PR #2493, opened 2024-11-13
APS-1254: Update Space Booking Departure API , moveOnCategory no longer mandatory

RobBoothMOJ:
Changed the Space Booking Departure API 

- moveOnCategory no longer mandatory 
- when no moveOnCategory is supplied the system defaults to the 'Not Applicable' move on category to ensure ongoing compatibility with Probation Integration for which this field is mandatory.


davidatkinsuk (approved):
LGTM 👍

PR #2496, opened 2024-11-14
Change CAS3 bedspace search to use property name characteristics

muhammad-elabdulla:
Change the CAS3 bedspace search to use property name characteristics column. when filter bedspaces by attributes


davidatkinsuk (approved):
LGTM 👍

PR #2499, opened 2024-11-14
APS-1536 Add new seed from Excel job.

danhumphreys-moj:
We want the ability to be able to update approved premises room characteristic data from an Excel spreadsheet.  This commit adds the functionality to run seed jobs from xlsx files and populated rooms, beds and characteristics.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/SeedFromExcelController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/SeedFromExcelController.kt
@@ -1,0 +1,19 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.controller
+
+import org.springframework.http.HttpStatus
+import org.springframework.http.ResponseEntity
+import org.springframework.stereotype.Service
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.SeedFromExcelApiDelegate
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.SeedFromExcelRequest
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.seed.SeedService
+
+@Service
+class SeedFromExcelController(private val seedService: SeedService) : SeedFromExcelApiDelegate {
+  override fun seedFromExcelPost(seedFromExcelRequest: SeedFromExcelRequest): ResponseEntity<Unit> {
+    throwIfNotLoopbackRequest()
+
+    seedService.seedExcelData(seedFromExcelRequest.seedType, seedFromExcelRequest.premisesId, seedFromExcelRequest.fileName)
+
+    return ResponseEntity(HttpStatus.ACCEPTED)
+  }
+}
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/SeedJob.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/SeedJob.kt
@@ -1,5 +1,8 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.seed
 
+import org.jetbrains.kotlinx.dataframe.DataFrame
+import java.util.UUID
+
 abstract class SeedJob<RowType>(
   val requiredHeaders: Set<String>? = null,
   val runInTransaction: Boolean = true,
@@ -22,6 +25,11 @@ abstract class SeedJob<RowType>(
   }
   abstract fun deserializeRow(columns: Map<String, String>): RowType
   abstract fun processRow(row: RowType)
+}
+
+@Suppress("TooGenericExceptionThrown")
+abstract class ExcelSeedJob {
+  abstract fun processDataFrame(dataFrame: DataFrame<*>, premisesId: UUID)
 }
 
 class SeedException(message: String) : RuntimeException(message)
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
@@ -1,0 +1,186 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.seed.cas1
+
+import org.jetbrains.kotlinx.dataframe.DataFrame
+import org.jetbrains.kotlinx.dataframe.api.getColumn
+import org.jetbrains.kotlinx.dataframe.name
+import org.slf4j.LoggerFactory
+import org.springframework.data.repository.findByIdOrNull
+import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ServiceName
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApprovedPremisesRepository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.BedEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.BedRepository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.CharacteristicEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.CharacteristicRepository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.PremisesEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.RoomEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.RoomRepository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.seed.ExcelSeedJob
+import java.util.UUID
+
+class SiteSurveyImportException(message: String) : Exception(message)
+
+@Component
+@Suppress("LongParameterList")
+class ApprovedPremisesRoomsSeedFromXLSXJob(
+  private val approvedPremisesRepository: ApprovedPremisesRepository,
+  private val roomRepository: RoomRepository,
+  private val bedRepository: BedRepository,
+  private val siteSurvey: SiteSurvey,
+) : ExcelSeedJob() {
+  private val log = LoggerFactory.getLogger(this::class.java)
+  override fun processDataFrame(roomsWorksheet: DataFrame<*>, premisesId: UUID) {
+    findExistingPremisesOrThrow(premisesId)
+
+    var rooms = buildRooms(roomsWorksheet, premisesId)
+    val characteristics = buildCharacteristics(roomsWorksheet)
+
+    rooms.forEach {
+      characteristics[it.code]?.let { roomCharacteristics -> it.characteristics.addAll(roomCharacteristics.toList()) }
+    }
+
+    rooms = createOrUpdateRooms(rooms)
+
+    val beds = buildBeds(roomsWorksheet, rooms)
+    createBedsIfNotExist(beds)
+  }
+
+  private fun buildRooms(dataFrame: DataFrame<*>, premisesId: UUID): MutableList<RoomEntity> {
+    val rooms = mutableListOf<RoomEntity>()
+
+    for (i in 1..<dataFrame.columnsCount()) {
+      val roomAnswers = dataFrame.getColumn(i)
+      val room = buildRoom(premisesId, roomCode = roomAnswers.name, roomName = roomAnswers[0].toString())
+
+      rooms.add(room)
+    }
+    return rooms
+  }
+
+  private fun buildCharacteristics(dataFrame: DataFrame<*>): MutableMap<String, MutableList<CharacteristicEntity>> {
+    var premisesCharacteristics = mutableMapOf<String, MutableList<CharacteristicEntity>>()
+
+    siteSurvey.questionToCharacterEntityMapping.forEach { (question, characteristic) ->
+      val rowId = dataFrame.getColumn(0).values().indexOf(question)
+
+      if (rowId == -1) throw SiteSurveyImportException("Characteristic question '$question' not found on sheet Sheet3.")
+
+      for (colId in 1..<dataFrame.columnsCount()) {
+        val roomCode = dataFrame.getColumn(colId).name
+        val answer = dataFrame[rowId][colId].toString().trim()
+
+        if (answer.equals("yes", ignoreCase = true)) {
+          premisesCharacteristics.computeIfAbsent(roomCode) { mutableListOf() }.add(characteristic!!)
+        } else if (!answer.equals("no", ignoreCase = true) && !answer.equals("N/A", ignoreCase = true)) {
+          throw SiteSurveyImportException("Expecting 'yes' or 'no' for question '$question' but is '$answer' on sheet Sheet3 (row = ${rowId + 1}, col = $colId).")
+        }
+      }
+    }
+    return premisesCharacteristics
+  }
+
+  private fun buildBeds(dataFrame: DataFrame<*>, rooms: MutableList<RoomEntity>): List<BedEntity> {
+    val beds = mutableListOf<BedEntity>()
+    for (i in 1..<dataFrame.columnsCount()) {
+      val roomAnswers = dataFrame.getColumn(i)
+      val bedNumber = roomAnswers[1].toString()
+      beds.add(
+        buildBed(
+          bedName = "${roomAnswers.name}",
+          bedCode = "${roomAnswers.name} - $bedNumber",
+          room = rooms.firstOrNull { it.code == roomAnswers.name }
+            ?: throw IllegalArgumentException("Room not found with id ${roomAnswers.name} and code ${roomAnswers.name}."),
+        ),
+      )
+    }
+    return beds
+  }
+
+  private fun createOrUpdateRooms(rooms: MutableList<RoomEntity>): MutableList<RoomEntity> {
+    rooms.forEachIndexed { index, room ->
+      var persistedRoom = roomRepository.findByCodeAndPremisesId(room.code!!, room.premises.id)
+
+      if (persistedRoom == null) {
+        roomRepository.save(room)
+        log.info("Created new room ${room.id} with code ${room.code} and name ${room.name} in premise ${room.premises.name}.")
+        log.info("Added characteristic(s) ${room.characteristics.joinToString()} to room code ${room.code}.")
+      } else {
+        rooms[index] = persistedRoom
+        persistedRoom.characteristics.clear()
+        persistedRoom.characteristics.addAll(room.characteristics)
+        roomRepository.save(persistedRoom)
+        log.info("Added characteristic(s) ${persistedRoom.characteristics.joinToString()} to room code ${persistedRoom.code}.")
+      }
+    }
+    return rooms
+  }
+
+  private fun createBedsIfNotExist(beds: List<BedEntity>) {
+    beds.forEach {
+      val existingBed = bedRepository.findByCodeAndRoomId(it.code!!, it.room.id)
+      if (existingBed != null) {
+        log.info("Bed ${it.id} with code ${it.code} already exists in room code ${it.room.code}.")
+      } else {
+        bedRepository.save(it)
+        log.info("Created new bed ${it.id} with code ${it.code} and name ${it.name} in room code ${it.room.code}.")
+      }
+    }
+  }
+
+  private fun buildRoom(premisesId: UUID, roomCode: String, roomName: String): RoomEntity = RoomEntity(
+    id = UUID.randomUUID(),
+    name = roomName,
+    code = roomCode,
+    notes = null,
+    premises = approvedPremisesRepository.findByIdOrNull(premisesId)!!,
+    beds = mutableListOf(),
+    characteristics = mutableListOf(),
+  )
+
+  fun buildBed(bedName: String, bedCode: String, room: RoomEntity): BedEntity = BedEntity(
+    id = UUID.randomUUID(),
+    name = bedName,
+    code = bedCode,
+    room = room,
+    endDate = null,
+    createdAt = null,
+  )
+
+  private fun findExistingPremisesOrThrow(premisesId: UUID): PremisesEntity {
+    return approvedPremisesRepository.findByIdOrNull(premisesId) ?: throw SiteSurveyImportException(
+      "No premises with id '$premisesId' found.",
+    )
+  }
+}
+
+@Component
+class SiteSurvey(characteristicRepository: CharacteristicRepository) {
+  private val questionToPropertyNameMapping = mapOf(
+    "Is this bed in a single room?" to "isSingle",
+    "Is this room located on the ground floor?" to "isGroundFloor",
+    "Is the room using only furnishings and bedding supplied by FM?" to "isFullyFm",
+    "Does this room have Crib7 rated bedding?" to "hasCrib7Bedding",
+    "Is there a smoke/heat detector in the room?" to "hasSmokeDetector",
+    "Is this room on the top floor with at least one external wall and not located directly next to a fire exit or a protected stairway?" to "isTopFloorVulnerable",
+    "Is the room close to the admin/staff office on the ground floor with at least one external wall and not located directly next to a fire exit or a protected stairway?"
+      to "isGroundFloorNrOffice",
+    "is there a water mist extinguisher in close proximity to this room?" to "hasNearbySprinkler",
+    "Is this room suitable for people who pose an arson risk? (Must answer yes to Q; 6 & 7, and 9 or  10)" to "isArsonSuitable",
+    "Is this room currently a designated arson room?" to "isArsonDesignated",
+    "If IAP - Is there any insurance conditions that prevent a person with arson convictions being placed?" to "hasArsonInsuranceConditions",
+    "Is this room suitable for people convicted of sexual offences?" to "isSuitedForSexOffenders",
+    "Does this room have en-suite bathroom facilities?" to "hasEnSuite",
+    "Are corridors leading to this room of sufficient width to accommodate a wheelchair? (at least 1.2m wide)" to "isWheelchairAccessible",
+    "Is the door to this room at least 900mm wide?" to "hasWideDoor",
+    "Is there step free access to this room and in corridors leading to this room?" to "hasStepFreeAccess",
+    "Are there fixed mobility aids in this room?" to "hasFixedMobilityAids",
+    "Does this room have at least a 1500mmx1500mm turning space?" to "hasTurningSpace",
+    "Is there provision for people to call for assistance from this room?" to "hasCallForAssistance",
+    "Can this room be designated as suitable for wheelchair users?   Must answer yes to Q23-26 on previous sheet and Q17-19 & 21 on this sheet)" to "isWheelchairDesignated",
+    "Can this room be designated as suitable for people requiring step free access? (Must answer yes to Q23 and 25 on previous sheet and Q19 on this sheet)" to "isStepFreeDesignated",
+  )
+
+  val questionToCharacterEntityMapping = questionToPropertyNameMapping.map { (key, value) ->
+    Pair(key, characteristicRepository.findByPropertyName(value, ServiceName.approvedPremises.value))
+  }.toMap()
+}
--- a/src/main/resources/static/_shared.yml
+++ b/src/main/resources/static/_shared.yml
@@ -3688,6 +3688,20 @@ components:
       required:
         - seedType
         - fileName
+    SeedFromExcelRequest:
+      type: object
+      properties:
+        seedType:
+          $ref: '#/components/schemas/SeedFromExcelFileType'
+        premisesId:
+          type: string
+          format: uuid
+        fileName:
+          type: string
+      required:
+        - seedType
+        - premisesId
+        - fileName
     SeedFileType:
       type: string
       enum:
@@ -3719,6 +3733,10 @@ components:
         - approved_premises_space_planning_dry_run
         - approved_premises_import_delius_booking_management_data
         - approved_premises_update_space_booking
+    SeedFromExcelFileType:
+      type: string
+      enum:
+        - approved_premises_room
     MigrationJobRequest:
       type: object
       properties:
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/seed/SeedFromXLSXApprovedPremisesRoomsTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/seed/SeedFromXLSXApprovedPremisesRoomsTest.kt
@@ -1,0 +1,436 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.integration.cas1.seed
+
+import org.assertj.core.api.Assertions.assertThat
+import org.jetbrains.kotlinx.dataframe.api.dataFrameOf
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.TestInstance
+import org.springframework.beans.factory.annotation.Autowired
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.SeedFromExcelFileType
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.integration.seed.SeedTestBase
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.seed.cas1.SiteSurvey
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.seed.cas1.SiteSurveyImportException
+import java.util.UUID
+
+@TestInstance(TestInstance.Lifecycle.PER_METHOD)
+class SeedFromXLSXApprovedPremisesRoomsTest : SeedTestBase() {
+
+  @Autowired
+  lateinit var siteSurvey: SiteSurvey
+
+  fun MutableList<String>.addCharacteristics(numberOfRooms: Int = 1, activeCharacteristics: Map<String, List<Int>> = emptyMap()) {
+    siteSurvey.questionToCharacterEntityMapping.keys.forEach { question ->
+      this.add(question)
+      val answers = MutableList(numberOfRooms) { "No" }
+      activeCharacteristics[question]?.forEach {
+        answers[it] = "Yes"
+      }
+      this.addAll(answers)
+    }
+  }
+
+  @Test
+  fun `Creating a new room and a new bed with a characteristic succeeds`() {
+    val localAuthorityArea = localAuthorityEntityFactory.produceAndPersist()
+    val probationRegion = probationRegionEntityFactory.produceAndPersist()
+    val premises = approvedPremisesEntityFactory.produceAndPersist {
+      withLocalAuthorityArea(localAuthorityArea)
+      withProbationRegion(probationRegion)
+    }
+    val premisesId = premises.id
+
+    val header = listOf("Unique Reference Number for Bed", "SWABI01NEW")
+    val rows = mutableListOf(
+      "Room Number / Name",
+      "1",
+      "Bed Number (in this room i.e if this is a single room insert 1.  If this is a shared room separate entries will need to be made for bed 1 and bed 2)",
+      "1",
+    )
+    rows.addCharacteristics(1, mapOf("Is this room located on the ground floor?" to listOf(0)))
+
+    val dataFrame = dataFrameOf(header, rows)
+
+    withXlsx("example", "Sheet3", dataFrame)
+
+    seedService.seedExcelData(
+      SeedFromExcelFileType.approvedPremisesRoom,
+      premisesId,
+      "example.xlsx",
+    )
+
+    val newRoom = roomRepository.findByCode("SWABI01NEW")
+    assertThat(newRoom!!.characteristics).anyMatch {
+      it.name == "Is this room located on the ground floor?" &&
+        it.propertyName == "isGroundFloor"
+    }
+
+    val newBed = bedRepository.findByCodeAndRoomId("SWABI01NEW - 1", newRoom.id)
+    assertThat(newBed!!.name).isEqualTo("SWABI01NEW")
+    assertThat(
+      newBed.room.id == newRoom.id &&
+        newBed.room.code == "SWABI01NEW",
+    )
+  }
+
+  @Test
+  fun `Creating three new rooms and new beds with a characteristic succeeds`() {
+    val localAuthorityArea = localAuthorityEntityFactory.produceAndPersist()
+    val probationRegion = probationRegionEntityFactory.produceAndPersist()
+    val premises = approvedPremisesEntityFactory.produceAndPersist {
+      withLocalAuthorityArea(localAuthorityArea)
+      withProbationRegion(probationRegion)
+    }
+    val premisesId = premises.id
+
+    val header = listOf("Unique Reference Number for Bed", "SWABI01NEW", "SWABI02NEW", "SWABI03NEW")
+    val rows = mutableListOf(
+      "Room Number / Name",
+      "1",
+      "2",
+      "3",
+      "Bed Number (in this room i.e if this is a single room insert 1.  If this is a shared room separate entries will need to be made for bed 1 and bed 2)",
+      "1",
+      "1",
+      "1",
+    )
+    rows.addCharacteristics(3, mapOf("Is this room located on the ground floor?" to listOf(1)))
+
+    val dataFrame = dataFrameOf(header, rows)
+
+    withXlsx("example", "Sheet3", dataFrame)
+
+    seedService.seedExcelData(
+      SeedFromExcelFileType.approvedPremisesRoom,
+      premisesId,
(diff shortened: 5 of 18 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
```diff
@@ -0,0 +1,166 @@
...
+
+      val characteristics = mutableListOf<CharacteristicEntity>()
+
+      questions.drop(2).forEachIndexed { i, question ->
+        val characteristic = siteSurvey.questionToCharacterEntityMapping.getValue(question.toString())!!
+        if (roomCharacteristics[i + 2].toString().equals("yes", true)) {
+          characteristics.add(characteristic)
+          log.info("Added characteristic ${characteristic.propertyName} to room code $roomCode.")
+        } else {
+          if (characteristics.remove(characteristic)) {
```
danhumphreys-moj, line 69: Thinking to delete all characteristics and then re-create from the Excel file.  Would need integration to check that we rollback if there are any problems.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
```diff
@@ -0,0 +1,166 @@
...
+
+      val bed = room?.beds?.find { it.code == bedCode } ?: createBed(
+        bedName,
+        bedCode,
+        room!!,
+      )
+
+      val characteristics = mutableListOf<CharacteristicEntity>()
+
+      questions.drop(2).forEachIndexed { i, question ->
```
danhumphreys-moj, line 63: Spent some time looking at whether we can go straight from dataframe to list of objects but not easy with the layout of the site survey data as it stands.  It would require characteristics as column headers and room ids as row indices.  So a transposition of the existing layout (room ids as column headers and characteristics as row indices).

More straightforward to loop through the columns (for now).

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
```diff
@@ -0,0 +1,166 @@
...
+        val characteristic = siteSurvey.questionToCharacterEntityMapping.getValue(question.toString())!!
+        if (roomCharacteristics[i + 2].toString().equals("yes", true)) {
```
danhumphreys-moj, line 65: First two rows in the Excel spreadsheet are room and bed ids so need to be dropped when comparing to characteristic data. Will add a comment in the code.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
```diff
@@ -0,0 +1,166 @@
...
+    return premisesRepository.findByIdOrNull(premisesId) ?: throw RuntimeException(
+      "Error: no premises with id '$premisesId' found. ",
+    )
+  }
+}
+
+@Component
+class SiteSurvey constructor(characteristicRepository: CharacteristicRepository) {
+  private val questionToPropertyNameMapping = mapOf(
+//    "Is this bed in a single room?" to "isSingle",
```
danhumphreys-moj, line 120: Will uncomment when we get to running for all characteristics.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
```diff
@@ -0,0 +1,166 @@
...
+//    "Is there step free access to this room and in corridors leading to this room?" to "hasStepFreeAccess",
+//    "Are there fixed mobility aids in this room?" to "hasFixedMobilityAids",
+//    "Does this room have at least a 1500mmx1500mm turning space?" to "hasTurningSpace",
+//    "Is there provision for people to call for assistance from this room?" to "hasCallForAssistance",
+//    "Can this room be designated as suitable for wheelchair users?   Must answer yes to Q23-26 on previous sheet and Q17-21 on this sheet)" to "isWheelchairDesignated",
+//    "Can this room be designated as suitable for people requiring step free access? (Must answer yes to Q23 and 25 on previous sheet and Q19 on this sheet)" to "isStepFreeDesignated",
+    "Is this room located on the ground floor?" to "IsGroundFloor",
+  )
+
+  val questionToCharacterEntityMapping = questionToPropertyNameMapping.map { (key, value) ->
```
danhumphreys-moj, line 163: Originally this class was an object with a static property but seemed more appropriate to have this mapping and since this required using characteristicRepository used @Component and getBean() which seems to work.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/seed/SeedFromXLSXApprovedPremisesRoomsTest.kt
```diff
@@ -0,0 +1,144 @@
...
+          it.throwable.message == "Unable to process XLSX file"
+      }
+  }
+
+  @Test
+  fun `Creating a new room for a premise that doesn't exist throws error`() {
+  }
+
+  @Test
+  fun `Removing a characteristic that exists succeeds`() {
```
danhumphreys-moj, line 134: Redundant if we remove all characteristics to begin with.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
```diff
@@ -0,0 +1,166 @@
...
+
+      val room = roomRepository.findByCode(roomCode) ?: createRoom(
+        roomCode,
+        roomName,
+      )
+
+      val bedCode = roomCharacteristics[1].toString()
+      val bedName = "$roomName - $bedCode"
+
+      val bed = room?.beds?.find { it.code == bedCode } ?: createBed(
```
danhumphreys-moj, line 55: Beds aren't actually used during this job, however I have presumed that if they are in the spreadsheet and not in the database then we need to create them as part of this import task.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/SeedJob.kt
```diff
@@ -27,4 +28,18 @@ abstract class SeedJob<RowType>(
...
+  val premisesId: UUID,
+  val sheetName: String,
+) {
+  init {
+    if (fileName.contains("/") || fileName.contains("\\")) {
+      throw RuntimeException("Filename must be just the filename of a .xlsx file in the /seed directory, e.g. for /seed/upload.xlsx, just `upload` should be supplied")
+    }
+  }
+
+  abstract fun processDataFrame(dataFrame: DataFrame<*>)
```
davidatkinsuk, line 42: Again, one for once everything is working but i think seed jobs should be spring managed beans and whatever arguments are required should be passed into the function doing the processing. I'm slowly working on changing the existing seed jobs to use this pattern

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
```diff
@@ -0,0 +1,166 @@
...
+  fileName = fileName,
+  premisesId = premisesId,
+  sheetName = sheetName,
+) {
+  private val log = LoggerFactory.getLogger(this::class.java)
+  override fun processDataFrame(dataFrame: DataFrame<*>) {
+    findExistingPremisesOrThrow(premisesId)
+
+    val questions = dataFrame.getColumn(0).values().toList()
+    if (questions.drop(2) != siteSurvey.questionToCharacterEntityMapping.keys.toList()) throw RuntimeException("Site survey file $fileName contains wrong questions on sheet $sheetName")
```
davidatkinsuk, line 40: Do we actually care if the number of questions mismatch, as long as we can find the rows we're interested in? It's easy in excel to add additional 'blank' rows and i think this check will fail in that case?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
davidatkinsuk: agreed

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
```diff
@@ -0,0 +1,166 @@
...
+      val roomCode = roomCharacteristics.name
+      val roomName = roomCharacteristics[0].toString()
+
+      val room = roomRepository.findByCode(roomCode) ?: createRoom(
+        roomCode,
+        roomName,
+      )
+
+      val bedCode = roomCharacteristics[1].toString()
+      val bedName = "$roomName - $bedCode"
```
davidatkinsuk, line 53: I think we should be using the value of the 'Room Number / Name' row for this. I _think_ this is the first time premises have been able to define their own names

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
```diff
@@ -0,0 +1,166 @@
...
+
+  private fun findExistingPremisesOrThrow(premisesId: UUID): PremisesEntity {
+    return premisesRepository.findByIdOrNull(premisesId) ?: throw RuntimeException(
+      "Error: no premises with id '$premisesId' found. ",
+    )
+  }
+}
+
+@Component
+class SiteSurvey constructor(characteristicRepository: CharacteristicRepository) {
```
davidatkinsuk, line 118: i don't _think_ you need the constructor keyword here?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
davidatkinsuk, line 163: I think this should be probably be a function so we resolve the characteristics everytime they are required (e.g. get hold of this list once in the seed job when processing the excel). These things rarely change but they are changed directly in the DB this list will be stale

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
davidatkinsuk: I think it may be cleaner to deal with the spreadsheet in two stages

Stage 1 - load the data into an internal model (use data classes). Do validation whilst loading. The model should use booleans instead of "Yes/No" strings
Stage 2 - use the internal model to update premise/room/bed

That way all of the awkward use of indexes is used in one place and should need less documentation?

davidatkinsuk, line 65: I think we need should trim and lower case any cell we read before comparisons as humans are likely to add junk in the fields. I also wonder if we should error if something other than "yes" or "no" is in the cell e.g. if someone added "sometimes" :)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
davidatkinsuk: Yep makes sense

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/SeedFromExcelController.kt
```diff
@@ -0,0 +1,19 @@
...
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.SeedFromExcelApiDelegate
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.SeedFromExcelRequest
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.seed.SeedService
+
+@Service
+class SeedFromExcelController(private val seedService: SeedService) : SeedFromExcelApiDelegate {
+  override fun seedFromExcelPost(seedFromExcelRequest: SeedFromExcelRequest): ResponseEntity<Unit> {
+    throwIfNotLoopbackRequest()
+
+    seedService.seedExcelData(seedFromExcelRequest.seedType, seedFromExcelRequest.premisesId, seedFromExcelRequest.fileName)
```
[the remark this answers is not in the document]
danhumphreys-moj: I was thinking to pass the premise id as an argument to the run seed script to begin with (haven't done this yet), seemed the most straightforward thing to do.

Then run the script for every spreadsheet rather than for every file in a directory?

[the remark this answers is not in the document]
davidatkinsuk: Probably one to return to later. We have 108 premises, most of which (presumably) we'd want to load in bulk. We could always just script a curl command to be fair!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
danhumphreys-moj: Logic updated to delete and re-create characteristics.  Ticket [APS-1732](https://dsdmoj.atlassian.net/browse/APS-1732) raised to add integration test to check that transactions are rolled back if seed job fails at the database stage.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
danhumphreys-moj: No need to investigate this - seed job works as is.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
danhumphreys-moj: Logic has been restructured to do this.  Load, validate and create rooms, beds and characteristics, then save to database.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
danhumphreys-moj: Beds are created if they don't exist.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/SeedFromExcelController.kt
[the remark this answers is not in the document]
danhumphreys-moj: Ticket [APS-1733](https://dsdmoj.atlassian.net/browse/APS-1733) raised to investigate simplifying the existing script.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/SeedJob.kt
danhumphreys-moj: This has been changed for all seed jobs now.  Have followed the same principle for this job.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
danhumphreys-moj: Have removed this check.  Updated the logic so that we run through the list of questions we expect to see in the site survey data.  If we do not find a match in the premise spreadsheet then the import fails.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
danhumphreys-moj: Raised ticket [APS-1734](https://dsdmoj.atlassian.net/browse/APS-1734) to do this.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
danhumphreys-moj: Added `.trim()`; `true` also passed to equals which ignores case.

Discussed addition of "N/A" to string comparison logic with @davidatkinsuk this morning.  If answer is "N/A" - answer is set to "No".

src/main/resources/static/_shared.yml
```diff
@@ -3719,6 +3733,10 @@ components:
         - approved_premises_space_planning_dry_run
         - approved_premises_import_delius_booking_management_data
         - approved_premises_update_space_booking
+    SeedFromExcelFileType:
+      type: string
+      enum:
+        - approved_premises_room_from_excel
```
davidatkinsuk, line 3739: can drop _from_excel here i think? given it's of type `SeedFromExcelFileType`

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
```diff
@@ -0,0 +1,210 @@
...
+//    "Is this AP Catered (no for Self Catered)?" to "isCatered",
+//    "Is there a step free entrance to the AP at least 900mm wide?" to "hasWideStepFreeAccess",
+//    "Are corridors leading to communal areas at least 1.2m wide?" to "hasWideAccessToCommunalAreas",
+//    "Do corridors leading to communal areas have step free access?" to "hasStepFreeAccessToCommunalAreas",
+//    "Does this AP have bathroom facilities that have been adapted for wheelchair users?" to "hasWheelChairAccessibleBathrooms",
+//    "Is there a lift at this AP?" to "hasLift",
+//    "Does this AP have tactile & directional flooring?" to "hasTactileFlooring",
+//    "Does this AP have signs in braille?" to "hasBrailleSignage",
+//    "Does this AP have a hearing loop?" to "hasHearingLoop",
+//    "Are there any additional restrictions on people that this AP can accommodate?" to "additionalRestrictions",
```
davidatkinsuk, line 204: I _think_ these are all premise characteristics? In which case they're safe to ignore and can probably remove this comment

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
```diff
@@ -0,0 +1,210 @@
...
+    var rooms = buildRooms(dataFrame, premisesId)
+    val characteristics = buildCharacteristics(dataFrame)
+
+    // add the characteristics to the rooms if they exist
+    rooms.forEach {
+      characteristics[it.code]?.let { roomCharacteristics -> it.characteristics.addAll(roomCharacteristics.toList()) }
+    }
+
+    // update the rooms
+    rooms = createOrUpdateRooms(rooms)
```
davidatkinsuk, line 45: I think the code is nicely organised so we don't need these comments anymore

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
```diff
@@ -0,0 +1,210 @@
...
+@Component
+@Suppress("LongParameterList")
+class ApprovedPremisesRoomsSeedFromXLSXJob(
+  private val premisesRepository: PremisesRepository,
+  private val roomRepository: RoomRepository,
+  private val bedRepository: BedRepository,
+  private val siteSurvey: SiteSurvey,
+) : ExcelSeedJob() {
+  private val log = LoggerFactory.getLogger(this::class.java)
+  override fun processDataFrame(dataFrame: DataFrame<*>, premisesId: UUID) {
```
davidatkinsuk, line 32: would a better name for the dataFrame variable be 'roomsWorksheet?'

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
```diff
@@ -0,0 +1,210 @@
...
+    val beds = buildBeds(dataFrame, rooms)
+    createBedsIfNotExist(beds)
+  }
+
+  private fun buildRooms(dataFrame: DataFrame<*>, premisesId: UUID): MutableList<RoomEntity> {
+    val rooms = mutableListOf<RoomEntity>()
+
+    for (i in 1..<dataFrame.columnsCount()) {
+      val roomAnswers = dataFrame.getColumn(i)
+      val room = createRoom(premisesId, roomAnswers.name, roomAnswers[0].toString())
```
davidatkinsuk, line 56: i think what these variables are (especially `roomAnswers[0].toString()`) would be clearer if you provided the function parameter names?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
```diff
@@ -0,0 +1,210 @@
...
+    id = UUID.randomUUID(),
+    name = roomName,
+    code = roomCode,
+    notes = null,
+    premises = premisesRepository.findByIdOrNull(premisesId)!!,
+    beds = mutableListOf(),
+    characteristics = mutableListOf(),
+  )
+
+  fun createBed(bedName: String, bedCode: String, room: RoomEntity): BedEntity = BedEntity(
```
davidatkinsuk, line 143: i think for consistency with code earlier on this should be called `buildBed` (as we're not actually creating in the database at this point?)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
```diff
@@ -0,0 +1,210 @@
...
+      if (existingBed != null) {
+        log.info("Bed ${it.id} with code ${it.code} already exists in room code ${it.room.code}.")
+      } else {
+        bedRepository.save(it)
+        log.info("Created new bed ${it.id} with code ${it.code} and name ${it.name} in room code ${it.room.code}.")
+      }
+    }
+  }
+
+  private fun createRoom(premisesId: UUID, roomCode: String, roomName: String): RoomEntity = RoomEntity(
```
davidatkinsuk, line 133: i think for consistency with code earlier on this should be called `buildRoom` (as we're not actually creating in the database at this point?)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
```diff
@@ -0,0 +1,210 @@
...
+    return rooms
+  }
+
+  private fun buildCharacteristics(dataFrame: DataFrame<*>): MutableMap<String, MutableList<CharacteristicEntity>> {
+    var premisesCharacteristics = mutableMapOf<String, MutableList<CharacteristicEntity>>()
+
+    siteSurvey.questionToCharacterEntityMapping.forEach { (question, characteristic) ->
+      val rowId = dataFrame.getColumn(0).values().indexOf(question)
+
+      if (rowId == -1) throw SiteSurveyImportException("Characteristic question '$question' not found on sheet Sheet3.")
```
davidatkinsuk, line 69: nice

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
```diff
@@ -0,0 +1,210 @@
...
+
+      for (colId in 1..<dataFrame.columnsCount()) {
+        val roomCode = dataFrame.getColumn(colId).name
+        val answer = dataFrame[rowId][colId].toString().trim()
+
+        if (answer.equals("yes", true)) {
```
davidatkinsuk, line 75: could you please adding a parameter name for the second argument (ignore) for readibility

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
```diff
@@ -0,0 +1,210 @@
...
+//    "Does this AP have bathroom facilities that have been adapted for wheelchair users?" to "hasWheelChairAccessibleBathrooms",
+//    "Is there a lift at this AP?" to "hasLift",
+//    "Does this AP have tactile & directional flooring?" to "hasTactileFlooring",
+//    "Does this AP have signs in braille?" to "hasBrailleSignage",
+//    "Does this AP have a hearing loop?" to "hasHearingLoop",
+//    "Are there any additional restrictions on people that this AP can accommodate?" to "additionalRestrictions",
+  )
+
+  val questionToCharacterEntityMapping = questionToPropertyNameMapping.map { (key, value) ->
+    Pair(key, characteristicRepository.findByPropertyName(value, ServiceName.approvedPremises.value))
```
davidatkinsuk, line 208: I think this needs to error if a characteristic can't be found e.g.

`characteristicRepository.findByPropertyName(value, ServiceName.approvedPremises.value)!!`

This will then remove the need to use `!!` when retrieving characteristics from here

davidatkinsuk: It's probably neater to do something like

```
characteristicRepository.findByPropertyName(value, ServiceName.approvedPremises.value) ?: error("Can't find characteristic")
```

Note that because this is being ran on spring startup this will stop the service from starting. This will be resolved by APS-1734

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
```diff
@@ -0,0 +1,210 @@
...
+      if (rowId == -1) throw SiteSurveyImportException("Characteristic question '$question' not found on sheet Sheet3.")
+
+      for (colId in 1..<dataFrame.columnsCount()) {
+        val roomCode = dataFrame.getColumn(colId).name
+        val answer = dataFrame[rowId][colId].toString().trim()
+
+        if (answer.equals("yes", true)) {
+          premisesCharacteristics.computeIfAbsent(roomCode) { mutableListOf() }.add(characteristic!!)
+        } else if (!answer.equals("no", true) && !answer.equals("N/A", true)) {
+          throw SiteSurveyImportException("Expecting 'yes' or 'no' for question '$question' but is '$answer' on sheet Sheet3.")
```
davidatkinsuk, line 78: great! it would be helpful to output rowId and colId here too?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
```diff
@@ -0,0 +1,210 @@
...
+    for (i in 1..<dataFrame.columnsCount()) {
+      val roomAnswers = dataFrame.getColumn(i)
+      val bedNumber = roomAnswers[1].toString()
+      beds.add(
+        createBed(
+          "${roomAnswers.name}",
+          "${roomAnswers.name} - $bedNumber",
+          rooms.firstOrNull { it.code == roomAnswers.name }
+            ?: throw IllegalArgumentException("Room not found with id ${roomAnswers.name} and code ${roomAnswers.name}."),
+        ),
```
davidatkinsuk, line 96: can you please add function param names here so it's clear what each value represents

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
```diff
@@ -0,0 +1,210 @@
...
+      if (persistedRoom == null) {
+        roomRepository.save(room)
+        log.info("Created new room ${room.id} with code ${room.code} and name ${room.name} in premise ${room.premises.name}.")
+        log.info("Added characteristic(s) ${room.characteristics.joinToString()} to room code ${room.code}.")
+      } else {
+        rooms[index] = persistedRoom
+        persistedRoom.characteristics.clear()
+        persistedRoom.characteristics.addAll(room.characteristics)
+        roomRepository.save(persistedRoom)
+        log.info("Added characteristic(s) ${persistedRoom.characteristics.joinToString()} to room code ${persistedRoom.code}.")
```
davidatkinsuk, line 115: nice!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
```diff
@@ -0,0 +1,210 @@
...
+      }
+    }
+    return rooms
+  }
+
+  private fun createBedsIfNotExist(beds: List<BedEntity>) {
+    beds.forEach {
+      val existingBed = bedRepository.findByCodeAndRoomId(it.code!!, it.room.id)
+      if (existingBed != null) {
+        log.info("Bed ${it.id} with code ${it.code} already exists in room code ${it.room.code}.")
```
davidatkinsuk, line 125: Something to to via another ticket - we need to decide whether to update any fields on existing rooms.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/seed/SeedFromXLSXApprovedPremisesRoomsTest.kt
```diff
@@ -0,0 +1,436 @@
...
+    assertThat(newRoom!!.characteristics).anyMatch {
+      it.name == "Is this room located on the ground floor?" &&
+        it.propertyName == "isGroundFloor"
+    }
+
+    val newBed = bedRepository.findByCodeAndRoomId("SWABI01NEW - 1", newRoom.id)
+    assertThat(newBed!!.name).isEqualTo("SWABI01NEW")
+    assertThat(
+      newBed.room.id == newRoom.id &&
+        newBed.room.code == newRoom.code,
```
davidatkinsuk, line 70: I think when we assert on `newBed.room.code` we can use a String literal? Same goes for anywhere else doing a similar assertions

davidatkinsuk (approved):
LGTM and thanks for the thorough testing 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
```diff
@@ -0,0 +1,210 @@
...
+    id = UUID.randomUUID(),
+    name = bedName,
+    code = bedCode,
+    room = room,
+    endDate = null,
+    createdAt = null,
+  )
+
+  private fun findExistingPremisesOrThrow(premisesId: UUID): PremisesEntity {
+    return premisesRepository.findByIdOrNull(premisesId) ?: throw SiteSurveyImportException(
```
davidatkinsuk, line 153: could you please change this use approvedPremisesRepository instead? That way we won't accidentally update CAS3 premises

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
davidatkinsuk: I was thinking about this last night and I think it's probably best to leave this until working on APS-1734 to avoid the possibility of breaking the entire service if a characteristic is missing :)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
danhumphreys-moj: If a room with the same code exists, the characteristics are cleared out and the ones from the site survey spreadsheet are used.
If a room doesn't exist with this code, it is created with the characteristics from the site survey data.

If a bed exists with the same code then nothing is done (apart from logging that this has happened).
If a bed doesn't exist with the same code then one is created.

danhumphreys-moj: Ahh ok - I see.  I wondered if this is what you meant.  Have raised [APS-1735](https://dsdmoj.atlassian.net/browse/APS-1735) to determine if this is a requirement.

src/main/resources/static/_shared.yml
danhumphreys-moj: Have removed `_from_excel`.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
danhumphreys-moj: Removed these commented characteristics now.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
danhumphreys-moj: Renamed to `roomsWorksheet`.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
danhumphreys-moj: Had added parameter names for clarity.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
danhumphreys-moj: Renamed to `buildBed`.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
danhumphreys-moj: Renamed to `buildRoom`.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
danhumphreys-moj: Added parameter name.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
danhumphreys-moj: Ok, no problem.  Will mark as resolved since will be addressed by APS-1734.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
danhumphreys-moj: Added parameter names.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/seed/SeedFromXLSXApprovedPremisesRoomsTest.kt
danhumphreys-moj: Changed to use string literals where applicable.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/ApprovedPremisesRoomsSeedFromXLSXJob.kt
danhumphreys-moj: Done!

danhumphreys-moj:
Addressed all PR comments where I could, and raised tickets for others that can be done after we run this seed job in pre-production.  Tested updated branch locally again and site survey import was successful.

davidatkinsuk:
> Addressed all PR comments where I could, and raised tickets for others that can be done after we run this seed job in pre-production. Tested updated branch locally again and site survey import was successful.

Great, merge at will 👍

PR #2501, opened 2024-11-15
Get offences endpoint to use AP and Delius API

gregkhawkins:
- refactor away from integration with Community API
- consume the GET case-detail endpoint in the AP and Delius API instead
- feature-flag using AP and Delius API (flag on in dev / test only for now)


helm_deploy/values-preprod.yaml
```diff
@@ -95,6 +95,7 @@ generic-service:
     FEATURE-FLAGS_CAS1-EXPIRED-APPLICATIONS-ENABLED: true
     FEATURE-FLAGS_INCLUDE-ASSESSMENT-UPDATED-DOMAIN-EVENTS: false
     FEATURE-FLAGS_GET-DOCUMENTS-FROM-AP-DELIUS: true
+    FEATURE-FLAGS_GET-OFFENCES-FROM-AP-DELIUS: false
```
davidatkinsuk, line 98: i think we're fine to enable here

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/OffenderService.kt
```diff
@@ -627,6 +628,20 @@ class OffenderService(
...
+      is ClientResult.Failure.StatusCode -> when (caseDetailResult.status) {
+        HttpStatus.NOT_FOUND -> return AuthorisableActionResult.NotFound()
+        HttpStatus.FORBIDDEN -> return AuthorisableActionResult.Unauthorised()
+        else -> caseDetailResult.throwException()
+      }
+
+      is ClientResult.Failure -> caseDetailResult.throwException()
+    }
+    return AuthorisableActionResult.Success(caseDetail)
+  }
```
davidatkinsuk, line 643: could we please update this to return CaseResult types instead? It should just be a matter of replacing `AuthorisableActionResult` with `CasResult`.

Then in the controller, instead of using the local `getSuccessEntityOrThrow` function, you can use the `extractEntityFromCasResult` function in `EntityUtils.kt`

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/OffenseTransformer.kt
```diff
@@ -4,9 +4,10 @@ import org.springframework.stereotype.Component
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ActiveOffence
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.community.Conviction
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.community.OffenceDetail
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.deliuscontext.CaseDetail
 
 @Component
-class ConvictionTransformer {
+class OffenseTransformer {
```
davidatkinsuk, line 10: typo in `Offense`

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/OffenseTransformer.kt
```diff
@@ -17,6 +18,16 @@ class ConvictionTransformer {
     )
   } ?: emptyList()
 
+  fun transformToApi(caseDetail: CaseDetail) = caseDetail.offences.map {
+    ActiveOffence(
+      deliusEventNumber = it.eventNumber,
+      offenceDescription = it.description,
+      offenceId = it.id.toString(),
+      convictionId = it.eventId,
+      offenceDate = it.date,
```
davidatkinsuk, line 27: can you please let me know once this gets to pre-prod, i'd like to do some 'like for like' comparison when creating an application for a CRN that already has an application against it that was using the 'old' endpoint

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/PersonOffencesTest.kt
```diff
@@ -124,7 +130,48 @@ class PersonOffencesTest : InitialiseDatabasePerClassTestBase() {
...
-            objectMapper.writeValueAsString(convictionTransformer.transformToApi(activeConviction)),
+            objectMapper.writeValueAsString(offenseTransformer.transformToApi(activeConviction)),
+          )
+      }
+    }
+  }
+
+  @Test
+  fun `Getting offences for a CRN - with get-offences-from-ap-delius feature-flag on - returns OK with correct body`() {
+    mockFeatureFlagService.setFlag("get-offences-from-ap-delius", true)
```
davidatkinsuk, line 141: nice

davidatkinsuk (approved):
LGTM just a few minor comments

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/OffenseTransformer.kt
gregkhawkins: will do!

PR #2503, opened 2024-11-15
Get offences endpoint to use AP and Delius API

gregkhawkins:
- refactor away from integration with Community API
- consume the GET case-detail endpoint in the AP and Delius API instead
- feature-flag using AP and Delius API (flag on in dev / test only for now)


davidatkinsuk (approved):
LGTM to me 👍 

(noting that my original comments were on this PR -> https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/2501)

PR #2507, opened 2024-11-15
APS-1524: Add Parent-Child relationship to Departure Reasons

RobBoothMOJ:
Add Parent-Child relationship to Departure Reasons

- Added parent_reason_id to API, entity and table
- Created a new parent 'Breach / Recall' departure reason
- Populated the parent_reason_id field for children of the new parent with the new record's Id 
- Updated the child records to remove 'Breach / recall' text
- Implemented a change to the Cas1SpaceBookingTransformer to provide a composite name of parent and child 'Breach / recall' departure resons

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/DepartureReasonEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/DepartureReasonEntity.kt
@@ -1,7 +1,10 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity
 
 import jakarta.persistence.Entity
+import jakarta.persistence.FetchType
 import jakarta.persistence.Id
+import jakarta.persistence.JoinColumn
+import jakarta.persistence.ManyToOne
 import jakarta.persistence.Table
 import org.hibernate.annotations.CacheConcurrencyStrategy
 import org.springframework.data.jpa.repository.JpaRepository
@@ -12,6 +15,7 @@ import java.util.UUID
 
 @Repository
 interface DepartureReasonRepository : JpaRepository<DepartureReasonEntity, UUID> {
+
   @Query("SELECT d FROM DepartureReasonEntity d WHERE d.serviceScope = :serviceName OR d.serviceScope = '*'")
   fun findAllByServiceScope(serviceName: String): List<DepartureReasonEntity>
 
@@ -34,8 +38,17 @@ data class DepartureReasonEntity(
   val isActive: Boolean,
   val serviceScope: String,
   val legacyDeliusReasonCode: String?,
+  @ManyToOne(fetch = FetchType.LAZY)
+  @JoinColumn(name = "parent_reason_id")
+  val parentReasonId: DepartureReasonEntity?,
 ) {
   override fun toString() = "DepartureReasonEntity:$id"
+
+  fun generateParentChildName(): Pair<UUID, String> {
+    return parentReasonId?.let { parentReason ->
+      return Pair(id, "${parentReason.name} - $name")
+    } ?: return Pair(id, name)
+  }
 
   companion object {
     fun valueOf(apiValue: WithdrawPlacementRequestReason): PlacementRequestWithdrawalReason? =
--- a/src/main/resources/db/migration/all/20241114110241__update_departure_reasons_for_breach_recall.sql
+++ b/src/main/resources/db/migration/all/20241114110241__update_departure_reasons_for_breach_recall.sql
@@ -1,0 +1,17 @@
+ALTER TABLE departure_reasons ADD parent_reason_id UUID NULL;
+ALTER TABLE departure_reasons ADD CONSTRAINT parent_reason_fk FOREIGN KEY (parent_reason_id) REFERENCES departure_reasons (id);
+
+INSERT INTO departure_reasons(id, name, is_active, service_scope, legacy_delius_reason_code, parent_reason_id)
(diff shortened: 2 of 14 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/DepartureReasonEntity.kt
```diff
@@ -34,14 +40,28 @@ data class DepartureReasonEntity(
...
   val legacyDeliusReasonCode: String?,
+  val parentReasonId: UUID?,
 ) {
   override fun toString() = "DepartureReasonEntity:$id"
 
   companion object {
     fun valueOf(apiValue: WithdrawPlacementRequestReason): PlacementRequestWithdrawalReason? =
       PlacementRequestWithdrawalReason.entries.firstOrNull { it.apiValue == apiValue }
+
+    fun generateParentChildName(reason: DepartureReasonEntity?): NamedId? =
```
davidatkinsuk, line 51: i think it's probably more flexible if we

a) converted parentReasonId into a nullable @ManyToOne relationship to DepartureReasonEntity
b) move the generateParentChildName function into the class definition itself (instead of it being a 'static' function) and then follow this relationship if defined (no need to hardcode the BREACH RECALL id then)?

src/main/resources/db/migration/all/20241114110241__update_departure_reasons_for_breach_recall.sql
```diff
@@ -0,0 +1,17 @@
...
+INSERT INTO departure_reasons(id, name, is_active, service_scope, legacy_delius_reason_code, parent_reason_id)
+SELECT 'd3e43ec3-02f4-4b96-a464-69dc74099259', 'Breach / Recall', TRUE, 'approved-premises', NULL, NULL
+WHERE NOT EXISTS (
+    SELECT id FROM departure_reasons where id = 'd3e43ec3-02f4-4b96-a464-69dc74099259'
+);
+
+
+UPDATE departure_reasons
+    SET
+        name = upper(substring(SUBSTRING(name, '\((.+)\)') from 1 for 1)) || substring(SUBSTRING(name, '\((.+)\)') from 2),
```
davidatkinsuk, line 12: fancy!

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/transformer/cas1/Cas1SpaceBookingTransformerTest.kt
```diff
@@ -283,6 +285,211 @@ class Cas1SpaceBookingTransformerTest {
...
+
+      val result = transformer.transformJpaToApi(personInfo, spaceBooking, emptyList())
+
+      assertThat(result.id).isEqualTo(spaceBooking.id)
+      assertThat(result.departureReason!!.id).isEqualTo(departureReason.id)
+      assertThat(result.departureReason!!.name).isEqualTo("Breach / Recall - ${departureReason.name}")
+    }
+
+    @Test
+    fun `Space booking is transformed correctly with departure reason when parent reason id is invalid`() {
```
davidatkinsuk, line 426: we should add a FK constraint into the database so we don't need to check this

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/transformer/cas1/Cas1SpaceBookingTransformerTest.kt
```diff
@@ -283,6 +285,211 @@ class Cas1SpaceBookingTransformerTest {
     }
   }
 
+  @Nested
+  inner class TransformJpaToApiWithDepartureReason {
```
davidatkinsuk, line 289: this is quite a lot of new test code to test one field. Can we update an existing test for at least one of these (e.g. add in an assertion for departure reason where no parent)

davidatkinsuk (approved):
looks good, a few proposals from me 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/DepartureReasonEntity.kt
```diff
@@ -34,9 +39,21 @@ data class DepartureReasonEntity(
   val isActive: Boolean,
   val serviceScope: String,
   val legacyDeliusReasonCode: String?,
+  @ManyToOne(fetch = FetchType.LAZY)
+  @JoinColumn(name = "parent_reason_id")
+  val parentReasonId: DepartureReasonEntity?,
 ) {
   override fun toString() = "DepartureReasonEntity:$id"
 
+  fun generateParentChildName(reason: DepartureReasonEntity?): NamedId? =
```
davidatkinsuk, line 48: i don't think we need to pass reason in here as you can directly access the fields you need (it's not longer a static function)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/DepartureReasonEntity.kt
```diff
@@ -34,9 +39,21 @@ data class DepartureReasonEntity(
...
+    reason?.let { reason ->
+      reason.parentReasonId?.let {
+        parentReasonId?.let { parentReason ->
+          NamedId(id = reason.id, name = "${parentReason.name} - ${reason.name}")
+        }
+      } ?: NamedId(reason.id, reason.name)
```
davidatkinsuk, line 54: I don't think API types should be 'leaking' into the domain model. Could this return the name and the receiver wrap it up on `NamedId`?

PR #2510, opened 2024-11-15
APS-1567 -  increase dev probation search buffer

davidatkinsuk:
An attempt to fix this deployment issue

```Caused by: org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'webClientConfiguration' defined in URL [jar:nested:/app/app.jar/!BOOT-INF/classes/!/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/WebClientConfiguration.class]: Unsatisfied dependency expressed through constructor parameter 4: Failed to convert value of type 'java.lang.String' to required type 'int'; For input string: "2.62144e+06"```


RobBoothMOJ (approved):
Approved

PR #2530, opened 2024-11-19
APS-1583 - Optimise planning model creation

davidatkinsuk:
Before this commit a database query was ran to retrieve space bookings applicable for each day being planned.

This commit instead loads all space bookings that are active across a given date range, and uses this to find each applicable booking for the day (as we already do for out of service bed records).

This significantly reduces the number of database queries required.

This change has been made in preparation for reusing the planning models to calculate capacity wrt. characteristics that are of interest for space booking planning


RobBoothMOJ (approved):
Approved

PR #2531, opened 2024-11-20
APS-1583 implement CAS1 get premise capacity endpoint

davidatkinsuk:
**API Definition**

This commit changes some fields related to capacity to remove some ambiguity.

- `capacity` is now `availableBedsCount`
- `available` has been removed and instead we provide `bookingsCount`
- The value previously returned for `available` can be calculated using `availableBedsCount - bookingsCount`.

This change has been made to remove some ambiguity around what `available` actually means

**Test Coverage**

Note that because we use integration tests to test the `SpacePlanningService` directly, the integration test is light weight


RobBoothMOJ (approved):
Approved

PR #2534, opened 2024-11-20
CAS-666 - Introduce an endpoint to return Timeline Events

daveawc:
This introduces an endpoint that will return the timeline elements of an assessment.

### DIFF
--- a/src/main/resources/static/cas3-api.yml
+++ b/src/main/resources/static/cas3-api.yml
@@ -48,3 +48,25 @@ paths:
               schema:
                 type: string
                 format: binary
+  /timeline/{assessmentId}:
+    get:
+      operationId: getTimelineEntries
+      tags:
+        - CAS3
+      summary: Returns the timeline entries for a given Assessment.
+      parameters:
+        - name: assessmentId
+          in: path
+          required: true
+          schema:
+            type: string
+            format: uuid
+      responses:
+        200:
+          description: Successfully retrieved the timeline events
+          content:
+            'application/json':
+              schema:
+                type: array
+                items:
+                  $ref: '_shared.yml#/components/schemas/ReferralHistoryNote'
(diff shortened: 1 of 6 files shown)


src/main/resources/static/cas3-api.yml
```diff
@@ -48,3 +48,25 @@ paths:
               schema:
                 type: string
                 format: binary
+  /timeline/{assessmentId}:
+    get:
+      operationId: getTimelineEntries
+      tags:
+        - CAS3
+      summary: Returns the timeline entries for a given Assessment'.
```
davidatkinsuk: unexpected `'`

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas3/TimelineController.kt
```diff
@@ -0,0 +1,30 @@
...
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.cas3.TimelineCas3Delegate
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ReferralHistoryNote
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.TemporaryAccommodationAssessmentEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.AssessmentService
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.UserService
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas3.DomainEventService
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.AssessmentTransformer
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.extractEntityFromCasResult
+
+@RestController("CAS3TimelineController")
```
davidatkinsuk, line 14: these aren't actually RestControllers (the generated code provides the controller implementation). Therefore, I think this should just be a `@Service`

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas3/TimelineController.kt
```diff
@@ -0,0 +1,30 @@
...
+class TimelineController(
```
davidatkinsuk, line 15: Do we think there's any value having the class exposed under a different name in spring to it's class name? How about we rename the class to `Cas3TimelineController` and drop the value from the annotation?

(on a wider note, i think we should be prefixing all casx-specific classes with the corresponding CAS code to make it easier to navigate the code)

davidatkinsuk (approved):
LGTM, just a few minor comments

PR #2545, opened 2024-11-20
APS-1583 - Fix planning bug related to single rooms

davidatkinsuk:
This commit fixes a bug where a booking that requires a single room is sometimes placed into a room with multiple beds vs a room with a single bed and the ‘single bed characteristic’.

Whilst this was a legitimate bug, it does highlight a potential conflict in the order in whcih the following rules are applied:

1. Bookings are placed into rooms with the least surplus characteristics
2. If there are multiple competing rooms (after above goals are satisfied), single occupancy bookings will be placed into rooms with the least beds (ensuring minimum 'wastage' in these scenarios)

This can result in a single booking being put into a room with many beds because the single bed room has unrequired characteristics. This may not be desirable


RobBoothMOJ (approved):
Approved

PR #2547, opened 2024-11-20
Feature flag on for offences from Delius (all envs but prod)

gregkhawkins:


src/main/resources/application-local.yml
```diff
@@ -56,7 +56,7 @@ user-allocations:
 
 feature-flags:
   get-documents-from-ap-delius: true
-  get-offences-from-ap-delius: true
+  get-offences-from-ap-delius: false
```
gregkhawkins, line 59: switched this off temporarily as this has been breaking the local e2e tests for `hmpps-temporary-accommodation-ui`. Have a fix for this locally now but want to do some thorough testing before I raise a PR on that and do not want to block anyone

davidatkinsuk (approved):
LGTM 👍

PR #2552, opened 2024-11-21
APS-1507 - Remove Space Booking Perms from Workflow Manager

davidatkinsuk:
Any user requiring these permissons should use the Future Manager and/Or CRU Member Roles. As this functionality isn’t in use in production, no actual users will be affected by this change


RobBoothMOJ (approved):
Approved

PR #2553, opened 2024-11-21
Refactor offence description for Ap and Delius offences

gregkhawkins:
**Background**
- We need to mirror the way we generated the description from `Community API`
- Work has been done on this `AP and Delius` API PR to supply these extra fields (this will need to be deployed first): https://github.com/ministryofjustice/hmpps-probation-integration-services/pull/4455


davidatkinsuk (approved):
LGTM 👍

PR #2554, opened 2024-11-21
Cas 665 exclude user notes from timeline for referrers

daveawc:
This filters out any user notes from being visible to referrers, for when they are able to view the CAS3 timeline.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas3/Cas3TimelineController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas3/Cas3TimelineController.kt
@@ -5,6 +5,7 @@ import org.springframework.stereotype.Service
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.cas3.TimelineCas3Delegate
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ReferralHistoryNote
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.TemporaryAccommodationAssessmentEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserRole
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.AssessmentService
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.UserService
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas3.DomainEventService
@@ -20,10 +21,14 @@ class Cas3TimelineController(
 ) : TimelineCas3Delegate {
   override fun getTimelineEntries(assessmentId: java.util.UUID): ResponseEntity<List<ReferralHistoryNote>> {
     val user = userService.getUserForRequest()
-    val assessmentResult = assessmentService.getAssessmentAndValidate(user, assessmentId)
+    val assessmentResult = assessmentService.getAssessmentAndValidate(user, assessmentId, forTimeline = true)
     val assessment = extractEntityFromCasResult(assessmentResult) as TemporaryAccommodationAssessmentEntity
     val domainEventNotes = domainEventService.getAssessmentUpdatedEvents(assessmentId = assessment.id)
-    val timelineEntries = assessmentTransformer.getSortedReferralHistoryNotes(assessment, domainEventNotes)
+    val timelineEntries = assessmentTransformer.getSortedReferralHistoryNotes(
+      assessment,
+      domainEventNotes,
+      includeUserNotes = user.hasAnyRole(UserRole.CAS3_ASSESSOR),
+    )
 
     return ResponseEntity.ok(timelineEntries)
   }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AssessmentService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AssessmentService.kt
@@ -152,7 +152,11 @@ class AssessmentService(
     }
   }
 
-  fun getAssessmentAndValidate(user: UserEntity, assessmentId: UUID): CasResult<AssessmentEntity> {
+  fun getAssessmentAndValidate(
+    user: UserEntity,
+    assessmentId: UUID,
+    forTimeline: Boolean = false,
+  ): CasResult<AssessmentEntity> {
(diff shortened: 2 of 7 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/AssessmentTransformer.kt
```diff
@@ -111,20 +111,23 @@ class AssessmentTransformer(
   }
 
   fun getSortedReferralHistoryNotes(
-    jpa: TemporaryAccommodationAssessmentEntity,
+    assessment: TemporaryAccommodationAssessmentEntity,
```
davidatkinsuk, line 114: yes! i've no idea why we insist on using the 'jpa' variable name :)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/AssessmentTransformer.kt
```diff
@@ -111,20 +111,23 @@ class AssessmentTransformer(
...
     cas3Events: List<DomainEventEntity>,
+    filterUserNotes: Boolean = false,
```
davidatkinsuk, line 116: in the kotlin `filter` closure we use 'true' to mean 'include in the response'. I think this variable is using 'true' to mean 'don't include in the response.

would a more consistent name be 'includeUserNotes' (and flip the boolean?). 

but it's possible i'm getting my booleans confused :)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/AssessmentTransformer.kt
```diff
@@ -111,20 +111,23 @@ class AssessmentTransformer(
...
+    val notes = assessment.referralHistoryNotes.map {
       if (it.id == lastReferralRejectedHistoryNote?.id) {
-        assessmentReferralHistoryNoteTransformer.transformJpaToApi(it, jpa)
+        assessmentReferralHistoryNoteTransformer.transformJpaToApi(it, assessment)
       } else {
         assessmentReferralHistoryNoteTransformer.transformJpaToApi(it)
       }
-    }.toMutableList()
+    }
+      .filter { it.type != "user" || !filterUserNotes }
```
davidatkinsuk, line 129: it would be nice if this was an enumerated type - if that's going to cause UI issues could we maybe have constants defined in assessmentReferralHistoryNoteTransformer for the different possible values and refer to that here instead?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AssessmentService.kt
```diff
@@ -168,7 +172,9 @@ class AssessmentService(
       else -> throw RuntimeException("Assessment type '${assessment::class.qualifiedName}' is not currently supported")
     }
 
-    if (!userAccessService.userCanViewAssessment(user, assessment)) {
+    val isAuthorised = userAccessService.userCanViewAssessment(user, assessment) || (isReferrer && userAccessService.userCanViewApplication(user, assessment.application))
```
davidatkinsuk, line 175: Is there a reason this logic isn't in the `userCanViewAssessment` function? At the moment the logic is possibly fragmented? If doing it that way, i think `userCanViewAssessment` can directly access the user role (instead of passing `isReferrer` in

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas3/Cas3TimelineController.kt
```diff
@@ -20,10 +21,11 @@ class Cas3TimelineController(
 ) : TimelineCas3Delegate {
   override fun getTimelineEntries(assessmentId: java.util.UUID): ResponseEntity<List<ReferralHistoryNote>> {
     val user = userService.getUserForRequest()
-    val assessmentResult = assessmentService.getAssessmentAndValidate(user, assessmentId)
+    val isReferrer = user.hasAnyRole(UserRole.CAS3_REFERRER)
```
davidatkinsuk, line 24: I guess it's not possible that an assessor would also be a referrer?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/transformer/AssessmentTransformerTest.kt
```diff
@@ -459,6 +460,15 @@ class AssessmentTransformerTest {
     }
   }
 
+  @Test
+  fun `getSortedReferralHistoryNotes correctly filters user notes`() {
```
davidatkinsuk, line 464: can we wrap this in an inner class pls.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/transformer/AssessmentTransformerTest.kt
```diff
@@ -459,6 +460,15 @@ class AssessmentTransformerTest {
...
+    val application = TemporaryAccommodationApplicationEntityFactory().withDefaults().produce()
+    val assessment = TemporaryAccommodationAssessmentEntityFactory().withApplication(application).produce()
+    val result = assessmentTransformer.getSortedReferralHistoryNotes(assessment, cas3Events = emptyList(), filterUserNotes = true)
+
+    assertThat(result.size).isEqualTo(0)
```
davidatkinsuk, line 469: I think we need some test notes to be filtered out (and some that are retained?)

davidatkinsuk (approved):
functionally great, some comments on code i'm not too familiar with so i'm probably missing something :)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AssessmentService.kt
davidatkinsuk: Thanks for the explanation. I agree `forTimeline` gives context whilst remaining pragmatic 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/AssessmentTransformer.kt
davidatkinsuk: Yes, may as well be called 'object' :)

PR #2558, opened 2024-11-22
APS-1607 - Add endpoint to get a space booking without premise id

davidatkinsuk:
This is required to retrieve information when withdrawing a space booking


RobBoothMOJ (approved):
Approved

PR #2560, opened 2024-11-22
Remove filename from SeedJob constructor

davidatkinsuk:
The filename was only being used to apply some validation on init, which can be dealt with in the `SeedService` itself.

Removing the filename from the constructor simplifies the code, and gets us closer to being able to declare seed job as Spring Beans, simplifying how we use them in the `SeedService`


RobBoothMOJ (approved):
Approved

PR #2563, opened 2024-11-22
Increasing the length of generated strings

daveawc:
some duplicate identifiers were causing test failures. They were only 5 characters long, so hoping increasing this will prevent it from happening.


davidatkinsuk (approved):
LGTM 👍

PR #2566, opened 2024-11-25
APS-1301 - Remove unused CAS1 appeal logic

davidatkinsuk:
This commit removes some logic that relates to a booking cancellation reason that was deactivated in 08/2023 (see ca0e4c5a).

The logic leads to new placement request being created when a booking is cancelled with the reason ‘Booking successfully appealed’. Given that that current logic will return any placement request with a cancelled booking as ‘unmatched’ (regardless of the reason), this logic is redundant and would result in multiple placement requests for a given Booking.


RobBoothMOJ (approved):
Approved

PR #2568, opened 2024-11-25
APS-1568: Ensure no active Bookings when creating a Space Booking

RobBoothMOJ:
Ensure there are no active legacy Bookings or active Space Bookings when creating a Space Booking.
Drop the cas1_space_bookings_premises_id_placement_request_id_idx index as it is not required.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1SpaceBookingService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1SpaceBookingService.kt
@@ -89,12 +89,16 @@ class Cas1SpaceBookingService(
     premises!!
     placementRequest!!
 
-    when (
-      val existingBooking =
-        cas1SpaceBookingRepository.findByPremisesIdAndPlacementRequestId(premisesId, placementRequestId)
-    ) {
-      null -> {}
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1SpaceBookingService.kt
```diff
@@ -89,12 +89,16 @@ class Cas1SpaceBookingService(
...
-    ) {
-      null -> {}
-      else -> return existingBooking.id hasConflictError "A Space Booking already exists for this premises and placement request"
+    placementRequest.booking?.let {
+      if (it.isActive()) {
+        return it.id hasConflictError "A legacy Booking already exists for this premises and placement request"
+      }
+    }
+
+    cas1SpaceBookingRepository.findByPremisesIdAndPlacementRequestId(premisesId, placementRequestId)?.let {
```
davidatkinsuk, line 98: I think this will throw an exception if we have multiple cancelled space bookings linked to this placement reuqest. Probably best to a return a list and do something like

```
if(cas1SpaceBookingRepository.findByPremisesIdAndPlacementREquestId(premiseId, placementRequestId).any {  it.isActive }) {
   return error
}
```

davidatkinsuk (approved):
LGTM other than one subtle issue if we have multiple cancelled space bookings

PR #2569, opened 2024-11-25
Seed Jobs should only use Spring Managed Instances

davidatkinsuk:
This commit refactors seed jobs such that they only required spring manage instances on construction. This will then allow us to simplify the `SeedService` by converting all seed jobs into spring-managed components and delegate to spring when retreiving the required seed job.

To achieve this the `UserSeedJob` has been made abstract and specific usages of it in `SeedService` are sastified using context-specific seed job definitions.


RobBoothMOJ (approved):
Approved

PR #2570, opened 2024-11-25
Ensure premises supports Space Bookings when creating a Space Booking

RobBoothMOJ:
Checks Premises supplied is compatible with a Space Booking.


davidatkinsuk (approved):
LGTM 👍

PR #2573, opened 2024-11-25
APS-1568: Change Space Booking find by Premises ad PlacementRequest

RobBoothMOJ:
Change Space Booking find by Premises and PlacementRequest to find by PlacementRequest only.
Cater for the possibility of multiple bookings satisfying search criteria rather than single record.


davidatkinsuk (approved):
perfect 👍

PR #2582, opened 2024-11-26
APS-1711 - Deprecate Placement Matching Outcome Reports

davidatkinsuk:
This report is not currently in use and will be superseded by the space bookings placement report


RobBoothMOJ (approved):
Approved

PR #2584, opened 2024-11-26
Feature flag in prod off for offences from Ap and Delius

gregkhawkins:


davidatkinsuk:
closing this as the feature flags have now been removed

PR #2586, opened 2024-11-26
APS-1605 - Add getStaffDetailByCode service function

davidatkinsuk:
Before this commit domain event services were directly calling the `ap-and-delius` API client to retrieve keyworker information. This commit introduces a service function to retrieve keyworker information that uses `CasResult`, and uses it in the domain event service.


gregkhawkins (approved):

davidatkinsuk:
Closing this as we no longer use this function for domain events (we've removed keyworkers from domain events)

PR #2587, opened 2024-11-26
Switch swagger documentation to auto generated.

daveawc:
This is the first PR in moving away from the manually writted openAPI spec files. This generates the swagger documentation automatically, from the code that is generated from the spec files. There are extra group filters added, which will be more refined later in the migration process. (once we have package structure etc).

There should be no noticeable difference to the user, however this  is needed so we can test the UI generation from this output.
image one is the auto-generated documents, image two is taken from existing in preprod.
image

image


davidatkinsuk (approved):
LGTM 👍

PR #2590, opened 2024-11-26
updated swagger config

daveawc:
This PR sets the enums to be generated as reusable references, rather than inline. It was an issue for how the UI generates it's types. 

Also removed erroneous dependencies, and updated the exclusion filtering in the config

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/swagger/SwaggerConfiguration.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/swagger/SwaggerConfiguration.kt
@@ -10,6 +10,12 @@ import org.springframework.context.annotation.Primary
 @Configuration
 class SwaggerConfiguration {
 
+  companion object {
+    init {
+      io.swagger.v3.core.jackson.ModelResolver.enumsAsRef = true
+    }
+  }
+
   @Bean
   @Primary
   fun customOpenAPI(): OpenAPI {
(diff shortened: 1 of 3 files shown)


davidatkinsuk (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/swagger/SwaggerConfiguration.kt
```diff
@@ -10,6 +10,12 @@ import org.springframework.context.annotation.Primary
 @Configuration
 class SwaggerConfiguration {
 
+  companion object {
+    init {
+      io.swagger.v3.core.jackson.ModelResolver.enumsAsRef = true
```
davidatkinsuk, line 15: nice!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/swagger/SwaggerConfiguration.kt
```diff
@@ -37,25 +43,72 @@ class SwaggerConfiguration {
...
+      .pathsToExclude("/**/cas1/**", "/**/cas2/**", "/**/cas3/**", "/**/events/**")
+      .packagesToExclude(
+        "**.uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.**",
+      )
       .build()
   }
 
   @Bean
-  fun events(): GroupedOpenApi {
+  fun domainEvents(): GroupedOpenApi {
```
davidatkinsuk, line 54: would this be better named as cas1And3DomainEvents?

davidatkinsuk (approved):
LGTM, one minor proposal

PR #2596, opened 2024-11-27
CAS-1258 Extend the response for API for bedspace search to include personType

muhammad-elabdulla:
This PR CAS-1258 is to:

- Extend the response for API for bedspace search to include personType 
- Change the apDeliusContextAddListCaseSummaryToBulkResponse to accept list of case summries


davidatkinsuk (approved):
LGTM 👍

PR #2597, opened 2024-11-28
APS-1638 - CAS1 Domain Event Improvements

davidatkinsuk:
* updates the CAS1 Space Booking Domain Event service to remove the need to force non-nullness on arrival and departure dates, instead having the non-null values passed in as part of the function call
* introduces a Cas1ApplicationFacade to simplify how certain application properties are retrieved when an entity could be linked to an application or an offline application


RobBoothMOJ (approved):
Approved

PR #2598, opened 2024-11-28
APS-1609: Add optimistic lock to Cas1 Space Bookings

RobBoothMOJ:
Add optimistic lock to Cas1 Space Bookings
Adds a version column to the entity and table which allows JPA @Version to detect concurrent amendments to the same record.  
This approach is already in use on Bookings, Placement Applications, Placement Requests.


RobBoothMOJ:
Commits detail the wrong ticket number.   Should have been APS-1610

PR #2602, opened 2024-11-28
PR to update the cas1 domain event package

daveawc:
updating the cas1 domain event model package structure for consistency.

All of the changes in the file are the updated package name in the imports, and the change of output directory in build.gradle.kts


davidatkinsuk (approved):
LGTM 👍

PR #2605, opened 2024-11-28
APS 1638 - Add separate arrival/departure date & time fields to CAS1 Space Booking API

davidatkinsuk:
* Add separate date/time fields to CAS1 arrival API
* Add separate date/time fields to space booking API
* Remove usage of deprecated space booking functions


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas1/Cas1SpaceBookingController.kt
```diff
@@ -161,12 +163,28 @@ class Cas1SpaceBookingController(
   ): ResponseEntity<Unit> {
     userAccessService.ensureCurrentUserHasPermission(UserPermission.CAS1_SPACE_BOOKING_RECORD_ARRIVAL)
 
+    val arrivalDateAndTime = if (cas1NewArrival.arrivalDateTime != null) {
+      Pair(
+        cas1NewArrival.arrivalDateTime!!.toLocalDate(),
+        cas1NewArrival.arrivalDateTime!!.toLocalDateTime().toLocalTime(),
```
gregkhawkins, line 169: do we need the double bangs? Think the `if` above will negate the need

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas1/Cas1SpaceBookingController.kt
```diff
@@ -161,12 +163,28 @@ class Cas1SpaceBookingController(
...
+    } else {
+      val arrivalDate = cas1NewArrival.arrivalDate
+        ?: throw BadRequestProblem(invalidParams = mapOf("arrivalDate" to "is required"))
+      val arrivalTime = cas1NewArrival.arrivalTime
+        ?: throw BadRequestProblem(invalidParams = mapOf("arrivalTime" to "is required"))
+      Pair(
+        arrivalDate,
+        LocalTime.parse(arrivalTime),
+      )
+    }
```
gregkhawkins, line 180: is this new block worth of a private function? 
The jury's out on this tbf as this function is not that big even with this - occurred to me as an option :)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas1/Cas1SpaceBookingController.kt
```diff
@@ -179,13 +197,29 @@ class Cas1SpaceBookingController(
...
+    } else {
+      val departureDate = cas1NewDeparture.departureDate
+        ?: throw BadRequestProblem(invalidParams = mapOf("departureDate" to "is required"))
+      val departureTime = cas1NewDeparture.departureTime
+        ?: throw BadRequestProblem(invalidParams = mapOf("departureTime" to "is required"))
+      Pair(
+        departureDate,
+        LocalTime.parse(departureTime),
+      )
+    }
```
gregkhawkins, line 214: in fact as we are repeating could go with this:
```
val departureDateAndTime = getDateTime(
      dateTime = cas1NewDeparture.departureDateTime,
      date = cas1NewDeparture.departureDate,
      time = cas1NewDeparture.departureTime
    )
```
to call this:

```
private fun getDateTime(dateTime: Instant?, date: java.time.LocalDate?, time: String?) = if (dateTime != null) {
      Pair(
        first = dateTime.toLocalDate(),
        second = dateTime.toLocalDateTime().toLocalTime(),
      )
    } else {
      time ?: throw BadRequestProblem(invalidParams = mapOf("departureTime" to "is required"))
      Pair(
        first = date ?: throw BadRequestProblem(invalidParams = mapOf("departureDate" to "is required")),
        second = LocalTime.parse(time),
      )
    }
```

gregkhawkins (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas1/Cas1SpaceBookingController.kt
davidatkinsuk: So ... we shouldn't, but the command line compiler (i.e. not intellij) was convinced we do. This is only temporary code (cas1NewArrival.arrivalDateTime is deprecated) so I decided to just leave as-is instead of digging into why it was complaining. Good spot!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas1/Cas1SpaceBookingController.kt
gregkhawkins: then above re-used for arrivalDateAndTime also - prob obvious that bit :)

davidatkinsuk: Yep definitely looks better, but as temporary code probably not worth the refactor. Note that once `cas1NewDeparture.departureDateTime` is removed (it's deprecated), we'll make the other two fields mandatory in the API so we won't need to do these checks here either.

(sorry, i'm using temporary code as an excuse for not doing anything :) )

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas1/Cas1SpaceBookingController.kt
davidatkinsuk: Yep I agree, probably would be neater but as this is temporary code until we remove cas1NewArrival.arrivalDateTime i think probably fine as is

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas1/Cas1SpaceBookingController.kt
gregkhawkins: haha, that's fine - temp code as you say :)

PR #2608, opened 2024-11-28
Fix cas1-schema refs

davidatkinsuk:
This commit fixes the cas1-schemas references in cas1-api.yml for out of service bed types. This incorrect reference is only an issue when viewing the yml file in intellij.


RobBoothMOJ (approved):
Approved

PR #2609, opened 2024-11-29
APS-1635 - Fix space booking timeline url

davidatkinsuk:
When a link to a view space booking is shown on the timeline the wrong url was being used for local deploys. In all other deploys the configuration was incorrectly named in helm so was not used. This commit fixes those issues


muhammad-elabdulla (approved):
LGTM

PR #2612, opened 2024-11-29
APS-1619: Add manager details to approved premises

RobBoothMOJ:
Add manager details to approved premises

- Added migration to add manager_details field to approved_premises table 
- Added managerDetails field to entity
- Seeded managerDetails data for local, dev and test environments
- Added managerDetails to Cas1PremisesSummary in API


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PremisesEntity.kt
```diff
@@ -270,6 +270,7 @@ class ApprovedPremisesEntity(
   @Enumerated(value = EnumType.STRING)
   var gender: ApprovedPremisesGender,
   var supportsSpaceBookings: Boolean,
+  var managerDetails: String,
```
davidatkinsuk, line 273: I think this needs to be `String?` (I think in its current state this will error when deployed to preprod)

davidatkinsuk (approved):
Looks great, just one comment from me related to null-ability

PR #2613, opened 2024-11-29
Add permissions to CRU_MEMBER_FIND_AND_BOOK_BETA

davidatkinsuk:
This commit ensures the CAS1_CRU_MEMBER_FIND_AND_BOOK_BETA role has all the same permissions as CAS1_CRU_MEMBER, with the following exceptions:

* Only CAS1_CRU_MEMBER has the CAS1_BOOKING_CREATE permission
* Only CAS1_CRU_MEMBER_FIND_AND_BOOK_BETA has the CAS1_SPACE_BOOKING_CREATE permission

Note that the CAS1_WORKFLOW_MANAGER still has the CAS1_BOOKING_CREATE permission. This will be removed in a subsequent commit.


RobBoothMOJ (approved):
Approved

PR #2615, opened 2024-11-29
APS-1498 - Fix Space Booking Migration Reason Code Resolution

davidatkinsuk:
This commit fixes an issue caused by there being multiple entries in the departure reason code table with the same legacy delius code. One entry is a enabled, the other is not. This results in a `Query did not return a unique result: 2 results were returned` error.

This commit fixes this issue by getting a list of matching reason codes, taking the enabled one if it exists.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/Cas1BookingToSpaceBookingSeedJob.kt
```diff
@@ -173,11 +174,23 @@ class Cas1BookingToSpaceBookingSeedJob(
...
+      departureReasonRepository
+        .findAllByServiceScope(ServiceName.approvedPremises.value)
+        .filter { it.legacyDeliusReasonCode == reasonCode }
+        .maxByOrNull {
+          if (it.isActive) {
+            1
+          } else {
+            0
+          }
+        }
```
muhammad-elabdulla, line 187: this code can be replaced with
        .maxByOrNull { it.isActive}

muhammad-elabdulla (approved):
LGTM. only small comment

PR #2625, opened 2024-12-02
Chore/rename generated enums to uppercase

daveawc:
This updates the generated enum names to UPPERCASE from camelCase, in preparation for the upcoming swagger changes, which means we'll avoid having to exclude loads of files from linting check failures because the naming conventions are wrong.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PlacementApplicationEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PlacementApplicationEntity.kt
@@ -218,16 +218,16 @@ enum class PlacementApplicationDecision(val apiValue: ApiPlacementApplicationDec
 }
 
 enum class PlacementApplicationWithdrawalReason(val apiValue: WithdrawPlacementRequestReason) {
-  DUPLICATE_PLACEMENT_REQUEST(WithdrawPlacementRequestReason.DUPLICATE_PLACEMENT_REQUEST),
-  ALTERNATIVE_PROVISION_IDENTIFIED(WithdrawPlacementRequestReason.ALTERNATIVE_PROVISION_IDENTIFIED),
-  WITHDRAWN_BY_PP(WithdrawPlacementRequestReason.WITHDRAWN_BY_PP),
-  CHANGE_IN_CIRCUMSTANCES(WithdrawPlacementRequestReason.CHANGE_IN_CIRCUMSTANCES),
-  CHANGE_IN_RELEASE_DECISION(WithdrawPlacementRequestReason.CHANGE_IN_RELEASE_DECISION),
-  NO_CAPACITY_DUE_TO_LOST_BED(WithdrawPlacementRequestReason.NO_CAPACITY_DUE_TO_LOST_BED),
-  NO_CAPACITY_DUE_TO_PLACEMENT_PRIORITISATION(WithdrawPlacementRequestReason.NO_CAPACITY_DUE_TO_PLACEMENT_PRIORITISATION),
-  NO_CAPACITY(WithdrawPlacementRequestReason.NO_CAPACITY),
-  ERROR_IN_PLACEMENT_REQUEST(WithdrawPlacementRequestReason.ERROR_IN_PLACEMENT_REQUEST),
-  RELATED_APPLICATION_WITHDRAWN(WithdrawPlacementRequestReason.RELATED_APPLICATION_WITHDRAWN),
+  DUPLICATE_PLACEMENT_REQUEST(WithdrawPlacementRequestReason.duplicatePlacementRequest),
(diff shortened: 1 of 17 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PlacementApplicationEntity.kt
```diff
@@ -218,16 +218,16 @@ enum class PlacementApplicationDecision(val apiValue: ApiPlacementApplicationDec
...
-  NO_CAPACITY_DUE_TO_LOST_BED(WithdrawPlacementRequestReason.noCapacityDueToLostBed),
-  NO_CAPACITY_DUE_TO_PLACEMENT_PRIORITISATION(WithdrawPlacementRequestReason.noCapacityDueToPlacementPrioritisation),
-  NO_CAPACITY(WithdrawPlacementRequestReason.noCapacity),
-  ERROR_IN_PLACEMENT_REQUEST(WithdrawPlacementRequestReason.errorInPlacementRequest),
-  RELATED_APPLICATION_WITHDRAWN(WithdrawPlacementRequestReason.relatedApplicationWithdrawn),
+  DUPLICATE_PLACEMENT_REQUEST(WithdrawPlacementRequestReason.DUPLICATE_PLACEMENT_REQUEST),
+  ALTERNATIVE_PROVISION_IDENTIFIED(WithdrawPlacementRequestReason.ALTERNATIVE_PROVISION_IDENTIFIED),
+  WITHDRAWN_BY_PP(WithdrawPlacementRequestReason.WITHDRAWN_BY_PP),
+  CHANGE_IN_CIRCUMSTANCES(WithdrawPlacementRequestReason.CHANGE_IN_CIRCUMSTANCES),
+  CHANGE_IN_RELEASE_DECISION(WithdrawPlacementRequestReason.CHANGE_IN_CIRCUMSTANCES),
```
davidatkinsuk, line 225: should be CHANGE_IN_RELEASE_DECISION?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/PersonTransformer.kt
```diff
@@ -155,9 +155,9 @@ class PersonTransformer {
...
 
   fun inmateStatusToPersonInfoApiStatus(status: InmateStatus?) = when (status) {
-    InmateStatus.IN -> PersonStatus.inCustody
-    InmateStatus.OUT -> PersonStatus.inCommunity
-    InmateStatus.TRN -> PersonStatus.inCustody
-    null -> PersonStatus.unknown
+    InmateStatus.IN -> PersonStatus.IN_CUSTODY
+    InmateStatus.OUT -> PersonStatus.IN_COMMUNITY
+    InmateStatus.TRN -> PersonStatus.IN_CUSTODY
+    null -> PersonStatus.IN_CUSTODY
```
davidatkinsuk, line 161: should be UNKNOWN

src/main/resources/static/_shared.yml
```diff
@@ -4833,10 +4833,10 @@ components:
...
         - person_risks_tier
       x-enum-varnames:
-        - duration
-        - expectedArrival
-        - createdAt
-        - applicationSubmittedAt
+        - DURATION
+        - EXPECTED_ARRIVAL
+        - CREATED_AT
+        - APPLICATION_SUBMITTED_AT
```
davidatkinsuk, line 4839: kind of weird there are only 4 in `x-enum-varnames` but more in `enum`. Any idea what happened here?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApplicationEntity.kt
```diff
@@ -460,7 +460,7 @@ class ApprovedPremisesApplicationEntity(
       else -> {}
     }
 
-    if (noticeType == Cas1ApplicationTimelinessCategory.emergency || noticeType == Cas1ApplicationTimelinessCategory.shortNotice) {
+    if (noticeType == Cas1ApplicationTimelinessCategory.EMERGENCY || noticeType == Cas1ApplicationTimelinessCategory.SHORT_NOTICE) {
```
davidatkinsuk, line 463: this is going to be an issue because we've used API types in the persistence model, meaning we'll have lowercase values in the database that will no longer resolve properly after this change. I think we probably need to 'fix' this by using an internally managed enumerated type. Let's discuss

davidatkinsuk: there's more of these that are less obvious because nothing has changed in the class e.g. Booking.status uses an API enum, i'm sure more do. These will all need fixing (sad face)

PR #2626, opened 2024-12-02
Convert seed jobs to spring beans

davidatkinsuk:
This simplifies the `SeedService` as it no longer has to manually inject dependencies.


danhumphreys-moj (approved):
Looks good.  I'll do the same for the characteristic seeding.

Makes me wonder why we didn't do this in the first place? 🤔

davidatkinsuk:
> Looks good. I'll do the same for the characteristic seeding.
> 
> Makes me wonder why we didn't do this in the first place? 🤔

Yes i'm not sure why it was done this way to be honest. I've had to do quite a bit of work to unpick the code to get it to this stage. I guess at some point seed jobs maybe retained state internally so we couldn't use singletons, although I did check each seed job and none are using instance-level internal state

PR #2633, opened 2024-12-03
APS-1613: Removed CAS1_LEGACY_MANAGER role from service

RobBoothMOJ:
Removed CAS1_LEGACY_MANAGER role from service
Removed role from API 
Removed LegacyManagerUserAccess Test


davidatkinsuk (approved):
Looks great, thankfully this is a lot more straight forward than removing `MANAGER` :)

PR #2635, opened 2024-12-03
CAS-1260 - add assessment ID to application response

daveawc:
There is work to enable a referrer to see non-user notes from the assessment timeline, so they need to have the assessment ID.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ApplicationTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ApplicationTest.kt
@@ -1307,6 +1307,64 @@ class ApplicationTest : IntegrationTestBase() {
     }
   }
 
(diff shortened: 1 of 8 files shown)


vbala-moj (approved):

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ApplicationTest.kt
```diff
@@ -1307,6 +1307,64 @@ class ApplicationTest : IntegrationTestBase() {
     }
   }
 
+  @Test
+  fun `GET submitted CAS3 application includes assessmentId in the response`() {
```
davidatkinsuk, line 1311: Could this be added to an existing 'generic' integration test? Arguably i'm not sure we need an integration test given we have a unit test too

davidatkinsuk (commented):
one minor comemnt from me

PR #2646, opened 2024-12-05
APS-1658 - Add ability to record a space booking id against a timeline note

davidatkinsuk:
This PR removes `ApplicationService. addNoteToApplication` and redirects all calls to the `ApplicationTimelineNoteService`, as the extra layer of indirection had no value and complicates changes

It then adds `cas1_space_booking_id` to the timeline note table


RobBoothMOJ (approved):
Approved

vbala-moj (approved):

PR #2653, opened 2024-12-05
APS-1642 - Remove staffIdentifier from domain events

davidatkinsuk:
Receivers (ap-and-delius-context) are no longer expecting domain events `staff member` to include a staff identifier value. This commit removes that value as it is not required.


gregkhawkins (approved):

gregkhawkins:
LGTM!

PR #2655, opened 2024-12-06
APS-1659 CAS1 update event number job should update space bookings

davidatkinsuk:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/SeedColumns.kt
```diff
@@ -12,6 +12,10 @@ data class SeedColumns(
 
   fun uuidOrNull(label: String) = columns[label]?.let { UUID.fromString(it.trim()) }
 
+  fun intOrNull(label: String) = columns[label]?.toIntOrNull()
+
+  fun longOrNull(label: String) = columns[label]?.toLongOrNull()
```
vbala-moj, line 17: The code `columns [label]? Used repeatedly so we could extract it into a function and use it instead.

Something like `    private fun getTrimmedStringValue(label: String): String? = columns[label]?.trim()`

vbala-moj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/SeedColumns.kt
vbala-moj: Also, all these functions return something, so it could be getXXX().

PR #2657, opened 2024-12-06
Chore/make characteristics consistent across all envs

danhumphreys-moj:
Make characteristics consistent across all environments.

The following discrepancies were identified:

1. preprod/prod: isSingle inserted into characteristics table presumably manually/via a seed job since it is not inserted when populating characteristic data is populated with [20230310000608__update_approved_premises_characteristics.sql](https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/df39c58095037bda4f3031d8867b0d02da7201fe/src/main/resources/db/migration/all/20230310000608__update_approved_premises_characteristics.sql#L25C92-L25C128).
2. preprod/prod: the name for isSingle is incorrect.  Is "Is this bed in a single room?Is this bed in a single room?" and should be "Is this bed in a single room?".
3. local/dev/test: there are two ground floor related properties "isGroundFloor" and "IsGroundFloor".  "IsGroundFloor" is redundant.
4. integration: the right number of characteristics are in the database but "IsGroundFloor" should be "isGroundFloor".
5. integration: the name for isSingle is incorrect, currently "Single room" and should be "Is this a single room?".

To resolve these discrepancies we need to:

1. Remove population of characteristic data from individual environments and use the sql files in db/migration/all.
2. Update IsGroundFloor to be isGroundFloor in local/integration/dev/test (correct in preprod/prod).
3. Change the isSingle name to "Is this bed in a single room?" in preprod/prod.
4. Insert the missing isSingle characteristic into local/integration/dev/test (exists in preprod/prod).
5. Remove redundant isSingle creation for integration env.


davidatkinsuk (approved):
LGTM, thanks for the details explanation 👍

PR #2662, opened 2024-12-09
Updated BedSearchAttribute enum

daveawc:
updated the BedSearchAttribute enum to use capital letters, and have the mapping as the same property name from the database. This will mean easier searching/filtering. The previous values remain until the UI updates, at which point they will be removed.


davidatkinsuk (approved):
LGTM 👍

PR #2663, opened 2024-12-09
chore/add hmpps-domain-events-topic to namespace_secrets in test env.

danhumphreys-moj:


danhumphreys-moj:
Thought that this would resolve issues deploying to the test environment because of the following warning: 

`28m         Warning   Failed              pod/hmpps-approved-premises-api-585d894b47-zbqsj             Error: secret "hmpps-domain-events-topic" not found`

However this is a red herring.  And the configuration as it stands is correct since we don't want to write to domain events topics in test.

davidatkinsuk:
closing this as test environment is now working

PR #2668, opened 2024-12-09
Remove integration-only migrations

davidatkinsuk:
This commit removes migrations specific to integraton (and inversely migrations that apply to all environments other than integration).

This separation was presumably added to ensure integration tests weren’t significantly slowed down by inserting postcode data. As we now use a template database to duplicate the integration database between tests, this is no longer a significant concern.


davidatkinsuk:
This currently fails because integration tests attempt to insert duplicate test data that previously didn't exist

davidatkinsuk:
closing this for now as additional work is required to do this

PR #2669, opened 2024-12-10
Removed community api client and its configuration.

vbala-moj:
PR#3 for [https://dsdmoj.atlassian.net/browse/APS-1650](url)


vbala-moj:
> also the OffenceFactory is no longer used, so that can go, plus can you take a look at the models in `model/community`? In theory they should all be able to be removed, or moved into `model/deliuscontext` if they are still used image

Intelij hangs when trying to move OffenderDetailsSummary, so I left that one.

davidatkinsuk (approved):
LGTM, great to see lots of code being removed.

Probably best to make sure @daveawc  is happy too before merging 👍

vbala-moj:
> Looks good!
> 
> Is there a plan for the OffenderDetailSummary that's now deprecated? I noticed this is the only remaining communtiy-api model, so would be good to get away from that. I appreciate that looks like a pretty big task though, and should be done separately

I tried to remove it, but it was all over the place, so I left it. I will do that part on a separate ticket.

PR #2671, opened 2024-12-10
Cas2bail/spike1

tobybatchmoj:
This is the second method of adding bail support to the CAS2 application without duplicating the controller/service/entity.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas2/ApplicationsTransformer.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas2/ApplicationsTransformer.kt
@@ -58,6 +58,7 @@ class ApplicationsTransformer(
         crn = jpaSummary.crn,
         nomsNumber = jpaSummary.nomsNumber,
         personName = personName,
+        applicationOrigin = jpaSummary.applicationOrigin,
       )
   }
 
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas2/ApplicationsTransformer.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas2/ApplicationsTransformer.kt
@@ -2,8 +2,10 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.cas2
 
 import com.fasterxml.jackson.databind.ObjectMapper
 import org.springframework.stereotype.Component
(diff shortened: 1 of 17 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas2/ApplicationsTransformer.kt
```diff
@@ -69,6 +71,13 @@ class ApplicationsTransformer(
     return ApplicationStatus.inProgress
   }
 
+  private fun getApplicationOriginFromJpa(summary: Cas2ApplicationSummaryEntity): ApplicationOrigin {
+    return when {
+      summary.applicationOrigin == "bail" -> ApplicationOrigin.bail
+      else -> ApplicationOrigin.court
```
adamhughes86, line 77: I think this should be `else -> ApplicationOrigin.hdc` as at this point it's which service is it coming from HDC or Bail

adamhughes86:
As a proof of concept this looks good. It's nice and simple and we can see the steps involved in adding one field to the main application data.

For anyone looking into the actual ApplicationOrigin data, this will still need to be considered

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas2/ApplicationsTransformer.kt
williamfalconeruk: would it be good to differentiate the bail types here also?

src/main/resources/static/_shared.yml
```diff
@@ -4921,3 +4931,8 @@ components:
         - sharedProperty
         - singleOccupancy
         - wheelchairAccessible
+    ApplicationOrigin:
+      type: string
+      enum:
+        - bail
```
williamfalconeruk, line 4937: as per previous comment - perhaps expand these to include hdc, and better differentiate the 2 types of bail e.g. `prisonBail,courtBail,hdc` ?

src/main/resources/static/codegen/built-api-spec.yml
```diff
@@ -9204,3 +9214,8 @@ components:
         - sharedProperty
         - singleOccupancy
         - wheelchairAccessible
+    ApplicationOrigin:
+      type: string
+      enum:
```
williamfalconeruk, line 9219: as per previous comment on hdc / bail enums

src/main/resources/static/_shared.yml
Rosie-Brigham: agree - though lets use `homeDetentionCurfew` instead of `hdc` as it's clearer

PR #2672, opened 2024-12-10
APS-1654: Space Search filter premises on application gender

RobBoothMOJ:
Space Search should filter premises based upon application's associated gender. 

`Cas1SpaceSearchRequirements` 
- `applicationId` is now mandatory 
- `genders` is deprecated

`Cas1SpaceSearchService.findSpaces`
- Gender is obtained from the associated application (via `applicationId`) `is_womens_application`
- modified the filter query to apply gender filter using `approved_premises.gender`

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1SpaceSearchService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1SpaceSearchService.kt
@@ -1,15 +1,18 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas1
 
+import org.springframework.data.repository.findByIdOrNull
 import org.springframework.stereotype.Service
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Cas1SpaceSearchParameters
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Gender
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ServiceName
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApprovedPremiseApplicationRepository
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.CharacteristicEntity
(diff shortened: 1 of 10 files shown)


vbala-moj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1SpaceSearchService.kt
```diff
@@ -19,11 +24,23 @@ import java.util.UUID
 class Cas1SpaceSearchService(
   private val characteristicService: CharacteristicService,
   private val spaceSearchRepository: Cas1SpaceSearchRepository,
+  private val applicationRepository: ApplicationRepository,
 ) {
   fun findSpaces(searchParameters: Cas1SpaceSearchParameters): List<Cas1SpaceSearchResult> {
+    val applicationId = searchParameters.applicationId
+    val application = applicationRepository.findByIdOrNull(searchParameters.applicationId)
+      ?: throw NotFoundProblem(applicationId, "Application")
+    if (application !is ApprovedPremisesApplicationEntity) {
```
davidatkinsuk, line 33: Optional - If we added an `ApprovedPremiseApplicationRepository` we wouldn't need to do the type check. We've done something similar for `ApprovedPremisesRepository`

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/givens/GivenAnApplication.kt
```diff
@@ -8,19 +8,23 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.randomInt
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.randomStringMultiCaseWithNumbers
 import java.time.OffsetDateTime
 
+@Suppress("LongParameterList")
 fun IntegrationTestBase.givenACas1Application(
   createdByUser: UserEntity,
   crn: String = randomStringMultiCaseWithNumbers(8),
   submittedAt: OffsetDateTime? = null,
   eventNumber: String = randomInt(1, 9).toString(),
+  isWomensApplication: Boolean = false,
```
davidatkinsuk, line 17: i think default should probably be null on this (as the default for 'submittedAt' is also null)

PR #2677, opened 2024-12-11
Remove camel case value from BedSearchAttribute enum

daveawc:
The UI has been updated to use the correct enum values, so we can remove these.


davidatkinsuk (approved):
LGTM 👍

PR #2681, opened 2024-12-11
Add `FullPersonSummary.isRestricted` to the API

davidatkinsuk:
This allows the UI to determine if an offender is an LAO, even if they have access to the offender. This matches a similar property on `FullPerson`


muhammad-elabdulla (approved):
LGTM

PR #2682, opened 2024-12-11
fix flaky booking search test

daveawc:
changed this test to set a precise name as part of the test to use for filtering, as this was only filtering on the last letter of the first name and first letter of the last name, this could return 1+ causing a failure. We can now filter on more characters/


davidatkinsuk (approved):
LGTM 👍

vbala-moj (approved):

PR #2685, opened 2024-12-12
Limit space booking characteristics in the API

davidatkinsuk:
This commit updates the create space booking api to make use of the new `Cas1SpaceBookingCharacteristic` type, which limits space bookings to 6 characteristics.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas1/Cas1SpaceBookingRequirementsTransformer.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas1/Cas1SpaceBookingRequirementsTransformer.kt
@@ -1,17 +1,17 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.cas1
 
 import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Cas1SpaceBookingCharacteristic
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Cas1SpaceBookingRequirements
-import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Cas1SpaceCharacteristic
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas1/Cas1SpaceBookingRequirementsTransformer.kt
```diff
@@ -1,17 +1,17 @@
...
 class Cas1SpaceBookingRequirementsTransformer {
   fun transformJpaToApi(cas1SpaceBookingEntity: Cas1SpaceBookingEntity) = Cas1SpaceBookingRequirements(
-    essentialCharacteristics = cas1SpaceBookingEntity.criteria.map { it.asCas1SpaceCharacteristic() },
+    essentialCharacteristics = cas1SpaceBookingEntity.criteria.map { it.asCas1SpaceBookingCharacteristic() },
   )
 
-  private fun CharacteristicEntity.asCas1SpaceCharacteristic() =
-    Cas1SpaceCharacteristic.valueOf(this.propertyName!!)
+  private fun CharacteristicEntity.asCas1SpaceBookingCharacteristic() =
+    Cas1SpaceBookingCharacteristic.entries.first { it.value == this.propertyName }
```
danhumphreys-moj, line 16: Presumably this will always find a value in `Cas1SpaceBookingCharacteristic.entries` i.e. it will never return null?

danhumphreys-moj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas1/Cas1SpaceBookingRequirementsTransformer.kt
davidatkinsuk: Yes it should always fine one. If it doesn't we've either messed up the reference data or API definition, so null pointer is probably a suitable error as it's something that'll need addressing immediately

PR #2686, opened 2024-12-12
updated pathParams template

daveawc:
updated so that enum values are not generated with extra escaped double quotes


davidatkinsuk (approved):
LGTM 👍

PR #2687, opened 2024-12-12
Cas2bail/cba56 add application origin

garethCAS2:
1. We've added applicationOrigin to the Cas2ApplicationEntity and Cas2ApplicationSummaryEntity. 
2. There is an API enum which has three values: courtBail, prisonBail, homeDetentionCurfew
3. The createApplication function in ApplicationService has a nullable parameter (applicationOrigin) which has a default value of null so as not to interfere with the current Cas2 service.
4. We added a migration to alter the table and summary views.
5. Added a some integration and unit tests.


adamhughes86:
ADR for this change - https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/5338431579/ADR+CAS2+Data+Change+-+Application+Origin

PR #2689, opened 2024-12-12
Arrival and Departure domain events displaying on timeline

RobBoothMOJ:
Corrects issue where the `occurredAt` timestamp was being set to the arrival date / departure date rather than the system's  timestamp. 
Updated the timeline output (DomainEventDescriber) to include the time in the generated output, to display, for an arrival / departure.


davidatkinsuk (approved):
perfect 👍

PR #2690, opened 2024-12-12
APS-1713 - Simplify CAS1 Premise Permissions

davidatkinsuk:
The user interface does not currently reference the removed, so they are safe to change and immediately deploy to prod


RobBoothMOJ (approved):
Approved

PR #2694, opened 2024-12-13
APS-1717 - Add CRU Member Enable OOSB Roles

davidatkinsuk:
This is a temporary role that will be used to enable out of service beds during rollout of the functionality.


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/Cas1OutOfServiceBedTest.kt
```diff
@@ -107,8 +107,14 @@ class Cas1OutOfServiceBedTest : InitialiseDatabasePerClassTestBase() {
         .isUnauthorized
     }
 
+    val test = arrayOf("blah")
```
gregkhawkins, line 110: is this used?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/transformer/UserTransformerTest.kt
```diff
@@ -195,39 +195,6 @@ class UserTransformerTest {
...
-      )
-
-      every { apAreaTransformer.transformJpaToApi(any()) } returns apArea
-
-      val result =
-        userTransformer.transformJpaToApi(user, approvedPremises) as ApprovedPremisesUser
-
-      assertThat(result.permissions).isEmpty()
-    }
-
```
gregkhawkins, line 230: why no need for this anymore?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/Cas1OutOfServiceBedTest.kt
davidatkinsuk: oops! i was trying something out and shouldn't have checked that in. good spot, i've removed it.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/transformer/UserTransformerTest.kt
davidatkinsuk: I don't think the test is really adding any value and we're constantly having to update it when we change permissions on roles

gregkhawkins (approved):

PR #2699, opened 2024-12-13
Updated code to return timeline events based on lao permission.

vbala-moj:
Timeline events based on lao permissions - [https://dsdmoj.atlassian.net/browse/APS-1671](url)


davidatkinsuk (approved):
Perfect! And thanks for tidying up the controller function 👍

PR #2702, opened 2024-12-13
APS-1213 - Remove CAS1_ADMIN Role

davidatkinsuk:
This is no longer in use by the UI so can be completely removed. Any users using this role in production that require access to user managment have already been assigned the user management role.


RobBoothMOJ (approved):
Approved

PR #2705, opened 2024-12-13
updating the hmpps dependencies.

daveawc:
updated gradle-spring-boot to 6.1.0
updated mpps-kotlin-spring-boot-starter to 1.1.0

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,7 +1,7 @@
 import org.apache.commons.io.FileUtils
 
 plugins {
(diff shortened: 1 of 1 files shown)


davidatkinsuk (commented):
LGTM. I'll defer approval until we've checked with Rosie on extent of testing we want to do for this. Given the minor upgrade I think relying on E2E tests is fine.

Rosie-Brigham (approved):
Agree with David on this - lets make sure that when it gets to pre-prod we have a good check to see that everything is ok before deploying to production however

PR #2706, opened 2024-12-13
Updates to the swagger config

daveawc:
Updates to the swagger config -

- set the generated source sets folder globally rather than just with the compile task
- add temporary dummy controllers for cas1 and cas3 to force create enums (as this uses allowedValues strings, so enums are not generated)
- added config to order documentation and prevent unused models being removed
- updated filtering on grouping


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas1/Cas1DummyController.kt
```diff
@@ -0,0 +1,57 @@
...
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.TaskSortField
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.UserSortField
+
+@RestController
+@RequestMapping(path = ["/cas1"])
+class Cas1DummyController {
+  private val log = LoggerFactory.getLogger(this::class.java)
+
+  @Operation(
+    description = "This is a dummy endpoint that adds models missing from autogenerated swagger documentation. This is to keep the UI happy.",
```
davidatkinsuk, line 29: presumably this is temporary. Given this will appear in the swagger it may be worth mentioning this? e.g.

`This is a dummy endpoint that adds models missing from autogenerated swagger documentation that are expected by the UI. This will be removed once we remove controller auto generation`

Same for other dummy controller

davidatkinsuk (approved):
LGTM, one minor documentation proposal

PR #2710, opened 2024-12-16
Removing unnecessary wrapper classes

daveawc:
these are just wrappers that create a class of `List<OASysQuestions>`, which is what's generated from the model. They are not used in the API, and the UI has been updated to user Array<OASysQuestion> instead. 

This removes a lot of unnecessary code, but will require a UI change. 

This is how the OASysSections class is generated with these changes

image

This is how it is generated as is in main: 

image


davidatkinsuk (approved):
LGTM 👍

PR #2718, opened 2024-12-17
Update add_new_endpoint.md

rossjones-moj:
Updates the section on security configuration as it now uses the the authorization DSL and not `mvcMatchers`


davidatkinsuk (approved):
Thanks for doing this, LGTM 👍

rossjones-moj:
Unfortunately, I can't merge as I don't have permission, so would appreciate if you could do that for me whenever is convenient.

PR #2720, opened 2024-12-17
APS-1727 - Optionally exclude booking from capacity calculations

davidatkinsuk:
This commit adds the option to exclude a given space booking id when calculating a premises occupancy. This will be used when retrieving capacity as part of amending the referenced booking.


RobBoothMOJ (approved):
Approved

gregkhawkins (approved):

PR #2724, opened 2024-12-18
APS-1730 remove cas1 from bookings extensions endpoint

davidatkinsuk:
This is not currently in use by CAS1 and will be replaced by space booking amendments


muhammad-elabdulla (approved):
LGTM

PR #2730, opened 2024-12-18
Remove unused availability info from space search

davidatkinsuk:
When the Space Search Service was originally written, it was expected that the results would include availability for each returned premise. This is no longer the case and is managed by a separate API.

This commit removes the ‘placeholder’ code added to provide capacity, simplifying the response from the space search service.

It doesn’t make any functional changes to the code


RobBoothMOJ (approved):
Approved

PR #2735, opened 2024-12-19
APS-1746 add premise site survey import seed job

davidatkinsuk:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/Cas1SeedPremisesFromSiteSurveyXlsxJob.kt
```diff
@@ -0,0 +1,289 @@
...
+      point = premisesInfo.point
+      gender = siteSurvey.maleFemale.toApprovedPremisesGender()
+    }
+
+    existingPremise.characteristics.clear()
+    existingPremise.characteristics.addAll(premisesInfo.characteristicEntities)
+
+    val afterChange = ApprovedPremisesForComparison.fromEntity(existingPremise)
+
+    val diff = javers.compare(beforeChange, afterChange)
```
danhumphreys-moj, line 193: Nice - looks like a useful library

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/Cas1SeedPremisesFromSiteSurveyXlsxJob.kt
```diff
@@ -0,0 +1,289 @@
...
+        status = entity.status,
+        point = entity.point,
+        gender = entity.gender,
+        supportsSpaceBookings = entity.supportsSpaceBookings,
+        managerDetails = entity.managerDetails,
+      )
+    }
+  }
+
+  @SuppressWarnings("TooGenericExceptionThrown")
```
danhumphreys-moj, line 261: There is a new `SiteSurveyImportException` class which might be applicable here.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/Cas1SiteSurveyPremiseFactory.kt
```diff
@@ -0,0 +1,160 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.seed.cas1
+
+import org.jetbrains.kotlinx.dataframe.AnyCol
+import org.jetbrains.kotlinx.dataframe.DataFrame
+import org.jetbrains.kotlinx.dataframe.api.getColumn
+import org.jetbrains.kotlinx.dataframe.name
+
+class Cas1SiteSurveyPremiseFactory {
```
danhumphreys-moj, line 8: We should be using factories for the room characteristic job too..

danhumphreys-moj (approved):
LGTM!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/Cas1SiteSurveyPremiseFactory.kt
davidatkinsuk: Yep, it helps split the logic up. Also this should make it easy for you to grab the qcode in that job from the first worksheet, with a little bit of refactoring

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/Cas1SeedPremisesFromSiteSurveyXlsxJob.kt
davidatkinsuk: I noticed that but i didn't really see the value given the seed service is just catching all exceptions anyway. I find `error("asdasd")` a more fluent way of raising unrecoverable errors.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/Cas1SeedPremisesFromSiteSurveyXlsxJob.kt
davidatkinsuk: Yes. If we refactored the room seed job to use an internal model before persisting (some data classes) we could then load the current entities into the same internal classes and generate a diff. This is less important in the short term as SWSC don't have any rooms defined to modify

PR #2743, opened 2024-12-23
APS-1768 - Allow Space Booking creation for LAOs

davidatkinsuk:
Before this commit a RestrictedPerson would be returned when retreiving a placement request, even if the calling user has the ‘Limited Access’ qualification.

This commit changes the PlacementRequest qualification to consider this qualification. This subsequently allows space bookings to be created for LAOs where the user has the ‘limited access’ qualification

To support this a new version of `OffenderService. getPersonInfoResult` is added that accepts the LimitedAccessStrategy, with the existing functions being deprecated.


RobBoothMOJ (approved):
Approved

PR #2746, opened 2024-12-24
Remove 'AnyValue' component.

daveawc:
This generates in kotlin as 'Any', which is then generated in documentation as 'Unit' by springdoc. The UI has been changed to reflect this, for when we switch over to autogenerated documentation


muhammad-elabdulla (approved):
LGTM

PR #2750, opened 2024-12-31
APS-1706: Add validation to ensure arrivals can only be in the past

RobBoothMOJ:
Validates presence of date and time 
Validates that the date and time must be in the past.


davidatkinsuk (approved):
just one minor comment from me

PR #2752, opened 2025-01-02
CAS-1325 Update the Create Arrival validation message in BookingService for CAS3

muhammad-elabdulla:
This PR CAS-1325 is to update the Create Arrival validation message in BookingService when the booking premises is Temporary accommodation


davidatkinsuk (approved):
LGTM although it's worth noting that we can remove createArrival/createDeparture in BookingService now as CAS1 doesn't use these (we only support this for space bookings)

davidatkinsuk:
Sorry ignore my previous comment, we appear to have a `NewCas2Arrival` type, so maybe CAS2 are using arrivals too (although I don't think CAS2 uses bookings so this could be unused code?)

muhammad-elabdulla:
@davidatkinsuk I was waiting to check with you if CAS1 still need them as I saw that the endpoint are not been used for last two months

PR #2755, opened 2025-01-02
APS-1733 - Simplify XLSX Seeding API Calls

davidatkinsuk:
Before this PR we had to provide the premises ID when updating/creating rooms from the site survey XLSX. This commit changes the ‘seed XLSX’ API, removing the need to provide the premises ID and instead finding the premises using the QCode defined within the spreadsheet itself

### DIFF
--- a/script/run_seed_from_excel_job
+++ b/script/run_seed_from_excel_job
@@ -6,12 +6,20 @@
 
 set -e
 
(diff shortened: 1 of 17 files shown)


RobBoothMOJ (approved):
Approved

PR #2758, opened 2025-01-03
APS-1773 - Placement Request with Space Booking should not appear as be unable to match

davidatkinsuk:
Before this commit if a placement request had been marked as ‘unable to match’ but then previously had a space booking created, it was still being returned on the placement request dashboard as unable to match.

This commit also improves the placement request dashboard API tests cover additional scenarios, improving regression testing


muhammad-elabdulla (approved):
LGTM

PR #2760, opened 2025-01-03
APS-1655 - Removed unused Cas1SpaceSearchResult.spacesAvailable

davidatkinsuk:


davidatkinsuk:
I'm closing this as we're going to look at creating a new version of this API, which won't include this

PR #2765, opened 2025-01-06
APS-1752 - Add get person info in batches function

davidatkinsuk:
Add a function to get offender info in batches and update Cas3ReportService to use this function for batching requests.

This is being added as batching functionality will be required for some new CAS1 seed jobs

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas3/Cas3ReportServiceTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas3/Cas3ReportServiceTest.kt
@@ -72,7 +72,7 @@ class Cas3ReportServiceTest {
 
     every { mockTransitionalAccommodationReferralReportRowRepository.findAllReferrals(any(), any(), any()) } returns listOf(testTransitionalAccommodationReferralReportData)
     every { mockUserService.getUserForRequest() } returns UserEntityFactory().withUnitTestControlProbationRegion().produce()
-    every { mockOffenderService.getPersonSummaryInfoResults(any<Set<String>>(), any()) } returns listOf(
+    every { mockOffenderService.getPersonSummaryInfoResultsInBatches(any<Set<String>>(), any(), batchSize = 2) } returns listOf(
       PersonSummaryInfoResult.Success.Full("", CaseSummaryFactory().produce()),
     )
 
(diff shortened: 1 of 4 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas3/Cas3ReportServiceTest.kt
```diff
@@ -156,42 +156,6 @@ class Cas3ReportServiceTest {
     verify(exactly = 0) { mockOffenderService.getPersonSummaryInfoResults(any<Set<String>>(), any()) }
   }
 
-  @Test
-  fun `createCas3ApplicationReferralsReport successfully generate report and call offender service 2 times when crn search limit exceed`() {
-    val startDate = LocalDate.of(2024, 1, 1)
```
muhammad-elabdulla, line 161: This test should remain as it make sure that if the numbers of Crns exceed the bath size is called multiple times

muhammad-elabdulla (approved):
LGTM except the comment in the test

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas3/Cas3ReportServiceTest.kt
davidatkinsuk: i've updated the code so that batching is dealt with in the OffenderService, so this test is now redundant?

muhammad-elabdulla: In case there are 2000 crns the new code should split them to batches of 500

davidatkinsuk: This is what we're now doing in `OffenderService.getPersonSummaryInfoResultsInBatches` and i've added unit tests to ensure this happens (albeit with a batch size of 400). Given this, there is no more batching/partitioning happening in the `Cas3ReportService`, so we don't need to test this specific behaviour in `Cas3ReportServiceTest`, hence the removal of this test function. Given me a shout on slack if you think something needs changing here, i may be misunderstanding.

Note that the check in the new code relates to the requested batch size, _not_ the number of crns requested

```
    if (batchSize > MAX_OFFENDER_REQUEST_COUNT) {
      throw InternalServerErrorProblem("Cannot request more than $MAX_OFFENDER_REQUEST_COUNT CRNs. A batch size of $batchSize has been requested.")
    }
```

PR #2776, opened 2025-01-09
APS-1750 Add Excel jobs to how-to and add example files.

danhumphreys-moj:


vbala-moj (approved):

doc/how-to/run_seed_job_remotely.md
```diff
@@ -144,4 +158,23 @@ Required fields:
 Required fields:
 
 - `ap_area_identifier`
-- `email_address`
\ No newline at end of file
+- `email_address`
+
+## Excel reference
```
davidatkinsuk, line 163: i'm not sure we need instructions for specific jobs. i'd vote to remove all of them and move documentation onto the seed jobs themselves. I think these instructions will invariably become stale because we won't update them (as i can see some of the ones above have become)

doc/how-to/run_seed_job_remotely.md
```diff
@@ -144,4 +158,23 @@ Required fields:
...
+There are two seed from Excel jobs which can be run to import premise and room data form site survey Excel xlsx files:
+
+- `cas1_import_site_survey_premises`
+- `cas1_import_site_survey_rooms`
+
+`cas1_import_site_survey_premises` will import and update premise characteristics which are required to be on the first sheet of the site survey data (named "Sheet2").
+
+`cas1_import_site_survey_rooms` can be used to import and update room characteristics which are required to be on the second sheet of the site survey data (named "Sheet3").
+
+Rather than check in an Excel file an example of the format of the two worksheets which are required to be in the Excel file can be found here:
```
davidatkinsuk, line 176: why not upload a sample XLSX? (with dummy data of course)

doc/how-to/run_seed_job_remotely.md
```diff
@@ -144,4 +158,23 @@ Required fields:
 Required fields:
 
 - `ap_area_identifier`
-- `email_address`
\ No newline at end of file
+- `email_address`
+
+## Excel reference
+
```
davidatkinsuk, line 164: i think our room/premise xlsx jobs supersede the CSV versions - maybe we should mark the csv versions as deprecated? We can remove them once we've applied all site surveys in production

davidatkinsuk (commented):
A few thoughts from me

doc/how-to/run_seed_job_remotely.md
danhumphreys-moj: Removed instructions and added javadoc comments to new seed from Excel jobs.

doc/how-to/run_seed_job_remotely.md
danhumphreys-moj: Marked 2 x legacy seed from CSV jobs as deprecated.

doc/how-to/run_seed_job_remotely.md
danhumphreys-moj: My logic for using the CSV examples of the spreadsheet worksheets was to enable tracking text changes of premise and room characteristic data.  Checking in a binary Excel file won't enable us to do this.

Also I was apprehensive about checking in an Excel file which (if copied from a real site survey spreadsheet) could contain information in it which isn't easily determined (for example hidden sheets, or file property information which could potentially contain sensitive information).

Following a discussion, it was decided adding an example Excel file (or example CSV equivalents) is not required since it would be unlikely these would be maintained, and there is little benefit of checking them into a repository.  Example site survey Excel files can be found on the local network and through asking the CAS team.

On the same basis and following discussion with @davidatkinsuk, the example CSV files (and links to the files in the instructions) have also been removed.

PR #2777, opened 2025-01-09
CAS-1343 Delete CAS1 lost beds from lost_beds and lost_bed_cancellations tables

muhammad-elabdulla:
This PR CAS-1343 is to delete CAS1 lost beds from lost_beds and lost_bed_cancellations tables


davidatkinsuk (approved):
LGTM 👍

PR #2779, opened 2025-01-09
CAS-1343 Rename to void bedspace

muhammad-elabdulla:
This PR CAS1343 is to rename functions and variables to void bedspace


davidatkinsuk (approved):
LGTM, i'm not sure why the unit test is failing

muhammad-elabdulla:
Its because BookingAndPersons is referring to BookingSerivce when should be Cas3BookingService. I fixed it

PR #2781, opened 2025-01-09
APS-1787 add seed job to create applications with pending request for placements

davidatkinsuk:
Add a seed job to create test applications that either pending submission, or pending matching

This job uses co-routines when creating applications. limitedParallelism is set to 10 to ensure we don’t exceed the size of the JDBC connection pool. Otherwise we start seeing timeouts waiting on connections in the pool. Adding additional parallelism didn’t produce any notable gains, although that my not be the case in other better-sized environments.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/Cas1CreateTestApplicationsSeedJob.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/Cas1CreateTestApplicationsSeedJob.kt
@@ -1,0 +1,53 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.seed.cas1
+
+import org.slf4j.LoggerFactory
+import org.springframework.stereotype.Service
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.seed.SeedColumns
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.seed.SeedJob
+
+@Service
+class Cas1CreateTestApplicationsSeedJob(
+  val cas1ApplicationSeedService: Cas1ApplicationSeedService,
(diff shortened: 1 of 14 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/Cas1CreateTestApplicationsSeedJob.kt
```diff
@@ -0,0 +1,72 @@
...
+  @OptIn(ExperimentalCoroutinesApi::class)
+  override fun processRow(row: Cas1CreateTestApplicationsSeedCsvRow) {
+    val deliusUsername = row.creatorUsername
+    val crn = row.crn
+    val count = row.count
+    val state = row.state
+
+    log.info("Creating $count applications for CRN $crn in state $state using creator username $deliusUsername")
+
+    runBlocking(Dispatchers.IO.limitedParallelism(10)) {
```
vbala-moj, line 50: Other option to consider is to use CoroutineScope if we dont need to block the thread.

vbala-moj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/Cas1CreateTestApplicationsSeedJob.kt
davidatkinsuk: i was playing around with this and i think `CoroutineScope` only works if being called from a `suspend` function, which eventually will need a `runBlocking` somewhere further up the call chain. But this is all quite new to me so there may be a way to do it without that. I did verify that parallelism is occurring with this configuration. My initial attempt didn't specify a dispatcher and that led to the default dispatched being used which just ran one co-routine at a time.

PR #2782, opened 2025-01-09
Add diff when updating room to log output.

danhumphreys-moj:
For the seed room characteristics from Excel site survey data, output the difference in room characteristics between the existing room entity and the updated room entity usinng the [JaVers](https://javers.org/) library.  This is [already done](https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/14f7408f3eae062a5b55df9694fddafc23ff4ba6/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/Cas1SeedPremisesFromSiteSurveyXlsxJob.kt#L197) when a premise is updated via the equivalent job.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/Cas1SeedRoomsFromSiteSurveyXlsxJob.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/Cas1SeedRoomsFromSiteSurveyXlsxJob.kt
@@ -1,5 +1,8 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.seed.cas1
 
+import org.javers.core.Javers
+import org.javers.core.JaversBuilder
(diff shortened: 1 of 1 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/Cas1SeedRoomsFromSiteSurveyXlsxJob.kt
```diff
@@ -161,11 +169,39 @@ class Cas1SeedRoomsFromSiteSurveyXlsxJob(
     log.info("Created new room with code ${room.roomCode} and name ${room.roomName} in premise ${premises.name}.")
   }
 
+  private data class ApprovedPremisesRoomForComparison(
```
davidatkinsuk, line 172: nice

davidatkinsuk (approved):
LGTM, very neat!

PR #2783, opened 2025-01-10
Re-add ApprovedPremisesBedSearchResult

davidatkinsuk:
This commit temporarily re-adds ApprovedPremisesBedSearchResult until it can be fully removed from the UI code

It also removes ApprovedPremisesBedSearchParameters which was incorrectly re-added


muhammad-elabdulla (approved):
LGTM

PR #2785, opened 2025-01-10
Cas3 bed search cleanup

daveawc:
cleaning up the bed search stuff now that cas1 and cas3 have been split, which will help ongoing work in the area.
- renamed TemporaryAccommodationBedSearchResult to Cas3SearchResult for the api models (UI models in the open api spec still refer to TemporaryAccommodation...
- removed the inheritance to simplify the code so there's just a simple cas3 model
- CAS1 UI changes in this PR https://github.com/ministryofjustice/hmpps-approved-premises-ui/pull/2279
- CAS3 UI changes in this PR https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/1191


muhammad-elabdulla:
We need to be careful when migrate this PR because it will break the UI

src/main/resources/static/cas1-schemas.yml
```diff
@@ -1035,15 +1035,3 @@ components:
         - updatedReferenceNumber
         - updatedReason
         - updatedNotes
-    ApprovedPremisesBedSearchResult:
```
davidatkinsuk, line 1038: we need this

davidatkinsuk (changes requested):
please see my comment around `ApprovedPremisesBedSearchResult`

src/main/resources/static/cas1-schemas.yml
davidatkinsuk: we're now safe to remove this 👍

davidatkinsuk (approved):
LGTM

PR #2788, opened 2025-01-10
APS-1787 pass user into application & assess functions

davidatkinsuk:
This PR removes usage of `userService.getUserForRequest()` from various service-level functions, instead taking the current user as a function argument.

This allows the functions to be used by seed jobs, where a user isn’t set on the request context


muhammad-elabdulla (approved):
LGTM

PR #2793, opened 2025-01-13
Cas2v2 user entity

rossjones-moj:
Adds a new entity for Cas2v2 users where all user types are represented in a single entity (Cas2v2UserEntity) and differentiated by the `userType` field which is a `Cas2v2UserType` containing the different allowed user types for Cas2v2. 

As a result of the change, all of the Cas2v2 classes that contain users have been changed to use the new type instead of NomisUserEntity/UserEntity/ExternalUserEntity.  This removes a dependence on the Cas2 entities but should still allow a future migration to use the entities in this namespace. 

This PR also introduces a CAS2V2 `ServiceName` which is used to ensure that a Cas2v2Application is used instead of Application or Cas2Application when calling APIs.

DomainEvents remain as cas2 domain events with no change

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2v2/Cas2v2SubmissionsController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2v2/Cas2v2SubmissionsController.kt
@@ -11,11 +11,9 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.SortDirection
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.SubmitCas2v2Application
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.cas2v2.Cas2v2ApplicationEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.cas2v2.Cas2v2ApplicationSummaryEntity
-import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.ExternalUserService
-import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.HttpAuthService
-import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.NomisUserService
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas2.OffenderService
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas2v2.Cas2v2ApplicationService
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas2v2.Cas2v2UserService
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.cas2v2.Cas2v2SubmissionsTransformer
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.PageCriteria
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.ensureEntityFromCasResultIsSuccess
@@ -24,21 +22,14 @@ import java.util.UUID
 
 @Service("Cas2v2SubmissionsController")
 class Cas2v2SubmissionsController(
-  private val httpAuthService: HttpAuthService,
   private val cas2v2ApplicationService: Cas2v2ApplicationService,
   private val cas2v2SubmissionsTransformer: Cas2v2SubmissionsTransformer,
   private val offenderService: OffenderService,
-  private val externalUserService: ExternalUserService,
-  private val nomisUserService: NomisUserService,
+  private val userService: Cas2v2UserService,
 ) : SubmissionsCas2v2Delegate {
 
   override fun submissionsGet(page: Int?): ResponseEntity<List<Cas2v2SubmittedApplicationSummary>> {
-    val principal = httpAuthService.getCas2v2AuthenticatedPrincipalOrThrow()
-    if (principal.isExternalUser()) {
-      ensureExternalUserPersisted()
-    } else {
-      ensureNomisUserPersisted()
-    }
+    userService.ensureUserPersisted()
 
     val sortDirection = SortDirection.asc
     val sortBy = "submittedAt"
@@ -51,12 +42,7 @@ class Cas2v2SubmissionsController(
   }
 
   override fun submissionsApplicationIdGet(applicationId: UUID): ResponseEntity<Cas2v2SubmittedApplication> {
-    val principal = httpAuthService.getCas2v2AuthenticatedPrincipalOrThrow()
-    if (principal.isExternalUser()) {
-      ensureExternalUserPersisted()
-    } else {
-      ensureNomisUserPersisted()
-    }
+    userService.ensureUserPersisted()
 
     val applicationResult = cas2v2ApplicationService.getSubmittedCas2v2ApplicationForAssessor(applicationId)
     val application = extractEntityFromCasResult(applicationResult)
@@ -68,19 +54,11 @@ class Cas2v2SubmissionsController(
   override fun submissionsPost(
     submitCas2v2Application: SubmitCas2v2Application,
   ): ResponseEntity<Unit> {
-    val user = nomisUserService.getUserForRequest()
+    val user = userService.getUserForRequest()
     val submitResult = cas2v2ApplicationService.submitCas2v2Application(submitCas2v2Application, user)
     ensureEntityFromCasResultIsSuccess(submitResult)
 
     return ResponseEntity(HttpStatus.OK)
-  }
-
-  private fun ensureExternalUserPersisted() {
-    externalUserService.getUserForRequest()
-  }
-
-  private fun ensureNomisUserPersisted() {
-    nomisUserService.getUserForRequest()
   }
 
   private fun getPersonNamesAndTransformToSummaries(
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas2v2/Cas2v2UserEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas2v2/Cas2v2UserEntity.kt
@@ -1,0 +1,79 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.cas2v2
+
+import jakarta.persistence.Convert
+import jakarta.persistence.Entity
+import jakarta.persistence.EnumType
+import jakarta.persistence.Enumerated
+import jakarta.persistence.Id
+import jakarta.persistence.OneToMany
+import jakarta.persistence.Table
+import org.hibernate.annotations.CreationTimestamp
+import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.stereotype.Repository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.converter.StringListConverter
+import java.time.OffsetDateTime
+import java.util.UUID
+
+enum class Cas2v2UserType(val authSource: String) {
+  DELIUS("delius"),
+  NOMIS("nomis"),
+  EXTERNAL("auth"),
+  ;
+
+  companion object {
+    fun fromString(authSource: String): Cas2v2UserType {
+      return entries.first { it.authSource == authSource }
+    }
+  }
+}
+
+@Repository
+interface Cas2v2UserRepository : JpaRepository<Cas2v2UserEntity, UUID> {
+  fun findByUsername(username: String): Cas2v2UserEntity?
+  fun findByUsernameAndUserType(username: String, type: Cas2v2UserType): Cas2v2UserEntity?
+}
+
+@Entity
+@Table(name = "cas_2_v2_users")
+data class Cas2v2UserEntity(
+  @Id
+  val id: UUID,
+  val username: String,
+
+  // Cas2v2User interface implementation
+  var email: String?,
+  var name: String,
+
+  @Enumerated(EnumType.STRING)
+  var userType: Cas2v2UserType,
+
+  // Nomis specific fields that are only expected to have values if the
+  // accountType is Cas2v2UserType.NOMIS
+  var nomisStaffId: Long?,
+  var activeNomisCaseloadId: String? = null,
+
+  // Delius specific fields that are only expected to have values if the
+  // accountType is Cas2v2UserType.DELIUS
+  @Convert(converter = StringListConverter::class)
+  var deliusTeamCodes: List<String>?,
+  var deliusStaffCode: String?,
+
+  var isEnabled: Boolean,
+  var isActive: Boolean,
+
+  @CreationTimestamp
+  private val createdAt: OffsetDateTime? = null,
+
+  @OneToMany(mappedBy = "createdByUser")
+  val applications: MutableList<Cas2v2ApplicationEntity> = mutableListOf(),
+) {
+  override fun toString() = "CAS2V2 user $id"
+
+  fun staffIdentifier() = when (userType) {
+    Cas2v2UserType.NOMIS -> nomisStaffId?.toString() ?: error("Couldn't resolve nomis ID for user $id")
+    Cas2v2UserType.DELIUS -> deliusStaffCode ?: error("Couldn't resolve delius ID for user $id")
+    Cas2v2UserType.EXTERNAL -> ""
+  }
+
+  fun isExternal() = userType == Cas2v2UserType.EXTERNAL
+}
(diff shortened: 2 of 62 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2v2/Cas2v2UserAccessService.kt
```diff
@@ -1,21 +1,30 @@
...
+    } else if (user.userType == Cas2v2UserType.NOMIS) {
+      offenderIsFromSamePrisonAsUser(application.referringPrisonCode, user.activeNomisCaseloadId)
     } else {
-      offenderIsFromSamePrisonAsUser(application.referringPrisonCode, user.activeCaseloadId)
+      false
     }
   }
 
+  fun userCanAddNote(user: Cas2v2UserEntity, application: Cas2v2ApplicationEntity): Boolean = when (user.userType) {
+    Cas2v2UserType.DELIUS -> user.id == application.createdByUser.id || offenderIsFromSamePrisonAsUser(application.referringPrisonCode, user.activeNomisCaseloadId)
```
rossjones-moj, line 23: This should be NOMIS

src/main/resources/static/cas2v2-schemas.yml
```diff
@@ -264,4 +265,144 @@ components:
...
+        id:
+          type: string
+          format: uuid
+        name:
+          type: string
+          example: 'Roger Smith'
+        username:
+          type: string
+          example: 'SMITHR_GEN'
+        authSource:
```
davidatkinsuk, line 281: could this be enumerated?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas2v2/Cas2v2UserEntity.kt
```diff
@@ -0,0 +1,87 @@
...
+  companion object {
+    fun fromString(authSource: String): Cas2v2UserType {
+      return entries.first { it.authSource == authSource }
+    }
+  }
+}
+
+sealed interface User
+
+sealed interface Cas2v2User : User {
```
davidatkinsuk, line 32: Now we have a single User entity, i'm not sure this interface provides any value? I think it was originally added to allow the receiver to handle users derived from more than one entity. Should it be deprecated at least?

If you used single table strategy (as mentioned below, i agree this would still be required (and maybe enhanced?)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas2v2/Cas2v2UserEntity.kt
```diff
@@ -0,0 +1,87 @@
...
+}
+
+@Repository
+interface Cas2v2UserRepository : JpaRepository<Cas2v2UserEntity, UUID> {
+  fun findByUsername(username: String): Cas2v2UserEntity?
+  fun findByUsernameAndUserType(username: String, type: Cas2v2UserType): Cas2v2UserEntity?
+}
+
+@Entity
+@Table(name = "cas_2_v2_users")
```
davidatkinsuk, line 45: Have you considered using Single Table Strategy for this?  That way you can have specific sub types for each of the three different sources, making the specific properties and requirements on those properties a bit clearer? e.g. can make deliusTeamCodes and deliusStaffCode mandatory in the kotlin model for delius users

https://www.geeksforgeeks.org/jpa-single-table-strategy/

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2v2/Cas2v2UserService.kt
```diff
@@ -0,0 +1,148 @@
...
+  private fun getEntityForDeliusUser(username: String): Cas2v2UserEntity {
+    val deliusUser: StaffDetail =
+      when (val staffUserDetailsResponse = apDeliusContextApiClient.getStaffDetail(username)) {
+        is ClientResult.Success<*> -> staffUserDetailsResponse.body as StaffDetail
+        is ClientResult.Failure<*> -> staffUserDetailsResponse.throwException()
+      }
+
+    val existingUser = getExistingUser(username, Cas2v2UserType.DELIUS)
+    if (existingUser != null) {
+      val teamsDiffer = deliusUser.teamCodes() != existingUser.deliusTeamCodes
```
davidatkinsuk, line 94: will this work if codes are in a different order?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/SeedService.kt
```diff
@@ -90,6 +90,8 @@ class SeedService(
         SeedFileType.approvedPremisesImportDeliusReferrals -> getBean(Cas1ImportDeliusReferralsSeedJob::class)
         SeedFileType.approvedPremisesUpdateSpaceBooking -> getBean(Cas1UpdateSpaceBookingSeedJob::class)
         SeedFileType.temporaryAccommodationReferralRejection -> getBean(Cas3ReferralRejectionSeedJob::class)
+        SeedFileType.approvedPremisesBackfillActiveSpaceBookingsCreatedInDelius -> TODO()
+        SeedFileType.approvedPremisesCreateTestApplications -> TODO()
```
davidatkinsuk, line 94: `main` should have the implementation for this at the same point the enumeration was added. i'm not sure why this is now TODO() unless you lost the implementation on rebase?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/reference/Cas2v2ApplicationStatusSeeding.kt
```diff
@@ -0,0 +1,253 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.model.reference
+
+import java.util.UUID
+
+object Cas2v2ApplicationStatusSeeding {
```
davidatkinsuk, line 5: did we need a new version of `Cas2PersistedApplicationStatus`? This is alot of duplicated code and i don't _think_ it's needed?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/MigrationJobService.kt
```diff
@@ -53,6 +53,7 @@ class MigrationJobService(
         MigrationJobType.cas3ApplicationOffenderName -> getBean(Cas3UpdateApplicationOffenderNameJob::class)
         MigrationJobType.cas3DomainEventTypeForPersonDepartedUpdated -> getBean(Cas3UpdateDomainEventTypeForPersonDepartureUpdatedJob::class)
         MigrationJobType.cas1ApplicationsLicenceExpiryDate -> getBean(Cas1UpdateApplicationLicenceExpiryDateJob::class)
+        MigrationJobType.cas1BackfillOfflineApplicationName -> TODO()
```
davidatkinsuk, line 56: see previous comment - if you have the API type you should have the implementation as these are both on main

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2v2/Cas2v2ApplicationService.kt
```diff
@@ -336,9 +336,9 @@ class Cas2v2ApplicationService(
             conditionalReleaseDate = application.conditionalReleaseDate,
             submittedBy = Cas2ApplicationSubmittedEventDetailsSubmittedBy(
               staffMember = Cas2StaffMember(
-                staffIdentifier = application.createdByUser.nomisStaffId,
+                staffIdentifier = application.createdByUser.nomisStaffId ?: 0,
                 name = application.createdByUser.name,
-                username = application.createdByUser.nomisUsername,
+                username = application.createdByUser.username,
```
davidatkinsuk, line 341: are we happy that the receiver will be ok with non-nomis usernames here? it's probably worth adding user type to the domain event definition even if receiver doesn't use it so we can make use of this information in the future

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2v2/Cas2v2UserAccessService.kt
```diff
@@ -1,21 +1,34 @@
...
       false
+    } else if (user.userType == Cas2v2UserType.NOMIS) {
+      offenderIsFromSamePrisonAsUser(application.referringPrisonCode, user.activeNomisCaseloadId)
     } else {
-      offenderIsFromSamePrisonAsUser(application.referringPrisonCode, user.activeCaseloadId)
+      false
     }
   }
 
+  fun userCanAddNote(user: Cas2v2UserEntity, application: Cas2v2ApplicationEntity): Boolean = when (user.userType) {
```
davidatkinsuk, line 22: nice

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2v2/Cas2v2SubmissionsController.kt
```diff
@@ -68,19 +60,15 @@ class Cas2v2SubmissionsController(
...
     return ResponseEntity(HttpStatus.OK)
   }
 
-  private fun ensureExternalUserPersisted() {
-    externalUserService.getUserForRequest()
-  }
-
-  private fun ensureNomisUserPersisted() {
-    nomisUserService.getUserForRequest()
+  private fun ensureUserPersisted() {
```
davidatkinsuk, line 70: (this is an existing issue) - i don't like how we have 'get' functions in user service that can have side effects (e.g. adding a user to the database). I believe the `ensureUserPersisted` function was added here to clarify intention. Could this be moved into your user service so we can use it more widely where applicable?

davidatkinsuk (changes requested):
The approach looks good, i've added some comments that need addressing

Also, i've had a chat with the dev lead and we both agree that we're going to have to revert the formatting changes somehow. I reckon more than 70% of this PR is irrelevant formatting changes which is making it really hard to review in detail. I'd like to re-review this once formatting changes have been reverted. Also note that as this is going into a 'mega PR' these unnecessary changes are going to make it hard to 'sanity check' before merging into main

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/SeedService.kt
rossjones-moj: The implementation wasn't in cas2v2/dev when I branched I think although I'm not sure how I ended up with the two new branches in that case.  I'll take a look and see if I can find them.

src/main/resources/static/cas2v2-schemas.yml
rossjones-moj: Updated in 4ff7a5b78

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas2v2/Cas2v2UserEntity.kt
```diff
@@ -0,0 +1,87 @@
...
+  var isEnabled: Boolean,
+  var isActive: Boolean,
+
+  @CreationTimestamp
+  private val createdAt: OffsetDateTime? = null,
+
+  @OneToMany(mappedBy = "createdByUser")
+  val applications: MutableList<Cas2v2ApplicationEntity> = mutableListOf(),
+) : Cas2v2User {
+  override fun toString() = "CAS2 $userType bail user $id"
```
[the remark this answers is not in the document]
rossjones-moj: Updated in e029dd05b

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas2v2/Cas2v2UserEntity.kt
rossjones-moj: Updated in e029dd05b. This wasn't actually used anywhere except this file, I think perhaps confusion with the api model with the same name.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas2v2/Cas2v2UserEntity.kt
rossjones-moj: Looked at this originally, and I think the added complexity made me pause, and my confusion around the ADR which mentions it (in relation to the original CAS2 work) made me take this simpler approach. The use of the entity is localised to a single service and so hopefully this will be okay until we've more time to look at this approach?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/reference/Cas2v2ApplicationStatusSeeding.kt
rossjones-moj: Removed with associated/dependent files in a06b64463

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2v2/Cas2v2SubmissionsController.kt
rossjones-moj: Updated in 64a7c011c

Even with this move it feels like a bad thing to be doing, and something that perhaps should have happened by the time of an authz check (if not authn check).

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2v2/Cas2v2ApplicationService.kt
rossjones-moj: Added in e57f70745 and default the cas2 sending to nomis as cas2 only supports nomis users.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/SeedService.kt
rossjones-moj: It appears I'd got the schema enum values from somewhere but not the actual implementations so I've put the schema back to how it is in the base branch (cas2v2/dev).

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/MigrationJobService.kt
rossjones-moj: As mentioned above, was an incorrect schema but now fixed.

src/main/resources/db/migration/all/20250122155323__cas2v2_user_entity.sql
```diff
@@ -0,0 +1,56 @@
+
+CREATE TABLE cas_2_v2_users
+(
+    id                       UUID    NOT NULL,
+    name                     TEXT    NOT NULL,
+    username                 TEXT    NOT NULL,
+    email                    TEXT    DEFAULT 'unknown@justice.gov.uk',
```
davidatkinsuk, line 7: I think this is probably a good opportunity to get rid of this default value given the value is nullable in the model

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas2v2/Cas2v2UserEntity.kt
davidatkinsuk: Have discussed offline and agreed this is something we could consider in the future, but it's acceptable right now as the usage of this class is quite limited

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas2v2/Cas2v2UserEntity.kt
```diff
@@ -0,0 +1,79 @@
...
+  @CreationTimestamp
+  private val createdAt: OffsetDateTime? = null,
+
+  @OneToMany(mappedBy = "createdByUser")
+  val applications: MutableList<Cas2v2ApplicationEntity> = mutableListOf(),
+) {
+  override fun toString() = "CAS2V2 user $id"
+
+  fun staffIdentifier() = when (userType) {
+    Cas2v2UserType.NOMIS -> nomisStaffId?.toString() ?: ""
```
davidatkinsuk, line 73: I think this should maybe crash out if id can't defined as that should never happen? e.g. use `nomisStaff!!.toString()`

davidatkinsuk: Or could do something like `nomisStaffId?toString() ?: error("Couldn't resolve nomis ID for user $id")` if you want a more useful message

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2v2/Cas2v2ApplicationService.kt
davidatkinsuk: looks good

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/SeedService.kt
davidatkinsuk: great, thanks

davidatkinsuk (approved):
LGTM, just a few minor optional comment 👍

PR #2794, opened 2025-01-14
CAS-1238 Don't return soft deleted applications for applications summary view

muhammad-elabdulla:
This [PR CAS-1238](https://dsdmoj.atlassian.net/browse/CAS-1238) is to not return soft deleted applications for applications summary view


vbala-moj (approved):

danhumphreys-moj (approved):
LGTM

PR #2797, opened 2025-01-14
APS-1792 - Use seed job to create CAS3 premise & rooms

davidatkinsuk:
This commit removes SQL scripts used in local, dev and test to create CAS3 premise, rooms and beds and replaces them with two seed job CSV files.

Whilst most of the configuration was identical in Local vs Dev/Test, the following changes have been made

* Added rooms/beds to PAVILION11, PAVILION12 and PAVILION13 that didn’t previously exist locally
* Changed the LAO and PDU for ‘Something Court’ to match its region

Tests have all passed locally

Screenshot 2025-01-14 at 16 15 42


danhumphreys-moj (approved):
Nice to see the removal of 4 x migration sql scripts.  Also nice to see we are heading towards making local, dev and test environments more consistent.

PR #2799, opened 2025-01-14
CAS-1238 get application returns not found when Cas3 application was deleted

muhammad-elabdulla:
This [PR CAS-1238](https://dsdmoj.atlassian.net/browse/CAS-1238) is to change get application to return not found response when Cas3 application was deleted


davidatkinsuk (approved):
LGTM 👍

PR #2801, opened 2025-01-14
CAS3-1238 To prevent submit Cas3 draft application that was deleted

muhammad-elabdulla:
This PR CAS-1238 is to prevent submit Cas3 draft application that was deleted


davidatkinsuk (approved):
LGTM 👍

PR #2807, opened 2025-01-15
turned off emitting cas 3 messages in dev.

daveawc:
There's an issue where the E2E tests use the same CRN to apply and place multiple times, which results in a lot of addresses being created in delius for the same offender. Last count was over 7k.

Messages still emit in the test environment, and can potentially be re-enabled here if we update the tests to create and use new test data per test run.


davidatkinsuk (approved):
LGTM

PR #2811, opened 2025-01-15
CAS-1205: Update Bed Search Filters

daveawc:
This updates the BedSearchParameters model to TemoraryAccommodationBedSearchParameters (renaming to come later as this will affect the UI), as only cas3 uses it now. It also refactors the controller and service to pass the filter object through instead of individual parameters, this is initial refactoring in order to maintain this endpoint during refactoring to add new filter options.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas3/Cas3BedspaceSearchService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas3/Cas3BedspaceSearchService.kt
@@ -4,6 +4,7 @@ import org.springframework.stereotype.Service
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.BedSearchAttributes
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.PersonType
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ServiceName
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.TemporaryAccommodationBedSearchParameters
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.BookingRepository
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.OverlapBookingsSearchResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ProbationDeliveryUnitRepository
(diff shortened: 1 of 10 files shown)


src/main/resources/static/_shared.yml
```diff
@@ -4125,44 +4125,62 @@ components:
...
       properties:
         serviceName:
           type: string
         startDate:
           type: string
           format: date
-          description: The date the Bed will need to be free from
         durationDays:
-          type: integer
+          type: long
```
muhammad-elabdulla, line 4137: Is there a reason that need to change durationDays from integer to long?

muhammad-elabdulla:
As I mention before creating a new controller and endpoint for CAS3 Bedspace Search in CAS3 namespace and maintain the current do the changes there is better than modify the current Api endpoints response which break the UI

src/main/resources/static/codegen/built-cas3-api-spec.yml
```diff
@@ -4224,44 +4224,62 @@ components:
         - rfap
         - mhapStJosephs
         - mhapElliottHouse
-    BedSearchParameters:
+    TemporaryAccommodationBedSearchParameters:
       type: object
```
muhammad-elabdulla, line 4228: I prefer to rename it to Cas3BedspaceSearchParameters to align with the new premises architecture work

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas3/Cas3BedspaceSearchService.kt
```diff
@@ -41,23 +42,20 @@ class Cas3BedspaceSearchService(
   @Suppress("detekt:CyclomaticComplexMethod")
   fun findBedspaces(
     user: UserEntity,
-    probationDeliveryUnits: List<UUID>,
-    startDate: LocalDate,
-    durationInDays: Int,
-    propertyBedAttributes: List<BedSearchAttributes>?,
+    queryParams: TemporaryAccommodationBedSearchParameters,
```
muhammad-elabdulla, line 45: can we rename this parameter to searchParams

muhammad-elabdulla (changes requested):

src/main/resources/static/_shared.yml
muhammad-elabdulla: Yes but as Api we only expect days as integer

muhammad-elabdulla: Forget my previous comment

muhammad-elabdulla:
What we discuss if you go back to the conversation is to create a new Api Cas3 BedspaceSearch endpoint and maintain the existing one. But you decided to merge the change in that affected the UI in the previous PR. 

I am only trying to align the changes that we are doing now without expanding the work to be aligned with the new architecture work

muhammad-elabdulla (approved):

PR #2818, opened 2025-01-16
APS-1756 - Use default replicaCount in `test` for load test

davidatkinsuk:
This ensures that the `test` environment has the same number of replicas as production


muhammad-elabdulla (approved):
LGTM

PR #2819, opened 2025-01-16
CAS-1303 Remove ReferralHistoryNotes from CAS3 assessment Api response

muhammad-elabdulla:
This [PR CAS-1303](https://dsdmoj.atlassian.net/browse/CAS-1303) is to remove ReferralHistoryNotes from Cas3 assessment Api response

### DIFF
--- a/src/main/resources/static/_shared.yml
+++ b/src/main/resources/static/_shared.yml
@@ -2537,10 +2537,6 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/ClarificationNote'
-        referralHistoryNotes:
-          type: array
-          items:
-            $ref: '#/components/schemas/ReferralHistoryNote'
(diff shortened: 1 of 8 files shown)


src/main/resources/static/_shared.yml
```diff
@@ -2607,6 +2603,10 @@ components:
               $ref: '#/components/schemas/ApprovedPremisesAssessmentStatus'
             createdFromAppeal:
               type: boolean
+            referralHistoryNotes:
+              type: array
+              items:
+                $ref: '#/components/schemas/ReferralHistoryNote'
```
davidatkinsuk, line 2609: we don't need this so happy for it to be completely removed

muhammad-elabdulla: done

PR #2820, opened 2025-01-16
APS-1756 - Update CAS1 Non Arrival Reasons

davidatkinsuk:
Active reasons after changes

```
Failed to arrive as planned
Arrested / In custody / Recalled (correct reason code to follow in subsequent commit)
Parole release not granted
Admitted to hospital (correct reason code to follow in subsequent commit)
Other - provide reasons
```


davidatkinsuk:
> 👍 , guessing those UUIDs are the same in each of the environments?

Good question - i just double checked and they are.

PR #2828, opened 2025-01-17
APS-1785 - Return all APs when space search apType=normal

davidatkinsuk:
Before this commit if normal was specified APs with a specific type defined (e.g. PIPE) would not be returned.

Whilst apType should ideally be undefined/null in this case, we are currently using ‘normal’ on the UI.

This commit also updates the locally configured approved premises to provide a mix of AP Types


muhammad-elabdulla (approved):
LGTM

PR #2831, opened 2025-01-20
CAS-1375 Upper case crn when call getPersonInfoResult in premises service

muhammad-elabdulla:
This [PR CAS-1375](https://dsdmoj.atlassian.net/browse/CAS-1375) is to fix an issue when calling get offender details and the crn is in lower case


aliuk2012:
Is it worth also trimming the crn? I know the UI is doing this at the moment but thinking of cases where we forget to trim or theres another api client thats calling this endpoint and ends up breaking because its not cleaned up.

davidatkinsuk (approved):
LGTM 👍

PR #2833, opened 2025-01-20
APS-1785 retain premise characteristics when converting a booking to space booking

davidatkinsuk:


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/seed/SeedCas1BookingToSpaceBookingTest.kt
```diff
@@ -68,9 +67,9 @@ class SeedCas1BookingToSpaceBookingTest : SeedTestBase() {
   lateinit var premises: ApprovedPremisesEntity
   lateinit var otherPremise: ApprovedPremisesEntity
   lateinit var otherUser: UserEntity
-  lateinit var roomCriteriaOfInterest: List<CharacteristicEntity>
+  lateinit var criteriaOfInterest: List<CharacteristicEntity>
   lateinit var roomCriterionNotOfInstant: CharacteristicEntity
-  lateinit var premiseCriterion: CharacteristicEntity
+  lateinit var premisesCriterionNotOfInterest: CharacteristicEntity
```
Rosie-Brigham, line 72: `premisesCriteriaNotOfInterest`?
(as premises is always plural)

Rosie-Brigham (approved):
Looks good, I like the change in naming convention - only one question on the `premisesCriterionNotOfInterest` - should that be Criteria?
(I am very near bounds of my grammatical knowledge here....)

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/seed/SeedCas1BookingToSpaceBookingTest.kt
davidatkinsuk: Fred has already schooled me on this :) Premises, in the context of buildings is used for both singular and plural (like sheep). Criterion is also singular. So premisesCriterion are both referring to single things, which i think is correct here :)

PR #2834, opened 2025-01-21
spike/CAS-1341-CAS2-Prison-Transfer-Event-Emit-POC

besscerule:
This is a POC branch to demonstrate how to set up a listener for the domain events topic.

Message Listener listens to the inboundQueue which is filtered by the allocation and handover event types and prints out the message.

You can run it locally and send messages using the scripts found in test-messages in the test resources folder.

It has a simple Integration test that checks that the listener was called and also a tests to check the health of the queue.


helm_deploy/hmpps-approved-premises-api/values.yaml
```diff
@@ -64,6 +64,10 @@ generic-service:
     elasticache-redis:
       SPRING_DATA_REDIS_HOST: "primary_endpoint_address"
       SPRING_DATA_REDIS_PASSWORD: "auth_token"
+    inbound-queue-sqs-instance-output:
+      HMPPS_SQS_QUEUES_INBOUNDQUEUE_QUEUE_NAME: "inboundqueue"
```
Rosie-Brigham, line 68: Is there a specific reason this isn't either camel or snake case?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/InboundQueueTest.kt
```diff
@@ -0,0 +1,39 @@
...
+  fun putMessageOnInboundQueue() = putMessageOnQueue(
+    inboundQueueClient,
+    inboundQueue.queueUrl,
+  )
+
+  @Test
+  fun `Put Message on Inbound Queue Request is successful`() {
+    putMessageOnInboundQueue()
+    TimeUnit.MILLISECONDS.sleep(10000)
+    verify(exactly = 1) { mockMessageListener.processMessage(any()) }
```
davidatkinsuk, line 37: again, probably not an issue as this is POC but a library like [awaitility](http://www.awaitility.org/) would let us remove this sleep. I'm not sure if there's a more kotlin way of doing it though!

besscerule: I'm not sure we even need the sleep tbf

helm_deploy/hmpps-approved-premises-api/values.yaml
Rosie-Brigham: Excellent!

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/InboundQueueTest.kt
besscerule: ahh yes we do! fails when i remove!

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/InboundQueueTest.kt
```diff
@@ -0,0 +1,39 @@
...
+import uk.gov.justice.hmpps.sqs.MissingQueueException
+import java.util.concurrent.TimeUnit
+
+class InboundQueueTest : IntegrationTestBase() {
+
+  @Autowired
+  private lateinit var hmppsQueueService: HmppsQueueService
+
+  @SpykBean
+  private lateinit var mockMessageListener: MessageListener
```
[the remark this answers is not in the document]
besscerule: i've added an extra verify test that hooks up a message service where we can do all of the business logic

PR #2837, opened 2025-01-21
APS-1829 - Enhance test app seed job to include bookings

davidatkinsuk:
This commit updates the create test application seed job to support seeding applications with a space bookings. This is required to support load testing.

Example config:

```
creator_username,crn,count,state,premises_qcode
JIMSNOWLDAP,X320741,50,AUTHORISED,
JIMSNOWLDAP,X320741,50,BOOKED,Q713
```

Example output

Screenshot 2025-01-21 at 15 58 30

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/Cas1ApplicationSeedService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/Cas1ApplicationSeedService.kt
@@ -1,5 +1,8 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.seed.cas1
 
+import com.google.common.cache.CacheBuilder
+import com.google.common.cache.CacheLoader
+import com.google.common.cache.LoadingCache
 import org.json.JSONObject
 import org.slf4j.LoggerFactory
 import org.springframework.stereotype.Service
@@ -42,6 +45,7 @@ import java.io.IOException
 import java.time.LocalDate
 import java.time.OffsetDateTime
 import java.util.UUID
+import java.util.concurrent.TimeUnit
 
 @SuppressWarnings("MagicNumber", "TooGenericExceptionCaught")
 @Service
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/Cas1ApplicationSeedService.kt
```diff
@@ -327,32 +393,37 @@ class Cas1ApplicationSeedService(
...
+@SuppressWarnings("MagicNumber")
+@Service
+class Cas1ApplicationSeedServiceCaches(
+  private val offenderService: OffenderService,
+) {
 
-  private fun loadFixtureAsResource(filename: String): String {
-    val path = "db/seed/local+dev+test/cas1_application_data/$filename"
+  var personInfoCache: LoadingCache<String, PersonInfoResult.Success.Full> = CacheBuilder.newBuilder()
+    .maximumSize(10000)
```
gregkhawkins, line 405: assume this is a safeguard and we are always below 10,000? Can see (if you did deep enough) we orchestrate calls to Delius which I assume is why we are doing this.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/Cas1ApplicationSeedService.kt
```diff
@@ -109,14 +144,24 @@ class Cas1ApplicationSeedService(
...
       }
+      ApplicationState.BOOKED -> {
+        val application = createApplicationPendingSubmission(deliusUserName, crn)
+        submitApplication(application)
+        assessAndAcceptApplication(application)
+        val premises = premisesRepository.findByQCode(premisesQCode!!)!!
+        createSpaceBooking(
+          application,
+          premises,
+        )
```
gregkhawkins, line 155: nice!

gregkhawkins (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/Cas1ApplicationSeedService.kt
davidatkinsuk: I think i just copied the example :) in reality i think cache size of 1 is sufficient because we keep using the same CRN :)

PR #2838, opened 2025-01-21
Disable notify in ‘test’

davidatkinsuk:
We disable notify in test as we use the environment for load testing, and notify is a 3rd party service outside of the scope of load testing


muhammad-elabdulla (approved):
LGTM

PR #2843, opened 2025-01-22
APS-1617 Updated job to use the correct json path.

vbala-moj:
[https://dsdmoj.atlassian.net/browse/APS-1617](url)


davidatkinsuk (approved):
LGTM 👍

PR #2845, opened 2025-01-22
Move common logic into CAS1TimelineService

davidatkinsuk:
This is in preparation for changes to the timeline response


muhammad-elabdulla (approved):
LGTM

PR #2850, opened 2025-01-23
CAS-1374 Migrate to the new prisoner alerts Api endpoint

muhammad-elabdulla:
This [PR CAS-1374](https://dsdmoj.atlassian.net/browse/CAS-1374) is to migrate to the new prisoner alerts api endpoint. This PR include:

- a new feature flag `get-alerts-from-prisoner-alerts-api` to use the new prisoner alerts Api endpoint will be activated in dev and preprod environments
- Remove the `expired` and `active` properties from `PersonAcctAlert` as is not used in CAS3 and CAS1 UI 
- Return alertId as zero value because it was replaced in the new prisoner alerts Api endpoint response by `AlertUuid` which is a UUID value. It will highly replaced by the new property `alertTypeDescription` added to the `PersonAcctAlert` Api response 
-  Return `description`  property to replace the `comment` property in the UI. We need to maintain both for now to not break the UI 
- The exiting code related to prison alerts will be removed after testing and  CAS3 and CAS1 UI changed to use the new properties


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PeopleController.kt
```diff
@@ -141,14 +145,27 @@ class PeopleController(
...
-    val acctAlerts = when (acctAlertsResult) {
-      is AuthorisableActionResult.NotFound -> throw NotFoundProblem(crn, "Inmate")
-      is AuthorisableActionResult.Unauthorised -> throw ForbiddenProblem()
-      is AuthorisableActionResult.Success -> acctAlertsResult.entity
-    }
+    val usePrisonerAlertsApi = featureFlagService.getBooleanFlag("get-alerts-from-prisoner-alerts-api")
+
+    if (usePrisonerAlertsApi) {
+      val acctAlertsResult = offenderService.getAcctPrisonerAlertsByNomsNumber(nomsNumber)
+      val acctAlerts = when (acctAlertsResult) {
```
davidatkinsuk, line 152: we should be using `CasResult` for this. Can then use `extractEntityFromCasResult` instead of the when block

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/client/PrisonerAlertsApiClient.kt
```diff
@@ -0,0 +1,18 @@
...
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.config.WebClientConfig
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.prisoneralertsapi.Alert
+
+@Component
+class PrisonerAlertsApiClient(
+  @Qualifier("prisonerAlertsApiWebClient") webClientConfig: WebClientConfig,
+  objectMapper: ObjectMapper,
+  webClientCache: WebClientCache,
+) : BaseHMPPSClient(webClientConfig, objectMapper, webClientCache) {
+  fun getAlerts(nomsNumber: String, alertCode: String) = getRequest<List<Alert>> {
```
davidatkinsuk, line 15: looking at the [schema](https://alerts-api-dev.hmpps.service.justice.gov.uk/swagger-ui/index.html#/RO%20Operations/retrievePrisonerAlerts_1) the top level response isn't a list of alerts, we need to drill down into the JSON to get this, so a different type is required here?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/PrisonerAlertTransformer.kt
```diff
@@ -0,0 +1,17 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer
+
+import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.PersonAcctAlert
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.prisoneralertsapi.Alert
+
+@Component
+class PrisonerAlertTransformer {
+  fun transformToApi(alert: Alert) = PersonAcctAlert(
```
davidatkinsuk, line 9: I know it's simple but can we have a unit test for this please, mainly because the integration test relies on it

davidatkinsuk (approved):
Looks good, just a few comments from me

Is there going to be a corresponding ap-tools update? Hopefully we can use wiremock for these responses instead of starting up the alerts service locally?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/factory/PrisonerAlertFactory.kt
```diff
@@ -0,0 +1,123 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.factory
```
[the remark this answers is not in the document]
davidatkinsuk: also would be nice to have a factory model that supports copy() so we can create baseline factories and reuse/tweak

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/client/PrisonerAlertsApiClient.kt
muhammad-elabdulla: You mean to add all the repose structure in CAS?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/PrisonerAlertTransformer.kt
muhammad-elabdulla: done

muhammad-elabdulla:
I think it will be a separate service as the url is different

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/client/PrisonerAlertsApiClient.kt
davidatkinsuk: The code assumes we're getting a list of alerts back, but the list is actually embedded in the results in a 'content' property:

```
{
  "totalElements": 9007199254740991,
  "totalPages": 1073741824,
  "first": true,
  "last": true,
  "size": 1073741824,
  "content": [
    {
      "alertUuid": "8cdadcf3-b003-4116-9956-c99bd8df6a00",
      "prisonNumber": "A1234AA",
      "alertCode": {
        "alertTypeCode": "A",
        "alertTypeDescription": "Alert type description",
        "code": "ABC",
        "description": "Alert code description"
      },
      "description": "Alert description",
      "authorisedBy": "A. Nurse, An Agency",
      "activeFrom": "2021-09-27",
      "activeTo": "2022-07-15",
      "isActive": true,
      "createdAt": "2021-09-27T14:19:25",
      "createdBy": "USER1234",
      "createdByDisplayName": "Firstname Lastname",
      "lastModifiedAt": "2022-07-15T15:24:56",
      "lastModifiedBy": "USER1234",
      "lastModifiedByDisplayName": "Firstname Lastname",
      "activeToLastSetAt": "2022-07-15T15:24:56",
      "activeToLastSetBy": "USER1234",
      "activeToLastSetByDisplayName": "Firstname Lastname"
    }
  ]
}
```

So I think the return type from this function would be something like

data class AlertResult(
   val content: List<Alert>
)

davidatkinsuk:
> I think it will be a separate service as the url is different

Assuming this is in response to

> Is there going to be a corresponding ap-tools update? Hopefully we can use wiremock for these responses instead of starting up the alerts service locally?

We should be able to use wiremock to return responses for this as we use wiremock to return responses for multiple services already. We could configure the new alert service in the docker-compose file but i think we need to avoid that if wiremock is good enough.

PR #2853, opened 2025-01-23
Increased max in memory size

daveawc:


davidatkinsuk (approved):
LGTM 👍 

We should revert this once we've migrated to the new API for alerts

PR #2855, opened 2025-01-23
chore/upgrade openapi plugin version from 7.7 to 7.11.

danhumphreys-moj:
Running the gradle openApiGenerate task currently results in the following issues:

"More than one inline schema specified in allOf:. Only the first one is recognized. All others are ignored."

Resolving these issues requires:

- upgrading the openAPI generator plugin version from 7.7 to 7.11
- setting new property `removeEnumValuePrefix = true`

See more information in the comment descriptions.

After these changes the gradle openApiGenerate task should run without complaining about the inline schema issues.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -3,7 +3,7 @@ import org.apache.commons.io.FileUtils
 plugins {
   id("uk.gov.justice.hmpps.gradle-spring-boot") version "6.1.2"
   kotlin("plugin.spring") version "2.0.21"
-  id("org.openapi.generator") version "7.7.0"
+  id("org.openapi.generator") version "7.11.0"
   id("org.jetbrains.kotlin.plugin.jpa") version "1.9.22"
   id("io.gatling.gradle") version "3.13.1"
   id("io.gitlab.arturbosch.detekt") version "1.23.7"
@@ -239,6 +239,7 @@ openApiGenerate {
   typeMappings.put("DateTime", "Instant")
   importMappings.put("Instant", "java.time.Instant")
   templateDir.set("$rootDir/openapi")
+  additionalProperties.put("removeEnumValuePrefix", "true")
 }
 
 registerAdditionalOpenApiGenerateTask(


danhumphreys-moj:
> @danhumphreys-moj i think this should be OK to go in, my only concern would be if there was an enum that had a different value to jsonproperty. I've been through and checked, and I don't think there is.
> 
> It might be an idea for you to have a double check thoufh as well, if you don't find anything, we should be ok to merge. So any enum in the generated package: `hmpps-approved-premises-api/build/generated/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi` should have the same values in bother the `@JsonProperty` annotation, and the enum value. I can't think of any way to change this, so it's probably overkill, but better to be safe.
> 
> image

When I ran the following shell script I couldn't find any cases where this was the case.  Think we're good to go then.

See Slack chat [here](https://mojdt.slack.com/archives/C03K0HB0LBE/p1738767967020249).

```
#!/bin/bash

# Directory to search
directory="../build/generated/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/api"

# Function to check if @JsonProperty value matches enum constructor value
check_enum_values() {
  local file=$1
  echo "Checking file: $file"
  while IFS= read -r line; do
    if [[ $line =~ @JsonProperty\(\"([^\"]+)\"\)\ ([a-zA-Z0-9_]+)\(\"([^\"]+)\"\) ]]; then
      json_property_value="${BASH_REMATCH[1]}"
      enum_value="${BASH_REMATCH[2]}"
      enum_constructor_value="${BASH_REMATCH[3]}"
      if [[ "$json_property_value" == "$enum_constructor_value" ]]; then
        echo "Match found: @JsonProperty(\"$json_property_value\") $enum_value(\"$enum_constructor_value\")"
      else
        echo "Mismatch found: @JsonProperty(\"$json_property_value\") $enum_value(\"$enum_constructor_value\")"
      fi
    fi
  done < "$file"
}

# Find .kt files containing "enum class" and process them
grep -rl "enum class" "$directory" --include \*.kt | while read -r file; do
  check_enum_values "$file"
done
```

PR #2857, opened 2025-01-23
Enable ap-and-oasys retry on timeout

davidatkinsuk:
This commit enables retry on timeout for ap-and-oasys and reduces its timeout threshold from 10 seconds to 4 seconds. Observations have shown that when we do get a timeout and a 500 is returned to the UI a retry triggered by the UI succeeds.

The timeout of 4 seconds is based upon the 99th percentil response time of 340 milliseconds, with only one request taking longer than 4 seconds (without timeout) in January so far.


muhammad-elabdulla (approved):
LGTM

PR #2862, opened 2025-01-24
APS-1857, Moved applications and people timeline end points to cas1 specific.

vbala-moj:
[https://dsdmoj.atlassian.net/browse/APS-1857](url)

### DIFF
--- a/src/main/resources/static/api.yml
+++ b/src/main/resources/static/api.yml
@@ -1346,6 +1346,8 @@ paths:
           $ref: '_shared.yml#/components/responses/500Response'
   /people/{crn}/timeline:
     get:
+      description: deprecated, use /cas1/people/{crn}/timeline
+      deprecated: false
       summary: Returns a timeline of all applications for a Person.
       operationId: peopleCrnTimelineGet
       parameters:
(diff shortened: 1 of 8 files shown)


src/main/resources/static/api.yml
```diff
@@ -1899,6 +1901,8 @@ paths:
       x-codegen-request-body-name: body
   /applications/{applicationId}/timeline:
     get:
+      description: deprecated, use /cas1/applications/{applicationId}/timeline
+      deprecated: true
```
davidatkinsuk, line 1905: nice

src/main/resources/static/cas1-api.yml
```diff
@@ -1283,3 +1283,78 @@ paths:
...
+      operationId: getApplicationTimeLine
+      parameters:
+        - name: applicationId
+          in: path
+          description: ID of the application
+          required: true
+          schema:
+            type: string
+            format: uuid
+        - name: X-Service-Name
```
davidatkinsuk, line 1300: we can drop X-Service-Name as this is implied through the path (we can also remove the corresponding check in the controller)

PR #2869, opened 2025-01-27
Capture Delius ID when creating/backfill Space Bookings

davidatkinsuk:
This PR captures the Delius ID provided in the delius referral export when backfilling booking information from Delius for space bookings. It does this in two places

1. When migrating CAS1 legacy `Bookings` to `SpaceBookings`
2. When creating `SpaceBookings` for accepted referrals created directly in Delius

This will be used to provide a report to Delius of CAS1 bookings ID for backfilled bookings (point 2 above), allowing domain events to be processed


danhumphreys-moj (approved):
LGTM

PR #2872, opened 2025-01-27
Cas2v2/dev

tobybatchmoj:
The sum of the new name space work without the search by CRN


src/main/resources/static/_shared.yml
```diff
@@ -1626,6 +1626,7 @@ components:
           Offline: '#/components/schemas/OfflineApplication'
           CAS1: '#/components/schemas/ApprovedPremisesApplication'
           CAS2: '#/components/schemas/Cas2Application'
+          CAS2V2: 'cas2v2-schemas.yml#/components/schemas/Cas2v2Application'
```
davidatkinsuk, line 1629: i don't _think_ `_shared.yml` should be referring to CAS1 specific schemas (the other way around is fine). This unfortunately means putting the `Cas2v2Application` type in this file until we untangle use of the common Application model. Sorry, I should have spotted this before

Rosie-Brigham (approved):
Carefully read half, skim read the second - couldn't see any glaring issues though

davidatkinsuk (approved):
Just one comment from me.

As i've already reviewed the constituent parts on various PRs i've just focused on changes to share code, which looks good 👍

PR #2876, opened 2025-01-27
APS-1830 Updated domain event describer to return TimelineEventContentPayload along with content description.

vbala-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/ApplicationTimelineTransformer.kt
```diff
@@ -25,13 +25,15 @@ class ApplicationTimelineTransformer(
 
   fun transformDomainEventSummaryToTimelineEvent(domainEventSummary: DomainEventSummary): TimelineEvent {
     val associatedUrls = generateUrlsForTimelineEventType(domainEventSummary)
+    val contentAndPayload = domainEventDescriber.getContentPayload(domainEventSummary)
 
     return TimelineEvent(
```
davidatkinsuk, line 30: this can be done on a different PR, but i think we also need to add schemaVersion to `TimelineEvent` which is taken from the domain event schemaVersion

src/main/resources/static/_shared.yml
```diff
@@ -4797,3 +4799,92 @@ components:
...
+          type: array
+          items:
+            $ref: '#/components/schemas/Cas1SpaceCharacteristic'
+        previousCharacteristics:
+          type: array
+          items:
+            $ref: '#/components/schemas/Cas1SpaceCharacteristic'
+      required:
+        - expectedArrival
+        - expectedDeparture
```
davidatkinsuk, line 4844: perfect!

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/transformer/ApplicationTimelineTransformerTest.kt
```diff
@@ -410,11 +415,63 @@ class ApplicationTimelineTransformerTest {
...
+    assertThat(result.createdBy).isEqualTo(userApi)
+    assertThat(result.triggerSource).isEqualTo(null)
+    assertThat(result.payload).isInstanceOf(BookingChangedContentPayload::class.java)
+    assertThat(payload.premises.id).isEqualTo(premisesId)
+    assertThat(payload.premises.name).isEqualTo("name")
+    assertThat(payload.type).isEqualTo(domainEventType.timelineEventType)
+    assertThat(payload.expectedArrival).isEqualTo(payload.expectedArrival)
+    assertThat(payload.expectedDeparture).isEqualTo(payload.expectedDeparture)
+    assertThat(payload.characteristics).isEqualTo(payload.characteristics)
+    assertThat(payload.previousCharacteristics).isEqualTo(payload.previousCharacteristics)
```
davidatkinsuk, line 475: very thorough 👍

davidatkinsuk (approved):
Looks great, just one comment from me which can be done on a new PR 👍

src/main/resources/static/_shared.yml
```diff
@@ -4797,3 +4799,92 @@ components:
...
+      allOf:
+        - $ref: "#/components/schemas/TimelineEventContentPayload"
+      properties:
+        expectedArrival:
+          type: string
+          format: date
+        expectedDeparture:
+          type: string
+          format: date
+        previousExpectedArrival:
```
davidatkinsuk, line 4828: could we document each previous value with a description saying that they're only populated if the new value is different, and where schema version = 2

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/ApplicationTimelineTransformer.kt
vbala-moj: Complelty forget that 🤦🏻‍♂️, will do that on this PR.

PR #2880, opened 2025-01-28
Update hmpps.gradle-spring-boot

davidatkinsuk:
Upgrade to the latest hmpps.gradle-spring-boot plugin, ensuring we ignore irrelevant vulnerability CVE-2024-50379 during trivy scans.

This will upgrade spring boot from 3.4.0 to 3.4.1. Hibernate Spatial has been upgraded to align with the newer version of hibernate used in 3.4.1 (note that it previously not aligned with the spring boot hibernate version)

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,8 +1,8 @@
 import org.apache.commons.io.FileUtils
 
 plugins {
-  id("uk.gov.justice.hmpps.gradle-spring-boot") version "6.1.0"
-  kotlin("plugin.spring") version "2.0.20"
+  id("uk.gov.justice.hmpps.gradle-spring-boot") version "6.1.2"
+  kotlin("plugin.spring") version "2.0.21"
   id("org.openapi.generator") version "7.7.0"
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -1,8 +1,8 @@
 import org.apache.commons.io.FileUtils
 
 plugins {
-  id("uk.gov.justice.hmpps.gradle-spring-boot") version "6.1.0"
-  kotlin("plugin.spring") version "2.0.20"
+  id("uk.gov.justice.hmpps.gradle-spring-boot") version "6.1.2"
```
aliuk2012, line 4: @davidatkinsuk Not sure if this makes any difference but it looks like 7.0 has been release yesterday afternoon

https://mojdt.slack.com/archives/C69NWE339/p1738079571632049

davidatkinsuk: Cheers - i'll get this pushed out now as 7.0 appears to have more breaking changes and might interfere with the open api work so we need to look into that

PR #2883, opened 2025-01-28
WIP CAS2V2 - FIE metrics collection

rossjones-moj:
This PR intends to extend the range of status options that are currently available for CAS2V2 to ensure that the reasons for each state contain more useful information for management info.  

Currently each status has a list of sub-statuses (the StatusDetail, actually a `Cas2v2ApplicationStatusDetail` ) which describe in slightly more detail the reason for the state change by external users.  This PR updates the model so that each StatusDetail can also contain a list of child StatusDetails, allowing for a tree structure with more deeper, and more detailed, reasons for the state change.


davidatkinsuk:
I think this most likely can be closed, but would be good to get a second opinion from someone more familiar with CAS2v2

davidatkinsuk:
closing this, can always revive from the branch if required

PR #2885, opened 2025-01-29
fix: includes a missed built api file for TimelineEventContentPayload

tobybatchmoj:
Just a miss, the TimeLineEventCntent Patyload gets pulled in by the domain events, the change was made by 'bala' in the _shared.yml and I guess our new cas2v2 api wasn't in yet.


davidatkinsuk (approved):
LGTM 👍

PR #2886, opened 2025-01-29
Preload & Simplify Open Api Docs

davidatkinsuk:
We have found that the intial load of open api docs takes quite some time leading to ‘broken pipe’ errors, presumably because the client gives up and times out.

This commit preloads the open docs in all environments other than local. Local tests suggests that the initilisation runs concurrent from start, so shouldn’t have too much of an effect on startup times. Regardless, we have disabled this for local deployments where the endpoints are rarely used.

This commit also removes the CAS-specific open api docs that don’t include shared elements.

### DIFF
--- a/src/test/resources/application-test.yml
+++ b/src/test/resources/application-test.yml
@@ -51,6 +51,9 @@ spring:
   cache:
     type: none
 
+springdoc:
+  pre-loading-enabled: false
+
 hmpps.sqs:
   useWebToken: false
   provider: localstack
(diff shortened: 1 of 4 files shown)


src/test/resources/application-test.yml
```diff
@@ -51,6 +51,9 @@ spring:
   cache:
     type: none
 
+springdoc:
+  pre-loading-enabled: false
```
gregkhawkins, line 55: is this intentional? From reading the PR description it seemed like the intention was to disable only for the local profile.

davidatkinsuk: ah yes this is kind of confusing. This is the profile used for integration tests, not the `test` environment. I did initially exclude this change (because it didn't occur to me) but pre-loading caused integration tests to fail (i think because it massively slowed them down maybe)

gregkhawkins: makes sense - approved!

gregkhawkins (approved):

PR #2891, opened 2025-01-30
CAS-1374 Map new prisoner api client to nomis credentials

muhammad-elabdulla:
This PR is to map new prisoner api client to nomis credentials


davidatkinsuk (approved):
LGTM 👍

PR #2893, opened 2025-01-30
Sync CAS3 Circle CI e2e config

aliuk2012:
We noticed that the config being used here was very different from the config in CAS UI repo itself. CAS3 UI Repo config looks fresher than this version and so we decided to update this one to match cas3 ui config


davidatkinsuk (approved):
LGTM

PR #2895, opened 2025-01-30
Clone CAS3 repo to run e2e in circle ci

aliuk2012:
We cant just call `checkout` in circle ci becaus we need to checkout cas 3 repo and not the api repo.

Reintroduce steps that were removed as part of from https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/2893

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -321,7 +321,16 @@ jobs:
         type: string
         default: dev
     steps:
-      - checkout
+      - run:
+          name: Checkout e2e repo
+          command: |
+            git clone https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui.git .
+      - run:
+          name: install
+          command: apt-get install xz-utils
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -321,7 +321,16 @@ jobs:
...
         default: dev
     steps:
-      - checkout
+      - run:
+          name: Checkout e2e repo
+          command: |
+            git clone https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui.git .
+      - run:
+          name: install
+          command: apt-get install xz-utils
```
aliuk2012, line 330: I'm not 100% sure if these are needed for cypress. CAS3 is the only repo thats using cypress for e2e. CAS1 & CAS2 use playwright. 
This run step was removed from the previous attempt at syncing the configs

brimchugh (approved):

PR #2897, opened 2025-01-30
Feature/aps 1879 add dont remove role option to user seed

davidatkinsuk:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/AbstractUsersSeedJob.kt
```diff
@@ -7,33 +7,64 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserRole
...
- *  using ApStaffUsersSeedJob.
- */
 @SuppressWarnings("TooGenericExceptionThrown", "TooGenericExceptionCaught")
 abstract class AbstractUsersSeedJob(
   private val useRolesForServices: List<ServiceName>,
   private val userService: UserService,
 ) : SeedJob<UsersSeedCsvRow>(
   requiredHeaders = setOf(
-    "deliusUsername",
+    "delius_username",
```
Rosie-Brigham, line 16: Why change this to a snake case (I am sure there is a good reason, just I can't immidiadetly see it)

davidatkinsuk: We are massively inconsistent in this area. I didn't want to add the new column `remove_existing_roles_and_qualifications` as camel case so i decided to make the existing columns consistent.  I need to go fix the existing pre-prod user seed file for this PR anyway so i'll be sure to fix the column name too

PR #2903, opened 2025-01-31
APS-1863 Tidy up Tasks Logic

davidatkinsuk:
* Move a tasks function from users service to tasks service
* Tidy up tasks integration tests
* Add a note on the TaskWrapper model around issue caused by extending user

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/TasksController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/TasksController.kt
@@ -170,7 +170,7 @@ class TasksController(
       taskInfo.requiredPermission,
     )
 
-    val workload = userService.getUserWorkloads(users.map { it.id })
+    val workload = taskService.getUserWorkloads(users.map { it.id })
     val transformedAllocatableUsers = users.map {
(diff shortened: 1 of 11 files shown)


muhammad-elabdulla (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/TasksController.kt
```diff
@@ -170,7 +170,7 @@ class TasksController(
       taskInfo.requiredPermission,
     )
 
-    val workload = userService.getUserWorkloads(users.map { it.id })
+    val workload = taskService.getUserWorkloads(users.map { it.id })
```
gregkhawkins, line 173: think we can delete the `userService. getUserWorkloads()` function now we are no longer using it. Had a quick look in code base and nothing else calling it

gregkhawkins (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/TasksController.kt
davidatkinsuk: ah good spot! i meant to do that :)

PR #2904, opened 2025-01-31
Updated Get Applications logic

daveawc:
Previously if a user with ASSESSOR role called the getApplications endpoint, it would return all applications in a region. This endpoint now returns all applications that were created by the user.


muhammad-elabdulla (approved):
LGTM

PR #2907, opened 2025-01-31
APS-1796 Re-enable concurrent process of delius import

davidatkinsuk:
This commit makes fixes to the seed job concurrent procesisng to avoid memory issues, and subsequently re-enables concurrent processing for the `CAS1 Improt Delius Referrals Seed Job`.

The fix uses a semaphore to ensure there are only 5 processing or pending co-routines at any moment in time when processing the CSV. We found that if a coroutine was created for every row in the CSV immediately we encountered out of memory exceptions. Note that the import CSV file has 298132 rows.

If we to use a CSV reader that allowed us to stream the CSV rows instead of loading them all in to memory, this restricton may not be required.

The commit also configures locally ran instances of the service to have the same heap space as production (512MB)

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -130,7 +130,7 @@ tasks.register("bootRunLocal") {
   doFirst {
     tasks.bootRun.configure {
       systemProperty("spring.profiles.active", "local")
-      jvmArgs("-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=32323")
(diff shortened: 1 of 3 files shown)


gregkhawkins (approved):

gregkhawkins:
thanks for the detailed explanation on the PR desc - may have struggled with that a little otherwise :)

davidatkinsuk:
> thanks for the detailed explanation on the PR desc - may have struggled with that a little otherwise :)

on reflection i think this may be a case where a comment in the code may be useful. i'll add it in

PR #2908, opened 2025-01-31
Departure reasons and Breach/recall changes

gregkhawkins:
JIRA: https://dsdmoj.atlassian.net/browse/APS-1869

**PR includes**
* Update departure reasons data inline with new parent and child options
* Add sort by name to the find query so that they are ordered when consumed by the front-end

**Screenshots after changes**

Screenshot 2025-01-31 at 14 33 47

Screenshot 2025-01-31 at 14 33 58


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/factory/PersistedFactory.kt
```diff
@@ -34,6 +34,15 @@ class PersistedFactory<EntityType : Any, PrimaryKeyType : Any, FactoryType : Fac
...
   }
 
+  fun produceAndSortAndPersistMultiple(amount: Int, configuration: FactoryType.() -> Unit, comparator: Comparator<EntityType>): List<EntityType> {
+    val factory = factoryProducer()
+    configuration(factory)
+    val producedAndSortedEntities = (1..amount).map {
+      factory.produce()
+    }.sortedWith(comparator)
+    return repository.saveAllAndFlush(producedAndSortedEntities)
+  }
```
gregkhawkins, line 44: hope this is not too radical just want a way to sort before persist so that we can do this the opposite way to what we expect to come back from the query. Hopefully a re-usable way to test the sort.
Let me know if not a fan :)

src/main/resources/db/migration/all/20250130153425__update_departure_reasons.sql
```diff
@@ -0,0 +1,15 @@
+-- changes to departure reason parent options
+UPDATE departure_reasons SET name = 'Absconded' WHERE id = 'ea6f21d9-e658-487f-9765-d8b09187df93'::uuid; -- updates "Absconded, still at large" -> "Absconded"
+DELETE FROM departure_reasons WHERE id = '1ecc1a09-725f-44d8-b03b-314d1ff4da62'::uuid; -- deletes "Other"
+
+-- changes to departure reason child options (for "Breach/Recall" parent)
+DELETE FROM departure_reasons WHERE id = '012f21fa-1c46-4dca-ab8d-2e73cbe5ec84'::uuid; -- deletes "Breach/Recall > Other"
+DELETE FROM departure_reasons WHERE id = '6761f2a5-4a83-4c2b-a9c8-d6906a7f6e8d'::uuid; -- deletes "Breach/Recall > House rules"
```
davidatkinsuk, line 7: we typically can't delete existing entries because they may already be linked to bookings in production. In this case, set `is_active` to false instead of deleting the entries

gregkhawkins: will do!

davidatkinsuk (approved):
LGTM, just one comment to address 👍

PR #2910, opened 2025-01-31
CAS-486 Remove duplicate calls to delius when getting applications

muhammad-elabdulla:
This PR is to remove duplicate calls to delius to check if the user have access to the applications when calling Get `/applications` api endpoint


davidatkinsuk (approved):
LGTM 👍

PR #2911, opened 2025-01-31
Create e2e test workflow

Rosie-Brigham:
Adding an experimental GH actions workflow that runs on HMPPS runner that may be able to run our E2E tests, on any branch, without merging into main.

This is entirely experimental at this moment, and which may not work.

Planning to get playwite working first and then getting to cypress.
Alas, this has to be deployed to main to check wether it works, BUT once it does - we should not have to merge to main to run our e2e tests. 🤞 🤞🤞


davidatkinsuk:
closing this as we're now fully migrated to github actions

PR #2913, opened 2025-01-31
feat: adds cas2v2 to the swagger doc as sections

tobybatchmoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/swagger/SwaggerConfiguration.kt
```diff
@@ -74,6 +74,33 @@ class SwaggerConfiguration {
...
+  }
+
+  @Bean
+  fun cas2v2DomainEvents(): GroupedOpenApi {
+    return GroupedOpenApi.builder()
+      .group("CAS2v2DomainEvents")
+      .displayName("CAS2v2 Domain Events")
+      .pathsToMatch("/**/events/cas2v2/**")
+      .build()
+  }
```
davidatkinsuk, line 102: i don't think you're going to have any v2 specific domain event so i don't think this is required?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/swagger/SwaggerConfiguration.kt
```diff
@@ -74,6 +74,33 @@ class SwaggerConfiguration {
...
   }
 
+  @Bean
+  fun cas2v2ALL(): GroupedOpenApi {
+    return GroupedOpenApi.builder()
+      .group("CAS2v2ALL")
+      .displayName("CAS2v2 All")
+      .pathsToMatch("/**/cas2v2/**")
+      .build()
+  }
```
davidatkinsuk, line 84: we decided a few days ago to drop having multiple versions of each CASs API here because the initialisation of this stuff is _really_ slow. Therefore, i don't think we need cas2v2ALL()

See https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/2886/files

davidatkinsuk (approved):
LGTM, just a few comments based upon some very recent changes on main

PR #2916, opened 2025-02-03
APS-1783 - Add criteria to placement request booking summary

davidatkinsuk:
To support this we’ve had to move the Cas1SpaceCharacteristics type back into _shared.yml. In time we’ll move placement request code into a /cas1 endpoint

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/PlacementRequestBookingSummaryTransformer.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/PlacementRequestBookingSummaryTransformer.kt
@@ -1,12 +1,20 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer
 
+import org.slf4j.LoggerFactory
 import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Cas1SpaceCharacteristic
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.PlacementRequestBookingSummary
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.BookingEntity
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/PlacementRequestBookingSummaryTransformer.kt
```diff
@@ -25,5 +34,13 @@ class PlacementRequestBookingSummaryTransformer {
...
+    characteristics = jpa.criteria.mapNotNull { it.propertyName }.mapNotNull {
+      try {
+        Cas1SpaceCharacteristic.valueOf(it)
+      } catch (e: IllegalArgumentException) {
+        log.warn("Couldn't find a Cas1SpaceCharacteristic enum entry for propertyName $it")
+        null
+      }
+    },
   )
 }
```
vbala-moj, line 46: Could use a simple function!

`  private fun toCas1SpaceCharacteristicOrNull(propertyName: String): Cas1SpaceCharacteristic? =
    Cas1SpaceCharacteristic.entries.find { it.name == propertyName } ?: run {
      log.warn("Couldn't find a Cas1SpaceCharacteristic enum entry for propertyName $propertyName")
      null
    }`

vbala-moj (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/PlacementRequestBookingSummaryTransformer.kt
davidatkinsuk: Good idea, i've used a similar approach but used an extension function 👍

vbala-moj: If you do not want any warning then simply ` Cas1SpaceCharacteristic.entries.firstOrNull { it.value == char }`

PR #2917, opened 2025-02-03
Fix/lower case crn in bookings fix

daveawc:
Fix an issue where crns were being saved with a lower case leading letter. This also reverts a change from a previous [PR ](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/2831)
 that was suspected to have caused it 
tested the query in preprod, only affects the three incorrect rows.
image

### DIFF
--- a/src/main/resources/db/migration/all/20250203152144__correct_booking_crn.sql
+++ b/src/main/resources/db/migration/all/20250203152144__correct_booking_crn.sql
@@ -1,0 +1,4 @@
+update bookings
+set crn = upper(crn)
+where crn ~ '[a-z]'
+  and service = 'temporary-accommodation'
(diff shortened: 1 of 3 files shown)


src/main/resources/db/migration/all/20250203152144__correct_booking_crn.sql
```diff
@@ -0,0 +1,3 @@
+update bookings
+set crn = upper(crn)
+where crn ~ '[a-z]'
```
davidatkinsuk, line 3: i'm probably being paranoid here but could you maybe isolate this to cas3 bookings :) e.g. where `service = 'temporary-accommodation' ?

davidatkinsuk (approved):
LGTM 👍

PR #2919, opened 2025-02-04
CAS-1205 : update characterstics endpoint

daveawc:
this adds functionality to the get reference data endpoint, so the characteristics can be requested for either a premises or a bedspace. It removes the "includeInactive" flag, as this isn't used.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ReferenceDataController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ReferenceDataController.kt
@@ -71,19 +71,28 @@ class ReferenceDataController(
   private val apAreaTransformer: ApAreaTransformer,
 ) : ReferenceDataApiDelegate {
 
-  override fun referenceDataCharacteristicsGet(xServiceName: ServiceName?, includeInactive: Boolean?): ResponseEntity<List<Characteristic>> {
-    val characteristics = when (xServiceName != null) {
-      true -> when (includeInactive) {
-        true -> characteristicRepository.findAllByServiceScope(xServiceName.value)
-        else -> characteristicRepository.findActiveByServiceScope(xServiceName.value)
-      }
-      false -> when (includeInactive) {
-        true -> characteristicRepository.findAll()
-        else -> characteristicRepository.findActive()
-      }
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ReferenceDataController.kt
```diff
@@ -71,19 +71,28 @@ class ReferenceDataController(
...
+        modelScope = toModelScopeEnum(modelScope).value,
+      ).map(characteristicTransformer::transformJpaToApi),
+    )
+  }
 
-    return ResponseEntity.ok(characteristics.map(characteristicTransformer::transformJpaToApi))
+  /*
+  this can be refactored to use the enum directly in the controller
+  when we move away from the manually written openApiSpec files
+   */
```
davidatkinsuk, line 89: we're using enumerated parameters elsewhere - is there a technical constraint i'm missing here?

For example look at /placement-requests/dashboard

```
        - name: tier
          in: query
          description: filter by tier of Offender at time Application was created
          schema:
            $ref: '_shared.yml#/components/schemas/RiskTierLevel'
```

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/ReferenceDataController.kt
```diff
@@ -71,19 +71,28 @@ class ReferenceDataController(
...
-    }
+  override fun referenceDataCharacteristicsGet(
+    xServiceName: ServiceName?,
+    modelScope: String?,
+  ): ResponseEntity<List<Characteristic>> {
+    return ResponseEntity.ok(
+      characteristicRepository.findActiveByServiceScopeAndModelScope(
+        serviceScope = xServiceName?.value ?: "*",
+        modelScope = toModelScopeEnum(modelScope).value,
+      ).map(characteristicTransformer::transformJpaToApi),
```
davidatkinsuk, line 82: much neater!

davidatkinsuk (approved):
LGTM, a few comments to consider 👍

davidatkinsuk:
Thanks for answering my questions, merge at will 👍

PR #2921, opened 2025-02-04
APS-1870 standardise site survey room import

danhumphreys-moj:
Ref: https://dsdmoj.atlassian.net/browse/APS-1870

- Change `roomCode` calculation to `$qCode-$roomNumber`. Currently we introduce a space before/after the hyphen, which is not consistent with the CSV import
- Change `bedName` calculation to `${row.roomNumber} - ${row.bedNumber}` (note that there are spaces before/after the hyphen). Currently we’re just using the bed number
- Add validation to ensure all beds have unique names in a given premise. If they don’t, throw an error, halting processing of the row
- When reading Room Name and Bed Number, if the value is numeric, we need to format it to remove any decimal points.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/Cas1SiteSurveyPremiseFactory.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/Cas1SiteSurveyPremiseFactory.kt
@@ -109,7 +109,11 @@ fun DataFrame<*>.resolveAnswer(question: String, answerCol: Int = 1): String {
 
   if (questionIndex == -1) error("Question '$question' not found on sheet Sheet3.")
 
-  val answer = answers[questionIndex].toString().trim()
+  fun removeDecimalPlaces() = answers[questionIndex].let {
+    if (it is Double) it.toInt() else it
+  }.toString().trim()
+
+  val answer = removeDecimalPlaces()
 
   if (answer.isBlank()) {
     error("Answer for question '$question' cannot be blank")
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/Cas1SeedRoomsFromSiteSurveyXlsxJob.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/Cas1SeedRoomsFromSiteSurveyXlsxJob.kt
@@ -147,7 +147,7 @@ class Cas1SeedRoomsFromSiteSurveyXlsxJob(
   private fun resolveBeds(qCode: String, siteSurveyBeds: List<Cas1SiteSurveyBed>): List<BedInfo> {
     return siteSurveyBeds.map {
       BedInfo(
-        bedName = it.bedNumber,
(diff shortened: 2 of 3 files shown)


danhumphreys-moj:
Tested locally and points 1, 2 and 4 visibly resolved.  Point 3 evidenced by additional integration test.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/Cas1SiteSurveyPremiseFactory.kt
```diff
@@ -109,7 +109,10 @@ fun DataFrame<*>.resolveAnswer(question: String, answerCol: Int = 1): String {
 
   if (questionIndex == -1) error("Question '$question' not found on sheet Sheet3.")
 
-  val answer = answers[questionIndex].toString().trim()
+  // toInt() on a double truncates decimal places
```
davidatkinsuk, line 112: We could use the '[extract method'](https://refactoring.com/catalog/extractFunction.html) refactor here to remove the need for a comment?

e.g. create a private function (or extension function) called removeDecimalPlaces?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/Cas1SiteSurveyPremiseFactory.kt
```diff
@@ -109,7 +109,10 @@ fun DataFrame<*>.resolveAnswer(question: String, answerCol: Int = 1): String {
...
+  val answer = answers[questionIndex].let {
+    if (it is Double) it.toInt() else it
+  }.toString().trim()
```
davidatkinsuk, line 115: would be good to see a test for this, maybe update an existing test's room number to emulate this scenario?

davidatkinsuk (approved):
The combination of clean code and breaking changes into commits made this _very_ easy to review. Just a few comments from me

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/Cas1SiteSurveyPremiseFactory.kt
danhumphreys-moj: Added new private function:
```
  fun removeDecimalPlaces() = answers[questionIndex].let {
    if (it is Double) it.toInt() else it
  }.toString().trim()

  val answer = removeDecimalPlaces()
```

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/seed/cas1/Cas1SiteSurveyPremiseFactory.kt
danhumphreys-moj: Yes I tried to do this yesterday and hit a blocker.  Revisited it again this morning and figured out why (`rows` needed an explicit type definition adding after changing the type of the `addCharacteristics` extension function from `<String>` to `<Any>`.  This worked and I have changed a few of the tests to populate the dataframe test data with double room and bed number types.  Didn't change all tests though since we also need it to work with Strings.

PR #2922, opened 2025-02-04
Removed premise from Cas1PremiseCapacity end point.

vbala-moj:
https://dsdmoj.atlassian.net/browse/APS-1715


davidatkinsuk (approved):
LGTM 👍

PR #2924, opened 2025-02-05
Refactor the getUser method to fallback to searching for the user in the system if not found in Delius

vbala-moj:
https://dsdmoj.atlassian.net/browse/APS-1715


davidatkinsuk (approved):
Looks great.

Something to consider going forward - can we please try and limit the first line of the commit to 50 characters so it isn't truncated in github. Thanks!

PR #2933, opened 2025-02-05
Re-enable domain events in dev

davidatkinsuk:
These are required by the probation-integration end to end tests


helm_deploy/values-dev.yaml
```diff
@@ -36,9 +36,11 @@ generic-service:
     SENTRY_ENVIRONMENT: dev
     SENTRY_TRACES-SAMPLE-RATE: 0.01
     CACHES_STAFFMEMBERS_EXPIRY-SECONDS: 20
+    # do not disable - these are used by probation-integration e2e tests
     DOMAIN-EVENTS_CAS1_EMIT-ENABLED: false
```
danhumphreys-moj, line 40: This value is false yet the comment above says not to disable.  Presumably false is correct - maybe the comment needs modifying?

danhumphreys-moj (approved):

helm_deploy/values-dev.yaml
davidatkinsuk: my brain needs modifying, hang on :)

PR #2947, opened 2025-02-07
APS-1887 Added exclude space booking id for day summary.

vbala-moj:
https://dsdmoj.atlassian.net/browse/APS-1887

### DIFF
--- a/src/main/resources/static/cas1-api.yml
+++ b/src/main/resources/static/cas1-api.yml
@@ -1078,6 +1078,13 @@ paths:
           description: The field to sort the booking by. If not defined, will sort on person name
           schema:
             $ref: 'cas1-schemas.yml#/components/schemas/Cas1SpaceBookingDaySummarySortField'
+        - name: excludeSpaceBookingId
+          in: query
+          description: ID of a booking to exclude from capacity calculations and the list of space bookings
+          required: false
+          schema:
+            type: string
+            format: uuid
       responses:
         200:
           description: successful operation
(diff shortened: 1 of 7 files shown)


src/main/resources/static/cas1-api.yml
```diff
@@ -1003,6 +1003,13 @@ paths:
           description: The field to sort the booking by. If not defined, will sort on person name
           schema:
             $ref: 'cas1-schemas.yml#/components/schemas/Cas1SpaceBookingDaySummarySortField'
+        - name: excludeSpaceBookingId
```
danhumphreys-moj, line 1006: I wonder whether a better name would be `excludedSpaceBookingId`?  

`excludeSpaceBookingId` initially made me think this was a Boolean type.

danhumphreys-moj (approved):
A couple of minor observations.

src/main/resources/static/cas1-api.yml
davidatkinsuk: I prefer `exclude` because it reads as a command. I think `excluded` reads as a response e.g. i have excluded this space booking. but i'm not sure it matters too much either way :) we have an existing endpoint using `excludeSpaceBookingId`

davidatkinsuk: actually i think `excludedSpaceBookingId` may be better but we'd have to update the UI for the existing endpoint so I propose we leave as is

vbala-moj: That is the name used in other places, so I followed it. Also, it is an API parameter that instructs the endpoint to perform an action (e.g., exclude a specific spaceBookingId from processing). Therefore, excludeSpaceBookingId is appropriate, I think.

src/main/resources/static/cas1-api.yml
```diff
@@ -1078,6 +1078,13 @@ paths:
...
+          in: query
+          description: ID of a booking to exclude from capacity calculations
```
davidatkinsuk, line 1083: this should maybe say

> ID of a booking to exclude from capacity calculations and the list of space bookings

davidatkinsuk (approved):
LGTM 👍

PR #2963, opened 2025-02-10
CAS-1205 - added working day postgres function

daveawc:
Added a new function in postgres that will be used in the bed search query improvements,

### DIFF
--- a/src/main/resources/db/migration/all/20250210162533__create_function_add_working_days.sql
+++ b/src/main/resources/db/migration/all/20250210162533__create_function_add_working_days.sql
@@ -1,0 +1,31 @@
+--function to add working days onto a given date.
+CREATE OR REPLACE FUNCTION add_working_days(
+    start_date DATE,
+    working_days_to_add INTEGER,
+    bank_holidays DATE[]
+) RETURNS DATE AS
+$$
+DECLARE
+    date_to_check DATE    := start_date;
+    added_days    INTEGER := 0;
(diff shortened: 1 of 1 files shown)


src/main/resources/db/migration/all/20250210162533__create_next_working_day_function.sql.sql
```diff
@@ -0,0 +1,22 @@
...
+        if extract(isodow from day_to_check) not in (6, 7)
+            and not exists (select 1 from unnest(bank_holidays) bh where bh::date = day_to_check) then
+            return day_to_check;
+        end if;
+
+        -- or else it IS a Saturday or Sunday, or IS a bank holiday, so move to the next day
+        day_to_check := day_to_check + interval '1 day';
+    end loop;
+end;
+$$ language plpgsql;
```
davidatkinsuk, line 22: I'm not sure on the context of this, but does the time of day matter? We do something in CAS1 (albeit in kotlin) in https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/bfa4a9775aa6b5c4a1be7b6dacd72fe452def6ab/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/TaskDeadlineService.kt#L63. See `slewedToWorkingPattern` for how we consider time of day

davidatkinsuk (approved):
LGTM 👍

PR #2964, opened 2025-02-10
CAS-1312 Prevent users from archiving a property with an existing bookings

muhammad-elabdulla:
This PR CAS1312 is to prevent users from archiving a property when there are a booking in provisional, confirmed, or active status. 

I added it as a feature until the UI done their change to handle the Api response

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BookingEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BookingEntity.kt
@@ -269,6 +269,11 @@ interface BookingRepository : JpaRepository<BookingEntity, UUID> {
     nativeQuery = true,
   )
   fun findAllIdsByPremisesId(premisesId: UUID): List<UUID>
+
+  @Query(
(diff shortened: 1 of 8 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BookingEntity.kt
```diff
@@ -269,6 +269,11 @@ interface BookingRepository : JpaRepository<BookingEntity, UUID> {
     nativeQuery = true,
   )
   fun findAllIdsByPremisesId(premisesId: UUID): List<UUID>
+
+  @Query(
+    "SELECT b FROM BookingEntity b WHERE b.service = :serviceName AND b.premises.id = :premisesId AND b.arrivalDate >= :date AND b.status in :statuses",
```
davidatkinsuk, line 274: is it a problem if there is a booking that arrived before today but is yet to depart? If yes I guess maybe this should be based upon departure date instead? (but i don't fully understand how CAS3 uses bookings)

davidatkinsuk (approved):
just one comment from me 👍

PR #2966, opened 2025-02-11
APS-1934 Miscellaneous Optimisations

davidatkinsuk:
* Log response times locally to aid performance investigations
 * Add missing index on foreign key
 * Fix a bug that blocked the API starting locally if lots of applications existed in the database


gregkhawkins (approved):
LGTM

PR #2967, opened 2025-02-11
APS-1885 Added assessment reporting properties.

vbala-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AssessmentService.kt
```diff
@@ -378,23 +369,62 @@ class AssessmentService(
...
-    if (assessment is ApprovedPremisesAssessmentEntity) {
-      if (assessmentData == null) {
-        validationErrors["$.data"] = "empty"
-      } else if (!jsonSchemaService.validate(assessment.schemaVersion, assessmentData)) {
-        validationErrors["$.data"] = "invalid"
-      }
+    return CasResult.Success(assessment)
+  }
 
+  private fun validateAssessmentData(
```
davidatkinsuk, line 375: thanks for removing this duplication of logic. In the future would you mind putting similar refactoring into a separate commit on the PR, it then makes it easier to untangle it from the functional change being made 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AssessmentService.kt
```diff
@@ -378,23 +369,62 @@ class AssessmentService(
...
+    assessment: AssessmentEntity,
+    placementRequirements: PlacementRequirements? = null,
+    isAcceptAssessment: Boolean = false,
+  ): CasResult<AssessmentEntity> {
+    val validationErrors = ValidationErrors()
+    if (assessment.data == null) {
+      validationErrors["$.data"] = "empty"
+    } else if (!jsonSchemaService.validate(assessment.schemaVersion, assessment.data!!)) {
+      validationErrors["$.data"] = "invalid"
+    }
```
davidatkinsuk, line 385: I think _all_ validations in here are CAS1 only, looking at the old version of this code? Maybe this function should be called `validateCas1AssessmentData` and take in an ApprovedPremisesAssessmentEntity?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AssessmentService.kt
```diff
@@ -378,23 +369,62 @@ class AssessmentService(
...
+    if (assessment is ApprovedPremisesAssessmentEntity && isAcceptAssessment) {
```
davidatkinsuk, line 386: given this only applies to accept assessment, i don't think it should be in a generic function (i.e. leave it in the accept function)?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AssessmentService.kt
```diff
@@ -340,28 +340,19 @@ class AssessmentService(
     return CasResult.Success(savedAssessment)
   }
 
-  fun acceptAssessment(
-    acceptingUser: UserEntity,
+  private fun validateAssessment(
```
davidatkinsuk, line 343: would you mind moving the two new private functions to the bottom of this file? I think it may make the diff easier to read

davidatkinsuk (changes requested):
A few comments from me, i'll finish reviewing once resolved as i think they'll make the diff more readable 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AssessmentService.kt
davidatkinsuk: that did the job, thank you

davidatkinsuk (approved):
This is great and the refactoring is going to make completely splitting the CAS1/CAS3 code much easier 👍

PR #2983, opened 2025-02-13
CAS-1411 - add sexual risk flag to overlapped bookings

daveawc:


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/BedSearchTest.kt
```diff
@@ -1915,7 +1923,7 @@ class BedSearchTest : IntegrationTestBase() {
       return beds
     }
 
-    private fun createAssessment(user: UserEntity, crn: String): Pair<TemporaryAccommodationApplicationEntity, AssessmentEntity> {
+    private fun createAssessment(user: UserEntity, crn: String, sexualRisk: Boolean? = null): Pair<TemporaryAccommodationApplicationEntity, AssessmentEntity> {
```
davidatkinsuk, line 1926: is it worth testing all 3 fields that are considered sexual risk?

src/main/resources/static/_shared.yml
```diff
@@ -4377,13 +4377,16 @@ components:
         assessmentId:
           type: string
           format: uuid
+        isSexualRisk:
```
davidatkinsuk, line 4380: the ticket says this should be 'poses sexual risk'. i'm not sure if it matters too much

davidatkinsuk (approved):
just a few minor comments from me 👍

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/BedSearchTest.kt
davidatkinsuk: given the three fields are checked in SQL i think it has to be in an integration test?

PR #2986, opened 2025-02-14
APS-1885 Update approved premises  assessment job.

vbala-moj:
[https://dsdmoj.atlassian.net/browse/APS-1942](url)

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/migration/cas1/Cas1UpdateAssessmentReportPropertiesJob.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/migration/cas1/Cas1UpdateAssessmentReportPropertiesJob.kt
@@ -1,0 +1,68 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.migration.cas1
+
+import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.data.jpa.repository.Modifying
+import org.springframework.data.jpa.repository.Query
+import org.springframework.stereotype.Component
+import org.springframework.stereotype.Repository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApprovedPremisesAssessmentEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.migration.MigrationJob
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.migration.MigrationLogger
+import java.util.UUID
+
+@Component
+class Cas1UpdateAssessmentReportPropertiesJob(
+  private val migrationLogger: MigrationLogger,
+  private val updateAssessmentReportPropertiesRepository: UpdateAssessmentReportPropertiesRepository,
+  override val shouldRunInTransaction: Boolean = true,
+) : MigrationJob() {
+
+  override fun process(pageSize: Int) = executeUpdateAssessmentReportProperties()
+
+  @SuppressWarnings("TooGenericExceptionCaught")
+  private fun executeUpdateAssessmentReportProperties() {
+    try {
+      updateAssessmentReportPropertiesRepository.updateAssessmentReportProperties()
+    } catch (exception: Exception) {
+      migrationLogger.error("Unable to run update assessment report properties job", exception)
+    }
+  }
+}
+
+@Repository
+interface UpdateAssessmentReportPropertiesRepository : JpaRepository<ApprovedPremisesAssessmentEntity, UUID> {
+
+  companion object {
+    private const val QUERY_UPDATE_ASSESSMENT_REPORT_PROPERTIES = """
+    WITH assessment_data AS (
+      SELECT
+          a.id AS id,
+          a.data -> 'suitability-assessment' -> 'application-timeliness' ->> 'agreeWithShortNoticeReason' AS agreeWithShortNoticeReason,
+          a.data -> 'suitability-assessment' -> 'application-timeliness' ->> 'agreeWithShortNoticeReasonComments' AS agreeWithShortNoticeReasonComments,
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/migration/cas1/Cas1UpdateAssessmentReportPropertiesJob.kt
```diff
@@ -0,0 +1,69 @@
...
+            a.data -> 'suitability-assessment' -> 'application-timeliness' ->> 'reasonForLateApplication' AS reasonForLateApplication
+        FROM 
+            assessments AS a
+        LEFT JOIN 
+            approved_premises_assessments AS apa
+        ON 
+            apa.assessment_id = a.id
+        WHERE 
+            a.service = 'approved-premises'
+            AND a.submitted_at IS NULL
```
davidatkinsuk, line 59: this isn't mentioned on the ticket, but i think this needs to be `IS NOT NULL` i.e. it should only apply to submitted assessments

vbala-moj: will update. Thanks.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/migration/cas1/Cas1UpdateAssessmentReportPropertiesJobTest.kt
```diff
@@ -0,0 +1,113 @@
...
+
+      val agreeWithShortNoticeReason = when (extractedData["agreeWithShortNoticeReason"]?.toString()?.lowercase()) {
+        "yes" -> true
+        "no" -> false
+        else -> null
+      }
+      assertThat(assessmentAfterUpdate.agreeWithShortNoticeReason).isEqualTo(agreeWithShortNoticeReason)
+      assertThat(assessmentAfterUpdate.agreeWithShortNoticeReasonComments).isEqualTo(extractedData["agreeWithShortNoticeReasonComments"])
+      assertThat(assessmentAfterUpdate.reasonForLateApplication).isEqualTo(extractedData["reasonForLateApplication"])
+    }
```
davidatkinsuk, line 73: i would rather that this test was deterministic, if possible (i.e. we know exactly which assessments will be in what state at the end from reading the code). At the moment i think it's random, and it's possible that on a given test run it only checks one scenario e.g. where agreeWithShortNoticeReason is "yes". We should then be able to get rid of `extractAssessmentData`?

Could we not instead setup a few distinct assessments that have this fields explicitly set to known values, and then in assertions explicitly assert that expected values?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/migration/cas1/Cas1UpdateAssessmentReportPropertiesJobTest.kt
```diff
@@ -0,0 +1,113 @@
...
+      mapOf(
+        "suitability-assessment" to mapOf(
+          "application-timeliness" to mapOf(
+            "agreeWithShortNoticeReason" to "no",
+            "agreeWithShortNoticeReasonComments" to "Applicant disagrees as they were not informed in advance.",
+            "reasonForLateApplication" to null,
+          ),
+        ),
+      ),
+    )
```
davidatkinsuk, line 110: i think it would be helpful to add a scenario where these fields don't exist in the JSON at all, this will happen if the application is not short notice (e.g. the questions are not asked at all)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/migration/cas1/Cas1UpdateAssessmentReportPropertiesJob.kt
```diff
@@ -0,0 +1,69 @@
...
+    reason_for_late_application = source.reasonForLateApplication
+    FROM (
+        SELECT 
+            a.id AS id,
+            a.data -> 'suitability-assessment' -> 'application-timeliness' ->> 'agreeWithShortNoticeReason' AS agreeWithShortNoticeReason,
+            a.data -> 'suitability-assessment' -> 'application-timeliness' ->> 'agreeWithShortNoticeReasonComments' AS agreeWithShortNoticeReasonComments,
+            a.data -> 'suitability-assessment' -> 'application-timeliness' ->> 'reasonForLateApplication' AS reasonForLateApplication
+        FROM 
+            assessments AS a
+        LEFT JOIN
```
davidatkinsuk, line 53: if you changed this to an inner join on approved_premises_assessments you can then get rid of the check `a.service = 'approved-premises'`

davidatkinsuk (approved):
Implementations looks great, it would be good if we could please make some tweaks to the test 👍

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/migration/cas1/Cas1UpdateAssessmentReportPropertiesJobTest.kt
```diff
@@ -0,0 +1,122 @@
...
+      withReasonForLateApplication(null)
+      withSubmittedAt(OffsetDateTime.now())
+      withData(
+        objectMapper.writeValueAsString(
+          mapOf(
+            "suitability-assessment" to emptyMap<String, Any>(),
+          ),
+        ),
+      )
+    }
```
davidatkinsuk, line 103: perfect!

davidatkinsuk (approved):
Looks great, thanks for the changes 👍

PR #2993, opened 2025-02-17
Disable premise overbooking banner

davidatkinsuk:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1PremisesService.kt
```diff
@@ -58,7 +58,7 @@ class Cas1PremisesService(
   private fun buildOverBookingSummary(premises: ApprovedPremisesEntity): List<Cas1OverbookingRange> {
     val premisesCapacitySummary = spacePlanningService.capacity(
       premises,
-      range = DateRange(LocalDate.now(), LocalDate.now().plusWeeks(OVERBOOKING_RANGE_DURATION_WEEKS)),
+      rangeInclusive = DateRange(LocalDate.now(), LocalDate.now().plusWeeks(OVERBOOKING_RANGE_DURATION_WEEKS).minusDays(1)),
```
gregkhawkins, line 61: should we test this shift to inclusiveness? Can't see anything related in unit or int test

gregkhawkins (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1PremisesService.kt
davidatkinsuk: it always was inclusive, i just renamed to clarify behaviour incase we want to re-enable this

gregkhawkins: but we are adding `.minusDays(1)),` which is new. Does that not have an impact?

PR #2996, opened 2025-02-17
Scheduled job to refresh inmate-details cache

gregkhawkins:
* The cas1-enable-scheduled-job-refresh-inmate-details feature-flag is used here to switch on the new InmateDetailsCacheRefreshScheduledJob scheduled job and switch off the pre-existing PreemptiveCacheRefresher job
* We found the PreemptiveCacheRefresher was slowing down the performance of the application as a whole (as it was locking redis with it's locking mechanism)
* The new InmateDetailsCacheRefreshScheduledJob replaces PreemptiveCacheRefresher and uses the spring SchedulerLock instead
* The implementation code in InmateDetailsCacheRefreshScheduledJob is the same as PreemptiveCacheRefresher but cleaned up a little
* Integration test included to test the new InmateDetailsCacheRefreshScheduledJob


src/test/resources/application-test.yml
```diff
@@ -213,5 +213,11 @@ scheduling:
 
 logging.level.org.springframework.jdbc.core.JdbcTemplate: debug
 
+refresh-inmate-details-cache:
+  lock-at-least-for: 1s
+  lock-at-most-for: 1s
```
gregkhawkins, line 218: This configuration is need so that the lock can be aquired for all tests. When this inherits `lock-at-most-for: 210m` only the first test passes

src/test/resources/application-test.yml
```diff
@@ -213,5 +213,10 @@ scheduling:
...
+  logging-enabled: false
```
gregkhawkins, line 219: This configuration is need so that the lock can be acquired for all tests. When this inherits lock-at-most-for: `210m` from `application.yml` only the first test passes

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/scheduled/InmateDetailsCacheRefreshScheduledJob.kt
```diff
@@ -0,0 +1,118 @@
...
+          val prisonsApiResult = prisonsApiClient.getInmateDetailsWithCall(nomsNumber)
+          incrementCacheEntryStats(inmateDetailsCacheRefreshResults, nomsNumber, prisonsApiResult)
+        }
+      }
+      log.info("Have completed refreshing inmate details cache. ${inmateDetailsCacheRefreshResults.stats()}")
+      return inmateDetailsCacheRefreshResults
+    }
+  }
+
+  private fun shouldLogProgress(entriesProcessed: Int) = entriesProcessed == 1 || entriesProcessed % 5000 == 0
```
davidatkinsuk, line 63: nice

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/scheduled/InmateDetailsCacheRefreshScheduledJob.kt
```diff
@@ -0,0 +1,118 @@
...
+      log.info("Cache status for $nomsNumber: $cacheEntryStatus")
+      if (cacheEntryStatus == PreemptiveCacheEntryStatus.EXISTS) {
+        log.info("No upstream call made when refreshing Inmate Details for $nomsNumber, stored result still within soft TTL")
+      }
+    }
+    return cacheEntryStatus
+  }
+
+  private fun incrementCacheEntryStats(inmateDetailsCacheRefreshResults: InmateDetailsCacheRefreshResults, nomsNumber: String, prisonsApiResult: ClientResult<InmateDetail>) {
+    if (prisonsApiResult is ClientResult.Failure.StatusCode) {
```
davidatkinsuk, line 88: could we use `when(prisonsApiResult)` instead of these 3 `if` checks, ensuring that the checks are exhaustive?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/InmateDetailsCacheRefreshScheduledJobTest.kt
```diff
@@ -0,0 +1,117 @@
...
+
+  @Autowired
+  lateinit var inmateDetailsCacheRefreshScheduledJob: InmateDetailsCacheRefreshScheduledJob
+
+  @Autowired
+  lateinit var domainEventService: Cas1DomainEventService
+
+  @AfterEach
+  fun afterEach() {
+    Thread.sleep(1000)
```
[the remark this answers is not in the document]
davidatkinsuk: i would assume locking only applies when spring scheduler invokes the annotated function, i'm surprised to see that it still applies when calling the function directly!

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/InmateDetailsCacheRefreshScheduledJobTest.kt
```diff
@@ -0,0 +1,117 @@
...
+
+  @Test
+  fun `successfully updates 1 cache entry`() {
+    mockFeatureFlagService.setFlag("cas1-enable-scheduled-job-refresh-inmate-details", true)
+    givenAProbationRegion { probationRegion ->
+      givenAUser(probationRegion = probationRegion) { user, _ ->
+        val inmateDetail = createInmateDetailAndApplication(user)
+        prisonAPIMockSuccessfulInmateDetailsCall(inmateDetail)
+
+        val results = inmateDetailsCacheRefreshScheduledJob.refreshInmateDetailsCache()!!
```
davidatkinsuk, line 48: nice usage of the result type to do some clean assertions in the test 👍

src/test/resources/application-test.yml
davidatkinsuk: i'm guessing it doesn't like values less than 1 second?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/InmateDetailsCacheRefreshScheduledJobTest.kt
```diff
@@ -0,0 +1,117 @@
...
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.integration.givens.givenAUser
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.integration.httpmocks.prisonAPIMockSuccessfulInmateDetailsCall
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.prisonsapi.InmateDetail
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.prisonsapi.InmateStatus
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas1.Cas1DomainEventService
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas1.scheduled.InmateDetailsCacheRefreshScheduledJob
+import java.util.UUID
+
+class InmateDetailsCacheRefreshScheduledJobTest : IntegrationTestBase() {
```
davidatkinsuk, line 21: thanks for the very thorough test

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/scheduled/InmateDetailsCacheRefreshScheduledJob.kt
```diff
@@ -0,0 +1,118 @@
...
+  private val prisonsApiClient: PrisonsApiClient,
+  private val cacheRefreshExclusionsInmateDetailsRepository: CacheRefreshExclusionsInmateDetailsRepository,
+  private val featureFlagService: FeatureFlagService,
+  @Value("\${refresh-inmate-details-cache.logging-enabled}") private val loggingEnabled: Boolean,
+) {
+  private val log = LoggerFactory.getLogger(this::class.java)
+
+  @Scheduled(cron = "0 0 0/4 * * *")
+  @SchedulerLock(
+    name = "cas1_inmate_details_cache_refresh",
```
davidatkinsuk, line 33: this cache is shared across cas1 and cas3, so it can maybe just be called `inmate_details_cache_refresh`?

davidatkinsuk (approved):
just a few very minor thoughts from me 👍

src/test/resources/application-test.yml
gregkhawkins: correct, I did try looking into millis but seconds is smallest supported. Makes sense I guess as what would a usage in production of millis realistically - frustrating for tests though

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/InmateDetailsCacheRefreshScheduledJobTest.kt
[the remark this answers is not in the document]
gregkhawkins: it is strange, especially as scheduled tasks are switched off in `application-test.yml` you would have though that also switched off the lock. 
There was quite a bit of head scratching by me initially when my debugger hit the a breakpoint in the implementation when i ran the test first time round and then did not hit the breakpoint second time. I then commented out the `@SchedulerLock` annotation and the debugger consistently hit the breakpoint with every run. 
So I set about building all the tests with the annotation commented out. When I put the annotation back in; the first test passed and all others tests failed. 
So, it looks like the lock applies independently of the scheduled task. Very weird.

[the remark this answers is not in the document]
davidatkinsuk: Yep it's surprising. I guess it was the easiest way to get shedlock interacting with spring scheduling.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/scheduled/InmateDetailsCacheRefreshScheduledJob.kt
gregkhawkins: will do!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/scheduled/InmateDetailsCacheRefreshScheduledJob.kt
gregkhawkins: thanks, will do!

PR #2999, opened 2025-02-17
CAS-1205 - Add filtering to bedspace search

daveawc:
This PR updates how filtering works on bed searches. You can now pass in a list of UUIDs for premises and bedspace (room), which can either be included or excluded. This is because, for example, we want to see beds that ARE wheelchair accessible, but ARE NOT suitable for sexual offenders.

There is going to be ongoing refactoring work which will remove the mapping from characteristicId to propertyName, and that will be removed in a future PR.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/CharacteristicService.kt
```diff
@@ -33,6 +34,11 @@ class CharacteristicService(
     }
   }
 
+  fun getTemporaryAccommodationCharacteristics(): List<CharacteristicEntity> = characteristicRepository.findActiveByServiceScopeAndModelScope(
```
davidatkinsuk, line 37: can we rename this to `getCas3Characteristics` please, it's less wordy :)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas3/Cas3BedspaceSearchService.kt
```diff
@@ -126,9 +106,69 @@ class Cas3BedspaceSearchService(
...
+    }
+
+    if (searchParams.bedspaceFilters == null && searchParams.premisesFilters == null) {
+      return results
+    }
+
+    val characteristicPropertyNames =
+      characteristicService.getTemporaryAccommodationCharacteristics().associateBy({ it.id }, { it.propertyName })
+
+    // we will use IDs in the query in future refactoring, but for now it only returns names, so map IDs to characteristic propertyNames.
```
davidatkinsuk, line 132: I'd suggest using propertyNames instead of IDs in the future. You can then also enumerate them in the spec.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas3/Cas3BedspaceSearchService.kt
```diff
@@ -126,9 +106,69 @@ class Cas3BedspaceSearchService(
...
+    val premisesToInclude =
+      searchParams.premisesFilters?.includedCharacteristicIds?.map { characteristicPropertyNames[it]!! }
+    val premisesToExclude =
+      searchParams.premisesFilters?.excludedCharacteristicIds?.map { characteristicPropertyNames[it]!! }
+
+    return results
+      .filter { result ->
+        val premisesCharacteristics = result.premisesCharacteristics.map { it.propertyName }
+        (premisesToInclude.isNullOrEmpty() || premisesToInclude.all { it in premisesCharacteristics }) &&
+          (premisesToExclude.isNullOrEmpty() || premisesToExclude.none { it in premisesCharacteristics })
```
davidatkinsuk, line 146: would it maybe be more readable to use `contains` instead e.g.

```
 (premisesCharacteristics.contains(premisesToInclude ?: emptyList()) &&
 (!premisesCharacteristics.contains(premisesToExclude ?: emptyList())
```

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas3/Cas3BedspaceSearchService.kt
```diff
@@ -126,9 +106,69 @@ class Cas3BedspaceSearchService(
...
+      }
   }
 
+  @Deprecated("adding to maintain functionality until the UI switches to use search filters")
+  private fun applyLegacyFilters(
+    searchParams: TemporaryAccommodationBedSearchParameters,
+    results: List<Cas3BedSearchResult>,
+  ): List<Cas3BedSearchResult> = results
+    .filter { bedspace ->
+      searchParams.attributes?.contains(BedSearchAttributes.WHEELCHAIR_ACCESSIBLE) != true ||
```
davidatkinsuk, line 160: the calling code checks if `searchPrams.attributes` is non null, so we (logically) know at this point it's not null. If you passed through a non null list of attributes to this function you wouldn't need the `?` and can then replace `!= true` with a simple `!` inversion on the predicate?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/BedSearchTest.kt
```diff
@@ -984,6 +989,318 @@ class BedSearchTest : IntegrationTestBase() {
...
+            premisesFilters = PremisesFilters(
+              includedCharacteristicIds = listOf(characteristicOne.id),
+            ),
+          ),
+        )
+
+        val returnedPremisesIds = result.results.map { it.premises.id }
+
+        assertThat(returnedPremisesIds).containsExactly(premisesOne.id)
+        assertThat(returnedPremisesIds).doesNotContain(premisesTwo.id)
```
davidatkinsuk, line 1041: I think `doesNotContain` is redundant in this case, same applies for other tests

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/BedSearchTest.kt
```diff
@@ -984,6 +989,318 @@ class BedSearchTest : IntegrationTestBase() {
...
+        // only rooms 1 and 2 should be returned, as both have RoomCharacteristicOne and neither have RoomCharacteristicThree
+        val returnedBedIds = result.results.map { it.bed.id }
+        assertThat(returnedBedIds).hasSize(4)
+          .containsExactlyInAnyOrder(
+            premisesOneBedOne.id,
+            premisesOneBedTwo.id,
+            premisesTwoBedOne.id,
+            premisesTwoBedTwo.id,
+          )
+      }
```
davidatkinsuk, line 1278: nice

davidatkinsuk (approved):
LGTM, just some minor comments

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas3/Cas3BedspaceSearchService.kt
davidatkinsuk: i'm not sure but as this is temporary, don't worry about it :)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas3/Cas3BedspaceSearchService.kt
davidatkinsuk: ah ok. in CAS1 our property names pretty static, maybe it's different in CAS3

PR #3001, opened 2025-02-18
Generate CAS2 Bail types on merging to main

libuk:
We have a newly created accommodation service which requires its types to be kept up to date with the API models.


davidatkinsuk (approved):
LGTM 👍

PR #3006, opened 2025-02-18
feat: created a cas2v2 reports subsystem

tobybatchmoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas2v2/Cas2v2ApplicationStatusUpdatesReportRepository.kt
```diff
@@ -24,10 +26,13 @@ interface Cas2v2ApplicationStatusUpdatesReportRepository : JpaRepository<DomainE
...
 
       FROM domain_events events
+      LEFT JOIN cas_2_v2_applications applications ON events.application_id = applications.id
       LEFT JOIN LATERAL jsonb_array_elements(events.data -> 'eventDetails' -> 'newStatus' -> 'statusDetails') as details ON true
       WHERE events.type = 'CAS2_APPLICATION_STATUS_UPDATED'
         AND events.occurred_at  > CURRENT_DATE - 365
-      GROUP BY events.id  
+      GROUP BY 
+        events.id,
+        applications.application_origin
```
davidatkinsuk, line 35: is this group by required? I would think the event.id is sufficient but i could be wrong

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas2v2/Cas2v2SubmittedApplicationReportRepository.kt
```diff
@@ -21,10 +22,10 @@ interface Cas2v2SubmittedApplicationReportRepository : JpaRepository<DomainEvent
         CAST(events.data -> 'eventDetails' ->> 'hdcEligibilityDate' as DATE) AS hdcEligibilityDate,
         CAST(events.data -> 'eventDetails' ->> 'conditionalReleaseDate' as DATE) AS conditionalReleaseDate,
         TO_CHAR(events.occurred_at,'YYYY-MM-DD"T"HH24:MI:SS') AS submittedAt,
-        TO_CHAR(applications.created_at, 'YYYY-MM-DD"T"HH24:MI:SS') AS startedAt
+        TO_CHAR(applications.created_at, 'YYYY-MM-DD"T"HH24:MI:SS') AS startedAt,
+        applications.application_origin as applicationOrigin
```
davidatkinsuk, line 26: you have a field `cas2v2ApplicationOrigin` on the `CAS2_APPLICATION_SUBMITTED` domain event. You could (should?) use that instead, default null cases to the original CAS2 application type i guess

(although on reflection i feel like it should just be called `applicationOrigin` on the domain event)

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2v2/Cas2v2ReportsTest.kt
```diff
@@ -264,6 +270,60 @@ class Cas2v2ReportsTest : Cas2v2IntegrationTestBase() {
           Assertions.assertThat(actual).isEqualTo(expectedDataFrame)
         }
     }
+
+    @Test
+    fun `streams spreadsheet of cas2v2 Cas2SubmittedApplicationEvents, with application origin`() {
```
davidatkinsuk, line 275: 3 reports have been updated, should we have a few more tests added/updated?

davidatkinsuk (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas2v2/Cas2v2ApplicationStatusUpdatesReportRepository.kt
tobybatchmoj: I was getting an error if I omitted it, I don't see why either byt the ORM didn't like it.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas2v2/Cas2v2SubmittedApplicationReportRepository.kt
tobybatchmoj: Those should (will) both be same, but i see your point

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2v2/Cas2v2ReportsTest.kt
tobybatchmoj: I'll create a ticket for that.

PR #3044, opened 2025-02-25
APS-1965 Handle empty string when update assessments.

vbala-moj:
[https://dsdmoj.atlassian.net/browse/APS-1965](url)


davidatkinsuk (approved):
LGTM 👍

PR #3048, opened 2025-02-26
Refactor Cas2 Applications to use CasResult

daveawc:
2nd PR refactoring Cas2 Applications to use CasResult


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/OffenderService.kt
```diff
@@ -236,7 +236,23 @@ class OffenderService(
...
+      is ClientResult.Failure.StatusCode -> if (offenderResponse.status == HttpStatus.NOT_FOUND) {
+        return CasResult.NotFound(
+          "OffenderDetailSummary",
+          crn,
+        )
+      } else {
+        offenderResponse.throwException()
+      }
+
+      is ClientResult.Failure -> offenderResponse.throwException()
```
davidatkinsuk, line 251: i appreciate you're maintaining existing behaviour, but could these not be returned as `CasResult` types instead of throwing exceptions? Or is too much for this PR? I don't mind either way

davidatkinsuk (approved):
LGTM 👍

PR #3053, opened 2025-02-26
refactored cas2 applications controller to use cas2 models.

daveawc:
updated operationId names, and also updated to return cas2 model implementations.
split cas2 yml out into its own schema.
added missing cas2v2 definitions
updated references to use cas2v2 models


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas2/ApplicationsController.kt
```diff
@@ -118,7 +118,7 @@ class ApplicationsController(
 
   private fun getPersonDetailAndTransform(
     application: Cas2ApplicationEntity,
-  ): Application {
+  ): Cas2Application {
     val personInfo = offenderService.getFullInfoForPersonOrThrow(application.crn)
 
     return applicationsTransformer.transformJpaToApi(application, personInfo)
```
davidatkinsuk, line 124: presumably this transformer needs updating? or am i missing something

davidatkinsuk: ah, inheritance 👍

davidatkinsuk (approved):
LGTM 👍

davidatkinsuk:
Changes moving cas2 schema to it's own file looks good 👍

PR #3081, opened 2025-02-28
Chore/refactor cas2 applications

daveawc:
- removed duplicated calls to get offender information
- corrected functionality to match [the documentation](https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/4680581265/CAS2+Access+to+offenders+with+restrictions+or+exclusions) that should only allow applications to be created on Full person types.
- replaced deprecated @OrderBy annotations

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/Cas2OffenderService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/Cas2OffenderService.kt
@@ -174,6 +174,10 @@ class Cas2OffenderService(
       }
     }
 
+    if (offender.currentExclusion || offender.currentRestriction) {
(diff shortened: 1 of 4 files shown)


davidatkinsuk (approved):
LGTM 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/Cas2OffenderService.kt
```diff
@@ -170,6 +174,10 @@ class Cas2OffenderService(
       }
     }
 
+    if (offender.currentExclusion || offender.currentRestriction) {
+      return PersonInfoResult.Success.Restricted(crn, offender.otherIds.nomsNumber)
+    }
```
davidatkinsuk, line 179: whilst i appreciate this correct, there are _alot_ of controller functions that no longer did this check that now do. I think that's correct, just flagging it in case we start seeing some issues.

PR #3084, opened 2025-03-03
Improve Limited Access Offender Strategy Names and Documentation

davidatkinsuk:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/LaoStrategy.kt
```diff
@@ -0,0 +1,67 @@
...
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.PersonInfoResult
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.PersonSummaryInfoResult
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.deliuscontext.CaseSummary
+
+/**
+ * A Limited Access Offender (LAO) is one for which access is limited to
+ * specific delius users.
+ *
+ * An Offender is an LAO if they are configured with either a restriction (allow list)
+ * or an exclusion (deny list) in Delius. When retrieving an offender boolean fields
```
Rosie-Brigham, line 16: Maybe an additional line to explain what the allow and deny list is?

davidatkinsuk: Good idea, i've renamed this paragraph to hopefully make it clearer:

```
 An Offender is an LAO if they are configured with either a restriction
 or an exclusion in Delius. For restrictions, the calling user is allowed
 to view the offender if they are in an 'allow list' configured in Delius.
 For exclusions, the calling user can view the offender unless they are
 in a 'deny list' configured in Delius.
 ```

PR #3085, opened 2025-03-03
Rip-out feature flag for refresh inmate-details scheduled job

gregkhawkins:
I previously deployed a new scheduled job for refreshing the inmates-details cache. As this has been switched on in production (and stable) for a week, I am now ripping out the old job and the feature flag to toggle between old and new


davidatkinsuk (approved):
Glad to see this gone! 👍

PR #3086, opened 2025-03-03
APS-2039: Split Report Date and Time Fields

vbala-moj:
[https://dsdmoj.atlassian.net/browse/APS-2039](url)

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas1/Cas1PlacementReportRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas1/Cas1PlacementReportRepository.kt
@@ -13,41 +13,59 @@ class Cas1PlacementReportRepository(
 
   companion object {
     private const val SELECT_DATASET_QUERY = """
-          SELECT csb.id AS placement_id,
-                 p.name AS premises_name,
-                 pr.name AS premises_region,
-                 csb.expected_arrival_date,
-                 COALESCE(csb.expected_departure_date - csb.expected_arrival_date, 0) AS expected_duration_nights,
-                 csb.expected_departure_date,
-                 (csb.actual_arrival_date + csb.actual_arrival_time) AS actual_arrival_date_time,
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas1/Cas1PlacementReportRepository.kt
```diff
@@ -13,41 +13,59 @@ class Cas1PlacementReportRepository(
...
+        LEFT JOIN non_arrival_reasons nar ON nar.id = csb.non_arrival_reason_id
+        LEFT JOIN cancellation_reasons cr ON cr.id = csb.cancellation_reason_id
+        LEFT JOIN (
+            SELECT 
+                criteria.space_booking_id,
+                STRING_AGG(c.property_name, ', ' ORDER BY c.property_name) AS criteria
+            FROM cas1_space_bookings_criteria criteria
+            LEFT JOIN "characteristics" c ON c.id = criteria.characteristic_id
+            GROUP BY criteria.space_booking_id
+        ) criteria_grouped ON criteria_grouped.space_booking_id = csb.id
```
vbala-moj, line 52: Moved the subquery to the JOIN caluse, this will improve the perofurmance.

davidatkinsuk: Nice! I think a lateral join _may_ make this slightly quicker, something like the following. Can maybe look at using it if performance is an issue:

```
LEFT OUTER JOIN LATERAL (
  SELECT STRING_AGG(c.property_name, ', ' ORDER BY c.property_name) AS criteria
  FROM cas1_space_bookings_criteria criteria WHERE criteria.space_booking_id = csb.id
  LEFT JOIN "characteristics" c ON c.id = criteria.characteristic_id
) initial_assessment ON TRUE -- ON condition is mandatory with LEFT OUTER JOIN, but satisfied already in lateral join subquery
```

davidatkinsuk (approved):
LGTM 👍

PR #3087, opened 2025-03-03
CAS-1434 - cas2 prisoner locations

daveawc:
Currently, the prisoner's prison is retrieved from the prison API and stored as part of the application when it is submitted. This PR adds functionality to store the location in the cas_2_prisoner_location table, which is needed for the listenders to process POM allocation and prison transfer messages.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas2PrisonerLocationEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas2PrisonerLocationEntity.kt
@@ -6,7 +6,6 @@ import jakarta.persistence.JoinColumn
 import jakarta.persistence.ManyToOne
 import jakarta.persistence.Table
 import org.springframework.data.jpa.repository.JpaRepository
-import org.springframework.data.jpa.repository.Query
 import org.springframework.stereotype.Repository
 import java.time.OffsetDateTime
 import java.util.UUID
@@ -14,8 +13,7 @@ import java.util.UUID
 @Repository
 interface Cas2PrisonerLocationRepository : JpaRepository<Cas2PrisonerLocationEntity, UUID> {
 
-  @Query("SELECT p FROM Cas2PrisonerLocationEntity p WHERE p.application.id = :applicationId and p.endDate is NULL")
-  fun findPrisonerLocation(applicationId: UUID): Cas2PrisonerLocationEntity?
+  fun findAllByStaffIdOrderByOccurredAtDesc(staffId: UUID): List<Cas2PrisonerLocationEntity>
 }
 
 @Entity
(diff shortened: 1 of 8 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas2PrisonerLocationEntity.kt
```diff
@@ -5,17 +5,17 @@ import jakarta.persistence.Id
...
 import org.springframework.stereotype.Repository
 import java.time.OffsetDateTime
 import java.util.UUID
 
 @Repository
 interface Cas2PrisonerLocationRepository : JpaRepository<Cas2PrisonerLocationEntity, UUID> {
 
-  @Query("SELECT p FROM Cas2PrisonerLocationEntity p WHERE p.application.id = :applicationId and p.endDate is NULL")
-  fun findPrisonerLocation(applicationId: UUID): Cas2PrisonerLocationEntity?
+  @SQLOrder("occurredAt DESC")
```
davidatkinsuk, line 17: Does @SQLOrder on a spring data function work? I can't figure out if the integration tests verify this

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/Cas2PrisonerLocationService.kt
```diff
@@ -16,4 +23,17 @@ class Cas2PrisonerLocationService {
...
+  fun createPrisonerLocation(application: Cas2ApplicationEntity) {
+    val location = Cas2PrisonerLocationEntity(
+      id = UUID.randomUUID(),
+      application = application,
+      prisonCode = application.referringPrisonCode!!,
+      staffId = application.createdByUser.id,
+      occurredAt = application.createdAt,
+      endDate = null,
+    )
+    prisonerLocationRepository.save(location)
```
davidatkinsuk, line 37: i guess it's a deliberate choice to maybe have this information duplicated if there are multiple applications for the same CRN? (although i guess like in CAS1 these would run sequentially instead of in parallel). If we had the offender table would we remove the application link here?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/Cas2PrisonerLocationService.kt
```diff
@@ -16,4 +23,17 @@ class Cas2PrisonerLocationService {
...
     log.info("Handle location changed event at ${event.occurredAt}")
   }
+
+  @Transactional
+  fun createPrisonerLocation(application: Cas2ApplicationEntity) {
+    val location = Cas2PrisonerLocationEntity(
+      id = UUID.randomUUID(),
+      application = application,
+      prisonCode = application.referringPrisonCode!!,
+      staffId = application.createdByUser.id,
```
davidatkinsuk, line 33: Is staffId the best name for this given it's populated with an internal user identifier? Would this be better named as createdByStaffId?

davidatkinsuk (approved):
Mostly questions from me :D

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/Cas2PrisonerLocationService.kt
davidatkinsuk: No worries I just thought i'd flag it before it got integrated any more. I think `userId` may actually make the most sense if you want to align with the new table?

PR #3091, opened 2025-03-03
APS-2021: Cas1 offender entity.

vbala-moj:
[https://dsdmoj.atlassian.net/browse/APS-2021](url)

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas1/Cas1OffenderEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas1/Cas1OffenderEntity.kt
@@ -1,0 +1,26 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.cas1
+
+import jakarta.persistence.Entity
+import jakarta.persistence.Id
+import jakarta.persistence.Table
+import java.time.OffsetDateTime
+import java.util.UUID
+
+@Entity
+@Table(name = "cas1_offenders")
+data class Cas1OffenderEntity(
+  @Id
+  val id: UUID,
+  val crn: String,
+  val nomsNumber: String?,
+  /**
+   * The offender name. This should only be used for search purposes (i.e. SQL)
+   * If returning the offender name to the user, use the [OffenderService], which
+   * will consider any LAO restrictions
+   */
+  val name: String,
+  val tier: String?,
+  val createdAt: OffsetDateTime,
+  val lastUpdatedAt: OffsetDateTime,
+
+)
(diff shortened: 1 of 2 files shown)


src/main/resources/db/migration/all/20250303081728__add_cas1_offender_table.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE cas1_offenders (
+    id UUID NOT NULL ,
+    crn TEXT NOT NULL ,
+    noms_number TEXT,
+    tier TEXT,
```
davidatkinsuk, line 5: for the tier can we please use the custom collation specified in APS-1782. I've tried this locally, hopefully it works via flyway SQL:

```
CREATE COLLATION cas1_tier (provider = icu, locale = 'en-GB', 
rules = $$ 
& ' ' <*a-z <*A-Z < 9 < 8 < 7 < 6 < 5 < 4 < 3 < 2 < 1 < 0 
$$);

CREATE TABLE collation_test(
 value text NULL COLLATE "cas1_tier"
);
```

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas1/Cas1OffenderEntity.kt
```diff
@@ -0,0 +1,26 @@
...
+@Entity
+@Table(name = "cas1_offenders")
+data class Cas1OffenderEntity(
+  @Id
+  val id: UUID,
+  val crn: String,
+  val nomsNumber: String?,
+  /**
+   * The offender name. This should only be used for search purposes (i.e. SQL)
+   * If returning the offender name to the user, use the [Cas1OffenderService], which
```
davidatkinsuk, line 18: I think the reference in the comment should be to `[OffenderService]`

davidatkinsuk (approved):
just a few comments from me 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas1/Cas1OffenderEntity.kt
vbala-moj: Sorry, the ticket mentioned this.

davidatkinsuk: Ah I see. In this case the `OffenderService` should still be used to retrieve information about an offender to present on the UI (i.e., the comment on this field is correct). `Cas1OffenderService` will only be used manage this offender service table.

I do think the naming is a bit confusing though so maybe it needs a rethink!

PR #3092, opened 2025-03-04
CAS-1385 Remove premises api endpoint no longer used by CAS services

muhammad-elabdulla:
This[ PR CAS-1385](https://dsdmoj.atlassian.net/browse/CAS-1385) is to:

- Remove GET `/premises` api endpoint as is no longer used by CAS services
- Remove CAS1 related code from GET `/premises/summary` as is no longer used


gregkhawkins (approved):
Nice, good to get rid of dead code! 👍 
All looks good (as long as we are certain no longer used)

PR #3103, opened 2025-03-04
CAS-1434/create cas2 offenders table

daveawc:
created cas2 offenders table, updated cas 2 prisoner locations table

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas2/Cas2OffenderEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas2/Cas2OffenderEntity.kt
@@ -1,0 +1,33 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.cas2
+
+import jakarta.persistence.Entity
+import jakarta.persistence.Id
+import jakarta.persistence.OneToMany
+import jakarta.persistence.Table
+import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.stereotype.Repository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas2PrisonerLocationEntity
+import java.time.OffsetDateTime
+import java.util.UUID
+
+@Repository
+interface Cas2OffenderRepository : JpaRepository<Cas2OffenderEntity, UUID> {
+  fun findByNomsNumber(nomsNumber: String): Cas2OffenderEntity?
+}
+
+@Entity
+@Table(name = "cas_2_offenders")
+data class Cas2OffenderEntity(
+  @Id
+  val id: UUID,
+  val nomsNumber: String,
+  val crn: String,
+  val createdAt: OffsetDateTime,
+  val updatedAt: OffsetDateTime,
+
+  @OneToMany(mappedBy = "offender")
+  val locations: List<Cas2PrisonerLocationEntity> = emptyList(),
+) {
+  fun currentLocation() = locations.maxByOrNull { it.createdAt }
+  fun hasCorrectCurrentLocation(prisonCode: String, userId: UUID): Boolean = currentLocation() != null && currentLocation()?.allocatedPomUserId == userId && currentLocation()?.prisonCode == prisonCode
+}
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/Cas2ApplicationService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/Cas2ApplicationService.kt
@@ -21,6 +21,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas2Applicati
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas2ApplicationSummaryEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas2LockableApplicationRepository
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.NomisUserEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.cas2.Cas2OffenderEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.DomainEvent
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.PaginationMetadata
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.PersonInfoResult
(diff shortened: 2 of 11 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas2/Cas2OffenderEntity.kt
```diff
@@ -0,0 +1,33 @@
...
+  val nomsNumber: String,
+  val crn: String,
+  val createdAt: OffsetDateTime,
+  val updatedAt: OffsetDateTime,
+
+  @OneToMany(mappedBy = "offender")
+  val locations: List<Cas2PrisonerLocationEntity> = emptyList(),
+) {
+  fun currentLocation() = locations.maxByOrNull { it.createdAt }
+  fun hasCorrectCurrentLocation(prisonCode: String, userId: UUID): Boolean = currentLocation() != null && currentLocation()?.allocatedPomUserId == userId && currentLocation()?.prisonCode == prisonCode
```
davidatkinsuk, line 32: each call to currentLocation() will loop through the list. not a huge deal but i'm not sure how big the list will get

if did something like the following could get rid of the `?`s

```
currentLocation()?.let {
...
} :? false
```

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/Cas2ApplicationService.kt
```diff
@@ -267,7 +268,18 @@ class Cas2ApplicationService(
 
     createAssessment(application)
 
-    prisonerLocationService.createPrisonerLocation(application)
+    // This to be moved earlier in the process of creating an application, when the applications table is updated to use offenderId.
+    // added this here as the call to get the offender prison details happens immediately before in `retrievePrisonCode()`.
+    val prisonCode = application.referringPrisonCode!!
```
davidatkinsuk, line 273: (note - see my last comment in this class before considering addressing anything)

I think the code could be refactored to make what the comment is saying clearer? save referringPrisonCode in a `val` and use that to update the application? You can then get rid of the `!!` too?

I can't express why but i find it confusing when a function sets a `val` on some object, and then the value from that object is used further down the function instead of using the original `val`. I think because it makes it hard to trace the origin of some values.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/Cas2ApplicationService.kt
```diff
@@ -267,7 +268,18 @@ class Cas2ApplicationService(
...
+    val offender: Cas2OffenderEntity =
+      offenderService.findByNomsNumber(application.nomsNumber!!) ?: offenderService.createCas2Offender(
+        nomsNumber = application.nomsNumber!!,
+        crn = application.crn,
+      )
```
davidatkinsuk, line 278: (note - see my last comment in this class before considering addressing anything)

in Cas1 we're adding a `Cas1OffenderService` to deal with this upsert logic

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/Cas2ApplicationService.kt
```diff
@@ -267,7 +268,18 @@ class Cas2ApplicationService(
...
+
+    if (!offender.hasCorrectCurrentLocation(prisonCode, user.id)) {
```
davidatkinsuk, line 280: it's not clear to me why the user id matters wrt. current location? If a different user called this function the location would be 'wrong'?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/Cas2ApplicationService.kt
```diff
@@ -267,7 +268,18 @@ class Cas2ApplicationService(
...
+      prisonerLocationService.createPrisonerLocation(prisonCode, user.id, OffsetDateTime.now(), offender)
```
davidatkinsuk, line 281: I feel like `prisonerLocationService` should own all the logic related to locations currently sitting in this class i.e. `PrisonerLocationService` should be telling this class what to do wrt. locations, not the other way around.

I think the root cause of this is having the `retrievePrisonCode` function in this class instead of `PrisonerLocationService`. Could we not have a function on prisonerLocationService that returns the latest prisoner location and internally updates its records if out of date at the same time? whatever that returns can then be saved on the application? I may be missing something!

src/main/resources/db/migration/all/20250304191645__create_cas_2_offenders_table.sql
```diff
@@ -0,0 +1,67 @@
...
+
+--back fill it with offender information from previous submitted applications
+INSERT INTO cas_2_offenders (id, noms_number, crn, created_at, updated_at)
+    (SELECT gen_random_uuid(),
+            app.noms_number,
+            app.crn,
+            app.submitted_at,
+            app.submitted_at
+     FROM cas_2_applications app
+     WHERE submitted_at IS NOT NULL);
```
davidatkinsuk, line 24: can you have multiple `cas_2_applications` with the same id - will that result in duplicate cas_2_offender entries?

we have a similar challenge on Cas1. we're using a backfill job to get the latest information for each CRN. I guess you could use the newest application for each CRN?

src/main/resources/db/migration/all/20250304191645__create_cas_2_offenders_table.sql
```diff
@@ -0,0 +1,67 @@
+--create the new table
+CREATE TABLE cas_2_offenders
+(
+    id          uuid                     NOT NULL,
+    noms_number text                     NOT NULL,
+    crn         text                     NOT NULL,
+    created_at  TIMESTAMP WITH TIME ZONE NOT NULL,
+    updated_at  TIMESTAMP WITH TIME ZONE NOT NULL,
+    PRIMARY KEY (id),
+    UNIQUE (noms_number),
```
davidatkinsuk, line 10: i'm a bit wary of this constraint, i suspect it may be possible for a CRN to have multiple noms so may cause issues. I guess something to monitor

src/main/resources/db/migration/all/20250304191645__create_cas_2_offenders_table.sql
```diff
@@ -0,0 +1,67 @@
...
+
+INSERT INTO cas_2_prisoner_locations (id, offender_id, prison_code, allocated_pom_user_id, created_at)
+    (SELECT gen_random_uuid()
+          , (SELECT id
+             FROM cas_2_offenders
+             WHERE noms_number = app.noms_number)
+          , app.referring_prison_code
+          , app.created_by_user_id
+          , app.submitted_at
+     FROM cas_2_applications app
```
davidatkinsuk, line 62: again, not sure what would happen if have multiple apps with same crn, if that's possible

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/Cas2ApplicationService.kt
davidatkinsuk: I guess you could have two sequently entries where the location is the same but the pom allocation changes? Given this prisoner_allocations does make more sense.

PR #3108, opened 2025-03-06
APS-2021: Link AP application with CAS1 offender.

vbala-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApplicationEntity.kt
```diff
@@ -420,6 +423,9 @@ class ApprovedPremisesApplicationEntity(
   @Enumerated(value = EnumType.STRING)
   var noticeType: Cas1ApplicationTimelinessCategory?,
   var licenceExpiryDate: LocalDate?,
+  @OneToOne(fetch = FetchType.LAZY, cascade = [CascadeType.PERSIST, CascadeType.MERGE])
```
davidatkinsuk, line 426: we can have multiple applications for a single offender (CRN) because applications expire after a year. Therefore, this should be a `@ManyToOne`. I think PERSIST/MERGE cascades on many to one may be the default?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1OffenderService.kt
```diff
@@ -0,0 +1,38 @@
...
+      id = UUID.randomUUID(),
+      crn = cas1OffenderDetails.crn,
+      name = cas1OffenderDetails.name,
+      tier = cas1OffenderDetails.tier,
+      nomsNumber = cas1OffenderDetails.nomsNumber,
+      createdAt = OffsetDateTime.now(),
+      lastUpdatedAt = OffsetDateTime.now(),
+    )
+    return cas1OffenderRepository.saveAndFlush(offender)
+  }
```
vbala-moj, line 30: Currently updating this to not update if things are not changed.

src/main/resources/db/migration/all/20250304114432__add_offender_to_ap_application.sql
```diff
@@ -0,0 +1,7 @@
+ALTER TABLE cas1_offenders ADD COLUMN version BIGINT NOT NULL DEFAULT 1;
+CREATE INDEX idx_cas1_offenders_crn ON cas1_offenders(crn);
+CREATE INDEX idx_cas1_offenders_noms_number ON cas1_offenders(noms_number);
+
```
davidatkinsuk, line 4: we may as well add the index on for name whilst adding these. Could you please have a check and find out what the most effective index type is for partial string matching

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/ApplicationService.kt
```diff
@@ -237,6 +241,14 @@ class ApplicationService(
         deliusEventNumber,
         offenceId,
         riskRatings,
+        cas1OffenderEntity = cas1OffenderService.getOrCreateOffender(
+          Cas1OffenderDetails(
+            crn = offenderDetails.otherIds.crn,
+            name = "${offenderDetails.firstName.uppercase()} ${offenderDetails.surname.uppercase()}",
+            tier = riskRatings?.tier?.value?.level,
+            nomsNumber = offenderDetails.otherIds.nomsNumber,
```
davidatkinsuk, line 249: ultimately i think we want the Cas1OffenderService to deal with retreiving offender details itself and working out how to convert them into the values shown here. But - until we start caching results from ap-and-delius this will result in duplicate calls being made for this information.

In the meantime, I propose we do the following:

* Instead of passing in crn, name, tier and noms number, pass in the 'root' objects (`offenderDetails` and `riskRatings`). Let `Cas1OffenderService` convert these into required values
* `OffenderDetails` is deprecated, we should instead be using `CaseSummary`. There is a function `OffenderDetailsUtils.kt` to convert `OffenderDetails` into a `CasSummary` that can be used when passing the information to `Cas1OffenderSummary`

davidatkinsuk (approved):
Looks good, a few comments from me to consider

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1OffenderService.kt
davidatkinsuk: hibnerate doesn't issue SQL if nothing has changed, but i guess because we set updated at it'll will always detect a change. Is there an annotation we can use on the update date column to delegate that to hibernate? Then we shouldn't have to do any check.

vbala-moj: Will look in to this.

src/main/resources/db/migration/all/20250304114432__add_offender_to_ap_application.sql
vbala-moj: [https://www.postgresql.org/docs/current/pgtrgm.html](url)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApplicationEntity.kt
vbala-moj: @ManyToOne relationships do not have cascading operations enabled unless explicitly specified.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1OffenderService.kt
vbala-moj: @PreUpdate was added to deal with lastUpdatedAt only when there is an update.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApplicationEntity.kt
davidatkinsuk: Thanks for checking. On second review, i don't think we want cascading here. The `Cas1OffenderService` manages persistence of these entities?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1OffenderService.kt
davidatkinsuk: looks great!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApplicationEntity.kt
vbala-moj: Agree. No neded.

PR #3115, opened 2025-03-07
CAS-1434 - prison transfers 

daveawc:
Rebrand the cas2_prisoner_locations table to cas2_application_assignments, to track the POM and prison where the prisoner is allocated. This makes it more clear for the intent.

Also a general clean up of existing code.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas2ApplicationEntity.kt
```diff
@@ -83,6 +88,10 @@ data class Cas2ApplicationEntity(
   @OneToOne(mappedBy = "application")
   var assessment: Cas2AssessmentEntity? = null,
 
+  @OneToMany(mappedBy = "application", fetch = FetchType.LAZY, cascade = [CascadeType.ALL])
```
davidatkinsuk, line 91: adding a cascade has an interesting side effect of clarifying who 'owns' the relationship (e.g. if i'm deleted, also delete this thing). i like it 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/Cas2ApplicationService.kt
```diff
@@ -249,13 +251,23 @@ class Cas2ApplicationService(
...
+        val prisonCode = retrievePrisonCode(application)
         submittedAt = OffsetDateTime.now()
         document = serializedTranslatedDocument
-        referringPrisonCode = retrievePrisonCode(application)
+        referringPrisonCode = prisonCode
         preferredAreas = submitApplication.preferredAreas
         hdcEligibilityDate = submitApplication.hdcEligibilityDate
         conditionalReleaseDate = submitApplication.conditionalReleaseDate
         telephoneNumber = submitApplication.telephoneNumber
+        applicationAssignments.add(
```
davidatkinsuk, line 262: would it slightly reduce duplication if we called Cas2ApplicationAssignmentService.createApplicationAssignment instead?

davidatkinsuk (approved):
I think the renaming has (at least for me) made the purpose of this new table clearer 👍 

LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/Cas2ApplicationService.kt
```diff
@@ -81,6 +81,8 @@ class Cas2ApplicationService(
     return Pair(response.content, metadata)
   }
 
+  fun findMostRecentApplication(nomsNumber: String): Cas2ApplicationEntity? = applicationRepository.findFirstByNomsNumberAndSubmittedAtIsNotNullOrderBySubmittedAtDesc(nomsNumber)
```
besscerule, line 84: why do we have to put this in an extra layer and call from application service - can't we call from the repo directly?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/Cas2ApplicationService.kt
```diff
@@ -249,13 +251,23 @@ class Cas2ApplicationService(
 
     try {
       application.apply {
+        val prisonCode = retrievePrisonCode(application)
```
besscerule, line 254: is this the current prison and the referring prison code is where it came from?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas2/Cas2DomainEventListenerTest.kt
```diff
@@ -1,68 +1,45 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.unit.service.cas2
```
besscerule, line 1: do we need a test to check that event types that do not match the requires types are ignored?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas2ApplicationEntity.kt
davidatkinsuk: yes i'm not aware of any other cascades other than on deleting CAS1 legacy bookings. although arguably these records should be immutable once created so maybe we shouldn't be cascading after the initial merge. i'm not sure it matters to much tbh as long as it works.

PR #3118, opened 2025-03-09
APS-1908 Return earliest space booking records

danhumphreys-moj:
Ref: https://dsdmoj.atlassian.net/browse/APS-1908

GET /cas1/placement-requests

The SQL `PlacementRequestEntity.allForCas1Dashboard` assumes that there is only one active booking/space booking associated with a placement request. Update it to use the space booking with the earliest arrival date when populating bookingArrivalDate and bookingPremisesName.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PlacementRequestEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PlacementRequestEntity.kt
@@ -166,7 +166,13 @@ interface PlacementRequestRepository : JpaRepository<PlacementRequestEntity, UUI
       LEFT JOIN applications application ON application.id = pq.application_id
       LEFT JOIN bookings legacyBookings ON pq.booking_id = legacyBookings.id
       LEFT JOIN premises legacyBookingPremises ON legacyBookings.premises_id = legacyBookingPremises.id
-      LEFT JOIN cas1_space_bookings spaceBookings ON spaceBookings.placement_request_id = pq.id AND spaceBookings.cancellation_occurred_at IS NULL
+      LEFT OUTER JOIN LATERAL (
+        SELECT id, premises_id, canonical_arrival_date
+        FROM cas1_space_bookings b
+        WHERE b.placement_request_id = pq.id AND b.cancellation_occurred_at IS NULL
+        ORDER BY b.canonical_arrival_date ASC
+        LIMIT 1
+      ) spaceBookings ON TRUE
       LEFT JOIN premises spaceBookingPremises ON spaceBookings.premises_id = spaceBookingPremises.id
       LEFT JOIN cancellations legacyBookingCancellations ON legacyBookingCancellations.booking_id = legacyBookings.id
       WHERE
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/PlacementRequestsTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/PlacementRequestsTest.kt
@@ -1352,6 +1352,72 @@ class PlacementRequestsTest : IntegrationTestBase() {
     }
 
     @Test
+    fun `should return a placement request with the earliest space booking`() {
+      givenAUser(roles = listOf(UserRole.CAS1_WORKFLOW_MANAGER)) { user, jwt ->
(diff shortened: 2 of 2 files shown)


vbala-moj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PlacementRequestEntity.kt
```diff
@@ -134,7 +134,12 @@ interface PlacementRequestRepository : JpaRepository<PlacementRequestEntity, UUI
 
   @Query(
     """
-    WITH UNFILTERED AS (
+    WITH RANKED_SPACE_BOOKINGS AS (
+        SELECT *,
+            ROW_NUMBER() OVER (PARTITION BY placement_request_id ORDER BY canonical_arrival_date ASC) AS rn
+        FROM cas1_space_bookings),
+    EARLIEST_SPACE_BOOKINGS AS (SELECT * FROM RANKED_SPACE_BOOKINGS WHERE rn = 1),
```
davidatkinsuk, line 141: this is an interesting approach, i wasn't aware of `ROW_NUMBER()` and `PARTITION BY`). i'm not sure this is the most efficient approach as it requires multiple queries

I think using a lateral join would be more efficient and maybe more readable too?

You can do this by replacing the existing `LEFT JOIN EARLIEST_SPACE_BOOKNGS` with something like this (you'll need to some additional columns to the select for whatever you need)

```
LEFT OUTER JOIN LATERAL (
         SELECT  id
         FROM cas1_space_bookings b
         WHERE b.placement_request_id = pq.id AND b.cancellation_occurred_at IS NULL
         ORDER BY b.canonical_arrival ASC
         LIMIT 1
      ) spaceBookings ON TRUE -- ON condition is mandatory with LEFT OUTER JOIN, but satisfied already in lateral join subquery
```

You can find more examples in `Cas1PlacementMatchingOutcomesV2ReportRepository`

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/PlacementRequestsTest.kt
```diff
@@ -1351,6 +1351,72 @@ class PlacementRequestsTest : IntegrationTestBase() {
...
+            assertThat(summaries[0].id).isEqualTo(placementRequest.id)
+            assertThat(summaries[0].person.crn).isEqualTo(unmatchedOffender.otherIds.crn)
+            assertThat(summaries[0].placementRequestStatus).isEqualTo(Cas1PlacementRequestSummary.PlacementRequestStatus.matched)
+            assertThat(summaries[0].isParole).isEqualTo(placementRequest.isParole)
+            assertThat(summaries[0].requestedPlacementDuration).isEqualTo(placementRequest.duration)
+            assertThat(summaries[0].requestedPlacementArrivalDate).isEqualTo(placementRequest.expectedArrival)
+            assertThat(summaries[0].personTier).isEqualTo(placementRequest.application.riskRatings?.tier?.value?.level)
+            assertThat(summaries[0].applicationId).isEqualTo(placementRequest.application.id)
+            assertThat(summaries[0].applicationSubmittedDate).isEqualTo(placementRequest.application.submittedAt!!.toLocalDate())
+            assertThat(summaries[0].firstBookingPremisesName).isEqualTo("earlier_space_booking_premises")
```
davidatkinsuk, line 1413: nice use of free text fields to make what we're expecting clear 👍

davidatkinsuk (approved):
LGTM, i have a proposal on maybe changing how the SQL works

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PlacementRequestEntity.kt
danhumphreys-moj: I have edited this PR to follow this approach.  I agree it is more readable in that we are just modifying the existing LEFT JOIN rather than adding a new element to the CTE.

danhumphreys-moj: @davidatkinsuk I know you've approved already but can you give this SQL change the thumbs up please?

davidatkinsuk: looks great 👍

PR #3123, opened 2025-03-10
APS-2064 & APS-2065 Add bed on hold

danhumphreys-moj:
Ref: https://dsdmoj.atlassian.net/browse/APS-2064
Ref: https://dsdmoj.atlassian.net/browse/APS-2065

- Add a new entry to `cas1_out_of_service_bed_reasons` with the name ‘Bed on hold’
- Update the out of service bed report (`Cas1OutOfServiceBedsReportGenerator`) to exclude any entries where the reason for the bed being out of service is ‘Bed on hold’. To do this hardcode the ‘Bed on hold’ ID in the code.

'Bed on Hold' appears as a reason in the UI automatically when creating an out of service bed:

image

The 'Bed on hold' reason is also shown when viewing the list of out of service beds:

image

'Bed on hold' is also shown as the reason when viewing the out of service bed record itself:

image

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/reporting/generator/Cas1OutOfServiceBedsReportGenerator.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/reporting/generator/Cas1OutOfServiceBedsReportGenerator.kt
@@ -26,7 +26,7 @@ class Cas1OutOfServiceBedsReportGenerator(
 
     val outOfServiceBedIds = outOfServiceBedRepository.findByBedIdAndOverlappingDate(this.id, startOfMonth, endOfMonth, null)
 
-    val outOfServiceBeds = outOfServiceBedRepository.findAllById(outOfServiceBedIds.map(UUID::fromString))
+    val outOfServiceBeds = outOfServiceBedRepository.findAllById(outOfServiceBedIds.map(UUID::fromString)).filter { it.reason.id != Cas1OutOfServiceBedRepository.BED_ON_HOLD_CANCELLATION_REASON_ID }
 
     outOfServiceBeds.map {
       val bed = it.bed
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/reporting/generator/Cas1OutOfServiceBedsReportGenerator.kt
```diff
@@ -26,7 +26,8 @@ class Cas1OutOfServiceBedsReportGenerator(
 
     val outOfServiceBedIds = outOfServiceBedRepository.findByBedIdAndOverlappingDate(this.id, startOfMonth, endOfMonth, null)
 
-    val outOfServiceBeds = outOfServiceBedRepository.findAllById(outOfServiceBedIds.map(UUID::fromString))
+    val bedOnHoldReasonUUID = UUID.fromString("199bef2d-0839-40c0-85e2-00b84fde7fde")
```
davidatkinsuk, line 29: could this please be moved into a constant on `OutOfServiceBedEntity`. See `CancellationReasonRepository` for an example of where we do this elsewhere

davidatkinsuk (approved):
LGTM, just one minor proposal. Thanks for checking it appears on the UI 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/reporting/generator/Cas1OutOfServiceBedsReportGenerator.kt
danhumphreys-moj: Nice!  Have created a new constant `Cas1OutOfServiceBedRepository.BED_ON_HOLD_CANCELLATION_REASON_ID`.

PR #3124, opened 2025-03-10
CAS-1385 Fix issue return duplicated premises for pdus

muhammad-elabdulla:
This [PR CAS-1385](https://dsdmoj.atlassian.net/browse/CAS-1385) is to fix an issue with get CAS3 premises query which is returning duplicated premises in pdus


davidatkinsuk (approved):
LGTM 👍

PR #3127, opened 2025-03-10
CAS-564 Create CAS3 booking departure api endpoint

muhammad-elabdulla:
This [PR CAS-564](https://dsdmoj.atlassian.net/browse/CAS-564) is to create CAS3 booking departure api endpoint

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
@@ -593,7 +593,7 @@ class PremisesController(
           notes = body.notes,
         )
 
-        val departure = extractResultEntityOrThrow(result)
+        val departure = extractEntityFromCasResult(result)
 
         return ResponseEntity.ok(departureTransformer.transformJpaToApi(departure))
       }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas3/Cas3PremisesController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas3/Cas3PremisesController.kt
@@ -4,23 +4,34 @@ import org.springframework.http.ResponseEntity
 import org.springframework.stereotype.Service
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.cas3.PremisesCas3Delegate
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.BookingStatus
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Cas3Departure
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Cas3NewDeparture
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Cas3PremisesSummary
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.FutureBooking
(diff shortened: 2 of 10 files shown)


src/main/resources/static/cas3-api.yml
```diff
@@ -121,6 +121,60 @@ paths:
...
+          required: true
+          schema:
+            type: string
+            format: uuid
+      requestBody:
+        description: details of the departure
+        content:
+          'application/json':
+            schema:
+              $ref: '_shared.yml#/components/schemas/NewDeparture'
```
davidatkinsuk, line 150: Should we create a `Cas3NewDeparture` type to remove usage of shared types? We've already done this for Cas1

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
```diff
@@ -972,6 +972,18 @@ class PremisesController(
     )
   }
 
+  private fun <EntityType> extractResultEntityOrThrow(result: CasResult<EntityType>) = when (result) {
```
davidatkinsuk, line 975: I don't think this isused?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas3/Cas3PremisesController.kt
```diff
@@ -41,4 +85,26 @@ class Cas3PremisesController(
...
+    val premises = cas3PremisesService.getPremises(premisesId)
+      ?: throw NotFoundProblem(premisesId, "Premises")
+
+    return when (val result = cas3BookingService.getBookingForPremises(premises, bookingId)) {
+      is GetBookingForPremisesResult.Success -> result.booking
+      is GetBookingForPremisesResult.BookingNotFound -> throw NotFoundProblem(bookingId, "Booking")
+    }
+  }
+
+  private fun <EntityType> extractResultEntityOrThrow(result: CasResult<EntityType>) = when (result) {
```
davidatkinsuk, line 99: you can use `extractEntityFromCasResult` from `EntityUtils.ts` instead

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas3/Cas3BookingService.kt
```diff
@@ -312,27 +313,29 @@ class Cas3BookingService(
     moveOnCategoryId: UUID,
     notes: String?,
     user: UserEntity,
-  ) = validated<DepartureEntity> {
+  ): CasResult<DepartureEntity> {
```
davidatkinsuk, line 316: FYI you should be able to use `validatedCasResult` from `ValidationErrors.kt` as a replacement for the deprecated `validated` function. Current approach also works fine 👍

davidatkinsuk (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
muhammad-elabdulla: Is used by createDeparture as I change the type is returned 
I will remove it as I will replace extractResultEntityOrThrow by extractEntityFromCasResult

davidatkinsuk: Ah yes makes sense 👍

davidatkinsuk: Ah yes, makes sense 👍

PR #3143, opened 2025-03-12
Fix/cas 1441 update validation error

SaintRhys:
As part of this [ticket](https://dsdmoj.atlassian.net/jira/software/c/projects/CAS/boards/1489?selectedIssue=CAS-1441), rename validation error group from `bookings` to `status`, to allow frontend access.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas3/Cas3PremisesService.kt
```diff
@@ -232,7 +232,7 @@ class Cas3PremisesService(
       )
 
       if (futureBookings.any()) {
-        validationErrors["$.bookings"] = "existingBookings"
+        validationErrors["$.status"] = "existingBookings"
```
muhammad-elabdulla, line 235: Can be renamed to archiveProperty instead of status

PR #3152, opened 2025-03-13
Run CAS2 Bail e2e tests after deploying to dev

libuk:
[JIRA](https://dsdmoj.atlassian.net/browse/CBA-140?atlOrigin=eyJpIjoiMmM1MzMxYTVkYjI2NDg5YTg3NGZlNDVmMmFjNTE3NzUiLCJwIjoiaiJ9)

As we have a new CAS2 Bail repo making use of the API (namespace `CAS2V2`) we want to add the Bail e2e tests to the list of tests that are run before deploying to pre-prod to avoid regressions.

As no pre-prod or prod environments are currently in use for CAS2 Bail, this step is not required to pass before deploying to pre-prod, but will be the case in the future.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -310,6 +310,35 @@ jobs:
           event: fail
           channel: << pipeline.parameters.alerts-slack-channel >>
           template: basic_fail_1
+  cas2_bail_e2e_tests:
+    docker:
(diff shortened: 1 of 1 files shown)


davidatkinsuk (approved):
LGTM 👍

PR #3163, opened 2025-03-14
enabling cas2 message processing in test

daveawc:


src/main/resources/application.yml
```diff
@@ -208,6 +208,10 @@ services:
     read-timeout: 5000
 
 hmpps:
+  sqs.queus:
```
davidatkinsuk, line 213: typo on queues?

davidatkinsuk: also shouldn't this be coming from secrets so these entries are redundant? (Given they're also set in `application-local.yml` and `application-test.yml`)

davidatkinsuk (approved):
LGTM 👍

PR #3170, opened 2025-03-17
APS-2107: cas1 user role json.

vbala-moj:
[https://dsdmoj.atlassian.net/browse/APS-2107](url)

[built-cas1-roles.json](https://github.com/user-attachments/files/19311452/built-cas1-roles.json)

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,4 +1,5 @@
 import org.apache.commons.io.FileUtils
+import java.io.File
 
 plugins {
   id("uk.gov.justice.hmpps.gradle-spring-boot") version "7.1.4"
@@ -87,6 +88,10 @@ dependencies {
   implementation("uk.gov.service.notify:notifications-java-client:5.2.1-RELEASE")
 
   gatlingImplementation("org.springframework.boot:spring-boot-starter-webflux")
+}
+
+springBoot {
+  mainClass.set("uk.gov.justice.digital.hmpps.approvedpremisesapi.ApplicationKt")
 }
 
 java {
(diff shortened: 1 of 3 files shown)


build.gradle.kts
```diff
@@ -189,6 +194,20 @@ tasks.register<Test>("unitTest") {
     excludeTags("integration")
   }
 }
+tasks.register<JavaExec>("generateCas1Roles") {
+  group = "build"
+  description = "Generates the built-cas1-roles.json file using Cas1RolesFileGenerator."
+
+  mainClass.set("uk.gov.justice.digital.hmpps.approvedpremisesapi.util.Cas1UserRoleJsonFileGenerator")
+
```
vbala-moj, line 202: My first approach was to import the UserRole in Gradle, but I was not able to import `UserRole `from the application. Gradle only understand `uk.gov.justice.digital.hmpps` but not `approvedpremisesapi`.


```
import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserRole
import java.nio.file.Files
import java.nio.file.Paths

tasks.register("generateCas1RolesJson") {
  group = "build"

  doLast {
    val outputPath = Paths.get("src/main/resources/static/codegen/built-cas1-roles.json")

    Files.createDirectories(outputPath.parent)

    val userRoles = UserRole.values()
      .filter { it.serviceName == "approvedPremises" }
      .map { userRole ->
        mapOf(
          "name" to userRole.cas1ApiValue,
          "permissions" to userRole.permissions
        )
      }

    // Output JSON to file
    val objectMapper = ObjectMapper()
    val jsonContent = objectMapper.writerWithDefaultPrettyPrinter().writeValueAsString(userRoles)
    Files.write(outputPath, jsonContent.toByteArray())
  }
}
```

davidatkinsuk: I think i prefer it with the code in kotlin myself

PR #3174, opened 2025-03-18
APS-1916 Add temporary change request role and permission.

danhumphreys-moj:
Ref: https://dsdmoj.atlassian.net/browse/APS-1916

Add `cas1_appeal_create` permission
Add `cas1_planned_transfer_create` permission
Add `change_request_dev` user role with above permissions

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/UserPermission.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/UserPermission.kt
@@ -64,4 +64,7 @@ enum class UserPermission(val cas1ApiValue: ApprovedPremisesUserPermission?) {
   CAS1_REPORTS_VIEW(ApprovedPremisesUserPermission.reportsView),
   CAS1_REPORTS_VIEW_WITH_PII(ApprovedPremisesUserPermission.reportsViewWithPii),
   CAS1_REQUEST_FOR_PLACEMENT_WITHDRAW_OTHERS(ApprovedPremisesUserPermission.requestForPlacementWithdrawOthers),
+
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/UserPermission.kt
```diff
@@ -64,4 +64,7 @@ enum class UserPermission(val cas1ApiValue: ApprovedPremisesUserPermission?) {
   CAS1_REPORTS_VIEW(ApprovedPremisesUserPermission.reportsView),
   CAS1_REPORTS_VIEW_WITH_PII(ApprovedPremisesUserPermission.reportsViewWithPii),
   CAS1_REQUEST_FOR_PLACEMENT_WITHDRAW_OTHERS(ApprovedPremisesUserPermission.requestForPlacementWithdrawOthers),
+
+  CAS1_APPEAL_CREATE(ApprovedPremisesUserPermission.appealCreate),
+  CAS1_PLANNED_TRANSFER_CREATE(ApprovedPremisesUserPermission.appealCreate),
```
davidatkinsuk, line 69: I think the argument to this is wrong? Should be `ApprovedPremisesUserPermission.plannedTransferCreate`?

davidatkinsuk (approved):
Looks great, just one minor comment.

I think we'll need to update the types in the UI before we can assign this new role automatically to our test/dev/local users

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/UserPermission.kt
danhumphreys-moj: Yes of course, good spot.  Have changed this to `ApprovedPremisesUserPermission.plannedTransferCreate` now.

PR #3195, opened 2025-03-20
add exception capturing to sentry

daveawc:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/Cas2DomainEventListener.kt
```diff
@@ -15,12 +17,19 @@ class Cas2DomainEventListener(
...
-    val (message) = objectMapper.readValue<SQSMessage>(msg)
-    val event = objectMapper.readValue<HmppsDomainEvent>(message)
-    handleEvent(event)
+    try {
+      val (message) = objectMapper.readValue<SQSMessage>(msg)
+      val event = objectMapper.readValue<HmppsDomainEvent>(message)
+      handleEvent(event)
+    } catch (e: Exception) {
+      log.error("Exception caught in Cas2DomainEventListener", e)
+      Sentry.captureException(e)
```
davidatkinsuk, line 31: if you use SentryService to record this you wouldn't need to use a nasty `mockkStatic` in the test :D

davidatkinsuk (approved):
LGTM 👍

PR #3198, opened 2025-03-20
CAS-1500 Fix the gap calculation in the CAS3 bookings gap report

muhammad-elabdulla:
This [PR CAS-1500](https://dsdmoj.atlassian.net/browse/CAS-1500) is to fix some issues in the CAS3 bookings gap report which include:
- Fix the gap calculation in the report
- Include void in the gap calculation
- Include bedspaces that don't have booking during the report period

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas3/Cas3BookingGapReportRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas3/Cas3BookingGapReportRepository.kt
@@ -7,59 +7,134 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.reporting.util.ReportJdb
 import java.time.LocalDate
 
 val GAP_RANGES_QUERY = """
-with bed_bookings as (select
-    probation_regions.name as probation_region,
-    probation_delivery_units.name as pdu_name,
-    premises.name as premises_name,
-    rooms.name as room_name,
-    arrival_date,
-    departure_date,
-    (select working_day_count
-    from turnarounds
-    where turnarounds.id = (
-        select id
-        from turnarounds
-        where  turnarounds.booking_id = bookings.id
-        order by turnarounds.created_at desc
-        limit 1)) turnaround_days
-  from bookings
-  left join premises on premises.id = bookings.premises_id
-  left join probation_regions on probation_regions.id = premises.probation_region_id
-  left join temporary_accommodation_premises on temporary_accommodation_premises.premises_id = bookings.premises_id
-  left join probation_delivery_units on probation_delivery_units.id = temporary_accommodation_premises.probation_delivery_unit_id
-  left join beds on beds.id = bookings.bed_id
-  left join rooms on rooms.id = beds.room_id
-  left join confirmations on confirmations.booking_id = bookings.id
-  left join cancellations on  cancellations.booking_id = bookings.id
-  where bookings.service = 'temporary-accommodation' and cancellations.id is null and arrival_date <= :endDate and departure_date >= :startDate),
-bed_booked_days as (select
-      probation_region,
-      pdu_name,
-      premises_name,
-      room_name,
-      range_agg(daterange(arrival_date, departure_date,'[]')) as booked_days,
-      lower(range_agg(daterange(arrival_date, departure_date))) as date_min,
-      upper(range_agg(daterange(arrival_date, departure_date))) as date_max
-    from bed_bookings
-    group by probation_region, pdu_name, premises_name, room_name
-    order by probation_region, pdu_name, premises_name, room_name),
-booking_gaps as (select
-      probation_region,
-      pdu_name,
-      premises_name,
-      room_name,
-      unnest(multirange(daterange(least(date_min,:startDate), greatest(date_max,:endDate))) - booked_days) as gap
-  from bed_booked_days)
-select
-    booking_gaps.probation_region,
-    booking_gaps.pdu_name,
-    booking_gaps.premises_name,
-    booking_gaps.room_name as bed_name,
-    booking_gaps.gap::text,
-    upper(gap) - lower(gap) as gap_days,
-    turnaround_days
-from booking_gaps
-left join bed_bookings on departure_date = lower(gap) - 1 and  booking_gaps.premises_name =  bed_bookings.premises_name and booking_gaps.room_name = bed_bookings.room_name
+with bedspaces as (select
+                        beds.id,
+                        premises.name as premises_name,
+                        rooms.name as room_name,
+                        probation_regions.name as probation_region,
+                        probation_delivery_units.name as pdu_name,
+                        beds.end_date
+                    from beds
+                    inner join rooms on beds.room_id = rooms.id
+                    inner join premises on rooms.premises_id = premises.id
+                    inner join probation_regions on probation_regions.id = premises.probation_region_id
+                    inner join temporary_accommodation_premises on temporary_accommodation_premises.premises_id = premises.id
+                    inner join probation_delivery_units on probation_delivery_units.id = temporary_accommodation_premises.probation_delivery_unit_id
+                    where (beds.end_date is null or beds.end_date >= :startDate) and beds.created_at <= :endDate
+),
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas3/Cas3BookingGapReportRepository.kt
```diff
@@ -7,59 +7,132 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.reporting.util.ReportJdb
...
+                        probation_regions.name as probation_region,
+                        probation_delivery_units.name as pdu_name,
+                        beds.end_date
+                    from beds
+                    inner join rooms on beds.room_id = rooms.id
+                    inner join premises on rooms.premises_id = premises.id
+                    inner join probation_regions on probation_regions.id = premises.probation_region_id
+                    inner join temporary_accommodation_premises on temporary_accommodation_premises.premises_id = premises.id
+                    inner join probation_delivery_units on probation_delivery_units.id = temporary_accommodation_premises.probation_delivery_unit_id
+                    where service = 'temporary-accommodation' and (beds.end_date is null or beds.end_date >= :startDate) and beds.created_at <= :endDate
```
davidatkinsuk, line 23: given the inner join on temporary_accommodation_premises i don't think the check on `service is required`?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas3/Cas3BookingGapReportRepository.kt
```diff
@@ -7,59 +7,132 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.reporting.util.ReportJdb
...
+                      arrival_date,
+                      departure_date,
+                      (select working_day_count
+                       from turnarounds
+                       where turnarounds.id = (
+                           select id
+                           from turnarounds
+                           where  turnarounds.booking_id = bookings.id
+                           order by turnarounds.created_at desc
+                           limit 1)) turnaround_days
```
davidatkinsuk, line 36: i think this may be more efficient as a lateral join as currently it is two subqueries per row

something like

```
LEFT OUTER JOIN LATERAL (
  SELECT working_day_count 
  FROM turnarounds
  WHERE turnarounds.booking_id = bookings.id
  ORDER BY turnarounds.created_at DESC
  LIMIT 1
) turnaround ON TRUE -- ON condition is mandatory with LEFT OUTER JOIN, but satisfied already in lateral join subquery     
```

and then you can output `turnaround.working_data_count as turnaround_days`

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas3/Cas3BookingGapReportRepository.kt
```diff
@@ -7,59 +7,132 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.reporting.util.ReportJdb
...
+                  from bookings
+                  left join confirmations on confirmations.booking_id = bookings.id
```
davidatkinsuk, line 38: `confirmations` doesn't appear to be used?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas3/Cas3BookingGapReportRepository.kt
```diff
@@ -7,59 +7,132 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.reporting.util.ReportJdb
...
+                 inner join all_bookings on bedspaces.id = all_bookings.bed_id),
+ bedspace_voids as(select
+                     bed_id,
+                     premises_name,
+                     room_name,
+                     probation_region,
+                     pdu_name,
+                     voids.start_date,
+                     voids.end_date
+                     --bedspaces.end_date beds_end_date
```
davidatkinsuk, line 60: is this commented out code needed?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas3/Cas3BookingGapReportRepository.kt
```diff
@@ -7,59 +7,132 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.reporting.util.ReportJdb
...
+ bedspace_unavailable_dates as(
+    select
+        bed_id,
+        premises_name,
+        room_name,
+        probation_region,
+        pdu_name,
+        arrival_date as start_date,
+        departure_date as end_date,
+        turnaround_days
```
davidatkinsuk, line 74: could you use `bedspace_bookings.*` here for consistency with how you pull columns out of `bedspace_voids`? I guess you'll need to change the column names in that query?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas3/Cas3BookingGapReportRepository.kt
```diff
@@ -7,59 +7,132 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.reporting.util.ReportJdb
...
+               select
+                   probation_region,
+                   pdu_name,
+                   premises_name,
+                   room_name,
+                   daterange(:startDate,:endDate) as gap_range,
+                   daterange(:startDate,:endDate) as gap
+               from bedspaces
+               where bedspaces.id not in (select bed_id from bedspace_bookings)
+                 and bedspaces.id not in (select bed_id from bedspace_voids)),
```
davidatkinsuk, line 113: i could be wrong but i wonder if these would be more efficient as left joins and checking the join results in null? I guess something to consider if `explain analyse` shows this is innefficient

davidatkinsuk (approved):
The use of CTEs makes this easy to follow 👍  Although i didn't fully understand all of it, i can see existing tests are providing good regression.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas3/Cas3BookingGapReportRepository.kt
muhammad-elabdulla: ups

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas3/Cas3BookingGapReportRepository.kt
muhammad-elabdulla: YEs should be removed from the query

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas3/Cas3BookingGapReportRepository.kt
muhammad-elabdulla: I want to rename same of the columns this why I don't used * to make the query more clearer

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas3/Cas3BookingGapReportRepository.kt
muhammad-elabdulla: I compared both and LEFT OUTER JOIN LATERAL is take ing 7263 ms vs including it in the select statement 4690 ms

davidatkinsuk: interesting, good to know!

PR #3205, opened 2025-03-21
APS-1915 - Get change requests

davidatkinsuk:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas1/Cas1ChangeRequestEntity.kt
```diff
@@ -10,16 +11,63 @@ import jakarta.persistence.ManyToOne
...
+          booking.actual_arrival_date as actualArrivalDate
+        FROM cas1_change_requests cr
+        INNER JOIN cas1_space_bookings booking on cr.cas1_space_booking_id = booking.id
+        INNER JOIN placement_requests pr on cr.placement_request_id = pr.id
+        INNER JOIN approved_premises_applications apa on apa.id = pr.application_id
+        WHERE
+          cr.decision IS NULL AND
+          ((CAST(:cruManagementAreaId AS pg_catalog.uuid) IS NULL) OR apa.cas1_cru_management_area_id = :cruManagementAreaId)
+      )
+      select * from results
```
davidatkinsuk, line 51: i've used a CTE here because spring data was struggling to set the correct 'sort by' column when using the SQL directly

vbala-moj (approved):

PR #3207, opened 2025-03-21
Throw error after capturing in sentry

daveawc:
Updating to make sure the error is re-thrown so that the messages return to the DLQ


davidatkinsuk (approved):
LGTM 👍

PR #3208, opened 2025-03-21
APS-2091: Sort by arrival date

vbala-moj:
[https://dsdmoj.atlassian.net/browse/APS-2091 ](url)

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/TaskEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/TaskEntity.kt
@@ -49,7 +49,8 @@ interface TaskRepository : JpaRepository<Task, UUID> {
         apa.name as person,
         u.name as allocated_to,
         assessment.submitted_at as completed_at,
-        assessment.decision as decision
+        assessment.decision as decision,
+        apa.arrival_date AS sorting_date
       FROM
         assessments assessment
         INNER JOIN applications application ON assessment.application_id = application.id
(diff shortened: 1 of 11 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/TaskEntity.kt
```diff
@@ -94,13 +95,15 @@ interface TaskRepository : JpaRepository<Task, UUID> {
...
-        placement_application.decision as decision
+        placement_application.decision as decision,
+        pad.expected_arrival AS sorting_date
       from
         placement_applications placement_application
         INNER JOIN applications application ON placement_application.application_id = application.id
         INNER JOIN approved_premises_applications apa ON application.id = apa.id
         LEFT JOIN ap_areas area ON area.id = apa.ap_area_id
         LEFT JOIN users u ON u.id = placement_application.allocated_to_user_id
+        LEFT JOIN placement_application_dates pad on placement_application.id = pad.placement_application_id
```
davidatkinsuk, line 106: I think this may cause an issue if a placement application has multiple dates. Whilst these shouldn't appear in this anymore (we stopped creating them many months ago), we should probably tweak the SQL to protect against it.

I think the easiest thing to do is use a later join to make sure we only join on the earliest arrival date:

```
LEFT OUTER JOIN LATERAL (
  SELECT pad.expected_arrival
  FROM placement_application_dates pad
  WHERE placement_application.id = pad.placement_application_id
  ORDER BY created_at ASC
  LIMIT 1
) initial_assessment ON TRUE -- ON condition is mandatory with LEFT OUTER JOIN, but satisfied already in lateral join subquery
```

(i'd love to find some time to get rid of this placement_application_dates table as it will massively simplify alot of code!)

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/TasksTest.kt
```diff
@@ -1471,6 +1474,12 @@ class TasksTest {
                 decision = REJECTED,
               )
 
+              val placementDate1 = placementDateFactory.produceAndPersist {
+                withExpectedArrival(LocalDate.of(2024, 5, 10))
+                withPlacementApplication(placementApplication1)
+              }
```
davidatkinsuk, line 1480: could you please add an additional date here that's after this date to make sure it isn't used

davidatkinsuk (approved):
Looks good, just one comment from me 👍

PR #3214, opened 2025-03-24
CAS-1434 - update client config

daveawc:
Change the client to use WebClient, as more config is needed to use RestClient.


besscerule:
Looks good - do we need to change the way the Cas2DomainEventListenerTest file uses those clients - can we set them up in the way they are normally used? currently they use base url config

besscerule:
LGTM

besscerule (approved):

PR #3227, opened 2025-03-25
Replace domain event url variables in dev with correct UI values

adamhughes86:
- Removed the previously added API template urls due to new domain events being added
- Add the frontend urls to correctly populate the urls in the contacts created by domain events in delius


davidatkinsuk (approved):
pre-approved for if/when proposed change is made

PR #3232, opened 2025-03-25
add createApplicationAssignment function

daveawc:
Added a function to assist creating an application assigment, and refactored code to use it.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas2ApplicationEntity.kt
```diff
@@ -122,3 +122,15 @@ class Cas2LockableApplicationEntity(
   @Id
   val id: UUID,
 )
+
+fun Cas2ApplicationEntity.createApplicationAssignment(prisonCode: String, allocatedPomUserId: UUID?) {
```
davidatkinsuk, line 126: does this need to be an extension function? could it not just be a regular old function on the entity class given it's defined in the same `kt` file?

davidatkinsuk (approved):
LGTM, one minor proposal

PR #3239, opened 2025-03-26
fix: only return submitted events in the bail submitted report

tobybatchmoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas2v2/Cas2v2SubmittedApplicationReportRepository.kt
```diff
@@ -27,7 +27,8 @@ interface Cas2v2SubmittedApplicationReportRepository : JpaRepository<DomainEvent
         CAST(applications.bail_hearing_date as DATE) as bailHearingDate
       FROM domain_events events
       INNER JOIN cas_2_v2_applications applications ON events.application_id = applications.id      
-      WHERE applications.submitted_at IS NOT NULL
+      WHERE events.type = 'CAS2_APPLICATION_SUBMITTED'
+        AND applications.submitted_at IS NOT NULL
```
davidatkinsuk, line 31: I think submitted_at IS NOT NULL is redundant given we have a event, but there's no harm in it being there, so up to if you want it

davidatkinsuk (approved):
LGTM 👍

PR #3240, opened 2025-03-26
APS-1907: Reject change request.

vbala-moj:
[https://dsdmoj.atlassian.net/browse/APS-1907](url)

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1ChangeRequestService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1ChangeRequestService.kt
@@ -7,15 +7,20 @@ import org.springframework.stereotype.Service
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Cas1ChangeRequestSortField
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Cas1ChangeRequestType
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Cas1NewChangeRequest
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Cas1RejectChangeRequest
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas1SpaceBookingRepository
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.PlacementRequestRepository
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.cas1.Cas1ChangeRequestEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.cas1.Cas1ChangeRequestReasonRepository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.cas1.Cas1ChangeRequestRejectionReasonRepository
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.cas1.Cas1ChangeRequestRepository
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.cas1.Cas1ChangeRequestRepository.FindOpenChangeRequestResult
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.cas1.ChangeRequestDecision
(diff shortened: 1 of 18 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1ChangeRequestService.kt
```diff
@@ -82,4 +89,38 @@ class Cas1ChangeRequestService(
...
+    changeRequest: Cas1ChangeRequestEntity,
+    cas1RejectChangeRequest: Cas1RejectChangeRequest,
+  ): CasResult<Unit> = validatedCasResult {
+    if (changeRequest.placementRequest.id != placementRequestId) {
+      return CasResult.GeneralValidationError("The change request does not belong to the specified placement request")
+    }
+
+    lockableCas1ChangeRequestEntityRepository.acquirePessimisticLock(changeRequest.id)
+
+    val changeRequestWithLock = getChangeRequest(changeRequest.id)!!
```
davidatkinsuk, line 105: i like the variable name here 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1ChangeRequestService.kt
```diff
@@ -82,4 +89,38 @@ class Cas1ChangeRequestService(
       },
     ),
   )
+
+  @Transactional
+  fun rejectChangeRequest(
+    placementRequestId: UUID,
+    changeRequest: Cas1ChangeRequestEntity,
```
davidatkinsuk, line 96: could maybe pass in the Id here to ensure no one accidentally uses the wrong entity?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1ChangeRequestService.kt
```diff
@@ -82,4 +89,38 @@ class Cas1ChangeRequestService(
...
+      return CasResult.GeneralValidationError("The change request does not belong to the specified placement request")
+    }
+
+    lockableCas1ChangeRequestEntityRepository.acquirePessimisticLock(changeRequest.id)
+
+    val changeRequestWithLock = getChangeRequest(changeRequest.id)!!
+
+    if (changeRequestWithLock.decision != null) {
+      return if (changeRequestWithLock.decision == ChangeRequestDecision.REJECTED) {
+        Success(Unit)
```
davidatkinsuk, line 109: nice

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1ChangeRequestService.kt
```diff
@@ -82,4 +89,38 @@ class Cas1ChangeRequestService(
...
+      } else {
+        CasResult.GeneralValidationError("A decision has already been made for the change request")
+      }
+    }
+
+    val changeRequestRejectReason = cas1ChangeRequestRejectionReasonRepository.findByIdOrNull(cas1RejectChangeRequest.rejectionReasonId)
```
davidatkinsuk, line 115: I think this should maybe ignore archived entries and also only return entries with the corresponding change request type?

We already have this function `findByChangeRequestTypeAndArchivedIsFalse` so it'll be similar to that, but also taking an ID?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/Cas1ChangeRequestTest.kt
```diff
@@ -552,4 +557,150 @@ class Cas1ChangeRequestTest {
...
+    fun `reject change request without JWT returns 401`() {
+      webTestClient.patch()
+        .uri("/cas1/placement-request/${placementRequest1.id}/change-requests/${changeRequest.id}")
+        .exchange()
+        .expectStatus()
+        .isUnauthorized
+    }
+
+    @Test
+    fun `return 200 when successfully reject change request`() {
```
davidatkinsuk, line 661: could we please add a test where the user doesn't have the correct roles, for both a planned transfer and appeal?

it will be similar to `FindOpen.Getting change requests without a valid role returns 403`

PR #3243, opened 2025-03-26
CAS-1434 - enable message processing in prod

daveawc:
This is a PR that enables the processing of domain events in prod as part of CAS-1434.

Both events have been tested in preprod, with calls to other clients to get the necessary data.


muhammad-elabdulla (approved):
LGTM

PR #3246, opened 2025-03-27
APS-2158: Bug fix.

vbala-moj:


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas1/Cas1SpaceBookingServiceTest.kt
```diff
@@ -686,6 +686,28 @@ class Cas1SpaceBookingServiceTest {
...
+        premisesId = UUID.randomUUID(),
+        bookingId = UUID.randomUUID(),
+        arrivalDate = arrivalDate,
+        arrivalTime = arrivalTime,
+      )
+
+      assertThat(result).isInstanceOf(CasResult.ConflictError::class.java)
+      result as CasResult.ConflictError
+
+      assertThat(result.message).isEqualTo("The booking has already been canceled")
```
davidatkinsuk, line 708: we've added some handy assertions for casResult so you can do something like

```
      assertThat(result)
        .isConflictError()
        .hasMessageContaining("The booking has already been canceled")
```

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1SpaceBookingService.kt
```diff
@@ -179,6 +179,10 @@ class Cas1SpaceBookingService(
 
     existingCas1SpaceBooking!!
 
+    if (existingCas1SpaceBooking.isCancelled()) {
+      return existingCas1SpaceBooking.id hasConflictError "The booking has already been canceled"
```
davidatkinsuk, line 183: Typo on cancelled

davidatkinsuk (approved):
Thanks for the quick turnaround 👍

PR #3253, opened 2025-03-28
APS-2156 - Use day-specific assessment auto allocation config

davidatkinsuk:
https://dsdmoj.atlassian.net/browse/APS-2156

This commit changes the emergency and short notice automatic assessment allocation logic to use the day-specific allocation configuration.

This commit deprecates the old allocation configuration. This will be removed once this change has gone live without any issues


danhumphreys-moj (approved):
LGTM

PR #3255, opened 2025-03-28
Add transfer column and update booking permissions

vbala-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/UserRole.kt
```diff
@@ -94,6 +94,7 @@ enum class UserRole(val service: ServiceName, val cas1ApiValue: ApprovedPremises
       UserPermission.CAS1_OUT_OF_SERVICE_BED_CREATE_BED_ON_HOLD,
       UserPermission.CAS1_OUT_OF_SERVICE_BED_CANCEL,
       UserPermission.CAS1_VIEW_OUT_OF_SERVICE_BEDS,
+      UserPermission.CAS1_EMERGENCY_TRANSFER_PERFORM,
```
vbala-moj, line 97: Let me know if this is not a correct role.

davidatkinsuk: looks good to me 👍

davidatkinsuk (approved):
LGTM 👍

PR #3257, opened 2025-03-28
Prison transfers - Handle Missing Staff User Details

gregkhawkins:
**Context**
* JIRA: https://dsdmoj.atlassian.net/browse/CAS-1527
* Currently when we do not have the staff details of the pom manager related to the `offender-management.allocation.changed` event we throw an exception
* We want to make a call to the `nomis-user-roles-api` service under these circumstances and write back to the database.
 
**PR Includes**
* Extra integration to `nomis-user-roles-api` configured for `oauth2` authentication (current integration is user JWT driven)
* Get user details by staff ID by integrating with: `GET /users/staff/$staffId`
* Get full user details  by integrating with: `GET /users/$username`
* We make the above calls to write back to the `nomis_users` when we are missing the record from the db


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/client/NomisUserRolesForRequesterApiClient.kt
```diff
@@ -0,0 +1,20 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.client
+
+import com.fasterxml.jackson.databind.ObjectMapper
+import org.springframework.beans.factory.annotation.Qualifier
+import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.config.WebClientConfig
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.nomisuserroles.NomisUserDetail
+
+@Component
+class NomisUserRolesForRequesterApiClient(
```
davidatkinsuk, line 10: +1 for clarity in naming

davidatkinsuk (approved):
LGTM, with a very good renaming/refactoring of services to clarify purpose

PR #3262, opened 2025-03-31
APS-2034 modify booking cancellation endpoint

danhumphreys-moj:
Ref: https://dsdmoj.atlassian.net/browse/APS-2034

Following a discussion with @davidatkinsuk on 23/04/25 and based on some of the comments during review it was agreed to change some of the controller logic related to this PR.  The changes were essentially to add the approve appeal logic as a separate space booking appeal endpoint:

- Add new end point /cas1/premises/{premisesId}/space-bookings/{bookingId}/appeal
- Add new type Cas1ApprovedPlacementAppeal which is same as Cas1NewSpaceBookingCancellation but with a mandatory placementAppealChangeRequestId
- Create new controller function appeal() (which cancels space booking and approves change request), similar to cancelSpaceBooking()
- Make new controller function transactional.

---------------------

The original requirements were:

**Desired Outcome**

- Extend the current booking cancellation endpoint to have an optional parameter of change_request_id
- If the change_request_id is defined, the user must have the permission CAS1_CHANGE_REQUEST_ASSESS (for now add this to the user CAS1_CHANGE_REQUEST_DEV)
- Ensure that the booking has not been marked as either arrived or non arrived before cancelling (these checks are already carried out, non arrival was recently added)
- Where there is a change request id defined in the cancel booking flow, add logic to automatically update the change request to 'approved'
- Where there is a change request id defined included in the cancel booking flow, use cancellation reason acba3547-ab22-442d-acec-2652e49895f2 (Booking successfully appealed)
- Change the cancellation_reason parameter to be optional, as if a change_request_id is is supplied, a cancellation reason is not needed. Add a check to ensure it is provided and return a useful error when change_request_id is not provided, as this is no longer managed at the API level

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas1/Cas1SpaceBookingController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas1/Cas1SpaceBookingController.kt
@@ -1,9 +1,11 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.controller.cas1
 
+import jakarta.transaction.Transactional
 import org.springframework.http.HttpStatus
 import org.springframework.http.ResponseEntity
 import org.springframework.stereotype.Service
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.cas1.SpaceBookingsCas1Delegate
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Cas1ApprovedPlacementAppeal
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Cas1AssignKeyWorker
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Cas1NewArrival
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Cas1NewDeparture
@@ -20,6 +22,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Cas1TimelineEv
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Cas1UpdateSpaceBooking
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ServiceName
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.SortDirection
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.CancellationReasonRepository
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas1SpaceBookingEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserPermission
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserPermission.CAS1_EMERGENCY_TRANSFER_PERFORM
@@ -32,6 +35,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.CharacteristicSe
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.OffenderService
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.UserAccessService
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.UserService
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas1.Cas1ChangeRequestService
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas1.Cas1SpaceBookingService
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas1.Cas1SpaceBookingService.DepartureInfo
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas1.Cas1SpaceBookingService.SpaceBookingFilterCriteria
@@ -60,6 +64,7 @@ class Cas1SpaceBookingController(
   private val cas1WithdrawableService: Cas1WithdrawableService,
   private val characteristicService: CharacteristicService,
   private val cas1TimelineService: Cas1TimelineService,
+  private val cas1ChangeRequestService: Cas1ChangeRequestService,
 ) : SpaceBookingsCas1Delegate {
 
   override fun getSpaceBookingTimeline(premisesId: UUID, bookingId: UUID): ResponseEntity<List<Cas1TimelineEvent>> {
@@ -306,6 +311,30 @@ class Cas1SpaceBookingController(
             cancelledAt = body.occurredAt,
             body.reasonId,
             body.reasonNotes,
+          ),
+        ),
+      )
+  }
+
+  @Transactional
+  override fun appeal(premisesId: UUID, bookingId: UUID, body: Cas1ApprovedPlacementAppeal): ResponseEntity<Unit> {
+    userAccessService.ensureCurrentUserHasPermission(UserPermission.CAS1_PLACEMENT_APPEAL_ASSESS)
+    val spaceBooking = extractEntityFromCasResult(
+      cas1SpaceBookingService.getBooking(premisesId, bookingId),
+    )
+
+    cas1ChangeRequestService.approvePlacementAppeal(body.placementAppealChangeRequestId, userService.getUserForRequest(), spaceBooking)
+
+    return ResponseEntity
+      .ok()
+      .body(
+        extractEntityFromCasResult(
+          cas1WithdrawableService.withdrawSpaceBooking(
+            spaceBooking = spaceBooking,
+            user = userService.getUserForRequest(),
+            cancelledAt = body.occurredAt,
+            userProvidedReason = CancellationReasonRepository.CAS1_BOOKING_SUCCESSFULLY_APPEALED_ID,
+            otherReason = body.reasonNotes,
           ),
         ),
       )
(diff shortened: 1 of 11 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas1/Cas1SpaceBookingController.kt
```diff
@@ -289,7 +292,12 @@ class Cas1SpaceBookingController(
     bookingId: UUID,
     body: Cas1NewSpaceBookingCancellation,
   ): ResponseEntity<Unit> {
-    userAccessService.ensureCurrentUserHasPermission(UserPermission.CAS1_SPACE_BOOKING_WITHDRAW)
+    body.changeRequestId?.let {
+      userAccessService.ensureCurrentUserHasPermission(UserPermission.CAS1_APPEAL_ASSESS)
+      cas1ChangeRequestService.approveChangeRequest(it, userService.getUserForRequest(), extractEntityFromCasResult(cas1SpaceBookingService.getBooking(premisesId, bookingId)))
```
danhumphreys-moj, line 297: I believe the test coverage is sufficient to address the acceptance test requirements for this ticket, but am not currently happy with the way this new method is introduced into the controller.  This really needs refactoring - am open to suggestions.

I haven't looked too much into the `cas1WithdrawableService.withdrawSpaceBooking()` method called later in controller method `cancelSpaceBooking()`.  Perhaps any refactoring might impact this?

davidatkinsuk: Yes I agree - it doesn't feel quite right but i'm not sure putting it directly into `cas1WithdrawableService.withdrawSpaceBooking` is quite right as that's purely about triggering the withdrawal. Let's have a chat about it to work out the best approach.

I also think anything that cleanly separates the different decisions we make for a regular withdrawan versus an appeal would be good. I think the multiple `?:` entries makes it a little had to reason on

src/main/resources/static/cas1-schemas.yml
```diff
@@ -690,9 +690,11 @@ components:
           format: uuid
         reasonNotes:
           type: string
+        changeRequestId:
```
davidatkinsuk, line 693: thinking out loud - i wonder if `appealChangeRequestId` would be more descriptive here, clarifying its intent?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1ChangeRequestService.kt
```diff
@@ -90,6 +92,32 @@ class Cas1ChangeRequestService(
...
 
+  @Transactional
+  fun approveChangeRequest(changeRequestId: UUID, user: UserEntity, spaceBooking: Cas1SpaceBookingEntity): CasResult<Unit> = validatedCasResult {
+    if (spaceBooking.hasArrival()) {
+      return CasResult.GeneralValidationError("Space booking with ID ${spaceBooking.id} has been marked as arrived")
+    }
+
+    if (spaceBooking.hasNonArrival()) {
+      return CasResult.GeneralValidationError("Space booking with ID ${spaceBooking.id} has been marked as non-arrived")
+    }
```
davidatkinsuk, line 103: i guess we should also check if it's cancelled and error

davidatkinsuk (commented):
logically looks good, let's have a chat about some potential refactoring for clarity as you mention on your comment

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1ChangeRequestService.kt
```diff
@@ -90,6 +92,32 @@ class Cas1ChangeRequestService(
...
+
+    val changeRequest = getChangeRequest(changeRequestId)
+
+    if (changeRequest == null) {
+      return CasResult.GeneralValidationError("Change request with ID $changeRequestId not found")
```
davidatkinsuk, line 108: I think we can use a Not Found Error here?

danhumphreys-moj: Changed to return a `NotFound` error.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1ChangeRequestService.kt
danhumphreys-moj: Added `isCancelled()` check and unit test.

src/main/resources/static/cas1-schemas.yml
danhumphreys-moj: Renamed `changeRequestId` to `placementAppealChangeRequestId`.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas1/Cas1SpaceBookingController.kt
danhumphreys-moj: This work has now been refactored this morning and a new endpoint added (see commit and PR description).

src/main/resources/static/cas1-schemas.yml
danhumphreys-moj: This work has now been refactored this morning and a new endpoint added (see commit and PR description).

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas1/Cas1SpaceBookingController.kt
```diff
@@ -311,6 +316,30 @@ class Cas1SpaceBookingController(
       )
   }
 
+  @Transactional
+  override fun appeal(premisesId: UUID, bookingId: UUID, body: Cas1ApprovedPlacementAppeal): ResponseEntity<Unit> {
```
davidatkinsuk, line 320: nice!

davidatkinsuk (approved):
Looks great, just a few minor comments from me 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1ChangeRequestService.kt
```diff
@@ -99,6 +100,36 @@ class Cas1ChangeRequestService(
...
+
+    val changeRequest = findChangeRequest(changeRequestId)
+
+    if (changeRequest == null) {
+      return CasResult.NotFound("change request", changeRequestId.toString())
+    } else if (changeRequest.decision == ChangeRequestDecision.APPROVED) {
+      return CasResult.GeneralValidationError("Change request with ID $changeRequestId is already approved")
+    } else {
+      changeRequest.decision = ChangeRequestDecision.APPROVED
+      changeRequest.resolvedAt = OffsetDateTime.now()
```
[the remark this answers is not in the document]
danhumphreys-moj: No problem, changed to `changeRequest.resolve()`.

src/main/resources/static/cas1-schemas.yml
```diff
@@ -693,6 +693,23 @@ components:
...
+      type: object
+      properties:
+        occurredAt:
+          type: string
+          format: date
+        reasonId:
+          type: string
+          format: uuid
+        reasonNotes:
+          type: string
```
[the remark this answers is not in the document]
danhumphreys-moj: Yes, great point.  Removed `reasonId`.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/UserRole.kt
```diff
@@ -124,6 +124,7 @@ enum class UserRole(val service: ServiceName, val cas1ApiValue: ApprovedPremises
       UserPermission.CAS1_PLANNED_TRANSFER_CREATE,
       UserPermission.CAS1_PLACEMENT_APPEAL_ASSESS,
       UserPermission.CAS1_PLANNED_TRANSFER_ASSESS,
+      UserPermission.CAS1_SPACE_BOOKING_WITHDRAW,
```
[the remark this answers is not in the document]
danhumphreys-moj: Originally I didn't include this permission however...
`cas1WithdrawableService.withdrawSpaceBooking()` calls `cas1WithdrawableService.withdraw()`
which requires `rootNode.status.userMayDirectlyWithdraw` to be `true` and since it isn't `withdraw()` returns `CasResult.Unauthorised()`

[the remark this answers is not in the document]
danhumphreys-moj: `userMayDirectlyWithdraw` is set here in getWithdrawableState():
https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/813893b32eb6e56c8522788cd0d45cb83a7c6b46/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1SpaceBookingService.kt#L475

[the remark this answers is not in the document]
davidatkinsuk: That's a really good point. I think we should update that check to allow either `CAS1_SPACE_BOOKING_WITHDRAW` or `CAS1_PLANNED_TRANSFER_ASSESS`. I'll raise a PR to do that now

PR #3264, opened 2025-03-31
Cba 431 create a sar endpoint

garethCAS2:
The SubjectAccessRequestService has an existing function getSarResult. That calls three other functions:

getCAS1Result
getCAS2Result
getCAS3Result

Each one calls a different repository to load data about applications for the respective service.

We have added getCAS2v2Result. Following the existing pattern this new function calls a specific Cas2v2SubjectAccessRequestRepository. Here we search for applications, assesments, notes, status updates using cas_2_v2_applications and cas_2_v2_users.

We have extended the existing Cas2v2ApplicationEntityFactory and created two new factories - Cas2v2NoteEntityFactory, Cas2v2StatusUpdateDetailFactory.

We found a test in CAS2SubjectAccessRequestServiceTest which had no annotation so wasn't being run. We've annotated it.

We extended the Cas2v2IntegrationTestBase. SubjectAccessRequestServiceTestBase now inherits from Cas2v2IntegrationTestBase rather than IntegrationTestBase.

We added the test class Cas2v2SubjectAccessRequestServiceTest


davidatkinsuk (approved):
LGTM 👍

PR #3279, opened 2025-04-02
APS-2178 - Check application status when creating an appeal

davidatkinsuk:
This PR ensures that the application has a status of ‘rejected’ when creating an appeal. This fixes a bug we have observed where multiple users can create an appeal for the same application if they don’t navigate away from the application page.

This PR also removes deprecated code and replaces it with `CasResult`


gregkhawkins (commented):
LGTM

gregkhawkins (approved):

PR #3280, opened 2025-04-03
CBA-434-implement-new-status-details-content-in-assessor-view

tobybatchmoj:
In the Bail UI we inherit the status updates from CAS2. These are defined in code [here](https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/main/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/reference/Cas2ApplicationStatusSeeding.kt)

This list needs to be changed as set out in [this ticket](https://dsdmoj.atlassian.net/browse/CBA-434). Since some fields don't make sense for the different services, e.g. HDC licence and CPP details (CAS2) or  Previous unspent convictions (Bail), we need to differentiate the updates. The updates are referenced by UUID in other areas of the code, e.g. the [live applications SQL for the application repository](https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/main/src/main/resources/db/migration/all/20250107134233__application_controller_and_linked_services_in_cas2v2.sql#L161) which creates a worrying number of unknown unknowns.

To solve this issue we will add a service `enum` to the `Cas2PersistedApplicationStatus` in the seeder mentioned above.  This will default to CAS1, CAS2, CAS3 on all the existing updates, and then we can add our new statuses for Bail as CAS2V2 and exclude those from existing services while including the the ones we need from the existing services.

The status updates are pulled by the UI via `/reference/status-updates`. In the all the services that use that (those) end points we need to add code to include/exclude the Bail status updates from that returned list.

## Plan

 * Add a ServiceName enum to the Status update data class
 * Modify the find to filter on service name
 * Update the reference end points to pass service name
 * if no service is passed default to existing list
 * Update tests

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/reference/Cas2ApplicationStatusSeeding.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/reference/Cas2ApplicationStatusSeeding.kt
@@ -1,11 +1,11 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.model.reference
 
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ServiceName
 import java.util.UUID
 
 object Cas2ApplicationStatusSeeding {
 
-  @SuppressWarnings("LongMethod")
-  fun statusList(): List<Cas2PersistedApplicationStatus> = listOf(
+  val statuses = listOf(
     Cas2PersistedApplicationStatus(
       id = UUID.fromString("f5cd423b-08eb-4efb-96ff-5cc6bb073905"),
       name = "moreInfoRequested",
@@ -16,6 +16,49 @@ object Cas2ApplicationStatusSeeding {
           id = UUID.fromString("fabbb8c0-344e-4a9d-a964-7987b22d09c6"),
           name = "personalInformation",
           label = "Personal information",
+          applicableToServices = listOf(ServiceName.cas2),
+        ),
+        Cas2PersistedApplicationStatusDetail(
+          id = UUID.fromString("3df29b1b-e2fc-4df7-b4b8-0527cd9e3a6f"),
+          name = "applicantDetails",
+          label = "Applicant details",
+          applicableToServices = listOf(ServiceName.cas2v2),
+        ),
+        Cas2PersistedApplicationStatusDetail(
+          id = UUID.fromString("28926b7b-2056-478c-95fe-a611967b8fab"),
+          name = "concernsToOthers",
+          label = "Concerns to others",
+          applicableToServices = listOf(ServiceName.cas2v2),
+        ),
+        Cas2PersistedApplicationStatusDetail(
+          id = UUID.fromString("76be117e-c496-489c-a69d-ef26a05a6f64"),
+          name = "concernsToTheApplicant",
+          label = "Concerns to the applicant",
+          applicableToServices = listOf(ServiceName.cas2v2),
+        ),
+        Cas2PersistedApplicationStatusDetail(
+          id = UUID.fromString("03718bee-a0f1-4a5d-b095-e23c2b1211ab"),
+          name = "currentAllegedOffences",
+          label = "Current alleged offences",
+          applicableToServices = listOf(ServiceName.cas2v2),
+        ),
+        Cas2PersistedApplicationStatusDetail(
+          id = UUID.fromString("53f17f5b-c3bb-4bd2-9ccb-b5f0664aa104"),
+          name = "previousUnspentConvictions",
+          label = "Previous unspent convictions",
+          applicableToServices = listOf(ServiceName.cas2v2),
+        ),
+        Cas2PersistedApplicationStatusDetail(
+          id = UUID.fromString("4c8e39ea-0f8b-4131-a781-cac706f38914"),
+          name = "probationAndOASys",
+          label = "Probation & OASys",
+          applicableToServices = listOf(ServiceName.cas2v2),
+        ),
+        Cas2PersistedApplicationStatusDetail(
+          id = UUID.fromString("0e78cbc4-4139-4402-8a8a-763d9ca5d93b"),
+          name = "fundingAndID",
+          label = "Funding & ID",
+          applicableToServices = listOf(ServiceName.cas2v2),
         ),
         Cas2PersistedApplicationStatusDetail(
           id = UUID.fromString("05669c8a-d65c-48d2-a5e4-0c3f6fc8977b"),
@@ -31,26 +74,31 @@ object Cas2ApplicationStatusSeeding {
           id = UUID.fromString("7ba5cd7d-8ae3-4fe5-bb27-9367197ea160"),
           name = "riskToSelf",
           label = "Risk to self",
+          applicableToServices = listOf(ServiceName.cas2),
         ),
         Cas2PersistedApplicationStatusDetail(
           id = UUID.fromString("8641d719-7356-42d3-8363-e323bf76caec"),
           name = "riskOfSeriousHarm",
           label = "Risk of serious harm",
+          applicableToServices = listOf(ServiceName.cas2),
         ),
         Cas2PersistedApplicationStatusDetail(
           id = UUID.fromString("90f075ae-0b9f-445b-a9b5-1095abca87dc"),
           name = "currentOffences",
           label = "Current offences",
+          applicableToServices = listOf(ServiceName.cas2),
         ),
         Cas2PersistedApplicationStatusDetail(
           id = UUID.fromString("8ce77ea1-324e-4ac8-be8c-33d6d4d927f8"),
           name = "offendingHistory",
           label = "Offending history",
+          applicableToServices = listOf(ServiceName.cas2),
         ),
         Cas2PersistedApplicationStatusDetail(
           id = UUID.fromString("d2c44705-2795-4610-9879-dcf26940e121"),
           name = "hdcAndCpp",
           label = "HDC licence and CPP details",
+          applicableToServices = listOf(ServiceName.cas2),
         ),
         Cas2PersistedApplicationStatusDetail(
           id = UUID.fromString("94631a70-6c51-43d6-9112-2b6d042b5aa0"),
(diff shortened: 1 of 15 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/reference/Cas2ApplicationStatusSeeding.kt
```diff
@@ -56,6 +100,7 @@ object Cas2ApplicationStatusSeeding {
           id = UUID.fromString("94631a70-6c51-43d6-9112-2b6d042b5aa0"),
           name = "other",
           label = "Other",
+          serviceName = listOf(ServiceName.cas2, ServiceName.approvedPremises, ServiceName.temporaryAccommodation, ServiceName.cas2v2)
```
davidatkinsuk, line 103: can we please define this in terms of the application types (not services it supports). e.g. 'cas2 bail' and whatever a 'regular' cas2 application is called

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/reference/Cas2v2PersistedApplicationStatusFinder.kt
```diff
@@ -0,0 +1,16 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.model.reference
+
+import org.springframework.stereotype.Component
+import java.util.UUID
+
+@Component
+class Cas2v2PersistedApplicationStatusFinder(
+  private val statusList: List<Cas2PersistedApplicationStatus> = Cas2ApplicationStatusSeeding.statusList(),
```
davidatkinsuk, line 8: i appreciate this is work in progress but i guess you'll be passing a value into the call to `statusList()` to limit it to cas2 bail

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/reference/Cas2ApplicationStatusSeeding.kt
tobybatchmoj: I'm not sure we can. The problem is that the list is service dependent, not origin dependent. So in a CAS2 application we want one list:

 * a
 * b
 * c
 * g
 * h

And when we have a Bail application we want

 * a
 * b
 * d
 * e
 * f
 * g
 * h

The application origin is potentially present in both situation but it's nature of the requesting service that determines the questions.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/reference/Cas2ApplicationStatusSeeding.kt
```diff
@@ -248,4 +293,27 @@ object Cas2ApplicationStatusSeeding {
       description = "The accommodation is arranged for the agreed dates.",
     ),
   )
+
+  @SuppressWarnings("LongMethod")
+  fun statusList(
+    serviceNames: List<ServiceName> = listOf(
+      ServiceName.cas2,
+      ServiceName.approvedPremises,
+      ServiceName.temporaryAccommodation,
```
davidatkinsuk, line 302: Can we please remove `approvedPremises` and `temporaryAccommodation` here as these will never be supported

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/reference/Cas2PersistedApplicationStatusDetail.kt
```diff
@@ -1,10 +1,12 @@
...
 
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ServiceName
 import java.util.UUID
 
 data class Cas2PersistedApplicationStatusDetail(
   val id: UUID,
   val name: String,
   val label: String,
   val isActive: Boolean = true,
+  val serviceName: List<ServiceName> = listOf(ServiceName.cas2, ServiceName.approvedPremises, ServiceName.temporaryAccommodation),
```
davidatkinsuk, line 11: Can we please remove approvedPremises and temporaryAccommodation here as these will never be supported

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/reference/Cas2v2PersistedApplicationStatusFinder.kt
```diff
@@ -0,0 +1,17 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.model.reference
+
+import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ServiceName
+import java.util.UUID
+
+@Component
+class Cas2v2PersistedApplicationStatusFinder(
+  private val statusList: List<Cas2PersistedApplicationStatus> = Cas2ApplicationStatusSeeding.statusList(listOf(ServiceName.cas2v2)),
```
davidatkinsuk, line 9: nice. I think we need to update the `Cas2PersistedApplicationStatusFinder` to only return CAS2 status details? I think at the moment it will be getting Cas2V2 values?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2v2/Cas2v2ReferenceDataTest.kt
```diff
@@ -0,0 +1,34 @@
...
+    val jwt = jwtAuthHelper.createValidExternalAuthorisationCodeJwt()
+
+    webTestClient.get()
+      .uri("/cas2v2/reference-data/application-status")
+      .header("Authorization", "Bearer $jwt")
+      .exchange()
+      .expectStatus()
+      .isOk
+      .expectBody()
+      .json(expectedStatusOptions)
```
davidatkinsuk, line 32: I think we need to test that only CAS2 V2 status details are being returned (and do a similar test for CAS2 to ensure V2 ones aren't returned). This could be via an integration test or unit tests for `Cas2v2PersistedApplicationStatusFinder` and `Cas2PersistedApplicationStatusFinder`

Currently this test trusts the output from `Cas2v2PersistedApplicationStatusFinder.active()` is correct, but we don't verify that.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/reference/Cas2PersistedApplicationStatusDetail.kt
davidatkinsuk, line 11: I think renaming this to `applicableToServices` may make its purpose clearer?

davidatkinsuk (approved):
Looks good, just a few comments from me 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/reference/Cas2ApplicationStatusSeeding.kt
davidatkinsuk: makes sense. Thanks for the explanation 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/reference/Cas2ApplicationStatusSeeding.kt
tobybatchmoj: Done

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/reference/Cas2v2PersistedApplicationStatusFinder.kt
tobybatchmoj: I think that the default settings in the finder and details mean that that will happen automatically

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2v2/Cas2v2ReferenceDataTest.kt
tobybatchmoj: I've added a test to make sure the lists are different, I'll add one that checks the content (that's more than it does now)

tobybatchmoj: This test won't work as some statuses are shared, we could implicitly the lists match, but we don't that now. The UI e2e tests cover this.

davidatkinsuk (commented):
A few comments from me

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/reference/Cas2ApplicationStatusSeeding.kt
```diff
@@ -16,6 +16,49 @@ object Cas2ApplicationStatusSeeding {
           id = UUID.fromString("fabbb8c0-344e-4a9d-a964-7987b22d09c6"),
           name = "personalInformation",
           label = "Personal information",
+          applicableToServices = listOf(ServiceName.cas2),
+        ),
+        Cas2PersistedApplicationStatusDetail(
+          id = UUID.fromString("3df29b1b-e2fc-4df7-b4b8-0527cd9e3a6f"),
+          name = "applicantDetails",
+          label = "Applicant details",
+          applicableToServices = listOf(ServiceName.cas2v2),
```
Rosie-Brigham, line 25: Referring back to my slack message - I think specifically its the implicit nature`ServiceName.cas2` == HDC and `ServiceName.cas2v2` == Bail that is tripping me up here. I feel like `ServiceName.cas2HDC`,etc would be clearer for the next person to unpick this

davidatkinsuk (approved):
LGTM 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/reference/Cas2ApplicationStatusSeeding.kt
tobybatchmoj: @Rosie-Brigham Those values for service name come out of the yaml and are used all over the services to distinguish them. We could add another set but then that in itself would be confusing for anyone used to seeing approvedPremises, cas2 etc.

I'm going to merge this as it's a critical path blocker before code freeze on Thursday. If you want us to add a set of servcie aliases then let me know and I'll add a backlog ticket.

PR #3282, opened 2025-04-03
Firebreak remove unused Api endpoints from Applications controller

muhammad-elabdulla:
This PR is to remove the following Api endpoints from Applications controller as they are not used by any CAS

- GET /applications/{applicationId}/requests-for-placement/{requestForPlacementId}
- GET /applications/{applicationId}/withdrawables


davidatkinsuk (approved):
Thanks for doing this. LGTM 👍

PR #3283, opened 2025-04-03
Add Get All Applications endpoint to CAS1

vbala-moj:


src/main/resources/static/api.yml
```diff
@@ -1506,6 +1506,8 @@ paths:
       x-codegen-request-body-name: body
   /applications/all:
     get:
+      description: deprecated, use /cas1/applications/all
+      deprecated: true
```
davidatkinsuk, line 1510: is it worth tagging the `ApprovedPremisesApplicationSummary` as deprecated too? (i can't remember if you can deprecate types in open api)

src/main/resources/static/cas1-schemas.yml
```diff
@@ -1377,3 +1377,56 @@ components:
         - id
         - roomName
         - bedName
+    Cas1ApplicationSummary:
+      type: object
+      properties:
+        type:
```
davidatkinsuk, line 1383: i think this can be removed now as it was only provided as a discriminator?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/ApplicationsTransformer.kt
```diff
@@ -157,6 +158,32 @@ class ApplicationsTransformer(
     else -> throw RuntimeException("Unrecognised application type when transforming: ${domain::class.qualifiedName}")
   }
 
+  fun transformDomainToCas1ApplicationSummary(
```
davidatkinsuk, line 161: i think we need at least one unit test for this. can presumably copy the `transformJpaToApiSummary transforms an Approved Premises application correctly` unit test?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/Cas1ApplicationTest.kt
```diff
@@ -0,0 +1,836 @@
...
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.RiskWithStatus
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.bodyAsListOfObjects
+import java.time.LocalDate
+import java.time.OffsetDateTime
+import java.util.UUID
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ApprovedPremisesApplicationStatus as ApiApprovedPremisesApplicationStatus
+
+class Cas1ApplicationTest : IntegrationTestBase() {
+
+  @SpykBean
```
davidatkinsuk, line 35: i appreciate you've copied this from existing tests, but i don't think we should be using spys in integration tests unless absolutely necessary. I'm not sure they're being used in this test?

davidatkinsuk (approved):
just a few minor comments from me 👍

PR #3284, opened 2025-04-03
CAS-1530 adjusted application selection for assignment to remove appl…

besscerule:
…ications with forbidden current status updates

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/Cas2ApplicationService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/Cas2ApplicationService.kt
@@ -20,7 +20,6 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas2Applicati
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas2ApplicationRepository
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas2ApplicationSummaryEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas2LockableApplicationRepository
-import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas2StatusUpdateEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.NomisUserEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.DomainEvent
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.PaginationMetadata
@@ -54,8 +53,6 @@ class Cas2ApplicationService(
   @Value("\${url-templates.frontend.cas2.submitted-application-overview}") private val submittedApplicationUrlTemplate: String,
 ) {
 
-  val ignorableStatusLabels = setOf("Awaiting arrival", "Offer declined or withdrawn", "Referral cancelled", "Referral withdrawn")
-
   val repositoryUserFunctionMap = mapOf(
     null to applicationSummaryRepository::findByUserId,
     true to applicationSummaryRepository::findByUserIdAndSubmittedAtIsNotNull,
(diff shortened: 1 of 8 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas2/Cas2ApplicationServiceTest.kt
```diff
@@ -81,6 +82,95 @@ class Cas2ApplicationServiceTest {
...
   )
 
+  @Nested
+  inner class FindApplicationToAssign {
+    private val newestSchema = Cas2ApplicationJsonSchemaEntityFactory().produce()
+    private val user = NomisUserEntityFactory().produce()
+    private val nomsNumber = "ABC123"
+    private val application = Cas2ApplicationEntityFactory()
+      .withApplicationSchema(newestSchema)
+      .withId(UUID.randomUUID())
```
gregkhawkins, line 92: probably don't need the `.withId(UUID.randomUUID())` as the factory will do this for us when we do not do this explicitely

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas2/Cas2ApplicationServiceTest.kt
```diff
@@ -81,6 +82,95 @@ class Cas2ApplicationServiceTest {
...
+      .withCreatedByUser(user)
+      .withNomsNumber(nomsNumber)
+      .withSubmittedAt(OffsetDateTime.now())
```
gregkhawkins, line 95: same again

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/Cas2ApplicationService.kt
```diff
@@ -82,6 +85,12 @@ class Cas2ApplicationService(
 
   fun findMostRecentApplication(nomsNumber: String): Cas2ApplicationEntity? = applicationRepository.findFirstByNomsNumberAndSubmittedAtIsNotNullOrderBySubmittedAtDesc(nomsNumber)
 
+  fun findApplicationToAssign(nomsNumber: String): Cas2ApplicationEntity? = applicationRepository.findFirstByNomsNumberAndSubmittedAtIsNotNullOrderBySubmittedAtDesc(nomsNumber).takeIf {
+    it?.statusUpdates != null && it.statusUpdates!!.isNotEmpty() && isStatusAllowed(it.statusUpdates!!.first())
```
gregkhawkins, line 89: My understanding here is that if for example; `applicationRepository.findFirstByNomsNumberAndSubmittedAtIsNotNullOrderBySubmittedAtDesc()` gets us the first application meeting that criteria but then the `takeIf` condition fails we return null overall. This null will be passed back to the`Cas2AllocationChangedService` and we will take no action.
**Question**: did I misunderstand in the discussion early that under these circumstances we go to the next application that meets the criteria?
If I did not misunderstand I guess we probably need to rfactor to this:
```
1. retrieve all applications in submittedAt order
Cas2ApplicationRepository.findByNomsNumberAndSubmittedAtIsNotNullOrderBySubmittedAtDesc(nomsNumber: String): List<Cas2ApplicationEntity>

2. change the above to:
fun findApplicationToAssign(nomsNumber: String): Cas2ApplicationEntity? = applicationRepository.findByNomsNumberAndSubmittedAtIsNotNullOrderBySubmittedAtDesc(nomsNumber).firstOrNull {
    it.statusUpdates?.isNotEmpty() ?: false && isStatusAllowed(it.statusUpdates!!.first())
}
```
It's also possible I misunderstood business logic we discussed earlier!

besscerule: will call to discuss

besscerule: discussed in call and agreed above was not relevant as only need latest application that is submitted to check

PR #3288, opened 2025-04-04
Add cas1/applications

vbala-moj:


src/main/resources/static/cas1-api.yml
```diff
@@ -2101,6 +2101,27 @@ paths:
           $ref: '_shared.yml#/components/responses/403Response'
         500:
           $ref: '_shared.yml#/components/responses/500Response'
+  /applications:
+    get:
+      tags:
+        - Applications
+      summary: Lists all applications that the user has created
+      operationId: getApplications
```
davidatkinsuk, line 2109: maybe `getApplicationsForUser` would be a more descriptive name for this?

davidatkinsuk (approved):
Just one tiny comment from me. 👍

src/main/resources/static/cas1-api.yml
vbala-moj: Yeh thought the same. will do .ta

PR #3289, opened 2025-04-04
refactor to use Cas2StatusUpdateNonAssignable enum 

gregkhawkins:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/StatusUpdateEntity.kt
```diff
@@ -27,6 +27,13 @@ interface Cas2StatusUpdateRepository : JpaRepository<Cas2StatusUpdateEntity, UUI
   fun findAllStatusUpdatesWithoutAssessment(pageable: Pageable?): Slice<Cas2StatusUpdateEntity>
 }
 
+enum class Cas2StatusUpdateNonAssignable(val label: String) {
```
gregkhawkins, line 30: "non-assignable" felt more accurate but happy to revert back to "ignorable" if you guys had discussed and though this best

PR #3292, opened 2025-04-04
Add a debug profile

davidatkinsuk:
This is temporary additionl whilst we’re making changes to ap-tools. It allows us to simply enable debug logging whilst using profiles other than ‘local’


muhammad-elabdulla (approved):
LGTM

PR #3296, opened 2025-04-07
Unit test fix.

vbala-moj:
Fix for a failing unit test in  the main branch.


davidatkinsuk (approved):
Would you mind updating the commit message to remove the typo and explain why this change fixes it (because hardcoded date in the test was previously in the future and now in the past). It took me a minute to figure out why the change fixes it.

Other than that, LGTM 👍

PR #3299, opened 2025-04-08
Add /cas1/assessment

vbala-moj:


src/main/resources/static/cas1-schemas.yml
```diff
@@ -1487,6 +1487,137 @@ components:
...
+        data:
+          $ref: '_shared.yml#/components/schemas/Unit'
+        clarificationNotes:
+          type: array
+          items:
+            $ref: '_shared.yml#/components/schemas/ClarificationNote'
+        application:
+          $ref: 'cas1-schemas.yml#/components/schemas/Cas1Application'
+        allocatedToStaffMember:
+          $ref: '_shared.yml#/components/schemas/ApprovedPremisesUser'
```
vbala-moj, line 1523: Moving this to Cas1 requires duplication of the enumeration. Once all the end points (application and assessment) are moved, this can be easily moved over.

src/main/resources/static/cas1-schemas.yml
```diff
@@ -1487,6 +1487,137 @@ components:
...
+        status:
+          $ref: 'cas1-schemas.yml#/components/schemas/Cas1AssessmentStatus'
```
vbala-moj, line 1525: Moving this to Cas1 requires duplication of the enumeration. Once all the end points (application and assessment) are moved, this can be easily moved over.

src/main/resources/static/cas1-schemas.yml
```diff
@@ -1487,6 +1487,137 @@ components:
...
+              type: string
+              format: uuid
+            outdatedSchema:
+              type: boolean
+            data:
+              $ref: '_shared.yml#/components/schemas/Unit'
+            document:
+              $ref: '_shared.yml#/components/schemas/Unit'
+            status:
+              $ref: '_shared.yml#/components/schemas/ApprovedPremisesApplicationStatus'
```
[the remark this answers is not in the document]
davidatkinsuk: I think it may be best to move whatever enumerations we don't want in _shared.yml over now, because once the UI is using this new endpoint it'll be more painful to do? Unless this makes updating the code a massive headache of course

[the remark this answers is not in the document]
davidatkinsuk: Actually ignore my above comment, this is fine because it's CAS1 specific so we can just move it :)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas1/Cas1AssessmentsController.kt
```diff
@@ -51,6 +55,18 @@ class Cas1AssessmentsController(
       .body(transformedSummaries)
   }
 
+  override fun getAssessment(assessmentId: UUID): ResponseEntity<Cas1Assessment> {
+    val user = userService.getUserForRequest()
+
+    val assessment = extractEntityFromCasResult(assessmentService.getAssessmentAndValidate(user, assessmentId)) as ApprovedPremisesAssessmentEntity
+
+    val personInfo = offenderService.getPersonInfoResult(assessment.application.crn, user.cas1LaoStrategy())
```
davidatkinsuk, line 63: thanks for removing deprecate code here 👍

davidatkinsuk (approved):
LGTM 👍

PR #3300, opened 2025-04-08
Cba 434 fix load status update

tobybatchmoj:
The status updates where not being filtered when an application was retrieved, this fixes that.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/StatusUpdateDetailEntity.kt
```diff
@@ -36,7 +38,13 @@ data class Cas2StatusUpdateDetailEntity(
 ) {
   override fun toString() = "Cas2StatusDetailEntity: $id"
 
-  fun statusDetail(statusId: UUID, detailId: UUID): Cas2PersistedApplicationStatusDetail = Cas2PersistedApplicationStatusFinder().getById(statusId).statusDetails
-    ?.find { detail -> detail.id == detailId }
-    ?: error("Status detail with id $detailId not found")
+  fun statusDetail(statusId: UUID, detailId: UUID, serviceName: ServiceName = ServiceName.cas2): Cas2PersistedApplicationStatusDetail = when (serviceName) {
```
davidatkinsuk, line 41: given there is a cas2v2 version of this class, what Cas2v2 code is using this?  This PR doesn't include any updates to other classes to pass in `serviceName: ServiceName.cas2v2` so is this new logic required?

If this _is_ required, can we please remove the default on serviceName and set it explicitly in the calling code. I'm not sure cas2 is a suitable default.

tobybatchmoj: It's re-used by the transformer [here](https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/1084fa6720145bd5e6b3bcbca4a7b7e5344ec60a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas2v2/Cas2v2StatusUpdateTransformer.kt#L31).

We could duplicate all the transformer code to v2 but we were asked to re-suse where [possible by Billy.

davidatkinsuk: Actually looking at the code the type it's working on is `jpa: Cas2v2StatusUpdateDetailEntity` so i don't think it is calling this code?

tobybatchmoj: You may be right, I'll write some more detailed tests to cover that

PR #3303, opened 2025-04-08
Make truncation optional for dev profile

davidatkinsuk:
Before this commit any environment using the `dev` profile would truncate various tables truncated whenever the API starts.

This commit moves the truncation SQL into a new folder and old applies that folder for deployments to the ‘dev’ environment on the MOJ cloud platform. Any local deployments using the `dev` profile wll no longer truncate tables on startup.

This will break any deployments using the `dev` profile. To fix this, run the following:

```
delete from flyway_migrations where script = 'R__3_clear_applications.sql';
delete from flyway_migrations where script = 'R__6_clear_cas2_applications.sql';
delete from flyway_migrations where script = 'R__7_clear_cas2_domain_events.sql';
```


muhammad-elabdulla (approved):
LGTM

PR #3308, opened 2025-04-08
Firebreak move confirmations to Cas3 namespace

muhammad-elabdulla:
This PR is to move confirmations to Cas3 namespace and rename table and entity


davidatkinsuk (approved):
LGTM 👍

PR #3310, opened 2025-04-09
Firebreak/parse api yml script

besscerule:


script/create_apis_csv
```diff
@@ -0,0 +1,139 @@
...
+          then
+            lowerCaseResponseType=$( echo "${BASH_REMATCH[4]}" | tr '[:upper:]' '[:lower:]' )
+            lowerUrl=$( echo $url | tr '[:upper:]' '[:lower:]' )
+            removedDashUrl=$( echo "${lowerUrl//-}" )
+            if ! [[ $removedDashUrl =~ $lowerCaseResponseType ]]
+            then
+              action="${action} Ensure response type is found in url."
+            fi
+          fi
+          echo "${action},${strippedMethod},${baseUrl},${strippedUrl},${queries[*]},${requestBody},${response},${strippedStatusCode},${summary},${description},${responseDescription}" >> apis.csv
```
davidatkinsuk, line 133: i'd suggest we write the output to the build directory as that's a temporary directory. at the moment it's possible to accidentally check this generated file in.

To 'find' the directory add the following to the top of the script

```
cd "$(dirname "$0")"
script_dir="$(pwd)"
```

Then you can create a variable for the output file as

```
output_file="$script_dir"../build/apis/apis.csv
```

you may want to do a similar thing for the 'output.json' file too incase the script fails before the end (you could chuck that in /tmp/output.json?)

script/create_apis_csv
```diff
@@ -0,0 +1,139 @@
+#!/bin/sh
+
+echo "ACTIONS,METHOD,BASE-URL,URL,QUERIES,REQUEST-BODY,RESPONSE,STATUS-CODE,SUMMARY,DESCRIPTION,RESPONSE-DESCRIPTION" > apis.csv
+
+curl -s http://localhost:8080/v3/api-docs > output.json
+urls=$( jq -c '.paths | keys[]' output.json )
+
+for url in $urls
+do
```
davidatkinsuk, line 9: it might be worth echoing the urls as you loop through them, it gives positive feedback when waiting for the script to finish

```
  echo "Processing $url"
```

doc/how-to/create_apis_csv.md
```diff
@@ -0,0 +1,10 @@
+# How to create a csv file containing all currently available APIs
```
davidatkinsuk, line 1: I think 'api summary' may be a good name for this? e.g. `create_api_summary_csv`, with the output called `api_summary.csv`?

davidatkinsuk (approved):
A few comments from me 👍

script/create_apis_csv
besscerule: done

doc/how-to/create_apis_csv.md
besscerule: done

davidatkinsuk (approved):
LGTM 👍

PR #3313, opened 2025-04-09
Add cas1 accept/reject assessment

vbala-moj:
End points covered in this PR

1. `POST /cas1/assessments/{assessmentId}/acceptance`
2. `POST /cas1/assessments/{assessmentId}/rejection`

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AssessmentService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AssessmentService.kt
@@ -32,12 +32,13 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.TemporaryAcco
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.TemporaryAccommodationAssessmentJsonSchemaEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserPermission
-import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserQualification
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserRole
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.listeners.AssessmentClarificationNoteListener
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.listeners.AssessmentListener
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.PaginationMetadata
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.PersonSummaryInfoResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.ValidationErrors
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.deliuscontext.CaseSummary
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.InternalServerErrorProblem
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.results.AuthorisableActionResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.results.CasResult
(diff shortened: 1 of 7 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AssessmentService.kt
```diff
@@ -587,6 +580,139 @@ class AssessmentService(
     }
   }
 
+  @SuppressWarnings("TooGenericExceptionThrown")
+  fun acceptCas1Assessment(
```
davidatkinsuk, line 584: is it possible to move the new accept / reject functions into a Cas1AssessmentService?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AssessmentService.kt
```diff
@@ -505,17 +502,13 @@ class AssessmentService(
...
-        }
+      val caseSummary = getOffenderDetails(application.crn, rejectingUser.cas1LaoStrategy())
+        ?: throw RuntimeException("Offender details not found for CRN: ${application.crn} when creating Application Assessed Domain Event")
 
       cas1AssessmentDomainEventService.assessmentRejected(
         application = application,
         assessment = assessment,
-        offenderDetails = offenderDetails,
+        offenderDetails = caseSummary.asOffenderDetailSummary(),
         rejectingUser = rejectingUser,
```
davidatkinsuk, line 512: IMO we should leave the code in this function as-is as it's going to removed soon anyway? It slightly reduces risk of change

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AssessmentService.kt
```diff
@@ -425,18 +425,14 @@ class AssessmentService(
...
-      }
+    val caseSummary = getOffenderDetails(application.crn, acceptingUser.cas1LaoStrategy())
+      ?: throw RuntimeException("Offender details not found for CRN: ${application.crn} when creating Application Assessed Domain Event")
 
     if (application is ApprovedPremisesApplicationEntity) {
       cas1AssessmentDomainEventService.assessmentAccepted(
         application = application,
         assessment = assessment,
-        offenderDetails = offenderDetails,
+        offenderDetails = caseSummary.asOffenderDetailSummary(),
```
davidatkinsuk, line 435: IMO we should leave the code in this function as-is as it's going to removed soon anyway? It slightly reduces risk of change

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AssessmentService.kt
```diff
@@ -900,4 +1026,19 @@ class AssessmentService(
       CasResult.Success(assessment)
     }
   }
+
+  private fun getOffenderDetails(offenderCrn: String, laoStrategy: LaoStrategy): CaseSummary? {
+    val offenderDetails = offenderService.getPersonSummaryInfoResults(
+      setOf(offenderCrn),
+      laoStrategy,
+    ).firstOrNull()
```
davidatkinsuk, line 1034: you could use `offenderService.getPersonSummaryInfoResult` instead, which will slightly simplify this

davidatkinsuk (approved):
Just a few comments from me. It's nice to see the complexity of the APIs reducing as we're no longer including fields specific to CAS3

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AssessmentService.kt
davidatkinsuk: makes sense 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AssessmentService.kt
vbala-moj: I had'nt changed this initially but that will result in mocking both calls in the unit test, if you are ok with that then will revert these changes.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AssessmentService.kt
```diff
@@ -193,13 +195,9 @@ class AssessmentService(
...
-    val offenderResult = offenderService.getOffenderByCrn(
-      assessment.application.crn,
-      user.deliusUsername,
-      user.hasQualification(UserQualification.LAO),
-    )
+    val offenderDetails = getOffenderDetails(assessment.application.crn, user.cas1LaoStrategy())
 
-    if (offenderResult !is AuthorisableActionResult.Success) {
+    if (offenderDetails == null) {
       return CasResult.Unauthorised()
```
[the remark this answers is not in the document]
davidatkinsuk: yep i think that's fine

davidatkinsuk (approved):
LGTM 👍

PR #3317, opened 2025-04-09
PI-2972 Migrate Delius search endpoints to integration service

marcus-bcl:
This change removes the dependency on Probation Offender Search, and instead uses the existing [/probation-cases/summaries](https://ministryofjustice.github.io/hmpps-probation-integration-services/tech-docs/projects/approved-premises-and-delius/api-reference.html#probation-cases-summaries) endpoint on the Approved Premises and Delius integration API to find a person on probation's details by CRN or NOMIS id.


davidatkinsuk (approved):
Thanks for doing this. I think there is now some unused config for probation offender search in `application.yml` and `application-test.yml`. Could that please be removed?

PR #3319, opened 2025-04-09
Fix built-cas2-api-spec.yml - duplicate assignmentDate required field

gregkhawkins:


src/main/resources/static/codegen/built-cas2-api-spec.yml
```diff
@@ -5256,7 +5256,6 @@ components:
         - assignmentDate
         - allocatedPomName
         - allocatedPomUserId
-        - assignmentDate
```
gregkhawkins, line 5259: This is failing schema validation due to the `assignmentDate` being listed on line `5256` above also:

gregkhawkins: haha! Fair point

PR #3332, opened 2025-04-11
allow submit no noms number

garethCAS2:


tobybatchmoj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2v2/Cas2v2ApplicationService.kt
```diff
@@ -274,11 +274,21 @@ class Cas2v2ApplicationService(
     application.schemaVersion as? Cas2v2ApplicationJsonSchemaEntity
       ?: throw RuntimeException("Incorrect type of JSON schema referenced by CAS2 v2 Application")
 
+    var prisonCode:String? = null
+    if (application.nomsNumber != null) {
+      try {
+        prisonCode = retrievePrisonCode(application)
+      }
+     catch (e: Exception) {
+       print("Cas2v2ApplicationSubmission Error: ${e.message}")
```
davidatkinsuk, line 283: Just in case you're expecting this to alert...

if this is something we want an alert for, you'll need to do something like

```
sentryService.captureException(e)
```

davidatkinsuk (approved):
LGTM with one note to consider 👍

tobybatchmoj:
And I'll fix the linting...

PR #3342, opened 2025-04-14
Add the domain event url variables for test, preprod, and prod

adamhughes86:
This change will allow the domain events in CAS2V2 to correctly display the URLs instead of localhost

This will be tested manually in ndelius preprod to verify it's working and monitored closely in prod


davidatkinsuk (approved):
LGTM

PR #3353, opened 2025-04-16
Add default responses to Swagger config

daveawc:
This adds default responses for 401, 403, and 500 responses. This removes duplicated code from the open API spec files, and means there will be less noise in the generated controllers, which will make things easier when moving away from the openApi generator.

This only adds in responses where there is NOT a defined response for a given HttpStatus code, so any non-standard responses or additional responses will still be added (eg, 404, see 2nd pic).

image

image


davidatkinsuk (approved):
LGTM 👍

PR #3356, opened 2025-04-16
CAS-1471 Added extra column 'number of transfers' in cas2 mi reports

besscerule:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas2ApplicationStatusUpdatesReport.kt
```diff
@@ -9,25 +9,29 @@ import java.util.UUID
...
-      ORDER BY updatedAt DESC;
+                CAST(events.data -> 'eventDetails' ->> 'updatedAt' AS TIMESTAMP),
+                'YYYY-MM-DD"T"HH24:MI:SS'
+        ) AS updatedAt,
+        CASE
+            WHEN COUNT(distinct assignments.id) = 0 THEN 0
+            ELSE COUNT(distinct assignments.id) - 1
+        END as numberOfTransfers
+    FROM domain_events events
+             LEFT JOIN cas_2_application_assignments as assignments on events.application_id = assignments.application_id and assignments.allocated_pom_user_id is NOT NULL
```
davidatkinsuk, line 29: can we change the `assignments` alias to something that better reflects that we're only considering assignments where `allocated_pom_user_id` is not null? (comment applies to other report too). You'll know better than me but maybe `pom_assignments`?

davidatkinsuk (approved):
Just one minor comment from me 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas2ApplicationStatusUpdatesReport.kt
besscerule: ooo good

PR #3375, opened 2025-04-23
added prisonApplicationTransfer field to yml and checked if it was a …

besscerule:
…transfer by comparing the 2 prison codes


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas2/ApplicationsTransformer.kt
```diff
@@ -46,6 +46,7 @@ class ApplicationsTransformer(
       allocatedPomName = currentUser?.name,
       allocatedPomEmailAddress = currentUser?.email,
       currentPrisonName = jpa.currentPrisonCode?.let { offenderManagementUnitRepository.findByPrisonCode(it)?.prisonName ?: it },
+      prisonApplicationTransfer = jpa.currentPrisonCode != jpa.referringPrisonCode,
```
gregkhawkins, line 49: I know in java we would need `!jpa.currentPrisonCode.equals(jpa.referringPrisonCode)` but maybe fine in kotlin (can't remember!). Guess the unit test will flush this out anyway :)

gregkhawkins: ignore me, fine as is:
Screenshot 2025-04-23 at 10 56 38

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ApplicationTest.kt
```diff
@@ -1302,12 +1304,80 @@ class Cas2ApplicationTest : IntegrationTestBase() {
...
+            assertThat(responseBody.prisonApplicationTransfer).isFalse()
 
             Assertions.assertThat(responseBody.timelineEvents!!.map { event -> event.label })
               .isEqualTo(listOf("Application submitted"))
           }
         }
       }
+
+      @Test
+      fun `Get single submitted application returns 200 with timeline events and transferred`() {
```
gregkhawkins, line 1316: if this is all duplciate code of `Get single submitted application returns 200 with timeline events` but without the `referringPrisonCode` set then it might be worth doing everything in a private function and just passing the `applicationEntity` to that function as a param? The advantages being that; it will make the difference between the two tests more obvious, and it will reduce code duplication

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/transformer/cas2/ApplicationsTransformerTest.kt
```diff
@@ -126,6 +128,7 @@ class ApplicationsTransformerTest {
       assertThat(result.allocatedPomName).isNull()
       assertThat(result.assignmentDate).isNull()
       assertThat(result.currentPrisonName).isNull()
+      assertThat(result.prisonApplicationTransfer).isNull()
```
gregkhawkins, line 131: do we need this to be a `nullable` field? For this case could iut not just be `false`? Means we do not need to handle the nullable on the UI

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/transformer/cas2/ApplicationsTransformerTest.kt
```diff
@@ -188,6 +194,40 @@ class ApplicationsTransformerTest {
...
+      assertThat(result.telephoneNumber).isEqualTo(application.telephoneNumber)
+      assertThat(result.assessment!!.id).isEqualTo(assessment.id)
+
+      // these are assigned after an application is submitted
+      assertThat(result.submittedAt).isEqualTo(application.submittedAt!!.toInstant())
+      assertThat(result.allocatedPomEmailAddress).isNull()
+      assertThat(result.allocatedPomName).isNull()
+      assertThat(result.assignmentDate).isEqualTo(application.currentAssignmentDate)
+      assertThat(result.currentPrisonName).isEqualTo(newPrison.prisonName)
+      assertThat(result.prisonApplicationTransfer).isTrue()
```
gregkhawkins, line 230: nice!

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/transformer/cas2/ApplicationsTransformerTest.kt
besscerule: hmm i guess i just wanted it to copy the way currentPrisonName etc are null until submitted

gregkhawkins: interesting, before they "submitted" does this affect the status of whether they qualify as a `Prison Application Transfer`? You understand the flow better than me! I guess the an important thing to know here is what is the behaviour on the UI when this field is consumed as it will have an impact on this attempt at it:
`IF` prisonApplicationTransfer == true `THEN` show Prison transfer banner
`IF` prisonApplicationTransfer == false THEN` show general banner
`IF` prisonApplicationTransfer == null `THEN`  show general banner

besscerule: so prison transfers can ONLY happen after the app have been submitted

besscerule: i've made the change anyway

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ApplicationTest.kt
```diff
@@ -1255,33 +1256,55 @@ class Cas2ApplicationTest : IntegrationTestBase() {
...
-              withCreatedByUser(userEntity)
-              withSubmittedAt(OffsetDateTime.now().minusDays(1))
-            }
+            val responseBody = objectMapper.readValue(
+              rawResponseBody,
+              Cas2Application::class.java,
+            )
 
-            val omu = offenderManagementUnitEntityFactory.produceAndPersist()
+            Assertions.assertThat(responseBody.assessment!!.statusUpdates).isEqualTo(emptyList<Cas2StatusUpdate>())
```
gregkhawkins, line 1279: Minor: don't need `Assertions.` prefix

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ApplicationTest.kt
```diff
@@ -1255,33 +1256,55 @@ class Cas2ApplicationTest : IntegrationTestBase() {
...
+            assertThat(responseBody.allocatedPomEmailAddress).isEqualTo(userEntity.email)
+            assertThat(responseBody.allocatedPomName).isEqualTo(userEntity.name)
+            assertThat(responseBody.assignmentDate).isEqualTo(applicationEntity.currentAssignmentDate)
+            assertThat(responseBody.currentPrisonName).isEqualTo(prisonName)
+            assertThat(responseBody.prisonApplicationTransfer).isFalse()
 
-            applicationEntity.createApplicationAssignment(prisonCode = omu.prisonCode, allocatedPomUser = userEntity)
-            cas2ApplicationRepository.save(applicationEntity)
+            Assertions.assertThat(responseBody.timelineEvents!!.map { event -> event.label })
```
gregkhawkins, line 1286: Minor: don't need `Assertions.` prefix

PR #3376, opened 2025-04-23
Tidy up CAS1 email template config

davidatkinsuk:
This commit moves all CAS1 email template IDs into a CAS1-specific object, converting the values into constants


gregkhawkins (approved):
LGTM! That must have been fun ;)

PR #3377, opened 2025-04-23
Add change request fields to summary

vbala-moj:
[https://dsdmoj.atlassian.net/browse/APS-2059](url)

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas1SpaceBookingEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas1SpaceBookingEntity.kt
@@ -110,7 +110,31 @@ interface Cas1SpaceBookingRepository : JpaRepository<Cas1SpaceBookingEntity, UUI
           LEFT OUTER JOIN characteristics ON characteristics.id = sbc.characteristic_id
           WHERE sbc.space_booking_id = b.id 
           GROUP BY sbc.space_booking_id
-        ) AS characteristicsPropertyNames
+        ) AS characteristicsPropertyNames,
+      """
+
+    private const val CHANGE_REQUESTS_INFO_SUBQUERY =
+      """
+      CASE
+          WHEN EXISTS (
+              SELECT 1
+              FROM cas1_change_requests c1
+              WHERE c1.cas1_space_booking_id = b.id
+                AND c1.type = 'PLANNED_TRANSFER'
+                AND c1.resolved = false
+          ) THEN true
+          ELSE false
+      END AS plannedTransferRequested,
+      CASE
+          WHEN EXISTS (
+              SELECT 1
+              FROM cas1_change_requests c2
+              WHERE c2.cas1_space_booking_id = b.id
+                AND c2.type = 'PLACEMENT_APPEAL'
+                AND c2.resolved = false
+          ) THEN true
+          ELSE false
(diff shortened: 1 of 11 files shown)


src/main/resources/static/_shared.yml
```diff
@@ -4411,6 +4411,10 @@ components:
           type: string
         isCancelled:
           type: boolean
+        plannedTransferRequested:
+          type: boolean
+        appealRequested:
+          type: boolean
```
davidatkinsuk, line 4417: I think these should be mandatory? (i.e. typically false)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas1/Cas1SpaceBookingTransformer.kt
```diff
@@ -240,6 +240,8 @@ class Cas1SpaceBookingTransformer(
     } ?: listOf(),
     deliusEventNumber = searchResult.deliusEventNumber,
     isCancelled = searchResult.cancelled,
+    plannedTransferRequested = searchResult.plannedTransferRequested,
+    appealRequested = searchResult.appealRequested,
```
davidatkinsuk, line 244: I think there's another function in this transformer that'll need updating.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas1SpaceBookingEntity.kt
```diff
@@ -110,7 +110,31 @@ interface Cas1SpaceBookingRepository : JpaRepository<Cas1SpaceBookingEntity, UUI
...
+      END AS plannedTransferRequested,
+      CASE
+          WHEN EXISTS (
+              SELECT 1
+              FROM cas1_change_requests c2
+              WHERE c2.cas1_space_booking_id = b.id
+                AND c2.type = 'PLACEMENT_APPEAL'
+                AND c2.resolved = false
+          ) THEN true
+          ELSE false
```
davidatkinsuk, line 136: out of interest, can you drop the `CASE` and just do something like this? I've no idea if it'll work :D 

```
EXISTS (
              SELECT 1
              FROM cas1_change_requests c2
              WHERE c2.cas1_space_booking_id = b.id
                AND c2.type = 'PLACEMENT_APPEAL'
                AND c2.resolved = false
          ) AS appealRequested
```

davidatkinsuk (approved):
looks good, just a few comments from me 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas1/Cas1SpaceBookingTransformer.kt
vbala-moj: Looks like we need a separate query as we only have a space booking entity available.

davidatkinsuk: I guess it's easiest to just add something to the change request repository to get all open change requests for the space booking? Or we could add a lazy loaded one to many relationship on the space booking entity to change request entity? i guess there are multiple ways to do it, i'll let you decide.

davidatkinsuk (approved):
LGTM 👍

PR #3383, opened 2025-04-24
Remove /premises/{premisesId}/staff endpoint

davidatkinsuk:
This has been replaced by /cas1/premises… and is no longer in use in prod


danhumphreys-moj (approved):
LGTM

PR #3387, opened 2025-04-24
APS-2083 Update timeframes from 2 days to 5 days.

danhumphreys-moj:
Ref: https://dsdmoj.atlassian.net/browse/APS-2083

**Background**

The timeliness expectations for assessments are within 10 working days of allocation to an assessor.  This includes up to 5 working days for an information request to be made to Probation Practitioners where additional information is required to make a decision on suitability. Assessments for emergency applications (those submitted for arrivals within 7 days) will be assigned to the CRU Manager.  Where applications are submitted by 1pm, these should be assessed within 2 hours.  If submitted after 1pm, they should be completed no later than 11am on the following working day.   Assessments for short notice applications (those submitted for arrival between 8 and 28 days) will also be assigned to the CRU Manager.  These should be completed within 5 working days of allocation.  The 'countdown' on assessors dashboard works as expected for all types except the short notice applications which currently shows 2 working days instead of 5.  Please could this be corrected.

**Design**

The setting `ASSESSMENT_SHORT_NOTICE_TIMEFRAME_WORKDAYS` in `TaskDeadlineService.kt `should be increased from 2 to 5.

Furthermore, update `PLACEMENT_APPLICATION_TIMEFRAME_WORKDAYS` to be 5 days

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/TaskDeadlineService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/TaskDeadlineService.kt
@@ -83,11 +83,11 @@ class TaskDeadlineService(
 
   companion object {
     private const val ASSESSMENT_STANDARD_TIMEFRAME_WORKDAYS = 10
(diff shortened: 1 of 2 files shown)


davidatkinsuk (approved):
I think it's fair to say that this class has thorough test coverage :D LGTM 👍

PR #3392, opened 2025-04-25
APS-2231 - Add `CAS1_SPACE_BOOKING_ID` metadata name

davidatkinsuk:
This is to allow us to performance test moving space booking id from a dedicated column to metadata, enabling one-to-many relationships between domain events and space bookings, required for transfer domain events


muhammad-elabdulla (approved):
LGTM

PR #3395, opened 2025-04-25
APS-2227 - Add transfer info to space booking

davidatkinsuk:
We need to record if a space booking is an emergency or planned transfer to know if emails should be triggered on arrivals. This commit captures that information.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas1SpaceBookingEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas1SpaceBookingEntity.kt
@@ -456,9 +456,16 @@ data class Cas1SpaceBookingEntity(
    * purposes
    */
   val deliusId: String?,
-  @OneToOne
+  @OneToOne(fetch = FetchType.LAZY)
   @JoinColumn(name = "transferred_to", referencedColumnName = "id")
   var transferredTo: Cas1SpaceBookingEntity?,
+  @OneToOne(mappedBy = "transferredTo", fetch = FetchType.LAZY)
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas1SpaceBookingEntity.kt
```diff
@@ -456,9 +456,16 @@ data class Cas1SpaceBookingEntity(
    * purposes
    */
   val deliusId: String?,
-  @OneToOne
+  @OneToOne(fetch = FetchType.LAZY)
   @JoinColumn(name = "transferred_to", referencedColumnName = "id")
   var transferredTo: Cas1SpaceBookingEntity?,
+  @OneToOne(mappedBy = "transferredTo", fetch = FetchType.LAZY)
+  val transferredFrom: Cas1SpaceBookingEntity? = null,
+  /**
```
vbala-moj, line 464: It seems that this field is not populated when creating a new space booking for an emergency or planned transfer, isn't it?

vbala-moj (approved):
Just one comment

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas1SpaceBookingEntity.kt
davidatkinsuk: That is intention because the value in this column is `owned` by the space booking we're transforming from (in the `transferred_to` column). So when we next retrieve this space booking from the database the value of `transferredFrom` should be automatically populated. That's why I left it as a `val`.

PR #3398, opened 2025-04-25
APS-2057 - Use generic domain event envelope for CAS1 GET

davidatkinsuk:
We currently define an envelope type for each individual domain event, despite the envelope always containing the same 4 fields (as defined by the `Cas1DomainEventEnvelopeInterface`). This was previously required when defining domain event schema in YML to differentiate between the different types.

This commit introduces a generic envelope (`Cas1DomainEventEnvelope`) that can be used for all domain events, removing the need to add individual envelope types. It updates all GET requests to use this generic envelope, and deprecates all bespoke envelopes.

Subsequent commits will use the envelope when persisting domain events, which should unlock removing some additional duplication of logic when building the envelope. Because this will be quite a bit of effort, this change will be made over time.

Swagger ui generates types 'on the fly' for the new generic envelope e.g.

Screenshot 2025-04-29 at 11 55 47

I've tested domain events against delius in the test environment and all continue to work and are readable by probation-integration

app submitted DONE
app assessed DONE
app withdrawn DONE
booking made DONE
booking changed DONE
booking cancelled DONE
not arrived DONE
arrived DONE
departed DONE


danhumphreys-moj (approved):
Nice!!

PR #3402, opened 2025-04-25
Add planned transfer approval

vbala-moj:
[https://dsdmoj.atlassian.net/browse/APS-2131](url)

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1ChangeRequestService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1ChangeRequestService.kt
@@ -141,10 +141,7 @@ class Cas1ChangeRequestService(
     } else if (changeRequest.decision == ChangeRequestDecision.APPROVED) {
       return CasResult.GeneralValidationError("Change request with ID $changeRequestId is already approved")
     } else {
-      changeRequest.decision = ChangeRequestDecision.APPROVED
-      changeRequest.resolve()
-      changeRequest.decisionMadeByUser = user
-      cas1ChangeRequestRepository.save(changeRequest)
+      approveChangeRequest(changeRequest, user)
     }
 
     cas1ChangeRequestEmailService.placementAppealAccepted(changeRequest)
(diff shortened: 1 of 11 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1SpaceBookingService.kt
```diff
@@ -644,19 +647,98 @@ class Cas1SpaceBookingService(
...
+
+    if (plannedTransfer.arrivalDate >= plannedTransfer.departureDate) {
+      return GeneralValidationError("The provided departure date (${plannedTransfer.departureDate}) must be after the arrival date (${plannedTransfer.arrivalDate})")
+    }
+
+    if (changeRequest.spaceBooking.id != booking.id) {
+      return GeneralValidationError("The booking is not associated with the specified change request ${changeRequest.id}")
+    }
+
+    if (booking.hasDeparted()) {
```
davidatkinsuk, line 737: we should also check that that booking has an arrival (if not planned transfer shouldn't be possible, instead the user should raise an appeal). i.e. booking should be arrived, but not departed.

This then logically ensures the booking hasn't been cancelled because you can't cancel after arrival

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1ChangeRequestService.kt
```diff
@@ -219,4 +216,11 @@ class Cas1ChangeRequestService(
   }
 
   fun findChangeRequest(changeRequestId: UUID): Cas1ChangeRequestEntity? = cas1ChangeRequestRepository.findByIdOrNull(changeRequestId)
+
+  fun approveChangeRequest(changeRequestEntity: Cas1ChangeRequestEntity, user: UserEntity): Cas1ChangeRequestEntity {
```
davidatkinsuk, line 220: nice

davidatkinsuk (approved):
Looks great, just one comment from me which is an additional validation that wasn't mentioned on the ticket

PR #3404, opened 2025-04-28
Add auto resolve

vbala-moj:
[https://dsdmoj.atlassian.net/browse/APS-2232](url)

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1ChangeRequestService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1ChangeRequestService.kt
@@ -21,6 +21,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.cas1.ChangeRe
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.cas1.LockableCas1ChangeRequestRepository
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.validatedCasResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.results.CasResult
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.results.CasResult.GeneralValidationError
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.results.CasResult.Success
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.UserService
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.PageCriteria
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1SpaceBookingService.kt
```diff
@@ -528,7 +537,7 @@ class Cas1SpaceBookingService(
       else -> throw InternalServerErrorProblem("Withdrawal triggered automatically is not supported")
     }
 
-    cas1ChangeRequestService.spaceBookingWithdrawn(spaceBooking)
+    cas1ChangeRequestService.resolveSpaceBooking(spaceBooking)
```
vbala-moj, line 540: We could name it as `resolveChangeRequestForSpaceBooking`.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1ChangeRequestService.kt
```diff
@@ -121,25 +122,13 @@ class Cas1ChangeRequestService(
...
-
+  fun approvePlacementAppeal(changeRequestId: UUID, user: UserEntity): CasResult<Unit> = validatedCasResult {
     val changeRequest = findChangeRequest(changeRequestId)
 
     if (changeRequest == null) {
       return CasResult.NotFound("change request", changeRequestId.toString())
-    } else if (changeRequest.decision == ChangeRequestDecision.APPROVED) {
-      return CasResult.GeneralValidationError("Change request with ID $changeRequestId is already approved")
+    } else if (changeRequest.resolved) {
+      return GeneralValidationError("A decision has already been made for the change request")
```
davidatkinsuk, line 131: I think the message should maybe be 'This change request is already resolved?'

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1ChangeRequestService.kt
```diff
@@ -208,7 +197,7 @@ class Cas1ChangeRequestService(
     return Success(changeRequest)
   }
 
-  fun spaceBookingWithdrawn(spaceBooking: Cas1SpaceBookingEntity) = cas1ChangeRequestRepository.findAllBySpaceBookingAndResolvedIsFalse(spaceBooking).forEach { it.resolve() }
+  fun resolveChangeRequestForSpaceBooking(spaceBooking: Cas1SpaceBookingEntity) = cas1ChangeRequestRepository.findAllBySpaceBookingAndResolvedIsFalse(spaceBooking).forEach { it.resolve() }
```
davidatkinsuk, line 200: A note here for my purposes:

I originally envisaged the logic here being specific for each change request type, but because of the logic around when a change request can be created, this one-size-fits-all approach works. e.g. a placement extension can only be created after an arrival has been recorded, therefore, an arrival 'event' would never close an open placement extension. We need to keep this logic in mind in case the logic around when change requests can be created changes. E.g. if we can create an extension before arrival, and arrival would close that extension request, which wouldn't be correct

davidatkinsuk (approved):
Looks great! One minor comment from me regarding an error message.

PR #3408, opened 2025-04-29
Add Space Booking Actions

davidatkinsuk:
This commit adds the concept of actions, which are used to determine if a user can do something on a given entity (in this case space bookings). Actions have two purposes:

1. To provide the UI a list of available actions, allowing it to selectively render options
2. To check that a given action is allowed when it is attempted by a given user

This allows us to remove duplication of logic across the API and UI and change this logic in the API without needing UI changes.

Actions have been added for creating change requests


danhumphreys-moj (approved):
LGTM - good stuff

PR #3410, opened 2025-04-29
Update to User Access Service

daveawc:
Added an update to the user access service so that if an application has been transferred into a new prison, the new user can view it.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/givens/GivenAUser.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/givens/GivenAUser.kt
@@ -99,6 +99,7 @@ fun IntegrationTestBase.givenAUser(
 }
 
 fun IntegrationTestBase.givenACas2PomUser(
+  mockCallToGetMe: Boolean = true,
   id: UUID = UUID.randomUUID(),
   nomisUserDetailsConfigBlock: (NomisUserDetailFactory.() -> Unit)? = null,
   block: (nomisUserEntity: NomisUserEntity, jwt: String) -> Unit,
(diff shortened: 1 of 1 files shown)


gregkhawkins (approved):

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ApplicationTest.kt
```diff
@@ -1260,7 +1259,8 @@ class Cas2ApplicationTest : IntegrationTestBase() {
             val prisonName = "PRISON"
             val prisonCode = "PRI"
```
gregkhawkins, line 1260: can probably get rid of `prisonCode` (no longer used)

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas2/Cas2ApplicationTest.kt
```diff
@@ -1294,45 +1294,48 @@ class Cas2ApplicationTest : IntegrationTestBase() {
...
+              assertThat(responseBody.assignmentDate).isEqualTo(applicationEntity.currentAssignmentDate)
+              assertThat(responseBody.currentPrisonName).isEqualTo(omuNew.prisonName)
+              assertThat(responseBody.isTransferredApplication).isTrue()
+              assertThat(responseBody.omuEmailAddress).isEqualTo(omuNew.email)
 
-            assertThat(responseBody.timelineEvents!!.map { event -> event.label })
-              .isEqualTo(listOf("Prison transfer from $prisonCode to ${omuNew.prisonCode}", "Application submitted"))
+              assertThat(responseBody.timelineEvents!!.map { event -> event.label })
+                .isEqualTo(listOf("Prison transfer from ${userEntity.activeCaseloadId!!} to ${omuNew.prisonCode}", "Application submitted"))
+            }
```
gregkhawkins, line 1338: nice so we are testing that the second user (the one it is transferred to) can authenticate and get the CAS2Application successfully. Was the original test responing with a `403` status as no longer got past the new access stuff? 
Could be wrong! But If so, wondering whether we make this a new test and keep the original but change the original  to assert on the 403 response? The advantage being we have an integration this was that checks the new authorisation. Just an idea - perhaps overkill :)

gregkhawkins (approved):

PR #3411, opened 2025-04-29
APS-2057 - Add Planned Transfer Domain Events

davidatkinsuk:
This commit is quite large because it applies some new approaches to creating and retreiving domain events that should minimise the code required going forward. Notably:

* Uses the generic `Cas1DomainEventEnvelope` instead of type-specific envelopes. Some new common code was required to support this
* Uses standalone domain event describers for the timeline. Making it easier to unit test describers. Some new common code was required to support this
* Uses a common `Booking` type for timeline content. Some new common code was required to support this


vbala-moj (approved):
LGTM

PR #3412, opened 2025-04-30
APS-2057 - Simplify Change Request Domain Events

davidatkinsuk:
Update all domain events relates to change requests (And emergency transfer) to remove bespoke envelopes and use the new pattern for providing timeline payload content. This PR also simplifies the domain events themselves, reusing the common booking model and making property names more consistent.


danhumphreys-moj (approved):
Huddled with @davidatkinsuk this morning to discuss this and previous domain event tech debt work.  This PR uses the new approach for creating domain events for recent change request work however more work to be done to use new method with older existing domain events (which have been marked as deprecated).

PR #3418, opened 2025-04-30
Allow duplicate entries for domain event metadata

davidatkinsuk:
We have a requirement to be able to link a single domain event to multiple space bookings (e.g. ‘placement transferred’).

This commit modifies the domain event to metadata relationship to allow multiple entries to be stored against a given metadata name.

Subsequent commits will then make use of this to store multiple space bookings against a single entity, and add required indexes to ensure retrieving such domain events is efficient


vbala-moj (approved):

davidatkinsuk:
pausing this for now as a change in approach to domain events may make this irrelevant

PR #3419, opened 2025-05-01
Add open change requests to api

vbala-moj:
[https://dsdmoj.atlassian.net/browse/APS-2243](url)

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/PlacementRequestDetailTransformer.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/PlacementRequestDetailTransformer.kt
@@ -6,7 +6,9 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.PlacementReque
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.PlacementRequestDetail
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.CancellationEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.PlacementRequestEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.cas1.Cas1ChangeRequestEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.PersonInfoResult
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.cas1.Cas1ChangeRequestTransformer
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.cas1.Cas1SpaceBookingTransformer
 
 @Component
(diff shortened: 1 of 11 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/PlacementRequestDetailTransformer.kt
```diff
@@ -63,11 +66,25 @@ class PlacementRequestDetailTransformer(
...
+  ): Cas1PlacementRequestDetail {
     val placementRequest = placementRequestTransformer.transformJpaToApi(jpa, personInfo)
     val personSummaryInfo = personTransformer.personInfoResultToPersonSummaryInfoResult(personInfo)
     val placementRequestBookingSummary = placementRequestBookingSummaryTransformer.getBookingSummary(jpa)
 
+    val openChangeRequests = if (changeRequests.isNotEmpty()) {
+      cas1ChangeRequestTransformer.transformToChangeRequestSummaries(
+        changeRequests,
+        personInfo,
+        changeRequests.first().spaceBooking,
```
davidatkinsuk, line 82: I don't think this is quite right as we can now have multiple space bookings against a single placement request. I propose we remove the `booking` argument on `transformToChangeRequestSummaries` and in the function use the booking related to each individual change request when looping through them in the `map` call (will need to rejig that function a bit)

I think you can then get rid of the `isNotEmpty()` check and always just call this function even if the change requests are empty

davidatkinsuk (commented):
One comment from me that i think needs addressing before approval 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/PlacementRequestDetailTransformer.kt
vbala-moj: Which one you prefer?  The first one affect previous changes but not much. 😉
```
  fun transformToChangeRequestSummaries(
    cas1ChangeRequests: List<Cas1ChangeRequestEntity>,
    personInfoResult: PersonInfoResult
  ): List<Cas1ChangeRequestSummary> {
    val personSummary = personTransformer.transformPersonInfoResultToPersonSummary(personInfoResult)

    return cas1ChangeRequests.map { entity ->
      val booking = entity.spaceBooking

      Cas1ChangeRequestSummary(
        id = entity.id,
        person = personSummary,
        type = Cas1ChangeRequestType.valueOf(entity.type.name),
        createdAt = entity.createdAt.toInstant(),
        lengthOfStayDays = booking.canonicalArrivalDate
          .getDaysUntilInclusive(booking.canonicalDepartureDate).size,
        tier = booking.application?.riskRatings?.tier?.value?.level,
        expectedArrivalDate = booking.expectedArrivalDate,
        actualArrivalDate = booking.actualArrivalDate,
      )
    }
  }
```
OR

```
  private fun getOpenChangeRequests(
    changeRequests: List<Cas1ChangeRequestEntity>,
    personInfo: PersonInfoResult
  ): List<Cas1ChangeRequestSummary> {
    return changeRequests.groupBy { it.spaceBooking }.flatMap { (spaceBooking, changeRequestsForBooking) ->
      cas1ChangeRequestTransformer.transformToChangeRequestSummaries(
        changeRequestsForBooking,
        personInfo,
        spaceBooking
      )
    }
  }
```

davidatkinsuk: I think the former as it's more reusable i think and easier to read? (sorry, only just saw this)

PR #3437, opened 2025-05-06
Add separate emergency & planned transfer actions

davidatkinsuk:
The logic for these are current identical, this will diverge in subsequent commits/PRs


danhumphreys-moj (approved):
LGTM

PR #3442, opened 2025-05-07
APS-2241 Don’t allow multiple change request same type

davidatkinsuk:
This commit updates the Space Booking Actions to only allow a change request of a given type to be created if there isn’t already an open change request of that type for the same placement


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1SpaceBookingActionsService.kt
```diff
@@ -37,14 +46,25 @@ class Cas1SpaceBookingActionsService(
       unavailable("Space booking has been marked as non arrived")
     } else if (spaceBooking.isCancelled()) {
       unavailable("Space booking has been cancelled")
+    } else if (openChangeRequests.any { it.type == ChangeRequestType.PLACEMENT_APPEAL }) {
+      unavailable("There is an existing open change request of this type")
```
danhumphreys-moj, line 50: Good stuff

danhumphreys-moj (approved):
LGTM

PR #3444, opened 2025-05-07
APS-2250 remove reason for shortening a booking

danhumphreys-moj:
Ref: https://dsdmoj.atlassian.net/browse/APS-2250

- ~~send booking changed domain event when booking is shortened~~
- remove "reason" from API type and associated logic since it is [no longer required](https://mojdt.slack.com/archives/C03K0HB0LBE/p1746523751781129?thread_ts=1745998375.954829&cid=C03K0HB0LBE)
- ~~refactor common update/shorten email logic~~

_NB. Previous PRs already merged have added in the feature to send a domain event when a booking is shortened, therefore this PR and ticket scope is reduced to removing reason for shortening a booking only._


src/main/resources/static/cas1-api.yml
```diff
@@ -1141,7 +1141,7 @@ paths:
         content:
           'application/json':
             schema:
-              $ref: 'cas1-schemas.yml#/components/schemas/Cas1ShortenSpaceBooking'
+              $ref: 'cas1-schemas.yml#/components/schemas/Cas1UpdateSpaceBooking'
```
davidatkinsuk, line 1144: I think we should retain the specific `Cas1ShortenSpaceBooking` as it provides a more precise API definition e.g. you can make the change in departure date mandatory, it doesn't include irrelevant fields related to arrival date and characteristics.

davidatkinsuk (approved):
Just one comment from me 👍

PR #3450, opened 2025-05-07
Use Spring Events for Change Request Logic

davidatkinsuk:
Update logic that triggers domain events and emails related to change requests to use spring events. This helps reduce the amount of mocking in unit tests and decouples code


muhammad-elabdulla (approved):
LGTM

PR #3454, opened 2025-05-08
APS-2267 introduce internal booking events

davidatkinsuk:
Add internal event data classes when creating, amending or cancelling a space booking, in preparation for using spring events for 'intra service' communication, and in preparation for adding additional fields to these calls


muhammad-elabdulla (approved):
LGTM

PR #3458, opened 2025-05-08
Add `Cas1ChangeRequestSummary.placementRequestId`

davidatkinsuk:
Also make expectedArrivalDate mandatory and remove lengthOfStay as it isn't required


froddd (approved):
👍

PR #3460, opened 2025-05-09
APS-2267 isolate create update booking logic

davidatkinsuk:


danhumphreys-moj (approved):
Nice to have these big - but not significant - changes broken down into discrete commits.

PR #3463, opened 2025-05-09
Remove unused domain events

davidatkinsuk:
* Emergency Transfer Created Domain Event isn't required as this information will be included in the booking made event
* Change Request Approved Domain Event isn't required as this information will be included in the domain event representing the outcome (e.g. booking made, booking changed, cancelled)


danhumphreys-moj (approved):
LGTM

PR #3473, opened 2025-05-13
APS-2279 - Add optional email logging

davidatkinsuk:
This commit adds a configuration option to log each email’s template, recipient, reply to address and personalisation.

This is managed seperately from whether sending emails is enabled.

The commit enables this functionality for dev/test/local and disables it for pre-prod and prod


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/EmailNotificationService.kt
```diff
@@ -35,14 +43,16 @@ class EmailNotificationService(
     personalisation: Map<String, *>,
     replyToEmailId: String?,
   ) {
-    applicationEventPublisher.publishEvent(SendEmailRequestedEvent(EmailRequest(recipientEmailAddress, templateId, personalisation, replyToEmailId)))
+    val emailRequest = EmailRequest(recipientEmailAddress, templateId, personalisation, replyToEmailId)
+    applicationEventPublisher.publishEvent(SendEmailRequestedEvent(emailRequest))
+
+    if (notifyConfig.logEmails) {
```
besscerule, line 49: why is this in a separate if condition? does it make more sense to have in the `notifyConfig.mode == NotifyMode.DISABLED` as the log should only be sent in place of an email being sent - or is it more of a general option to have logging in whatever circumstances?

davidatkinsuk: And you raise a good point i hadn't considered - it's helpful to log the emails even if sending the, which is the setup we'll have in test. We can then grep through the emails in app insights, if required

besscerule: maybe you could make the personalisation log part optional - so it doesn't log sensitive stuff in preprod?

But happy with your current way too.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/EmailNotificationService.kt
```diff
@@ -22,6 +23,13 @@ class EmailNotificationService(
 ) : EmailNotifier {
   var log: Logger = LoggerFactory.getLogger(this::class.java)
 
+  companion object {
+    fun resolveTemplateName(templateId: String) = Cas2NotifyTemplates::class.memberProperties
+      .firstOrNull { it.get(Cas2NotifyTemplates) == templateId }?.name
```
besscerule, line 28: why is it get for one and getter.call for the other?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/util/LoggerExtension.kt
```diff
@@ -0,0 +1,44 @@
...
+
+import ch.qos.logback.classic.Logger
+import ch.qos.logback.classic.spi.ILoggingEvent
+import ch.qos.logback.classic.spi.ThrowableProxy
+import ch.qos.logback.core.read.ListAppender
+import org.junit.jupiter.api.extension.AfterEachCallback
+import org.junit.jupiter.api.extension.BeforeEachCallback
+import org.junit.jupiter.api.extension.ExtensionContext
+import org.slf4j.LoggerFactory
+
```
besscerule, line 11: argh - this is so useful, i remember wanting to assert logs earlier with emails - this is much better

besscerule (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/EmailNotificationService.kt
davidatkinsuk: `Cas1NotifyTemplates` value are constants (`const`), where-as `Cas2NotifyTemplates` are not. Ideally we'd update `Cas2NotifyTemplates` values to be constants but i'd already burnt too much time converting `Cas2NotifyTemplates` into an object on 18d42b9f24052357ceead7e56b1e12cca15dc68d

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/EmailNotificationService.kt
davidatkinsuk: We can always change to this approach in the future if needs be. At the moment i don't think there's an environment where email is disabled and logging isn't enabled.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/EmailNotificationService.kt
besscerule: would you be able to pop a tech debt ticket in the backlog to do this?

davidatkinsuk: Done - https://dsdmoj.atlassian.net/jira/software/c/projects/CAS/issues/CAS-1644D

PR #3474, opened 2025-05-13
APS-2298 - Show parent departure reason in placement report

davidatkinsuk:
If defined, show parent departure reason in the placement report


muhammad-elabdulla (approved):
LGTM

PR #3491, opened 2025-05-15
Add start/end date to reports.

vbala-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas1/Cas1DailyMetricsReportRepository.kt
```diff
@@ -0,0 +1,72 @@
...
+          COALESCE(applications_submitted, 0) AS applications_submitted,
+          COALESCE(unique_users_submitting_applications, 0) AS unique_users_submitting_applications,
+          COALESCE(assessments_completed, 0) AS assessments_completed,
+          COALESCE(unique_users_completing_assessments, 0) AS unique_users_completing_assessments,
+          COALESCE(bookings_made, 0) AS bookings_made,
+          COALESCE(unique_users_making_bookings, 0) AS unique_users_making_bookings
+      FROM applications_by_day
+      FULL OUTER JOIN domain_events_by_day
+          ON applications_by_day.report_date = domain_events_by_day.report_date
+      ORDER BY report_date
```
davidatkinsuk, line 56: I don't think this will happen, but will anything be output if there are no domain events for a given day?

To solve that you could use a CTE that just outputs all days within the range and use that as your `FROM` value here? Maybe using something like `generate_series('2021-03-20'::date, '2021-04-02'::date, interval '1 day') `

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas1/Cas1ReportsController.kt
```diff
@@ -34,105 +37,140 @@ class Cas1ReportsController(
...
+    val endDate = reportDateRange.end
 
+    // validateMonth(month) TODO()
+
+    // temp fix
     val monthSpecificReportParams = MonthSpecificReportParams(
-      year = year,
-      month = month,
+      year = reportDateRange.start.year,
+      month = reportDateRange.start.monthValue,
```
davidatkinsuk, line 60: I think it would be cleaner/easier to review if the date change work is handled in a separate PR

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/DailyMetricsReportTest.kt
```diff
@@ -246,15 +253,20 @@ class DailyMetricsReportTest : IntegrationTestBase() {
         .isOk
         .expectHeader().valuesMatch(
           "content-disposition",
-          "attachment; filename=\"daily-metrics-$year-${month.toString().padStart(2, '0')}-[0-9_]+.xlsx\"",
+          "attachment; filename=\"daily-metrics-$startDate-to-$endDate-\\d{8}_\\d{4}.xlsx\"",
         )
         .expectBody()
-        .consumeWith {
+        .consumeWith { response ->
```
davidatkinsuk, line 259: This test isn't actually verifying the output, it's just checking that the endpoint is returning whatever the report generator creates (this is an existing problem). I propose this test is improved on a separate PR merged before this one updating the assertions to check the actual values. Similar to what i had to do here - https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/3449

davidatkinsuk (changes requested):
A few comments from me

PR #3494, opened 2025-05-15
APS-2138 -  fix caching test

davidatkinsuk:
Tidies up the cache logic readability

Fixes a caching integration test that wasn't automatically running.


danhumphreys-moj (approved):
Discussed with @davidatkinsuk.  Work required in the future to agree how the cache refresh should actually work as we want it to.

PR #3495, opened 2025-05-15
APS-2318 Automatically remove corrupted cache metadata

davidatkinsuk:
If cache metadata states that a responseBody is available but that responseBody entry doesn’t exist in the cache, automatically remove the metadata from the cache as inconsistent state will not automatically resolve and we’ll always receive a `CachedValueUnavailable` error when attempting to retrieve from the cache until it’s hard TTL is reached.

We now also raise a sentry alert when this happens, as this really _shouldn’t_ happen


danhumphreys-moj (approved):
Discussed with @davidatkinsuk.  Work required in the future to agree how the cache refresh should actually work as we want it to.

PR #3509, opened 2025-05-21
CAS-1659 Remove redundant script

danhumphreys-moj:
Ref: https://dsdmoj.atlassian.net/browse/CAS-1659 & https://dsdmoj.atlassian.net/browse/CAS-1658

Following on from Firebreak work, in AP tools we do not run a local Dockerised version of HMPPS Auth any more, and therefore this script is now redundant, so removing.


muhammad-elabdulla (approved):
LGTM

PR #3512, opened 2025-05-21
updated to not throw error if no application status found

daveawc:
There can be applications without any relevant status - this returns a default value of 'submitted' instead of throwing an error.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/Cas2EmailService.kt
```diff
@@ -22,14 +22,17 @@ class Cas2EmailService(
   @Value("\${notify.emailaddresses.nacro}") private val nacroEmail: String,
 ) {
 
+  fun getApplicationStatusOrDefault(applicationId: UUID): String = statusUpdateRepository.findFirstByApplicationIdOrderByCreatedAtDesc(applicationId)?.label ?: "Submitted"
```
besscerule, line 25: Can we confirm that @raykhan12345 is aware of this an OK with it.

besscerule (approved):

gregkhawkins (approved):

PR #3519, opened 2025-05-22
Bedspace model refactor - migrate data to new cas3_premises table

gregkhawkins:
JIRA: https://dsdmoj.atlassian.net/browse/CAS-1569

## Context

We are trying to gradually create the tables required for this new database architecture and migrate the data:

new premises tables with cas3 void bedspace

## PR includes

* Create the new `cas3_premises` table using flyway
* Create an entity and repository for this new table
* Migrate the CAS3 related data from the `premises` and `temporary_accomodation_premises` into our new new `cas3_premises` table

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas3/Cas3PremisesEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas3/Cas3PremisesEntity.kt
@@ -1,0 +1,46 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.cas3
+
+import jakarta.persistence.Entity
+import jakarta.persistence.EnumType
+import jakarta.persistence.Enumerated
+import jakarta.persistence.Id
+import jakarta.persistence.Inheritance
+import jakarta.persistence.InheritanceType
+import jakarta.persistence.JoinColumn
+import jakarta.persistence.ManyToOne
+import jakarta.persistence.Table
+import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.stereotype.Repository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.PropertyStatus
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.LocalAuthorityAreaEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ProbationDeliveryUnitEntity
+import java.util.UUID
+
+@SuppressWarnings("LongParameterList")
+@Entity
+@Table(name = "cas3_premises")
+@Inheritance(strategy = InheritanceType.JOINED)
+class Cas3PremisesEntity(
+  @Id
+  val id: UUID,
+  var name: String,
+  var postcode: String,
+  var addressLine1: String,
+  var addressLine2: String?,
+  var town: String?,
+
+  @ManyToOne
+  @JoinColumn(name = "probation_delivery_unit_id")
+  var probationDeliveryUnit: ProbationDeliveryUnitEntity,
+
+  @ManyToOne
+  @JoinColumn(name = "local_authority_area_id")
+  var localAuthorityArea: LocalAuthorityAreaEntity?,
+
+  @Enumerated(value = EnumType.STRING)
+  var status: PropertyStatus,
+  var notes: String,
+)
+
+@Repository
+interface Cas3PremisesRepository : JpaRepository<Cas3PremisesEntity, UUID>
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/migration/cas3/Cas3MigrateNewBedspaceModelDataJob.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/migration/cas3/Cas3MigrateNewBedspaceModelDataJob.kt
@@ -1,0 +1,68 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.migration.cas3
+
+import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.data.jpa.repository.Query
+import org.springframework.stereotype.Component
+import org.springframework.stereotype.Repository
+import org.springframework.transaction.support.TransactionTemplate
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.TemporaryAccommodationPremisesEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.cas3.Cas3PremisesEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.cas3.Cas3PremisesRepository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.migration.MigrationInBatchesJob
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.migration.MigrationLogger
+import java.util.UUID
+
+@Component
+class Cas3MigrateNewBedspaceModelDataJob(
+  private val temporaryAccommodationPremisesRepository: TemporaryAccommodationPremisesRepository,
+  private val cas3PremisesRepository: Cas3PremisesRepository,
+  private val migrationLogger: MigrationLogger,
+  transactionTemplate: TransactionTemplate,
+) : MigrationInBatchesJob(migrationLogger, transactionTemplate) {
+  override val shouldRunInTransaction: Boolean = java.lang.Boolean.FALSE
+
+  override fun process(pageSize: Int) {
+    migrationLogger.info("Starting migration process...")
+    val cas3PremisesIds = temporaryAccommodationPremisesRepository.findTemporaryAccommodationPremisesIds()
+    super.processInBatches(cas3PremisesIds, batchSize = 100) { batchIds ->
+      migrateDataToNewBedspaceModelTables(batchIds)
+    }
+    migrationLogger.info("Completed migration process...")
+  }
+
+  private fun migrateDataToNewBedspaceModelTables(premiseIds: List<UUID>) {
+    migrationLogger.info("Starting bedspace model migration with batch size of ${premiseIds.size}...")
+    migrateDataToCas3Premises(premiseIds)
+    migrationLogger.info("Migrated batch size of ${premiseIds.size} to new cas3_premises table - data migrated from the premises and temporary_accommodation tables.")
+  }
+
+  private fun migrateDataToCas3Premises(premiseIds: List<UUID>) {
+    val temporaryAccommodationPremisesBatch = temporaryAccommodationPremisesRepository.findTemporaryAccommodationPremisesByIds(premiseIds)
+    val cas3Premises = transformCas3PremisesBatch(temporaryAccommodationPremisesBatch)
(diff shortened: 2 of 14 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PremisesEntity.kt
```diff
@@ -61,6 +61,12 @@ SELECT
   )
   fun findAllCas3PremisesSummary(regionId: UUID, postcodeOrAddress: String?, postcodeOrAddressWithoutWhitespace: String?, propertyStatus: String?): List<TemporaryAccommodationPremisesSummary>
 
+  @Query("SELECT tam.id FROM TemporaryAccommodationPremisesEntity tam")
```
muhammad-elabdulla, line 64: Can you change this to tap instead of tam

src/main/resources/db/migration/all/20250516172300__cas3_premises_table.sql
```diff
@@ -0,0 +1,18 @@
+CREATE TABLE cas3_premises(
+    id                              UUID not null,
+    name                            TEXT not null,
+    postcode                        TEXT not null,
+    probation_delivery_unit_id      UUID,
```
muhammad-elabdulla, line 5: should be not null

src/main/resources/db/migration/all/20250516172300__cas3_premises_table.sql
```diff
@@ -0,0 +1,18 @@
...
+    local_authority_area_id         UUID not null,
```
muhammad-elabdulla, line 6: Local authority can be null

src/main/resources/db/migration/all/20250516172300__cas3_premises_table.sql
```diff
@@ -0,0 +1,18 @@
...
+    address_line1                   TEXT not null,
+    address_line2                   TEXT,
+    town                            TEXT,
+    status                          TEXT default 'active'::TEXT not null,
+    turnaround_working_day_count    INTEGER default 2 not null,
```
muhammad-elabdulla, line 11: I will not set a default value for turnaround

muhammad-elabdulla (commented):
We agreed that create table will be separate commit

muhammad-elabdulla:
- What happened if the migration process fail in the middle of the migration?
- If the migration process will run another time will have duplicated record in `cas3_premises` table!!
- How we will make sure that all the premises was migrated correctly?

src/main/resources/db/migration/all/20250516172300__cas3_premises_table.sql
gregkhawkins: see response to above comment - happy to change if you are sure :)

src/main/resources/db/migration/all/20250516172300__cas3_premises_table.sql
gregkhawkins: good catch

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/migration/cas3/Cas3MigrateNewBedspaceModelDataJob.kt
```diff
@@ -0,0 +1,58 @@
...
+
+  private fun migrateDataToCas3Premises(premiseIds: List<UUID>) {
+    val temporaryAccommodationPremisesBatch = premisesRepository.findTemporaryAccommodationPremisesIdIn(premiseIds)
+    val cas3Premises = transformCas3PremisesBatch(temporaryAccommodationPremisesBatch)
+    cas3PremisesRepository.saveAllAndFlush(cas3Premises)
+  }
+
+  private fun transformCas3PremisesBatch(temporaryAccommodationPremisesBatch: List<TemporaryAccommodationPremisesEntity>) = temporaryAccommodationPremisesBatch.map { premise ->
+    Cas3PremisesEntity(
+      id = UUID.randomUUID(),
```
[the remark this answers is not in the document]
gregkhawkins: okay, I did wonder about this. If we retain the `premise.id` as `cas3_premises.id` then the records are linked which is helpful.
In addition, this may get us over the issue you flagged with `if the migration process will run another time will have duplicated record in cas3_premises table!!`. Will have a look to see if the `cas3PremisesRepository.saveAllAndFlush(cas3Premises)` is smart enough to do a `CREATE or UPDATE` under the hood.

Would quite like to retain the batchy-ness of this solution if possible (instead of iterating through one by one to create records). The rationale being that Inserting in batches should be more efficient and kinder on the DB (with the way the saveAll() is configured and interacts with the DB).

Anyway, will investigate sharing the id and running multiple times to see how it behaves.
Thanks!

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/migration/cas3/Cas3MigrateNewBedspaceModelDataJobTest.kt
```diff
@@ -0,0 +1,62 @@
...
+    migrationJobService.runMigrationJob(MigrationJobType.cas3BedspaceModelData, 1)
+    assertThatAllCas3PremisesDataWasMigratedSuccessfully()
+  }
+
+  private fun assertThatAllCas3PremisesDataWasMigratedSuccessfully() {
+    val migratedPremises = cas3PremisesRepository.findAll()
+    assertThat(migratedPremises.size).isEqualTo(NO_OF_PREMISES_MIGRATED)
+
+    migratedPremises.forEach { migratedPremise ->
+      val tap = temporaryAccommodationPremises.firstOrNull { it.name == migratedPremise.name }!!
```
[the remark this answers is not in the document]
gregkhawkins: yep, this will be an improvement now moving to using the same id. 
I did check whether this would lead to flaky tests btw, but the way we generate the name in the test setup data means practically impossible to get clashes.
Will be better to use id anyway so will move to that now we are sharing IDs

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/migration/cas3/Cas3MigrateNewBedspaceModelDataJob.kt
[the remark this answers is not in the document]
gregkhawkins: yep, retaining the `premise.id` as `cas3_premises.id` resolves the duplicate row issue. Nice and easy - thanks again for flagging :)

gregkhawkins:
> * What happened if the migration process fail in the middle of the migration?
> * If the migration process will run another time will have duplicated record in `cas3_premises` table!!
> * How we will make sure that all the premises was migrated correctly?

1. Good catch on the duplicate records! 
* resolved by retaining the `premise.id` as `cas3_premises.id` - now does an update or a create on the second/third/fourth run as a result of this change

2. How we will make sure that all the premises was migrated correctly?
* my view is we write a sql query to validate this to give us counts on all the tables to prove it. 
* if you strongly feel this count validation should be included in the migration job then happy to have a chat on this to discuss prod and cons :)

gregkhawkins:
> > * What happened if the migration process fail in the middle of the migration?
> > * If the migration process will run another time will have duplicated record in `cas3_premises` table!!
> > * How we will make sure that all the premises was migrated correctly?
> 
> 1. Good catch on the duplicate records!
> 
> * resolved by retaining the `premise.id` as `cas3_premises.id` - now does an update or a create on the second/third/fourth run as a result of this change
> 
> 2. How we will make sure that all the premises was migrated correctly?
> 
> * my view is we write a sql query to validate this to give us counts on all the tables to prove it.
> * if you strongly feel this count validation should be included in the migration job then happy to have a chat on this to discuss prod and cons :)


Think runnning this SQL against the prod db (post running migration job) should guarantee everything has been migrated:

```
select count(*)
from premises p
join temporary_accommodation_premises tap on tap.premises_id = p.id;

select count(c3p.*)
from cas3_premises c3p
join premises p on c3p.id = p.id
join temporary_accommodation_premises tap on tap.premises_id = p.id
where c3p.name = p.name
and c3p.postcode = p.postcode
and (c3p.probation_delivery_unit_id = tap.probation_delivery_unit_id OR (c3p.probation_delivery_unit_id IS NULL AND tap.probation_delivery_unit_id IS NULL))
and (c3p.local_authority_area_id = p.local_authority_area_id OR (c3p.local_authority_area_id IS NULL AND p.local_authority_area_id IS NULL))
and c3p.address_line1 = p.address_line1
and (c3p.address_line2 = p.address_line2 OR (c3p.address_line2 IS NULL AND p.address_line2 IS NULL))
and (c3p.town = p.town OR (c3p.town IS NULL AND p.town IS NULL))
and c3p.status = p.status
and c3p.turnaround_working_day_count = tap.turnaround_working_day_count
and (c3p.notes = p.notes OR (c3p.notes IS NULL AND p.notes IS NULL));
```

I think this is preferable to checking programmatically in code as otherwise we are essentially lifting the integration test into the implementation (same code to validate) and slows down the job. The integration test should give us confidence that the code does the right thing. Then, to be double sure it worked we run our count queries against the prod db when the job finishes to validate the data was migrated and be sure we are happy.

Also, I cannot see any other migration jobs in the codebase that validate in code whether the migration was successful - they all jsut have integration tests to validate it :)

My suggestion would be to add the validation queries against the data migration tickets so we can run pull them out of the tickets easily to validate when we run in prdcution. Can use these to validate all working when we run iteratively in preprod also.

Good shout though to have in scope of the ticket a way to validate the job is doing the right thing.

gregkhawkins:
> We agreed that create table will be separate commit

Can you clarify if you mean separate PR or separate commit? Also, what's the rationale for this? Happy to do it! It's just I have reviewed lots of PRs since being here with flyway scripts as well as Kotlin code changes together. So, just interested in why this is a standard for this case. Thanks

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas3/Cas3PremisesEntity.kt
```diff
@@ -0,0 +1,47 @@
...
+  val id: UUID,
+  var name: String,
+  var postcode: String,
+  var addressLine1: String,
+  var addressLine2: String?,
+  var town: String?,
+
+  @ManyToOne
+  @JoinColumn(name = "probation_delivery_unit_id")
+  var probationDeliveryUnit: ProbationDeliveryUnitEntity?,
```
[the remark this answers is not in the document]
muhammad-elabdulla: I know but for CAS3 premises PDU is always mandatory. This is the downside for sharing BE between CASs

muhammad-elabdulla:
I flagged that _How we will make sure that all the premises was migrated correctly?_

I don't have any objections to using an SQL query for check the integrity of the data. 

However, it would be good to include it as part of the PR so that we can refer back to it and run it in any environment. 

Also, it can be executed by any developer who runs the migration job.

muhammad-elabdulla:
I prefer to have separate commit for the creation table and another for the migration process in the same PR

gregkhawkins:
> I prefer to have separate commit for the creation table and another for the migration process in the same PR

Done! It was easiest to make the flyway script the second commit (to avoid too much git wizardry!). Will commit the flyway scripts as a first commit from now on and separate out. Logical to do it that way round

PR #3530, opened 2025-05-27
Suppress irrelevant POI vulnerability

davidatkinsuk:
As we trust the XLSX files we import, we are safe to ignore CVE-2025-31672

Note that there isn't currently a version of dataframe-excel that use a version of POI that resolves this issue. Once there is, we should upgrade to it and remove this suppression


muhammad-elabdulla (approved):
LGTM

PR #3533, opened 2025-05-27
Disable booking to space booking in prod

davidatkinsuk:
This adds an additional layer of safety whilst we test this process in pre-prod.


besscerule (approved):
LGTM

PR #3544, opened 2025-05-29
APS-2348 - Ensure JPA result nullability is correct

davidatkinsuk:
This commit updates various JPA ‘result row’ interface definitions to correctly indicate if a value is nullable.

It also renames some result rows that only apply to CAS3 to make this clear (nullability on certain shared tables differes for CAS1/CAS3).

It also removes some unused repository code


besscerule (approved):

muhammad-elabdulla (approved):
LGTM

PR #3552, opened 2025-05-30
Remove overloaded assertThat(CasResult)

davidatkinsuk:
Whilst technically this is fine, it often leads to the IDE importing it by mistake when we really want the default assertj `assertThat` function. To assert on CasResult the `assertThatCasResult` function should be used instead.


besscerule (commented):
This always used to annoy me!

There are a lot of Assertions.assertThat that can be changed to just assertThat now - could you change in this PR?

PR #3555, opened 2025-06-02
CAS-1684: Add offender_name to bookings table

rashadmughal:
This adds a migration script only to add the offender_name to the bookings table in the DB and updated code to handle new column, setting to null for the time being.


muhammad-elabdulla (approved):
LGTM

PR #3560, opened 2025-06-02
Refactor reports date range

vbala-moj:
Updated reports to use date-time range instead of just date, as it will have an impact on other report generation.


davidatkinsuk (approved):
LGTM 👍

PR #3562, opened 2025-06-02
feature/APS-2261-reports-name-refactor

vbala-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas1/Cas1ReportsController.kt
```diff
@@ -158,23 +150,16 @@ class Cas1ReportsController(
...
-  private fun createCas1ReportName(name: String, year: Int, month: Int, contentType: ContentType): String {
-    val timestamp = LocalDateTime.now().format(TIMESTAMP_FORMAT)
-    return "$name-$year-${month.toString().padStart(2, '0')}-$timestamp.${contentType.extension}"
-  }
-
   private fun createCas1ReportName(
     name: String,
     startDate: LocalDateTime,
     endDate: LocalDateTime,
-    contentType: ContentType,
```
vbala-moj, line 171: We may need to revert this change if we go back to using Excel for the OOSB report.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas1/Cas1ReportsController.kt
```diff
@@ -54,85 +53,78 @@ class Cas1ReportsController(
...
-    val monthSpecificReportParams = MonthSpecificReportParams(
-      year = year,
-      month = month,
-    )
 
     return when (reportName) {
       Cas1ReportName.applicationsV2 -> generateStreamingResponse(
         contentType = ContentType.CSV,
-        fileName = createCas1ReportName("applications", year, month, ContentType.CSV),
+        fileName = createCas1ReportName("applications", startDate, endDate),
```
davidatkinsuk, line 60: I guess it may be a tiny bit neater if we passed the dateRange directly into the `createCas1ReportName` function? I think we can then get rid of startDate/endDate variables declared on lines 54 & 55?

davidatkinsuk (approved):
Just one minor comment from me. This is looking super neat 👍 . As with the OOSB PR, we can only merged once we're happy that consumers can ingest the CSV version of the report

PR #3591, opened 2025-06-05
CAS-1725 - fix for incorrect email address

daveawc:
the incorrect email is being used when notes were added by nacro. 

This PR updates that to use either the most recent assigned POM, or to use the OMU email address if they are unallocated.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas2/Cas2EmailService.kt
```diff
@@ -101,6 +101,17 @@ class Cas2EmailService(
     )
   }
 
+  fun getReferrerEmail(application: Cas2ApplicationEntity): String? {
+    // currently with cas2 bail there is no POM, and the offender is not in prison.
+    // in this instance we should fall back to use the created by user (a delius user)
+    application.currentAssignment?.let {
```
gregkhawkins, line 107: presumably is there no `currentAssignment` for `cas2 bail` either then?

gregkhawkins (approved):

PR #3606, opened 2025-06-09
SPIKE - Controller -> Service refactor

gregkhawkins:
**Background**
The idea behind the work done for the spike is that we will create 1 JIRA ticket per new CAS3 endpoint. The scope of each ticket will be:
1. Decouple cas3 functionality from the original controller (discussed in comment below in depth)
2. Add new endpoint to the cas3 controller and integrate that new endpoint with the new cas3 serivce class (and cas3 repos etc)

**Delivered**
* A work in progress of doing the above for the `GET /premises/{premisesId}` endpoint
* Integration tests and unit tests pass which proves the decoupling bit has not broken the shared endpoint
* Have commented out functionality (not much!) that is not do-able yet (as new CAS3 datamodel in progress!)
* Have not yet delivered an integration tests for WIP new cas3 endpoint (or unit tests) - only a spike!
* Have made comments in the PR to justify / discuss certain aspects of this approach


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
```diff
@@ -88,10 +90,12 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas3.GetBookingF
 
 @Service
 class PremisesController(
+  private val requestContextService: RequestContextService,
+  private val cas3PremisesController: Cas3PremisesController,
   private val usersService: UserService,
   private val userAccessService: UserAccessService,
   private val premisesService: PremisesService,
-  private val cas3PremisesService: Cas3PremisesService,
+  private val temporaryAccommodationPremisesService: TemporaryAccommodationPremisesService,
```
gregkhawkins, line 98: I thought it might be a good idea to rename the pre-existing `Cas3PremisesService` -> `TemporaryAccommodationPremisesService`. The reason for this was I started adding the new functions to the pre-existing `Cas3PremisesService` and realised this would fast get out of control as the pre-existing service is already 500 lines of code.

Another good reason is that if you continue adding your functions to this renamed `TemporaryAccommodationPremisesService` class, then we can meanwhile add our new functions exclusively to the new `Cas3PremisesService. This will mean we reduce conflicts. If there is stuff we need to reuse from`TemporaryAccommodationPremisesService` my suggestion would be to move the reusable code to the new `Cas3PremisesService` and call it from `TemporaryAccommodationPremisesService` then we have access to it also.

The good thing about this approach is that at the end (once the UI has switched to the new cas3 endpoints) we can just delete `TemporaryAccommodationPremisesService` and keep the  `Cas3PremisesService`. This will save us from looking at conditional logic to figure out what we can delete that was used for the old shared endpoints.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas3/Cas3PremisesController.kt
```diff
@@ -60,9 +69,38 @@ class Cas3PremisesController(
...
+
+    if (!userAccessService.currentUserCanViewPremises(premises)) {
+      throw ForbiddenProblem()
+    }
+    val totalBeds = temporaryAccommodationPremisesService.getBedCount(premises)
+    val availableBedsForToday =
+      temporaryAccommodationPremisesService.getAvailabilityForRange(premises, LocalDate.now(), LocalDate.now().plusDays(1))
+        .values.first().getFreeCapacity(totalBeds)
+    return ResponseEntity.ok(premisesTransformer.transformJpaToApi(premises, totalBeds, availableBedsForToday))
+  }
```
gregkhawkins, line 84: the above is called by the `PremisesContoller` to do the cas3 stuff currently in production

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas3/Cas3PremisesController.kt
```diff
@@ -60,9 +69,38 @@ class Cas3PremisesController(
...
+    val premises = cas3PremisesService.getPremises(premisesId)
+      ?: throw NotFoundProblem(premisesId, "Premises")
+
+    if (!userAccessService.currentUserCanViewPremises(premises)) {
+      throw ForbiddenProblem()
+    }
+    val totalBeds = cas3PremisesService.getBedspaceCount(premises)
+    val availableBedsForToday = 0 // todo: replace with below equivalent
+//      cas3PremisesService.getAvailabilityForRange(premises, LocalDate.now(), LocalDate.now().plusDays(1))
+//        .values.first().getFreeCapacity(totalBeds)
```
gregkhawkins, line 96: this is commented out because what happens in the implementation of the `cas3PremisesService.getAvailabilityForRange()` depends on `cas3_void_bedspaces` which @daveawc is currently refactoring

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserAccessService.kt
```diff
@@ -56,6 +57,10 @@ class UserAccessService(
     else -> false
   }
 
+  fun currentUserCanViewPremises(premises: Cas3PremisesEntity) = userCanViewPremises(userService.getUserForRequest(), premises)
+
+  fun userCanViewPremises(user: UserEntity, premises: Cas3PremisesEntity) = userCanAccessRegion(user, ServiceName.temporaryAccommodation, probationRegionId = premises.id) && user.hasRole(UserRole.CAS3_ASSESSOR)
```
gregkhawkins, line 62: A keen eye might notice that I pass `premises.id` for the `probationRegionId` parameter here (to the `userCanAccessRegion()` function. This is obviously temporary as in the implementation of the `userCanAccessRegion()` function we check the the `user.probationRegion.id` is the same as this to make sure they have access.

My understanding is that we intentionally removed the join to the `ProbationRegionEntity` on the new `Cas3PremisesEntity` so not sure how we mimic this business logic?

Let me know if this has already been thought out. :)

PR #3611, opened 2025-06-10
APS-2092 enable prometheus alerts for prod

danhumphreys-moj:
Ref: https://dsdmoj.atlassian.net/browse/APS-2092

This PR enables Prometheus alerting in `prod` and enables database alerts to be triggered in the `dev` environment when we have at least one connection.  This is to check that database alerting works as expected.

Once this is confirmed we can set the alerting threshold values to those appropriate for each environment.


muhammad-elabdulla (approved):
LGTM

davidatkinsuk (approved):
LGTM 👍

PR #3630, opened 2025-06-12
CAS-1686: C3 BE Change bookings search to use new offender name field

rashadmughal:
Now that the `offender_name` field is populated by the migration job ``Cas3UpdateBookingOffenderNameJob` this PR uses this field to sort by name when required.


muhammad-elabdulla (approved):
LGTM

PR #3663, opened 2025-06-16
APS-2379 Isolate CAS2 OASys transform logic

davidatkinsuk:
Move all CAS2 OASys Transform logic into a CAS2 specific class. This does create some limited duplication, but is required for subsequent changes to the CAS2 logic that will diverge from CAS1 & CAS 3


besscerule (approved):
But wait to merge until we work out what is going on in the UI

davidatkinsuk:
> But wait to merge until we work out what is going on in the UI

Note that there is no functional change here, so it _should_ be safe to merge

PR #3675, opened 2025-06-18
Use simple PR template by default

davidatkinsuk:


.github/workflows/PULL_REQUEST_TEMPLATE/full_template.md
```diff
@@ -0,0 +1,33 @@
...
+  infrastructure, third-party integrations and any secrets or environment variables)
+- [ ] Has a required data migration been included in this change or been done in
+  advance?
+
+## Post merge checklist
+
+Once we've merged we now need to release this through to production before
+considering it done.
+
+[Find the build-and-deploy job in CircleCI](https://app.circleci.com/pipelines/github/ministryofjustice/hmpps-temporary-accommodation-ui)
```
besscerule, line 22: it's not circleCI anymore

besscerule:
> > Whats happened to the pull_request_template.md in root - won't that override this and does it need to be deleted?
> 
> As far as i can tell if we had templates just in the directory, none would be used unless you specify them via the URL (i don't think a drop down will appear from everything i've read).
> 
> I can try removing the one in root, merge this, and see what happens?

I think we should only have one way of doing templates so we shouldn't have multiple floating around - so maybe remove?

davidatkinsuk:
> > > Whats happened to the pull_request_template.md in root - won't that override this and does it need to be deleted?
> > 
> > 
> > As far as i can tell if we had templates just in the directory, none would be used unless you specify them via the URL (i don't think a drop down will appear from everything i've read).
> > I can try removing the one in root, merge this, and see what happens?
> 
> I think we should only have one way of doing templates so we shouldn't have multiple floating around - so maybe remove?

as discussed on slack, we'll try putting two templates in the folder and nothing else and see what happens

PR #3680, opened 2025-06-18
CAS-1669: C3 BE Migration process to assign start date to bedspaces

rashadmughal:
This PR adds a new migration script that sets the start date for beds, taking them from either the `created_date` or `arrival_date`

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BedEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BedEntity.kt
@@ -10,6 +10,8 @@ import jakarta.persistence.NamedNativeQuery
 import jakarta.persistence.SqlResultSetMapping
 import jakarta.persistence.Table
 import org.hibernate.annotations.CreationTimestamp
+import org.springframework.data.domain.Pageable
+import org.springframework.data.domain.Slice
 import org.springframework.data.jpa.repository.JpaRepository
 import org.springframework.data.jpa.repository.Modifying
 import org.springframework.data.jpa.repository.Query
@@ -54,6 +56,18 @@ interface BedRepository : JpaRepository<BedEntity, UUID> {
     nativeQuery = true,
   )
   fun findAllCas1BedSummariesForPremises(premisesId: UUID): List<Cas1PremisesBedSummary>
+
+  @Query(
+    """
+        SELECT b.*
+        FROM beds b
+        INNER JOIN rooms r ON b.room_id = r.id
+        INNER JOIN premises p ON r.premises_id = p.id
+        WHERE p.service = :service
+    """,
+    nativeQuery = true,
+  )
+  fun findAllByService(service: String, pageable: Pageable): Slice<BedEntity>
 
   @Modifying
   @Query("UPDATE BedEntity b SET b.code = :code WHERE b.id = :id")
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/migration/cas3/Cas3UpdateBedSpaceStartDateJob.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/migration/cas3/Cas3UpdateBedSpaceStartDateJob.kt
@@ -1,0 +1,68 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.migration.cas3
+
+import jakarta.persistence.EntityManager
+import org.springframework.data.domain.Limit
+import org.springframework.data.domain.PageRequest
+import org.springframework.data.domain.Slice
+import org.springframework.data.domain.Sort
+import org.springframework.data.domain.Sort.Direction.ASC
+import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ServiceName.temporaryAccommodation
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.BedEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.BedRepository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.BookingRepository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.migration.MigrationJob
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.migration.MigrationLogger
+
+@Component
+class Cas3UpdateBedSpaceStartDateJob(
+  private val entityManager: EntityManager,
+  private val migrationLogger: MigrationLogger,
+  private val bookingRepository: BookingRepository,
+  private val bedRepository: BedRepository,
+) : MigrationJob() {
+  override val shouldRunInTransaction = false
+
+  @SuppressWarnings("MagicNumber", "TooGenericExceptionCaught", "NestedBlockDepth")
+  override fun process(pageSize: Int) {
+    var page = 0
+    var hasNext = true
+    var slice: Slice<BedEntity>
+    var currentSliceBeds = setOf<String>()
+
+    try {
+      while (hasNext) {
+        migrationLogger.info("Getting page $page for max page size $pageSize")
+        slice = bedRepository.findAllByService(temporaryAccommodation.value, PageRequest.of(page, pageSize))
+
+        currentSliceBeds = slice.map { it.id.toString() }.toSet()
+
+        for (bed in slice.content) {
+          if (bed.startDate == null) {
+            migrationLogger.info("Updating bed start_date for bed id ${bed.id}")
+            if (bed.createdAt != null) {
+              migrationLogger.info("Using created_date as start_date for bed id ${bed.id}")
+              val updatedBed = bed.copy(startDate = bed.createdAt!!.toLocalDate()!!)
+              bedRepository.save(updatedBed)
+            } else {
+              migrationLogger.info("Using oldest booking arrival_date as start_date for bed id ${bed.id}")
+              val oldestBooking = bookingRepository.findByBedId(bed.id, Sort.by(ASC, "arrivalDate"), Limit.of(1))
+              if (oldestBooking.isEmpty()) {
+                migrationLogger.info("No bookings found for bed id ${bed.id}")
+              } else {
+                val updatedBed = bed.copy(startDate = oldestBooking[0].arrivalDate)
+                bedRepository.save(updatedBed)
+                migrationLogger.info("Updated start_date for bed id ${bed.id} from booking arrival_date")
+              }
+            }
+          }
+        }
+        entityManager.clear()
+        hasNext = slice.hasNext()
+        page++
+      }
+    } catch (exception: Exception) {
+      migrationLogger.error("Unable to update bed(s) start_date with id ${currentSliceBeds.joinToString()}", exception)
+    }
+  }
+}
--- a/src/main/resources/static/_shared.yml
+++ b/src/main/resources/static/_shared.yml
@@ -3361,6 +3361,7 @@ components:
         - update_cas1_backfill_user_ap_area
         - update_cas3_application_offender_name
         - update_cas3_booking_offender_name
+        - update_cas3_bedspace_start_date
         - update_cas3_domain_event_type_for_person_departed_updated
         - update_cas1_applications_licence_expiry_date
         - update_cas1_backfill_offline_application_name
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/migration/cas3/Cas3UpdateBedSpaceStartDateJobTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/migration/cas3/Cas3UpdateBedSpaceStartDateJobTest.kt
@@ -1,0 +1,176 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.integration.migration.cas3
+
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.MigrationJobType
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ServiceName.temporaryAccommodation
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.integration.givens.givenAUser
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.integration.migration.MigrationJobTestBase
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.BedEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.randomDateAfter
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.randomDateBefore
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.randomDateTimeBefore
+import java.time.LocalDate
+import java.time.OffsetDateTime
+
+class Cas3UpdateBedSpaceStartDateJobTest : MigrationJobTestBase() {
+
+  @Test
+  fun `beds with null startDate are updated using createdAt date`() {
+    givenAUser { user, jwt ->
+
+      val premisesTemporaryAccommodation = temporaryAccommodationPremisesEntityFactory.produceAndPersist {
+        withProbationRegion(user.probationRegion)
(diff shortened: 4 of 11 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/migration/cas3/Cas3UpdateBedSpaceStartDateJob.kt
```diff
@@ -0,0 +1,65 @@
...
+          migrationLogger.info("Updating bed start_date for bed id ${bed.id}")
+          if (bed.createdAt != null) {
+            migrationLogger.info("Using created_date as start_date for bed id ${bed.id}")
+            val updatedBed = bed.copy(startDate = bed.createdAt!!.toLocalDate()!!).apply {}
+            bedRepository.save(updatedBed)
+          } else {
+            migrationLogger.info("Using oldest booking arrival_date as start_date for bed id ${bed.id}")
+            val oldestBooking = bookingRepository.findByBedId(bed.id, Sort.by(ASC, "arrivalDate"), Limit.of(1))
+            if (oldestBooking.isEmpty()) {
+              migrationLogger.info("No bookings found for bed id ${bed.id}")
```
davidatkinsuk, line 49: If this occurs for more beds than the page size, i think we'll be stuck in an infinite loop. Is this likely to happen?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/migration/cas3/Cas3UpdateBedSpaceStartDateJobTest.kt
```diff
@@ -0,0 +1,127 @@
...
+        withPremises(premises)
+        withName("Test Room")
+      }
+
+      val createdAt = OffsetDateTime.now()
+      val beds = (1..5).map {
+        bedEntityFactory.produceAndPersist {
+          withRoom(room)
+          withName("Bed $it")
+          withStartDate(null)
```
davidatkinsuk, line 32: for future ref we have a bunch of 'givenA...' functions for populating entities in integration tests which are slightly more fluent than the clunky factories e.g. `givenABed()`.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/migration/cas3/Cas3UpdateBedSpaceStartDateJobTest.kt
```diff
@@ -0,0 +1,127 @@
...
+
+      val room = roomEntityFactory.produceAndPersist {
+        withPremises(premises)
+        withName("Test Room")
+      }
+
+      val createdAt = OffsetDateTime.now()
+      val beds = (1..5).map {
+        bedEntityFactory.produceAndPersist {
+          withRoom(room)
```
davidatkinsuk, line 30: i think it may be clearer if you explicitly set the createdAt value like you do in the subsequent test. Could use a shared function for it e.g. `setCreatedAtForBeds(beds, createdAt)`

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/migration/cas3/Cas3UpdateBedSpaceStartDateJobTest.kt
```diff
@@ -0,0 +1,127 @@
...
+          withName("Bed $it")
+          withStartDate(null)
+        }
+      }
+
+      beds.map {
+        it.createdAt = null
+        bedRepository.save(it)
```
davidatkinsuk, line 67: i guess we have to do this because we're using @CreationTimestamp?

rashadmughal: yup

rashadmughal:
> Looks good ... a few comments from me for consideration.
> 
> It doesn't help too much now (sorry), and you may have discussed this already - but i think i would of done this with two SQL scripts in `migration/all` e.g.
> 
> script 1 - set startDate to createdAt where startDate is null script 1 - set startDate to oldest booking arrival date where startDate is null (could use lateral join with limit 1, or subquery to find the oldest booking date)
> 
> I appreciate that's not automatically testable, but given this is a one time thing, i think that's fine and obviously less effort

yep, 100% agree, a couple of lines of SQL would have been quicker, but the argument was the SQL could potentially modify more records than we intended to - and without any logging about what was done.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/migration/cas3/Cas3UpdateBedSpaceStartDateJob.kt
davidatkinsuk: Yes, you're right - you're passing a page number into the query, i misread it. But, i think that highlights a different issue ...

```slice = bedRepository.findAllByStartDateIsNull(PageRequest.of(page, pageSize))```

Because you're returning entries where start date is null, what's in each page will change after every iteration, meaning you'll miss entries (i think)

e.g. imagine you have these beds that need updating:

A, B, C, D, E ,F

with a page size of 2.

First iteration, request page 1, returns A, B

You 'fix' A,B

We now have these beds that need updating

C, D, E, F

 request page 2, returns E,F

So C,D have been missed.

Typically in these migration jobs the page size is left at 1, and you modifying records naturally modifies what appears in the first page on every iteration (which can lead to the issue i originally thought you had).

I think you may be better off just getting a complete list of IDs to fix and just for loop through them, ignoring the paging configuration.  I guess there's only a few thousand to deal with?

File: src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/migration/cas3/Cas3UpdateBedSpaceStartDateJob.kt [hunk 5]
```diff
@@ -0,0 +1,65 @@
...
+  override fun process(pageSize: Int) {
+    var page = 0
+    var hasNext = true
+    var slice: Slice<BedEntity>
+    var currentSliceBeds = setOf<String>()
+
+    try {
+      while (hasNext) {
+        migrationLogger.info("Getting page $page for max page size $pageSize")
+        slice = bedRepository.findAllByStartDateIsNull(PageRequest.of(page, pageSize))
```
davidatkinsuk, line 35: This should be limited to CAS3 beds (where premises are CAS3)

muhammad-elabdulla: Can you filter only CAS3 bedspaces

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/migration/cas3/Cas3UpdateBedSpaceStartDateJob.kt
```diff
@@ -0,0 +1,73 @@
...
+  override fun process(pageSize: Int) {
+    var page = 0
+    var hasNext = true
+    var slice: Slice<BedEntity>
+    var currentSliceBeds = setOf<String>()
+
+    try {
+      while (hasNext) {
+        migrationLogger.info("Getting page $page for max page size $pageSize")
+        slice = bedRepository.findAll(PageRequest.of(page, pageSize))
```
muhammad-elabdulla, line 35: You need to get only CAS3 bedspaces. If you create a function in BedRepository to findAllByService and pass the service as a parameter to the function

```
select *
from beds b
inner join rooms r on b.room_id = r.id
inner join premises p on r.premises_id = p.id
where p.service = 'temporary-accommodation'
```
instead of findAll

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/migration/cas3/Cas3UpdateBedSpaceStartDateJob.kt
```diff
@@ -0,0 +1,73 @@
...
+
+        currentSliceBeds = slice.map { it.id.toString() }.toSet()
+
+        for (bed in slice.content) {
+          if (bed.startDate == null) {
+          migrationLogger.info("Updating bed start_date for bed id ${bed.id}")
+            if (bed.createdAt != null) {
+              migrationLogger.info("Using created_date as start_date for bed id ${bed.id}")
+              val updatedBed = bed.copy(startDate = bed.createdAt!!.toLocalDate()!!).apply {}
```
muhammad-elabdulla, line 44: do you need .apply {} ??

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/migration/cas3/Cas3UpdateBedSpaceStartDateJob.kt
```diff
@@ -0,0 +1,73 @@
...
+              bedRepository.save(updatedBed)
+            } else {
+              migrationLogger.info("Using oldest booking arrival_date as start_date for bed id ${bed.id}")
+              val oldestBooking = bookingRepository.findByBedId(bed.id, Sort.by(ASC, "arrivalDate"), Limit.of(1))
+              if (oldestBooking.isEmpty()) {
+                migrationLogger.info("No bookings found for bed id ${bed.id}")
+              } else {
+                if ("temporary-accommodation".equals(oldestBooking[0].service)) {
```
muhammad-elabdulla, line 52: Don't need to check by service as CAS3 not share bedspaces with other CASs

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/migration/cas3/Cas3UpdateBedSpaceStartDateJob.kt
```diff
@@ -0,0 +1,73 @@
...
+                  val updatedBed = bed.copy(startDate = oldestBooking[0].arrivalDate)
+                  bedRepository.save(updatedBed)
+                  migrationLogger.info("Updated start_date for bed id ${bed.id} from booking arrival_date")
+                } else {
+                  migrationLogger.info("skipping non-cas3 booking for bed id ${bed.id}")
+                }
+              }
+            }
+          } else {
```
muhammad-elabdulla, line 61: don't need this else statment

src/main/resources/static/_shared.yml
```diff
@@ -3360,6 +3360,7 @@ components:
         - update_cas1_backfill_user_ap_area
         - update_cas3_application_offender_name
         - update_cas3_booking_offender_name
+        - update_cas3_bed_space_start_date
```
muhammad-elabdulla, line 3363: can be changed to `update_cas3_bedspace_start_date`

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/migration/cas3/Cas3UpdateBedSpaceStartDateJobTest.kt
```diff
@@ -0,0 +1,132 @@
...
+import java.time.LocalDate
+import java.time.OffsetDateTime
+
+class Cas3UpdateBedSpaceStartDateJobTest : MigrationJobTestBase() {
+
+  @Test
+  fun `beds with null startDate are updated using createdAt date`() {
+    givenAUser { user, jwt ->
+      val probationRegion = givenAProbationRegion()
+      val premises = givenAnApprovedPremises(region = probationRegion)
```
muhammad-elabdulla, line 21: Can you change this to Premises Temporary Accommodation

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/migration/cas3/Cas3UpdateBedSpaceStartDateJobTest.kt
```diff
@@ -0,0 +1,132 @@
...
+
+      val createdAt = OffsetDateTime.now()
+      val beds = (1..5).map {
+        bedEntityFactory.produceAndPersist {
+          withRoom(room)
+          withName("Bed $it")
+          withStartDate(null)
+        }
+      }
+
```
muhammad-elabdulla, line 36: Can you add random createAt dates to bedspaces 

      beds.map { 
        it.createdAt = OffsetDateTime.now().randomDateTimeBefore(360)
        bedRepository.save(it)
      }

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/migration/cas3/Cas3UpdateBedSpaceStartDateJob.kt
```diff
@@ -0,0 +1,73 @@
...
+          migrationLogger.info("Updating bed start_date for bed id ${bed.id}")
+            if (bed.createdAt != null) {
+              migrationLogger.info("Using created_date as start_date for bed id ${bed.id}")
+              val updatedBed = bed.copy(startDate = bed.createdAt!!.toLocalDate()!!).apply {}
+              bedRepository.save(updatedBed)
+            } else {
+              migrationLogger.info("Using oldest booking arrival_date as start_date for bed id ${bed.id}")
+              val oldestBooking = bookingRepository.findByBedId(bed.id, Sort.by(ASC, "arrivalDate"), Limit.of(1))
+              if (oldestBooking.isEmpty()) {
+                migrationLogger.info("No bookings found for bed id ${bed.id}")
```
muhammad-elabdulla, line 50: Can you add a test to cover this situation

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/migration/cas3/Cas3UpdateBedSpaceStartDateJobTest.kt
```diff
@@ -0,0 +1,132 @@
...
+      val oldestArrivalDate = LocalDate.now().minusDays(30)
+      val newerArrivalDate = LocalDate.now().minusDays(15)
+
+      beds.forEachIndexed { index, bed ->
+        bookingEntityFactory.produceAndPersist {
+          withPremises(premises)
+          withServiceName(temporaryAccommodation)
+          withBed(bed)
+          withArrivalDate(oldestArrivalDate)
```
muhammad-elabdulla, line 78: Can be random date for the arrival date for each bedspace

muhammad-elabdulla (approved):
LGTM I left some comments

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BedEntity.kt
```diff
@@ -55,6 +58,18 @@ interface BedRepository : JpaRepository<BedEntity, UUID> {
...
+    """
+        SELECT b.*
+        FROM beds b
+        INNER JOIN rooms r ON b.room_id = r.id
+        INNER JOIN premises p ON r.premises_id = p.id
+        WHERE p.service = :service
+    """,
+    nativeQuery = true,
+  )
+  fun findAllByService(service: String, pageable: Pageable): Slice<BedEntity>
```
davidatkinsuk, line 71: i'm slightly wary of using a native query but then loading the results into an Entity. I don't think this will be leveraging JPA mapping information so if a column isn't mapped correctly (e.g. value is null when it shouldn't be) and you persist back to the database you may inadvertently remove values. But if you're happy this isn't happening, it works for me given this is just temporary code.

muhammad-elabdulla: This is used in many places in the project and as I know we don't had issues with this approach

rashadmughal:
This query will let you know if the `start_date` and `created_at` dates are the same or not:

```
SELECT
    b.start_date,
    b.created_at,
    CASE
      WHEN (b.start_date != b.created_at::date) THEN 'NOT_SAME'
      ELSE 'SAME'
    END AS pvc
FROM beds b
INNER JOIN rooms r ON b.room_id = r.id
INNER JOIN premises p ON r.premises_id = p.id
WHERE p.service = 'temporary-accommodation';
```

PR #3685, opened 2025-06-18
Tidy up UI type generation workflows

froddd:
This splits the workflows so changes to the relevant files for each UI are taken into account. The workflows are renamed so they can be differentiated easily in the Github Actions UI.


davidatkinsuk (approved):
LGTM 👍

PR #3720, opened 2025-06-23
CAS-1669: Changed SQL order by for bedspace migration job

rashadmughal:
Order by bed.id to ensure the paging is predictable.


muhammad-elabdulla (approved):
LGTM

PR #3728, opened 2025-06-24
Add `ensureUserCanAccessOffender` func

davidatkinsuk:
This helps reduce usage of deprecated offender service functions, and removes duplication of this logic


muhammad-elabdulla (approved):
LGTM

PR #3731, opened 2025-06-24
fix: switches the annotaion for order by in cas2v2 application entity

tobybatchmoj:
No Ticket.

This fixes the the SQL order by in the application entity to match the cas2 application

All tests pass


davidatkinsuk (approved):
Looking at https://docs.jboss.org/hibernate/orm/6.6/javadocs/org/hibernate/annotations/SQLOrder.html i'd expect this to require `created_at` instead of `createdAt`, but i can see you've just followed the usage already established on CAS2 so presumably `createdAt` is fine. LGTM 👍

PR #3734, opened 2025-06-24
fix: dates where an hour out in status update reports

tobybatchmoj:
CAS2v2 Only

https://dsdmoj.atlassian.net/browse/CBA-689

This fixes the timezone on status updates being an hour out respective of the the other reports.


davidatkinsuk (approved):
LGTM 👍

PR #3742, opened 2025-06-24
CAS-1273: added job to trigger workflow on deploy to dev

daveawc:
CAS-1273: This changes how the generate-types job is called in the cas3 api repo - currently it's triggered when branch is merged to main, however this moves it to run after a successful deploy to dev and after the e2e tests have ran. It also adds some filtering in to exclude cas2v2 paths from the generated swagger docs, and also adds a model into the dummy controller that is needed by the UI.

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -41,6 +41,12 @@ jobs:
     uses: ./.github/workflows/e2e.yml
     secrets: inherit
 
+  cas3_generate_types:
+    needs: e2e
+    name: Run the generate types job for the CAS3 UI
+    uses: ./.github/workflows/generate-cas3-ui-types.yml@main
+    secrets: inherit
+
   deploy_preprod:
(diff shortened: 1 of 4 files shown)


.github/workflows/pipeline.yml
```diff
@@ -34,6 +34,12 @@ jobs:
       environment: 'development'
       app_version: '${{ needs.docker_build.outputs.app_version }}'
 
+  cas3_generate_types:
+    name: Run the generate types job for the CAS3 UI
+    needs: deploy_dev
+    uses: ministryofjustice/hmpps-temporary-accommodation-ui/.github/workflows/generate-types.yml@main
+    secrets: inherit
+
```
froddd, line 42: Move this after e2e to ensure endpoint is available, and ensure `needs` is set to `e2e`

davidatkinsuk (approved):
LGTM 👍

PR #3743, opened 2025-06-24
Simplify Task integration tests

danhumphreys-moj:
Considering the integration tests required to test task sorting in `TaskTests.GetTasksTest.SortByTest` the following was noted:

- changing the sort direction from descending to ascending in `Get all tasks sorts by createdAt in descending order` [here](https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/450565e5a3fd67758f1ba7f315f99989adee220a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/TasksTest.kt#L1591) resulted in the tests still passing
- changing the sort direction from ascending (default) to descending (`SortDirection.desc`) in [here](https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/450565e5a3fd67758f1ba7f315f99989adee220a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/TasksTest.kt#L1609) resulted in the tests still passing
- presumably there are other instances where the tests are passing but don't appear to be testing the right conditions
- test logic is complicated with numerous private functions dedicated to sorting lists and a `sortTasks()` function

To make the tests more readable and understandable and to check they are working, this PR:

- removes the private functions and `sortTasks()` function
- returns responses from the API calls as `bodyAsListOfObjects<Task>`
- explicitly asserts that the responses contain the expected sorted values
- adds required `page` and `perPage` parameters to the URL strings (otherwise pageable returns null)
- sets `completed_at` to be `decision_made_at` (was `completed_at`) in the `PLACEMENT_APPLICATION_QUERY`.  `decision_made_at` is consistent with the value used in the `transformPlacementApplicationToTask()` function [here](https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/560f0b776e21957a39a15fd7c08fed74f79a5d9f/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/TaskTransformer.kt#L87).
- adds two private functions to reduce duplicate code in the created at and decision sorting tests

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/TaskEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/TaskEntity.kt
@@ -95,7 +95,7 @@ interface TaskRepository : JpaRepository<Task, UUID> {
         'PLACEMENT_APPLICATION' AS type,
         apa.name as person,
         u.name as allocated_to,
-        placement_application.submitted_at as completed_at,
+        placement_application.decision_made_at as completed_at,
         placement_application.decision as decision,
         pad.expected_arrival AS sorting_date,
         apa.ap_type AS ap_type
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/TaskEntity.kt
```diff
@@ -95,7 +95,7 @@ interface TaskRepository : JpaRepository<Task, UUID> {
         'PLACEMENT_APPLICATION' AS type,
         apa.name as person,
         u.name as allocated_to,
-        placement_application.submitted_at as completed_at,
+        placement_application.decision_made_at as completed_at,
```
davidatkinsuk, line 98: good spot! I guess this maybe isn't shown anywhere currently otherwise i'd expect it to have been raised by now!

PR #3747, opened 2025-06-25
Revert cas2v2 dev+test user+app seeding

davidatkinsuk:
The change made in 1d351f6139ed6480180344a2d5321055057e5da1 has led to breakages in cas2v2 E2Es. As a temporary measure to unblock the pipeline, this commit reverts changes in 1d351f6139ed6480180344a2d5321055057e5da1


danhumphreys-moj (approved):
Reverting shouldn't cause us any other problems and will unblock the pipeline and give us time to figure out what the issue is.

PR #3753, opened 2025-06-26
Replace space booking day summary

vbala-moj:
(https://dsdmoj.atlassian.net/browse/APS-2030)

**DO NOT MERGE**

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1SpaceBookingService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1SpaceBookingService.kt
@@ -421,7 +421,7 @@ class Cas1SpaceBookingService(
     bookingsSortDirection: SortDirection,
     excludeSpaceBookingId: UUID? = null,
     user: UserEntity,
-  ): CasResult<List<Cas1SpaceBookingSummary>> {
+  ): List<Cas1SpaceBookingSummary> {
     val sort = Sort.by(
       when (bookingsSortDirection) {
         SortDirection.desc -> Sort.Direction.DESC
@@ -448,9 +448,7 @@ class Cas1SpaceBookingService(
           personSummaryInfo = offenderSummaries.forCrn(bookingSummary.crn),
         )
       }
-    return Success(
-      spaceBookingSummaries,
-    )
+    return spaceBookingSummaries
   }
 
   private fun getBookingCharacteristicIds(bookingsCriteriaFilter: List<Cas1SpaceBookingCharacteristic>?) = bookingsCriteriaFilter?.let { bookingCriteria ->
--- a/src/main/resources/static/cas1-schemas.yml
+++ b/src/main/resources/static/cas1-schemas.yml
@@ -483,6 +483,7 @@ components:
         - spaceBookingSummaries
         - outOfServiceBeds
     Cas1SpaceBookingDaySummary:
+      deprecated: true
       type: object
       properties:
         id:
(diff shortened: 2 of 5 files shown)


src/main/resources/static/cas1-schemas.yml
```diff
@@ -462,9 +462,14 @@ components:
         capacity:
           $ref: '#/components/schemas/Cas1PremiseCapacityForDay'
         spaceBookings:
+          deprecated: true
```
davidatkinsuk, line 465: Could we also deprecate the type `Cas1SpaceBookingDaySummary`?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1SpaceBookingService.kt
```diff
@@ -401,6 +413,57 @@ class Cas1SpaceBookingService(
...
 
+  fun getSpaceBookingsByPremisesIdAndCriteriaForDate(
+    premises: ApprovedPremisesEntity,
+    date: LocalDate,
+    bookingsCriteriaFilter: List<Cas1SpaceBookingCharacteristic>?,
+    bookingsSortBy: Cas1SpaceBookingDaySummarySortField,
+    bookingsSortDirection: SortDirection,
+    excludeSpaceBookingId: UUID? = null,
+    user: UserEntity,
+  ): CasResult<List<Cas1SpaceBookingSummary>> {
```
davidatkinsuk, line 424: I don't think the response needs to be wrapped in a `CasResult`? (i think it's always success?)

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/Cas1PremisesTest.kt
```diff
@@ -785,6 +785,16 @@ class Cas1PremisesTest : IntegrationTestBase() {
...
+      assertThat(summaries.spaceBookingSummaries).hasSize(3)
+
+      summaries.spaceBookings.zip(summaries.spaceBookingSummaries).forEach { (dayBooking, bookingSummary) ->
+        assertThat(dayBooking.id).isEqualTo(bookingSummary.id)
+        assertThat(dayBooking.person.crn).isEqualTo(bookingSummary.person.crn)
+        assertThat(dayBooking.essentialCharacteristics).isEqualTo(bookingSummary.characteristics)
+        assertThat(dayBooking.tier).isEqualTo(bookingSummary.tier)
+        assertThat(dayBooking.canonicalArrivalDate).isEqualTo(bookingSummary.canonicalArrivalDate)
+        assertThat(dayBooking.canonicalDepartureDate).isEqualTo(bookingSummary.canonicalDepartureDate)
+      }
```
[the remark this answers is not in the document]
davidatkinsuk: Could we please do this as an additional commit on this PR? (e.g. do all assertions on both sets of space booking summaries?). i'd like to try and minimise how long we're returning both lists in prod because it's a bit inefficient (we're doing the space booking search and retrieving the offender list plus permissions checks twice, which is definitely the most pragmatic way to make this change of course).

davidatkinsuk (commented):
A few minor comments from me.

As for merging ... because this change makes the endpoint temporarily inefficient, i propose we ask the UI to prepare a change against a branch (if possible) and only merge this PR once UI are in a position to quickly release their change to prod, at which point we'll follow up with a commit to remove the old property (which we can prepare in advance)

davidatkinsuk (changes requested):
Looks great. Marking as 'request changes' as requested, as we don't want this merging yet

davidatkinsuk (approved):
LGTM ,merge at will

PR #3767, opened 2025-06-27
CAS-1809 Downgrade spring and jpa plugin versions

muhammad-elabdulla:
This [PR CAS-1809](https://dsdmoj.atlassian.net/browse/CAS-1809) is to downgrade spring and jpa plugin versions as many developers had issues in Intellij complaining about Unresolved reference


davidatkinsuk (approved):
LGTM 👍

PR #3771, opened 2025-06-27
update generate types job

daveawc:
CAS-1807 / CAS-1808
This moves the generate-types jobs to run after successful E2E runs. This is pre-work for switching over to using the open API spec files from the dev environment.

The trade off is that this job runs one each successful completion of the E2Es, rather than only if the build-spec files were changed.


davidatkinsuk (approved):
LGTM 👍

PR #3773, opened 2025-06-27
APS-1280 Don't return unsubmitted applications in the timeline

danhumphreys-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas1/Cas1PeopleController.kt
```diff
@@ -70,8 +70,9 @@ class Cas1PeopleController(
     return cas1PersonalTimelineTransformer.transformApplicationTimelineModels(personInfo, applicationTimelineModels)
   }
 
-  private fun getRegularApplications(crn: String) = cas1ApplicationService
+  private fun getRegularSubmittedApplications(crn: String) = cas1ApplicationService
```
danhumphreys-moj, line 73: Couldn't see which tests I needed to update to assert this is working as expected so added test coverage to the ticket description to evidence this change.

In my opinion these private functions should be moved to a service where they can be properly unit tested:
- `buildPersonInfoWithTimeline()`
- `getRegularSubmittedApplications()`
- `getOfflineApplications()`

Happy to do this, but will do outside of this ticket as a chore PR if required.

davidatkinsuk: I think the integration test `Cas1PersonalTimelineTest` is testing this. We should really rename to `Cas1PeopleTest` IMO as that aligns with the controller name.

If you update the existing test to ensure unsubmitted apps aren't returned i think that's sufficient. I agree ideally we'd unit test here but I think the integration test coverage good enough for now as (I think) it checks all fields returned, so i'd suggest unit testing is low priority.

davidatkinsuk: (I think we can update `Getting a personal timeline for a CRN returns OK with correct body` to add an additional application for the CRN that's unsubmitted. the existing assertion should prove it's not returned)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas1/Cas1PeopleController.kt
```diff
@@ -70,8 +70,9 @@ class Cas1PeopleController(
...
     .getApplicationsForCrn(crn, limit = TIMELINE_APPLICATION_LIMIT)
+    .filter { it.submittedAt != null }
```
davidatkinsuk, line 75: maybe we should update the `getApplicationsForCrn` function to instead be `getSubmittedApplicationsForCrn`, as this is the only place using it?

not that it really matters, but by filtering the results from that function you're slightly undercutting the application limit being applied (but in reality no user will have the number of applications defined by that limit, it was only added in as a protection against load tests creating loads of applications and causing issues on the timeline :) )

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas1/Cas1PeopleController.kt
danhumphreys-moj: Have updated the tests that needed updating, so this should be covered off now.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas1/Cas1PeopleController.kt
danhumphreys-moj: Yes I have removed the `.filter` and changed the repository function from `findByCrn` to `findByCrnAndSubmittedAtIsNotNull` JPA query method.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/Cas1PersonalTimelineTest.kt
```diff
@@ -127,10 +128,18 @@ class Cas1PersonalTimelineTest : InitialiseDatabasePerClassTestBase() {
   fun `Getting a personal timeline for a CRN returns OK with correct body`() {
     givenAUser { userEntity, jwt ->
       givenAnOffender { offenderDetails, inmateDetails ->
-        givenAnApplication(userEntity, crn = offenderDetails.otherIds.crn) { application ->
+        givenAnApplication(
+          userEntity,
+          crn = offenderDetails.otherIds.crn,
```
davidatkinsuk, line 133: would you mind explicitly setting

`submittedAt = null` 

to make it clear it isn't submitted

danhumphreys-moj: Sure - have done this, good idea.  I did call this object `unsubmittedApplication` originally, but detekt complained that I wasn't using it and steered away from writing a comment.  This solves the problem.

davidatkinsuk (approved):
LGTM 👍

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/Cas1PersonalTimelineTest.kt
davidatkinsuk: Agreed, comments are an acceptable fall back IMO if it can't be made clear via a variable

PR #3776, opened 2025-06-30
moved cas3v2 premises controller in to repo.

daveawc:
This moves the Cas3V2 premises controller into the repo, rather than it being automatically generated.

The standard responses (200, 401, 403, 500) are automatically included, and don't need to be added indivi

image
image


src/main/resources/static/codegen/built-cas3v2-api-spec.yml
```diff
@@ -1,5391 +0,0 @@
...
-          type: boolean
-        dateStarted:
-          type: string
-          format: date-time
-        dateCompleted:
-          type: string
-          format: date-time
-      required:
-        - hasApplicableAssessment
-    Cas3Booking:
```
gregkhawkins, line 5100: This comes from `cas3-schemas.yml` which `cas3-api.yml` and `cas3v2-api.yml` currently use to geenrate the models.

_I think_ this will still get generated for us even though unused by `cas3-api.yml` but begs a broader question of how we are creating / maintaining these going forward? All the models generted below are specific to this new endpoint that you are refactoring. Should we copy the built models into a new `uk.gov.justice.digital.hmpps.approvedpremisesapi.model.cas` package and then delete our models from `cas3-schemas.yml`? Or is this a later task

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/swagger/SwaggerConfiguration.kt
```diff
@@ -87,6 +87,7 @@ class SwaggerConfiguration {
   fun cas3Shared(): GroupedOpenApi = GroupedOpenApi.builder()
     .group("CAS3Shared")
     .displayName("CAS3 & Shared")
+    .pathsToMatch("/**/cas3/**")
```
davidatkinsuk, line 90: doesn't this also exclude 'shared'?

PR #3782, opened 2025-06-30
CAS1703 - Create Premises Api endpoint

muhammad-elabdulla:
This [PR CAS-1703](https://dsdmoj.atlassian.net/browse/CAS-1703) is to create a Api endpoint to create a premises
Also it rename turnaroundWorkingDayCount to turnaroundWorkingDays

### DIFF
--- a/src/main/resources/static/cas3-api.yml
+++ b/src/main/resources/static/cas3-api.yml
@@ -236,6 +236,27 @@ paths:
                 type: array
                 items:
                   $ref: '_shared.yml#/components/schemas/ReferralHistoryNote'
+  /premises:
+    post:
+      tags:
+        - Premises
+      summary: Add a new premises
+      operationId: createPremises
+      requestBody:
+        content:
+          'application/json':
+            schema:
+              $ref: 'cas3-schemas.yml#/components/schemas/Cas3NewPremises'
+        required: true
+      responses:
+        201:
+          description: new premises added
+          content:
+            'application/json':
+              schema:
+                $ref: 'cas3-schemas.yml#/components/schemas/Cas3Premises'
+
+      x-codegen-request-body-name: body
   /premises/summary:
     get:
       tags:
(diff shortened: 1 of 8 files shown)


src/main/resources/static/cas3-api.yml
```diff
@@ -236,6 +236,27 @@ paths:
...
+            schema:
+              $ref: 'cas3-schemas.yml#/components/schemas/Cas3NewPremises'
+        required: true
+      responses:
+        201:
+          description: new premises added
+          content:
+            'application/json':
+              schema:
+                $ref: 'cas3-schemas.yml#/components/schemas/Cas3Premises'
```
besscerule, line 257: this does not reference the Cas3Premises outlined in the ticket - is this intentional?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas3/Cas3PremisesTest.kt
```diff
@@ -54,6 +56,133 @@ class Cas3PremisesTest : Cas3IntegrationTestBase() {
...
+        webTestClient.post()
+          .uri("/cas3/premises")
+          .header("Authorization", "Bearer $jwt")
+          .header("X-Service-Name", ServiceName.temporaryAccommodation.value)
+          .bodyValue(newPremises)
+          .exchange()
+          .expectStatus()
+          .isForbidden
+      }
+    }
```
besscerule, line 183: do we need a tests for currentUserCanManagePremises this too?

src/main/resources/static/cas3-api.yml
besscerule: ignore

muhammad-elabdulla: Its contain the totals as is also used by the get premises Api endpoint

PR #3800, opened 2025-07-02
Feature/cas 1815 commit cas 3 files into repo 2

daveawc:
This moves the CAS3 files into the repository, and removes the gradle tasks associated with them.

Moving the annotations the 'controllers' annotated with @Service will be in a following PR


davidatkinsuk (approved):
LGTM. Thanks for splitting it up into easy to follow commits 👍

PR #3801, opened 2025-07-02
CAS-1815 - refactoring CAS3 Controllers

daveawc:
This PR refactors the CAS3 controllers annotated with @Service to be actual controllers, and removes the generated files that were previously checked in.


davidatkinsuk (approved):
LGTM. Would mind squashing the last 4 commits into one? 👍

besscerule (commented):
At some point can you add the how-to docs to one of these prs to detail how to add an endpoint

PR #3820, opened 2025-07-03
bug/CAS-1626-CAS3-LAO-inconsistent-visibility

besscerule:
https://dsdmoj.atlassian.net/browse/CAS-1626

Bug was replicated in preprod - appears to be because we filter out the restricted offenders instead of switching the name to LAO.

Fix involves applying an already existing name switch function instead of filtering.

- [x] e2e tests run successfully


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/BookingSearchTest.kt
```diff
@@ -1059,15 +1102,20 @@ class BookingSearchTest : IntegrationTestBase() {
     },
   )
 
-  private fun getExpectedResponseWithoutPersonName(
+  private fun getExpectedResponse(
     expectedBookings: List<BookingEntity>,
     crn: String,
+    withPersonName: Boolean = false,
```
muhammad-elabdulla, line 1108: Can you pass the person name instead of boolean

PR #3821, opened 2025-07-03
Feature/cas 1816 move cas 2 classes into repo

daveawc:
This moves the CAS2 and CAS2 domain event controllers and models into the repo, and removes the gradle tasks and spec files. 
Cas2 domain events needed to be in this PR as it shared cas2 schema files, and considering the low risk and low number of files (cas 1/3 domain events have already moved across, and the cas3 code), it was easier to include in this PR.

The files containing auto-generated/"do not edit" comments will be removed in the next PR.


davidatkinsuk (commented):
LGTM but i feel like a CAS2 dev should approve this one :)

PR #3822, opened 2025-07-03
Feature/cas 1815 move models into cas3 package

daveawc:
The 4 commits are:
* formatting (as intellij formatted code when moving packages)
* moving the classes to the cas3 models package
* excluding this package as enums are in the incorrect format, so linting fails. These will require updating and using Enum.value() rather than relying on the name
* formatting of the files that changed due to the import changes.
* added suppressions for detekt issues (previously these were excluded)
* removed the built-spec-file that should no longer be used


davidatkinsuk (approved):
LGTM 👍 

Going forward, would it be possible to add some context in your commit messages so they make sense outside of the PR e.g. 'formatting' becomes 'formatting checked in generated cas3 code'. Something to consider 👍

PR #3830, opened 2025-07-07
Remove bi-direction space booking transfer link

davidatkinsuk:
Despite being marked as lazy, the non-owning side of the transferredFrom/transferredTo `OneToOne` relationship was triggering eager loads. This is a common problem, dicussed here - https://thorben-janssen.com/hibernate-tip-lazy-loading-one-to-one/

The simplest solution to this problem was to make the relationship uni-directional, and use a query to find the ‘transferredFrom’ value in the few cases it is required.

This change was specifically made to optimise calculating capacity across multiple premises, which could involve 100s of space bookings.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas1SpaceBookingEntity.kt
```diff
@@ -362,6 +362,18 @@ interface Cas1SpaceBookingRepository : JpaRepository<Cas1SpaceBookingEntity, UUI
     nativeQuery = true,
   )
   fun countActiveSpaceBookings(premisesId: UUID): Long
+
+  fun findByTransferredFrom(spaceBooking: Cas1SpaceBookingEntity): Cas1SpaceBookingEntity?
```
vbala-moj, line 366: Also, enabling the 2nd level cache here reduces the number of repeated database hits as the date doesn't change much. @QueryHint().

vbala-moj (approved):

PR #3835, opened 2025-07-07
CAS-1682: C3 BE Make bestpace start date not nullable - script fixes

rashadmughal:
This Pull Request modifies a database migration script to include an additional column, "start_date," for the beds table and provides updated data inserts with corresponding values for this new field.

Main Changes

- Added a new column, start_date, to the beds table insert statements.
- Updated all beds data insert statements to include a value for the start_date column, with data such as dates ranging from 01/06/2024 to 01/01/2025.


davidatkinsuk (approved):
LGTM 👍

PR #3837, opened 2025-07-07
CAS-1709: C3 BE Make start date in premises non nullable

rashadmughal:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PremisesEntity.kt
```diff
@@ -304,7 +304,7 @@ class TemporaryAccommodationPremisesEntity(
   rooms: MutableList<RoomEntity>,
   characteristics: MutableList<CharacteristicEntity>,
   status: PropertyStatus,
-  var startDate: LocalDate?,
+  var startDate: LocalDate,
```
gregkhawkins, line 307: what about the premises with start null dates in production? Or have we already backfilled?

gregkhawkins (commented):
Looks good - one question on there

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/PremisesEntity.kt
gregkhawkins: also what about `CAS1` approved premises as think they extend this entity too right? If we have backfilled, have they?

gregkhawkins (approved):

PR #3838, opened 2025-07-07
added an @PaginationHeaders annotation

daveawc:
This can be added to each endpoint that uses paginatoin. Springdoc cannot generate specs for headers as they are dynamically populated. A better solution might be to move to using a PaginatedResults wrapper with a PaginationData object, which spring doc will be able to generate the specs from.

It also fixes a number of copy and paste errors in the specs file, which had the incorrect reference. 

Usage:
image

image

### DIFF
--- a/src/main/resources/static/api.yml
+++ b/src/main/resources/static/api.yml
@@ -1160,7 +1160,7 @@ paths:
             X-Pagination-TotalResults:
               $ref: '_shared.yml#/components/headers/X-Pagination-TotalResults'
             X-Pagination-PageSize:
-              $ref: '_shared.yml#/components/headers/X-Pagination-TotalResults'
+              $ref: '_shared.yml#/components/headers/X-Pagination-PageSize'
 
   /applications:
     post:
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/swagger/PaginationHeaders.kt
```diff
@@ -0,0 +1,51 @@
...
+@Target(AnnotationTarget.CLASS, AnnotationTarget.FUNCTION)
+@Retention(AnnotationRetention.RUNTIME)
+internal annotation class PaginationHeaders {
+    companion object {
+        const val CURRENT_PAGE = "X-Pagination-CurrentPage"
+        const val TOTAL_PAGES = "X-Pagination-TotalPages"
+        const val TOTAL_RESULTS = "X-Pagination-TotalResults"
+        const val PAGE_SIZE = "X-Pagination-PageSIze"
+    }
+}
```
davidatkinsuk, line 51: would this make more sense in the `controller` package (arguably we could maybe merge the swagger package into controller?). Or maybe have `controller.swagger`?

PR #3843, opened 2025-07-08
Refactor CAS2 generated files into actual controllers

daveawc:
This PR moves the annotations from the generated files onto the `@Service` annotated controllers, and condenses them to the correct request mapping type. It also adds an `@Cas2Controller` annotation, to match the @Cas1Controller one.


davidatkinsuk (approved):
LGTM 👍

PR #3847, opened 2025-07-08
APS-2466 associate application id with emails

danhumphreys-moj:
Ref: https://dsdmoj.atlassian.net/browse/APS-2466

Update `EmailNotificationService.sendEmail` and `EmailNotificationService.sendEmails` to take in an optional ‘reference: String?’ value. Default it to null. This should then be passed to `guestListNotificationClient!!.sendEmail` and `normalNotificationClient!!.sendEmail` functions as reference.

Then update `Cas1EmailNotificationService.sendEmail` and `Cas1EmailNotificationService.sendEmails` to populate the reference using `application.id.toString()`


davidatkinsuk (approved):
LGTM 👍

PR #3854, opened 2025-07-09
Chore/aps 2464 move cas1 generated files in to repo

daveawc:
This moves the generated controllers in to the repo, and removes them from the spec files. The models will follow in an additional PR, as well as the removal of the spec files and the gradle task


davidatkinsuk (approved):
LGTM 👍

PR #3855, opened 2025-07-09
CAS-1840 move shared generated controllers in to repo

daveawc:
This moves the shared controllers into the repo and removes them from the api spec.

It also adds in some PaginationHeaders annotations, which is currently breaking contract tests with the UI. 

Shared models will follow in a future PR.


besscerule (approved):
LGTM

PR #3860, opened 2025-07-09
Removed schema version from Cas2v2ApplicationEntity.

vbala-moj:
This PR removes the schema version from the CAS2V2 application data models and the entity but leaves them in the database to ensure there are no errors during deploy. A subsequent commit will then remove the columns and also the corresponding schema entry along with the related files from the code-base.


davidatkinsuk (commented):
LGTM but i'll let someone from CAS2 approve 👍

PR #3882, opened 2025-07-11
add @Cas3Controller annotation

daveawc:
this adds the annotation to save duplicating boilerplate code, as well as tidying up the annotations in the cas3 controllers.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas3/Cas3DomainEventsController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas3/Cas3DomainEventsController.kt
@@ -1,9 +1,9 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.controller.cas3
 
 import org.springframework.http.ResponseEntity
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas3/Cas3DomainEventsController.kt
```diff
@@ -18,113 +17,77 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.NotFoundProblem
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas3.Cas3DomainEventService
 import java.util.UUID
 
-@RestController
-@RequestMapping("\${api.base-path:}")
+@Cas3Controller
```
davidatkinsuk, line 20: Given the path for this _isn't_ /cas3, i wonder it's better to not use the @Cas3Controller here to avoid any confusion (I wasn't sure which @RequestMapping would be used in this case, but we have tests so it must be using the overridden one.

davidatkinsuk (approved):
one comment from me 👍

PR #3906, opened 2025-07-16
updating the reports controller

daveawc:
A previous change added the service name header to the spec file, which is not sent from the UI, and causes a failure in the contract tests.


davidatkinsuk (approved):
LGTM 👍

PR #3916, opened 2025-07-17
APS-2514 - Remove API to create legacy bookings

davidatkinsuk:
This logic is no longer required as legacy bookings have been fully replaced by space bookings


davidatkinsuk:
> Looks good. (I noticed there's still reference's to legacybookings throughout the code, but I guess there's a plan for them)

Yep, we have an epic to chip away at all that 👍

PR #3921, opened 2025-07-17
tidied up cas2application model.

daveawc:
this removes the auto-generated annotations, and also removes Cas2User field, which doesn't appear to be used anywhere in code. This also referenced entities, which was bringing in all of the nested entities into the UI. If needed in the future, there should be a model created instead.


davidatkinsuk (approved):
LGTM 👍

PR #3926, opened 2025-07-17
Add validation to Nomis User

daveawc:
This adds a null check on the primary email address field from the nomis user roles api call. This is a required field for sending emails within cas2, and raises errors in domain event processing, meaning messages go to the DLQ. This will prevent that from happening by ensuring the account always has an email address attached when the user logs in.


davidatkinsuk (approved):
Have agreed that it's acceptable for users with no email addresses to get a generic error as it's a tiny number of users and letting these users continue is causing lots of issues. 👍

PR #3934, opened 2025-07-18
Extract Alerts logic into standalone logic

davidatkinsuk:
This makes the overloaded OffenderService a little easier to deal with


vbala-moj (approved):
Just one minor comment.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PeopleController.kt
```diff
@@ -71,7 +73,7 @@ class PeopleController(
 
     val nomsNumber = offenderDetails.otherIds.nomsNumber
 
-    val prisonCaseNotesResult = offenderService.getFilteredPrisonCaseNotesByNomsNumber(
+    val prisonCaseNotesResult = caseNotesService.getFilteredPrisonCaseNotesByNomsNumber(
```
vbala-moj, line 76: Nice

PR #3944, opened 2025-07-21
Add local restriction to get premises

vbala-moj:
https://dsdmoj.atlassian.net/browse/APS-2519

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/api/model/Cas1Premises.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/api/model/Cas1Premises.kt
@@ -55,4 +55,10 @@ data class Cas1Premises(
 
   @Schema(example = "null", description = "")
   @get:JsonProperty("managerDetails") val managerDetails: kotlin.String? = null,
+
+  @Schema(
+    example = "No hate based offences",
+    description = "A list of restrictions that apply specifically to this approved premises.",
+  )
+  val localRestrictions: List<String>? = emptyList(),
 )
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas1/Cas1PremisesLocalRestrictionEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas1/Cas1PremisesLocalRestrictionEntity.kt
@@ -4,6 +4,7 @@ import jakarta.persistence.Entity
 import jakarta.persistence.Id
 import jakarta.persistence.Table
 import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.data.jpa.repository.Query
 import org.springframework.stereotype.Repository
 import java.time.OffsetDateTime
 import java.util.UUID
@@ -13,6 +14,17 @@ interface Cas1PremisesLocalRestrictionRepository : JpaRepository<Cas1PremisesLoc
   fun findAllByApprovedPremisesId(approvedPremisesId: UUID): List<Cas1PremisesLocalRestrictionEntity>
 
   fun findAllByApprovedPremisesIdAndArchivedFalseOrderByCreatedAtDesc(approvedPremisesId: UUID): List<Cas1PremisesLocalRestrictionEntity>
+
+  @Query(
+    """
+    SELECT r.description
+    FROM Cas1PremisesLocalRestrictionEntity r
+    WHERE r.approvedPremisesId = :premisesId
+    AND r.archived = false
+    order by r.createdAt DESC
+  """,
+  )
+  fun findAllActiveRestrictionDescriptionsByPremisesId(premisesId: UUID): List<String>
 }
 
 @Entity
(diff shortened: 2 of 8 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1PremisesService.kt
```diff
@@ -69,6 +69,7 @@ class Cas1PremisesService(
     val bedCount = premisesService.getBedCount(premise)
     val outOfServiceBedsCount = outOfServiceBedService.getCurrentOutOfServiceBedsCountForPremisesId(premisesId)
     val spaceBookingCount = spaceBookingRepository.countActiveSpaceBookings(premisesId).toInt()
+    val localRestrictions = cas1PremisesLocalRestrictionRepository.findAllActiveRestrictionDescriptionsByPremisesId(premisesId)
```
vbala-moj, line 72: Without refactoring this function we have to do this way. 😞

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/api/model/Cas1Premises.kt
```diff
@@ -55,4 +55,10 @@ data class Cas1Premises(
 
   @Schema(example = "null", description = "")
   @get:JsonProperty("managerDetails") val managerDetails: kotlin.String? = null,
+
+  @Schema(
+    example = "No hate based offences",
+    description = "A list of restrictions that apply specifically to this approved premises.",
+  )
+  @get:JsonProperty("localRestrictions") val localRestrictions: kotlin.collections.List<String>? = emptyList(),
```
davidatkinsuk, line 63: I think we can drop the `kotlin.collections` part

davidatkinsuk, line 63: We don't need to specify `@get:JsonProperty("localRestrictions")`, the variable name will be used by default (we really need to go through and remove all these at some point :) )

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas1/Cas1PremisesLocalRestrictionEntity.kt
```diff
@@ -13,6 +14,17 @@ interface Cas1PremisesLocalRestrictionRepository : JpaRepository<Cas1PremisesLoc
   fun findAllByApprovedPremisesId(approvedPremisesId: UUID): List<Cas1PremisesLocalRestrictionEntity>
 
   fun findAllByApprovedPremisesIdAndArchivedFalseOrderByCreatedAtDesc(approvedPremisesId: UUID): List<Cas1PremisesLocalRestrictionEntity>
+
+  @Query(
+    """
+    SELECT r.description
+    FROM Cas1PremisesLocalRestrictionEntity r
+    WHERE r.approvedPremisesId = :premisesId
```
davidatkinsuk, line 22: can we please add an index to `cas1_premises_local_restrictions.approved_premises_id` (although in reality i suspect the table will have so few entries postgres may just scan it)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1PremisesService.kt
davidatkinsuk: I think that's acceptable given this is (presumably) only called once by the UI when viewing information about a specific premises. Can always refactor if it's causing performance issues (also, there's load of other queries in here already :) )

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/Cas1PremisesTest.kt
```diff
@@ -213,6 +227,10 @@ class Cas1PremisesTest : IntegrationTestBase() {
       assertThat(summary.overbookingSummary).containsExactly(
         Cas1OverbookingRange(LocalDate.now(), LocalDate.now().plusWeeks(12).minusDays(1)),
       )
+      assertThat(summary.localRestrictions).containsExactly(
+        "No hate based offences",
+        "No offence against sex workers",
+      )
```
davidatkinsuk, line 233: cool

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/transformer/cas1/Cas1PremisesTransformerTest.kt
```diff
@@ -66,6 +66,7 @@ class Cas1PremisesTransformerTest {
           outOfServiceBeds = 2,
           availableBeds = 8,
           overbookingSummary = emptyList(),
+          localRestrictions = emptyList(),
```
davidatkinsuk, line 69: can we chuck a few entries in here please and assert on them

davidatkinsuk (approved):
Looks great, just a few minor comments from me

PR #3951, opened 2025-07-22
APS-2547 Create 'automatic' placement application on application approval

davidatkinsuk:
This PR changes what happens when application containing an implicit request for placement is assessed and approved.

Previously - Create an entry in `placement_requests` that is then treated as a 'standard' request for placement for reporting, withdrawal and request for placement API logic. Various workarounds are needed in the code to treat this `placement_request` in a particular way.

Now - Create an entry in `placement_applications` (with automatic = true) AND link it to a new entry in placement_requests. This means we no longer need to treat the placement_requests as as a 'standard' request for placement, and instead we can use the entry in `placement_applications`, which is consistent with all other requests for placements lodged against an application.

Because we still have existing data in production using the `placement_request` entry to represent the 'standard' request for placement we can't remove a lot of the aforementioned workarounds. This commit does allow us to remove _some_ duplicate domain event logic for new requests for placements. Once we've updated existing `placement_requests` in production to also have a corresponding `standard` placement application, we can remove the workarounds.

More detail is provided against each commit.


danhumphreys-moj (approved):
Reviewed with @davidatkinsuk on 25/07/25.

PR #3957, opened 2025-07-23
Feature/cas 1573 get void bedspaces

daveawc:
PR to add the v2 void bedspaces endpoint `cas3/v2/{premisesId}/void-bedspaces`. Functionality was based on the existing `premises/${premises.id}/lost-beds")` endpoint.

It also adds a cas3 user access controller and begins moving cas3 specific functions out.

[https://dsdmoj.atlassian.net/browse/CAS-1573](url)

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas3/Cas3UserAccessService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas3/Cas3UserAccessService.kt
@@ -1,0 +1,19 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas3
+
+import org.springframework.stereotype.Service
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserRole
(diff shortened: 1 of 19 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas3/Cas3UserAccessService.kt
```diff
@@ -0,0 +1,24 @@
...
+
+import org.springframework.stereotype.Service
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ServiceName
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserRole
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.UserAccessService
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.UserService
+import java.util.UUID
+
+@Service
+class Cas3UserAccessService(
```
davidatkinsuk, line 11: FYI i think CAS1 is using very little of the existing user access service. We could look at moving out the CAS1 stuff and renaming that CAS3UserAccessService. But i guess creating a fresh one for the new endpoints may be easier to manager / less effort

davidatkinsuk: makes sense

PR #3958, opened 2025-07-23
bug/CAS-1863-Cannot-submit-info-on-a-prisoner-that-has-moved-prisons

besscerule:


maccalsa (commented):
Looks good to me

maccalsa (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/service/Cas2AssessmentNoteService.kt
```diff
@@ -48,81 +44,67 @@ class Cas2AssessmentNoteService(
...
+      val user = userService.getUserForRequest()
+      if (userAccessService.offenderIsFromSamePrisonAsUser(assessment.application.currentPrisonCode, user.activeCaseloadId)) {
+        val savedNote = saveNote(assessment, note.note, user)
+        sendEmailToAssessors(savedNote)
+
+        return AuthorisableActionResult.Success(
+          ValidatableActionResult.Success(
+            savedNote,
+          ),
+        )
```
maccalsa, line 66: we are almost writing the same code here twice for different types of users. More of an observation

maccalsa (approved):
Only some code duplication, but it reads well, would be nice if we had less exit points from the following function, buts this is just me being picky

```kotlin
 fun createAssessmentNote(assessmentId: UUID, note: NewCas2ApplicationNote): AuthorisableActionResult<ValidatableActionResult<Cas2ApplicationNoteEntity>> {
```

PR #3962, opened 2025-07-24
APS-2480 Refactor tasks integration tests

danhumphreys-moj:
Ref: https://dsdmoj.atlassian.net/browse/APS-2480

So that we can test the Task permissions appropriately, this PR essentially refactors task integration tests for CAS 1.

- Created `TasksTest` (a copy of `Cas1TasksTest`).
- Remove `/tasks` related tests from `Cas1TasksTest`.
- Remove `/cas1/tasks` related test from `TasksTest`
- Remove `GET` `task` and `GET` `tasks` tests
- Removed `GET` `task` and `GET` `tasks` functionality from `TasksController`
- Remove legacy OpenAPI generated code from CAS 1.

Subsequent PR will isolate CAS 3 task related functionality from CAS 1:

- create a `Cas3AssessmentController` (for allocating/unallocating assessments)
- add functions to `Cas3AssessementService`
- make TasksTest `Cas3TasksTest`
- remove `TasksController` and legacy OpenAPI generated code.

Once all this is done we can push a PR which will test CAS 1 task permissions.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas1/Cas1TasksController.kt
```diff
@@ -195,12 +213,18 @@ class Cas1TasksController(
...
   @Transactional
-  override fun tasksTaskTypeIdAllocationsPost(
-    id: UUID,
-    taskType: String,
-    xServiceName: ServiceName,
-    body: NewReallocation?,
+  fun tasksTaskTypeIdAllocationsPost(
+    @PathVariable id: UUID,
+    @PathVariable taskType: String,
+    @RequestHeader("X-Service-Name") xServiceName: ServiceName,
```
davidatkinsuk, line 226: this shouldn't be needed, which means you can get rid of some CAS3 specific logic on line 234

davidatkinsuk (approved):
This is excellent work. Just one minor comment from me 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas1/Cas1TasksController.kt
danhumphreys-moj: I don't think I can do this yet.  CAS 3 still require `userId = user.id` since they are still using the reallocate task endpoint and CAS 1 controller methods.  I'll be able to do this once the CAS 1 controller methods CAS 3 are using are moved across to a CAS 3 controller.  I also see that I have left some CAS 3 tests in the CAS 1 tests (but they call `/cas1/tasks` which is wrong).  Looking at both these things now.

davidatkinsuk: My concern is we'll never remember to come back and remove it (of course this is a very minor piece of code so i don't think it matters too much, regardless...)

If there are no calls from CAS3 in the `Cas1TasksTest`, then presumably it's fine to change the code in the `Cas1TaskController` (and only in the `Cas1TasksController`)

```
    val userId = when {
      xServiceName == ServiceName.temporaryAccommodation -> user.id
      else -> {
        body?.userId ?: throw BadRequestProblem(invalidParams = mapOf("$.userId" to ParamDetails("empty")))
      }
    }
```

to

```
val userId = body?.userId ?: throw BadRequestProblem(invalidParams = mapOf("$.userId" to ParamDetails("empty")))
```

And then we no longer need the xServiceName header? 

CAS3 will be using the 'old' /tasks endpoint currently and in the future a /cas3 specific endpoint for this functionality, so this code should never be required in the `Cas1TasksController`?

PR #3964, opened 2025-07-24
test(bail): added document and data files for new noms numbers

tobybatchmoj:
https://dsdmoj.atlassian.net/browse/CBA-730

The seed data was injecting empty documents which was causing the ui to fail. While this could not happen in prod we need to to fix it for testing purposes.


davidatkinsuk (approved):
LGTM 👍

PR #3970, opened 2025-07-25
Add experimental permissions to CRU member

vbala-moj:
[https://dsdmoj.atlassian.net/browse/APS-2523](url)

[https://dsdmoj.atlassian.net/browse/APS-2473](url)

`These permissions are restricted to non-production environments, as experimental permissions are not allowed in production`


davidatkinsuk (approved):
LGTM 👍

PR #3981, opened 2025-07-28
APS-2586 - Add 'reallocate booking not mades' migration job

davidatkinsuk:
This migration job moves booking not made records from reallocated placement requests to their non reallocated equivalents.

This will affect a small number of placement requests as the ‘placement request allocation’ functionality has not been used for over 18 months.

This is a pre-requisite for deleting reallocated placement requests from the database, which is a pre-requisite to making the placement-request to placement-application relationships one-to-one.


davidatkinsuk:
> More of a comment to help me understand, but this is the logic here isn't it?
> 
> Get all the placement requests which have been reallocated and have booking not mades.
> 
> For each of these placement requests:
> 
> * Get the associated application and then from the application get all the placement requests associated with this application
> * Filter this list of all placement requests associated with an application where reallocated at is null and arrival date and duration are the same.  Effectively finding the placement request that hasn't been reallocated but may or may not have a booking not made.
> * Error if there are 0 non reallocated matching placement requests
> * Error if there are multiple non reallocated matching placement requests
> * Assumed both of these scenarios have been checked so we are confident there is only ever 1 matching non reallocated placement request.
> * Loop through the booking not made records for the matching non-reallocated placement request and set the corresponding placement request to the non re-allocated equivalent

That's correct.

Regarding this..

> Assumed both of these scenarios have been checked so we are confident there is only ever 1 matching non reallocated placement request.

I haven't checked this, but i'm reasonably confident it won't happen. I'll find out once we're in preprod and will have to tweak the job to work around it.

PR #3983, opened 2025-07-28
Optimize task performance and user data retrieval

vbala-moj:
[https://dsdmoj.atlassian.net/browse/APS-2539](url)

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas1/Cas1TasksController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas1/Cas1TasksController.kt
@@ -23,6 +23,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Task
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.TaskSortField
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.TaskType
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.TaskWrapper
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.UserWithWorkload
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.convert.EnumConverterFactory
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApprovedPremisesAssessmentEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.AssessmentEntity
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/cas1/Cas1TasksController.kt
```diff
@@ -127,67 +128,77 @@ class Cas1TasksController(
...
-        if (!placementApplication.isSubmitted()) {
-          throw NotFoundProblem(id, "Task")
-        }
+  private fun getTaskInfoByType(taskType: TaskType, id: UUID, user: UserEntity): TaskInfo = when (taskType) {
+    TaskType.assessment -> getAssessmentTaskInfo(id, user)
+    TaskType.placementApplication -> getPlacementApplicationTaskInfo(id, user)
+  }
 
-        val offenderSummaries = getOffenderSummariesForCrns(listOf(placementApplication.application.crn), user)
+  private fun getAssessmentTaskInfo(id: UUID, user: UserEntity): TaskInfo {
```
davidatkinsuk, line 150: going forward, if refactoring code for readability (which is great), could you please do it in a separate commit from whatever subsequent functional change you make, it makes it much easier to review the functional change bit (I think in this case nothing has functionally changed in this controller?)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/UserEntity.kt
```diff
@@ -38,6 +39,7 @@ interface UserRepository :
   @Query("SELECT DISTINCT u FROM UserEntity u join u.roles r where r.role = :role and u.isActive = true")
   fun findActiveUsersWithRole(role: UserRole): List<UserEntity>
 
+  @EntityGraph(attributePaths = ["qualifications"])
   @Query("SELECT DISTINCT u FROM UserEntity u join u.roles r where r.role in (:roles) and u.isActive = true")
   fun findActiveUsersWithAtLeastOneRole(roles: List<UserRole>): List<UserEntity>
```
vbala-moj, line 44: If the performance isn't being improved, we could switch to using a native query instead.

davidatkinsuk: You should be able to observe if lazy loads are still happening locally (or even via the integration tests i think). Would you mind verifying this? Ideally we wouldn't have any user-specific SQL running for the endpoint

davidatkinsuk (approved):
looks good, great to see roles being removed.

A few comments from me

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/UserEntity.kt
vbala-moj: We could add `probationRegion & cruManagementArea` to the entity graph as well.

PR #4025, opened 2025-07-31
CAS-1625 Extend the bad request response to include id and value

muhammad-elabdulla:
This [PR CAS-1625](https://dsdmoj.atlassian.net/browse/CAS-1625) is to extend the bad request response to include id and value

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/Cas3ValidationErrors.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/Cas3ValidationErrors.kt
@@ -6,6 +6,7 @@ value class Cas3ValidationErrors(private val errorMap: MutableMap<String, Cas3Va
 }
 
 data class Cas3ValidationMessage(
+  val entityId: String,
   val message: String,
   val value: String,
 )
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/BadRequestProblem.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/BadRequestProblem.kt
@@ -20,7 +20,8 @@ class BadRequestProblem(
         ParamError(
           propertyName = it.key,
           errorType = it.value.errorType,
-          errorDetail = it.value.errorDetail,
+          entityId = it.value.entityId,
+          value = it.value.value,
         )
       } ?: emptyList()
       ),
@@ -29,11 +30,13 @@ class BadRequestProblem(
 
 data class ParamDetails(
   val errorType: String,
-  val errorDetail: String? = null,
+  val entityId: String? = null,
+  val value: String? = null,
 )
 
 data class ParamError(
   val propertyName: String,
   val errorType: String,
-  val errorDetail: String? = null,
+  val entityId: String? = null,
+  val value: String? = null,
 )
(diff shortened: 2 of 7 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/BadRequestProblem.kt
```diff
@@ -29,11 +29,16 @@ class BadRequestProblem(
 
 data class ParamDetails(
   val errorType: String,
-  val errorDetail: String? = null,
+  val errorDetails: ErrorDetails?,
```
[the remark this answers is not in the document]
muhammad-elabdulla: I will put it back

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/model/Cas3ValidationErrors.kt
```diff
@@ -6,6 +6,7 @@ value class Cas3ValidationErrors(private val errorMap: MutableMap<String, Cas3Va
 }
 
 data class Cas3ValidationMessage(
+  val id: String,
```
davidatkinsuk, line 9: Would `entityId` be a better name? I would assume this was a 'message id' given the current name.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/BadRequestProblem.kt
```diff
@@ -29,11 +29,16 @@ class BadRequestProblem(
...
 
 data class ParamError(
   val propertyName: String,
   val errorType: String,
-  val errorDetail: String? = null,
+  val errorDetails: ErrorDetails?,
+)
+
+data class ErrorDetails(
+  val id: String,
```
davidatkinsuk, line 42: as per other comment, would `entityId` be a better name?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/BadRequestProblem.kt
```diff
@@ -29,11 +29,16 @@ class BadRequestProblem(
...
 )
 
 data class ParamError(
   val propertyName: String,
   val errorType: String,
-  val errorDetail: String? = null,
+  val errorDetails: ErrorDetails?,
+)
+
+data class ErrorDetails(
```
davidatkinsuk, line 41: `ParamDetails` - this is an internal type
`ParamError` - this is an API type
`ErrorDetails` - this appears to be an internal and API type

I find this is little confusing.

Do we actual need an `ErrorDetails` class which is adding another level to the API response? Could we not just add the 'id' parameter to `ParamDetails` and `ParamError`. This would then leave one type used internally (`ParamDetails`) and the other used in the API (`ErrorDetails`)?

davidatkinsuk (approved):
A few things from me to consider

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/BadRequestProblem.kt
davidatkinsuk: That makes sense. I still think add the new field to the main class as it then solves the 'class shared between internal/API problem too'. If it becomes too much we could maybe just have top level CAS1/CAS2/CAS3 versions of these errors instead

PR #4032, opened 2025-07-31
integration test fix.

daveawc:
This puts all PDU names to lowercase in the comparison, because of case sensitivity meaning lowercase would come before uppercase.


muhammad-elabdulla (approved):
LGTM

PR #4033, opened 2025-07-31
feature/CAS-1705-adding-reference-update

besscerule:


muhammad-elabdulla (approved):
Validation for unique reference added and fix issues when update PDU and turnaround working days

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/service/Cas3PremisesService.kt
```diff
@@ -376,7 +377,7 @@ class Cas3PremisesService(
     characteristicIds: List<UUID>,
     notes: String?,
     probationDeliveryUnitId: UUID,
-    turnaroundWorkingDayCount: Int,
+    turnaroundWorkingDays: Int,
```
gregkhawkins, line 380: how come we have decided to call the variable name different to the request-body field name. Agree that `turnaroundWorkingDays` is better and guess you do not want to break the contract but just wanted to make sure intentional

gregkhawkins (approved):

PR #4039, opened 2025-08-01
Update assessment annotations

daveawc:
This moves the annotations into the actual controller and removes the previously committed auto-generated files.

This is still split between CAS's, so the annotations are just copied and pasted, and not refactored (to @GetMapping etc). This can be in a future commit when the classes are split into cas specific ones.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/AssessmentController.kt
```diff
@@ -50,20 +58,29 @@ class AssessmentController(
...
+    @RequestParam(defaultValue = "asc") sortDirection: SortDirection,
+    @RequestParam(defaultValue = "arrivalDate") sortBy: AssessmentSortField,
+    @RequestParam statuses: List<AssessmentStatus>?,
+    @RequestParam crnOrName: String?,
+    @RequestParam page: Int?,
+    @RequestParam perPage: Int?,
   ): ResponseEntity<List<AssessmentSummary>> {
     val user = userService.getUserForRequest()
-    val resolvedSortDirection = sortDirection ?: SortDirection.asc
-    val resolvedSortBy = sortBy ?: AssessmentSortField.assessmentArrivalDate
```
davidatkinsuk, line 66: nice!

davidatkinsuk (approved):
LGTM 👍

PR #4040, opened 2025-08-01
CAS-1726 Create archive premises api endpoint

muhammad-elabdulla:
This [PR CAS-1726](https://dsdmoj.atlassian.net/browse/CAS-1726) is to create archive premises api endpoint

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/controller/Cas3PremisesController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/controller/Cas3PremisesController.kt
@@ -14,6 +14,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.BookingStatus
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.PropertyStatus
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ServiceName
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas3.model.Cas3ArchiveBedspace
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas3.model.Cas3ArchivePremises
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas3.model.Cas3PremisesBedspaceTotals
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/controller/Cas3PremisesController.kt
```diff
@@ -175,6 +176,43 @@ class Cas3PremisesController(
     return ResponseEntity.ok(result)
   }
 
+  @PostMapping("/premises/{premisesId}/archive")
```
rashadmughal, line 179: This requires @Transactional here too as you are passing premises.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/jpa/entity/Cas3VoidBedspaceEntity.kt
```diff
@@ -67,13 +67,25 @@ interface Cas3VoidBedspacesRepository : JpaRepository<Cas3VoidBedspaceEntity, UU
     """
     SELECT vb 
     FROM Cas3VoidBedspaceEntity vb
-    WHERE vb.bedspace.id = :bedspaceId
+    WHERE vb.bed.id = :bedspaceId
```
rashadmughal, line 70: Is this intentional?

rashadmughal (changes requested):

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/jpa/entity/Cas3VoidBedspaceEntity.kt
muhammad-elabdulla: Yes with the refactoring they create `bedspace` property but the current approach use `bed`

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/controller/Cas3PremisesController.kt
muhammad-elabdulla: Good point

PR #4046, opened 2025-08-01
map cas3 applications requests to cas3 applications controller

daveawc:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/api/model/Application.kt
```diff
@@ -1,37 +1,8 @@
...
-  val type: kotlin.String
-
-  @get:Schema(example = "null", requiredMode = Schema.RequiredMode.REQUIRED, description = "")
+  val type: String
   val id: java.util.UUID
-
-  @get:Schema(example = "null", requiredMode = Schema.RequiredMode.REQUIRED, description = "")
   val person: Person
-
-  @get:Schema(example = "null", requiredMode = Schema.RequiredMode.REQUIRED, description = "")
```
davidatkinsuk, line 35: nice

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/controller/Cas3ApplicationsController.kt
```diff
@@ -37,7 +39,8 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.extractEntityFromCa
 import java.net.URI
 import java.util.UUID
 
-@Cas3Controller
+@RestController
+@RequestMapping(headers = ["X-Service-Name=temporary-accommodation"])
```
davidatkinsuk, line 43: not for this PR, but i wonder if we should add this to all CAS-specific controllers (e.g. in the `Cas3Controller`) annotation? Would be a good proof that we've completely isolated APIs

davidatkinsuk (dismissed):
LGTM 👍

muhammad-elabdulla:
LGTM

davidatkinsuk:
@daveawc can you remember why we didn't proceed with this? I seem to recall we came up with some scenarios where this may cause issues? Maybe it was a logging problem (e.g. hard to tell what's going on when header isn't there, or something like that?)

PR #4049, opened 2025-08-04
APS-2594 Remove pending from PropertyStatus

danhumphreys-moj:
Ref: https://dsdmoj.atlassian.net/browse/APS-2594

Remove `pending` from `PropertyStatus`.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/api/model/PropertyStatus.kt
```diff
@@ -5,12 +5,11 @@ import com.fasterxml.jackson.annotation.JsonValue
 
 /**
 *
-* Values: pending,active,archived
+* Values: active,archived
 */
```
davidatkinsuk, line 9: going forward we can start removing this redundant generated javadoc completely

davidatkinsuk (approved):
LGTM 👍

PR #4050, opened 2025-08-04
APS-2594 Set pending property status to active.

danhumphreys-moj:
Ref: https://dsdmoj.atlassian.net/browse/APS-2594

Only one premises has a `pending` status which will be changed to `active` once this migration job has run.


davidatkinsuk (approved):
LGTM 👍

PR #4056, opened 2025-08-04
Feature/cas 1868 introduce arrival date validation

besscerule:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/service/v2/Cas3v2BookingService.kt
```diff
@@ -192,6 +194,12 @@ class Cas3v2BookingService(
       "$.expectedDepartureDate" hasValidationError "beforeBookingArrivalDate"
       return@validatedCasResult fieldValidationError
     }
+
```
besscerule, line 197: @gregkhawkins asked me to make the same changes to CAS3 v2

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/BookingTest.kt
```diff
@@ -1862,6 +1870,108 @@ class BookingTest : IntegrationTestBase() {
...
+          .expectBody()
+          .jsonPath("$.title").isEqualTo("Bad Request")
+          .jsonPath("$.invalid-params[0].propertyName").isEqualTo("\$.arrivalDate")
+          .jsonPath("$.invalid-params[0].errorType").isEqualTo("arrivalAfterLatestDate")
+      }
+    }
+  }
+
+  @Test
+  fun `Create Arrival returns field validation error when the arrival date is more than 14 days in the past CAS3`() {
```
besscerule, line 1925: tests the CAS3 route

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/BookingTest.kt
```diff
@@ -1862,6 +1870,108 @@ class BookingTest : IntegrationTestBase() {
     }
   }
 
+  @Test
```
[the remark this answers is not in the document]
besscerule: ignore above

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/PremisesController.kt
```diff
@@ -404,23 +403,6 @@ class PremisesController(
     }
 
     val result = when (body) {
-      is NewCas2Arrival -> {
```
besscerule, line 407: removed as this route is never hit due to validate above it

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/service/Cas3BookingService.kt
```diff
@@ -74,6 +75,7 @@ class Cas3BookingService(
   private val workingDayService: WorkingDayService,
   private val cas3DomainEventService: Cas3DomainEventService,
   private val featureFlagService: FeatureFlagService,
+  @Value("\${bookings.arrival-after-latest-date-limit-days}") private val arrivalAfterLatestDateLimitDays: Long,
```
muhammad-elabdulla, line 78: Do we need to add this in application resource or create a constraint inside the class?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/BookingTest.kt
```diff
@@ -1862,6 +1868,57 @@ class BookingTest : IntegrationTestBase() {
...
+          withDepartureDate(LocalDate.parse("2022-08-30"))
+          withCreatedAt(OffsetDateTime.parse("2022-07-01T12:34:56.789Z"))
+        }
+
+        val arrivalDate = LocalDate.now().minusDays(15)
+        val expectedDepartureDate = arrivalDate.plusDays(2)
+
+        webTestClient.post()
+          .uri("/premises/${booking.premises.id}/bookings/${booking.id}/arrivals")
+          .header("Authorization", "Bearer $jwt")
```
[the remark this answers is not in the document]
besscerule: done

besscerule:
> Can you move `premisesPremisesIdBookingsBookingIdArrivalsPost` from `PremisesController` to `Cas3PremisesController` may be you need to maintain both until UI switch to the new endpoint but it will be a good opportunity to move it as its only used by CAS3

That feels like quite a big chance for one ticket given we are going to be asking the FE to switch endpoints - do you think we should put it in another pr? @muhammad-elabdulla

muhammad-elabdulla:
Ok leave it 
Let me know when you merged your PR so I can do it after that

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/service/Cas3BookingService.kt
davidatkinsuk: IMO...

1. If the value will be the same in all environments, we should always keep it 'in code' as a constant using a companion object. Use @SuppressWarnings("MagicNumber") in this case. This has the added benefits of compile time checking of type, nullability etc.
2. If the value needs to differ between environments, I agree there should be a default in `applications.yml`, but it should be name-spaced to make it clear it's a cas3 thing. Our `applications.yml` is currently a message in this respect but i'd expect something like `cas3.bookings.arrival-after-latest-date-limit-days`

besscerule: so just so i'm clear - do i change the above and go for a magic number or keep it as is?

muhammad-elabdulla: change it to companion object
```
  companion object {
    const val ARRIVAL_AFTER_LATEST_DATE_LIMIT_DAYS: Long = 14
  }
```

PR #4061, opened 2025-08-05
APS-2620 Remove `PlacementApplicationEntity.duration`

davidatkinsuk:
This has been replaced by `requestedDuration` and `authorisedDuration`

It will be removed from the database in a subsequent commit


muhammad-elabdulla (approved):
LGTM

PR #4065, opened 2025-08-05
CAS-1731 Archive the premises when the last online bedspace has been archived

muhammad-elabdulla:
This [PR CAS-1731](https://dsdmoj.atlassian.net/browse/CAS-1731) is to archive the premises when the last online bedspace has been archived

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/service/Cas3PremisesService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/service/Cas3PremisesService.kt
@@ -555,18 +555,14 @@ class Cas3PremisesService(
     }
 
     // archive premises
-    premises.endDate = endDate
-    premises.status = PropertyStatus.archived
-    val updatedPremises = premisesRepository.save(premises)
+    val archivedPremises = archivePremisesAndSaveDomainEvent(premises, endDate)
 
     // archive all online bedspaces
     activeBedspaces.forEach { bedspace ->
       archiveBedspaceAndSaveDomainEvent(bedspace, endDate)
     }
 
-    cas3DomainEventService.savePremisesArchiveEvent(premises, endDate)
-
-    return success(updatedPremises)
+    return success(archivedPremises)
   }
 
   fun getBedspaceCount(premises: PremisesEntity): Int = premisesRepository.getBedCount(premises)
@@ -731,8 +727,10 @@ class Cas3PremisesService(
     return success(premisesRepository.save(premises))
   }
 
+  @Transactional
   fun archiveBedspace(
     bedspaceId: UUID,
+    premises: TemporaryAccommodationPremisesEntity,
     endDate: LocalDate,
   ): CasResult<BedEntity> = validatedCasResult {
     val bedspace = bedspaceRepository.findByIdOrNull(bedspaceId)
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/service/Cas3PremisesService.kt
```diff
@@ -731,8 +727,10 @@ class Cas3PremisesService(
     return success(premisesRepository.save(premises))
   }
 
+  @Transactional
```
rashadmughal, line 730: Do we need a transaction inside another transaction here?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/service/Cas3PremisesService.kt
```diff
@@ -1165,6 +1167,21 @@ class Cas3PremisesService(
     return null
   }
 
+  private fun isLastActiveBedspace(premisesId: UUID, endDate: LocalDate): Boolean {
+    val bedspaces = bedspaceRepository.findByRoomPremisesId(premisesId)
+
+    return (bedspaces.count { it.isCas3BedspaceArchived() } == bedspaces.count() && endDate <= LocalDate.now()) ||
+      (bedspaces.count { it.isCas3BedspaceArchived() } == bedspaces.count() - 1 && endDate > LocalDate.now())
+  }
```
davidatkinsuk, line 1175: I think the intention here is if the recently archived bedspace is ending in the future, we want to archive the premises on the same date,

What if there are multiple bedspaces that are ending in the future? In that case i'm not sure the premises will be archived? Maybe this could be tweaked to count bed spaces where an end date is set. If count() == bedspace.count(), then find the one with the latest end date and use that for the premises enddate.

You then no longer need to pass in the `endDate` to this function and the function can be called something like 'archivePremisesIfAllBedsArchived`?

rashadmughal (approved):
One comment.

davidatkinsuk (commented):
one comment from me about a potential logic trap (if i've understood it correctly)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/service/Cas3PremisesService.kt
davidatkinsuk: I guess you may need to flush the change to the bedspace for it to come back in the query at the start of this function

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/service/Cas3PremisesService.kt
muhammad-elabdulla: We are passing the premises to the `archiveBedspace` function

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/service/Cas3PremisesService.kt
muhammad-elabdulla: This is a good observation

PR #4066, opened 2025-08-05
APS-2547 - Add function to create placement app automatic

davidatkinsuk:
Rationale for this is described in the function’s javadoc


danhumphreys-moj (approved):
Thanks for adding the Javadoc explanation.

PR #4073, opened 2025-08-05
Return 204 when no results are found

vbala-moj:
test


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/integration/CAS2SubjectAccessRequestServiceTest.kt
```diff
@@ -78,15 +56,17 @@ class CAS2SubjectAccessRequestServiceTest : SubjectAccessRequestServiceTestBase(
...
     val expectedJson = """
    {
       "Applications": [${cas2ApplicationsJson(application)}],
-      "ApplicationNotes": [ ],
-      "Assessments": [ ],
-      "StatusUpdates": [ ],
-      "StatusUpdateDetails": [ ],
-      "DomainEvents":  [ ],
-      "DomainEventsMetadata": [ ]
+      "ApplicationNotes": null,
```
davidatkinsuk, line 64: i'm not sure we should be changing this. Could we still return an empty list for all these entries if they are null?

I guess could do something like the following when compiling the complete json?

```
"Applications": nullToEmptyList($approvedPremisesApplicationsJson),
```

davidatkinsuk (approved):
I've one comment, hopefully it's not too much of a headache to address.

Also - the exsisting tests are at the service level, how difficult would it be to add a single test calling the actual endpoint to ensure a 204 status code is returned if there's no backend data? There's an example of testing the endpoints here - https://github.com/ministryofjustice/hmpps-kotlin-lib/blob/da35df417cc60db48c3c77e81e0e2d2f484e114b/test-app/src/test/kotlin/uk/gov/justice/digital/hmpps/testapp/integration/resource/SubjectAccessRequestSampleIntegrationTest.kt

PR #4074, opened 2025-08-05
CAS-1735 Unarchive the premises when a bedspace back online

muhammad-elabdulla:
This [PR CAS-1735](https://dsdmoj.atlassian.net/browse/CAS-1735) is to unarchive the premises after it was archived as one of its bedspaces goes back online

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/service/Cas3PremisesService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/service/Cas3PremisesService.kt
@@ -877,11 +877,11 @@ class Cas3PremisesService(
   }
 
   fun unarchiveBedspace(
-    premises: PremisesEntity,
+    premises: TemporaryAccommodationPremisesEntity,
     bedspaceId: UUID,
     restartDate: LocalDate,
   ): CasResult<BedEntity> = validatedCasResult {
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/service/Cas3PremisesService.kt
```diff
@@ -906,20 +906,13 @@ class Cas3PremisesService(
...
-        startDate = restartDate,
-        endDate = null,
-      ),
-    )
+    val unarchivedBedspace = unarchiveBedspaceAndSaveDomainEvent(bedspace, restartDate)
 
-    cas3DomainEventService.saveBedspaceUnarchiveEvent(updatedBedspace, originalStartDate, originalEndDate)
+    if (premises.status == PropertyStatus.archived && premises.endDate != null && premises.endDate!! <= LocalDate.now()) {
+      unarchivePremisesAndSaveDomainEvent(premises, restartDate)
+    }
```
davidatkinsuk, line 913: I believe the logic for archiving a premises was 'if all beds have an end date, set the premises end date to the latest end date'

In this context, what happens here if i've previously archived the final bedspace with a future end date (meaning the premises also has a future end date) and i'm now un-archiving it using this function? I think the premises will remain archived? Is that correct?

also, regarding the premises archive logic - is it correct that the premises status would be 'archived' immediately even if the end date is in the future?

davidatkinsuk (approved):
LGTM 👍

PR #4079, opened 2025-08-06
Feature/cas 1883 prison bail implement search for prison number crn on view all prison bail applications page

besscerule:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2v2/jpa/entity/Cas2v2ApplicationSummarySpecifications.kt
```diff
@@ -12,13 +12,11 @@ object Cas2v2ApplicationSummarySpecifications {
...
   }
 
-  fun hasCrn(crn: String): Specification<Cas2v2ApplicationSummaryEntity> = Specification { root, _, criteriaBuilder ->
-    criteriaBuilder.equal(root.get<String>("crn"), crn)
-  }
-
-  fun hasOffenderPrisonNumber(nomsNumber: String): Specification<Cas2v2ApplicationSummaryEntity> = Specification { root, _, criteriaBuilder ->
-    criteriaBuilder.equal(root.get<String>("nomsNumber"), nomsNumber)
-  }
+  fun hasCrnOrNomsNumber(crnOrNomsNumber: String): Specification<Cas2v2ApplicationSummaryEntity> = Specification<Cas2v2ApplicationSummaryEntity> { root, _, criteriaBuilder ->
```
besscerule, line 15: combines the query builder so it is an OR as opposed to 2 ANDs

besscerule:
> 👍
> 
> what's the plan for switching the UI over? would it be better to add this as new functionality alongside the old values to maintain functionality, and then remove them once the UI is switched over?

@daveawc the ui doesn't use the crn of nomsNumber queries - so we can remove them

PR #4089, opened 2025-08-07
Remove use of deprecated offender func in CAs1ApplicationDomainEventService

davidatkinsuk:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/api/events/cas1/model/ApplicationSubmitted.kt
```diff
@@ -1,83 +1,50 @@
...
+  val releaseType: String,
 
-  @Schema(example = "43", required = true, description = "")
-  @get:JsonProperty("age", required = true) val age: kotlin.Int,
+  @Schema(example = "43")
+  val age: Int,
 
-  @Schema(example = "null", required = true, description = "")
-  @get:JsonProperty("gender", required = true) val gender: ApplicationSubmitted.Gender,
+  val gender: Gender?,
```
muhammad-elabdulla, line 25: I have a question this property is required before but accept null values?

davidatkinsuk: Good chat, hopefully i've addressed this in my commit message:

> AS part of this change the ‘gender’ value on the ApplicaitonSubmitted payload can be null, matching the response from probation-integration.

I’ve checked usage of this field and the only domain event consumer (probation-integration) do not use this field

PR #4098, opened 2025-08-08
Add expected_arrival_flexible column

vbala-moj:
Part 1 &2 for  [https://dsdmoj.atlassian.net/browse/APS-2631](url)

- Added expected_arrival_flexible to placement applications.
- Updated service to persist the same.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/api/model/SubmitPlacementApplication.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/api/model/SubmitPlacementApplication.kt
@@ -1,7 +1,7 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model
 
-import com.fasterxml.jackson.annotation.JsonProperty
 import io.swagger.v3.oas.annotations.media.Schema
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.cas1.Cas1RequestedPlacementPeriod
 
 /**
  *
@@ -11,12 +11,14 @@ import io.swagger.v3.oas.annotations.media.Schema
  */
 data class SubmitPlacementApplication(
 
-  @Schema(example = "null", required = true, description = "Any object")
-  @get:JsonProperty("translatedDocument", required = true) val translatedDocument: kotlin.Any,
+  @Schema(required = true)
+  val translatedDocument: Any,
 
-  @Schema(example = "null", required = true, description = "")
-  @get:JsonProperty("placementType", required = true) val placementType: PlacementType,
+  @Schema(required = true)
+  val placementType: PlacementType,
 
-  @Schema(example = "null", required = true, description = "")
-  @get:JsonProperty("placementDates", required = true) val placementDates: kotlin.collections.List<PlacementDates>,
+  @Schema(deprecated = true, description = "Please use requestedPlacementPeriods instead")
+  val placementDates: List<PlacementDates>?,
+
+  val requestedPlacementPeriods: List<Cas1RequestedPlacementPeriod>?,
 )
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1PlacementApplicationService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1PlacementApplicationService.kt
@@ -6,6 +6,7 @@ import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 import org.springframework.data.repository.findByIdOrNull
 import org.springframework.stereotype.Service
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.cas1.Cas1RequestedPlacementPeriod
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.PlacementApplicationDecisionEnvelope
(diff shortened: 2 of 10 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/api/model/SubmitPlacementApplication.kt
```diff
@@ -11,12 +11,13 @@ import io.swagger.v3.oas.annotations.media.Schema
...
-  @Schema(example = "null", required = true, description = "")
-  @get:JsonProperty("placementType", required = true) val placementType: PlacementType,
+  @Schema(required = true)
+  val placementType: PlacementType,
 
-  @Schema(example = "null", required = true, description = "")
-  @get:JsonProperty("placementDates", required = true) val placementDates: kotlin.collections.List<PlacementDates>,
+  val placementDates: List<PlacementDates>,
+
+  val requestedPlacementPeriods: List<Cas1RequestedPlacementPeriod>,
```
davidatkinsuk, line 22: I think both of these lists should be nullable.  i.e.

```
val placementDates: List<PlacementDates>?
val requestedPlacementPeriods: List<Cas1RequestedPlacementPeriod>?
```

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1PlacementApplicationService.kt
```diff
@@ -317,18 +317,14 @@ class Cas1PlacementApplicationService(
...
   ): CasResult<List<PlacementApplicationEntity>> {
     val placementApplicationAuthorisationResult = getApplicationForUpdateOrSubmit<List<PlacementApplicationEntity>>(id)
 
     if (placementApplicationAuthorisationResult is Either.Left) {
       return placementApplicationAuthorisationResult.value
     }
 
-    if (apiPlacementDates.isEmpty()) {
-      return CasResult.GeneralValidationError("At least one placement date is required")
-    }
```
davidatkinsuk, line 330: This is going to be required again once we remove the deprecated API type 'placementDates', so I propose we leave it in

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/PlacementApplicationsTest.kt
```diff
@@ -689,10 +724,11 @@ class PlacementApplicationsTest : IntegrationTestBase() {
           ) { _, _ ->
             govUKBankHolidaysAPIMockSuccessfullCallWithEmptyResponse()
 
-            val placementDates = listOf(
-              PlacementDates(
-                expectedArrival = LocalDate.of(2025, 3, 10),
+            val cas1RequestedPlacementPeriod = listOf(
```
davidatkinsuk, line 727: I think we need tests using both date collections as we'll be using the deprecated versions in production for a while yet. Could you please make copies of the following tests and update them to use the new collection? (retaining the original versions of the tests using the old collection)

```
submitting a placement application with multiple dates returns successfully and produces multiple placement apps
submitting a placement application with a single date returns successfully, sends emails, raises domain event and updates the application
```

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas1/Cas1PlacementApplicationServiceTest.kt
```diff
@@ -372,6 +381,7 @@ class Cas1PlacementApplicationServiceTest {
       val updatedPlacementApp3 = updatedPlacementApplications[2]
       assertThat(updatedPlacementApp3.expectedArrival).isEqualTo(LocalDate.of(2024, 6, 3))
       assertThat(updatedPlacementApp3.requestedDuration).isEqualTo(15)
+      assertThat(updatedPlacementApp3.expectedArrivalFlexible).isTrue
```
davidatkinsuk, line 384: nice

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/api/model/SubmitPlacementApplication.kt
```diff
@@ -11,12 +11,13 @@ import io.swagger.v3.oas.annotations.media.Schema
...
+  val translatedDocument: Any,
 
-  @Schema(example = "null", required = true, description = "")
-  @get:JsonProperty("placementType", required = true) val placementType: PlacementType,
+  @Schema(required = true)
+  val placementType: PlacementType,
 
-  @Schema(example = "null", required = true, description = "")
-  @get:JsonProperty("placementDates", required = true) val placementDates: kotlin.collections.List<PlacementDates>,
+  val placementDates: List<PlacementDates>,
```
davidatkinsuk, line 20: could we mention in the @Schema description that this is deprecated and to use `requestedPlacementPeriods`  instead? I think there may be a deprecated boolean attribute too that we can use.

davidatkinsuk (approved):
Looks great, just a few minor comments from me

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1PlacementApplicationService.kt
davidatkinsuk: It is for now but once we remove the deprecated 'placementDates' the check in the controller logic should probably be removed too (services should be validating where possible),  so I still think the service should be checking this as well?

PR #4109, opened 2025-08-11
Feature/aps 2658 optionally capture keyworker as user id part 2

davidatkinsuk:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserService.kt
```diff
@@ -60,6 +60,8 @@ class UserService(
 
   fun findByIdOrNull(id: UUID) = userRepository.findByIdOrNull(id)
 
+  fun findByStaffCode(staffCode: String) = userRepository.findByDeliusStaffCode(staffCode)
```
vbala-moj, line 63: > Not a big issue, but perhaps we should match the repository method name here for consistency
>

vbala-moj (approved):

PR #4111, opened 2025-08-11
APS-2028 Add count query to allForCas1Dashboard method.

danhumphreys-moj:
Ref: https://dsdmoj.atlassian.net/browse/APS-2028?focusedCommentId=643028

Fixes a bug whereby Spring Data was truncating the count query which resulted in a pagination SQL error.


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/Cas1PlacementRequestTest.kt
```diff
@@ -439,29 +439,21 @@ class Cas1PlacementRequestTest : IntegrationTestBase() {
...
-                  cas1PlacementRequestSummaryTransformer.transformPlacementRequestJpaToApi(
-                    placementRequest,
-                    PersonInfoResult.Success.Full(offenderDetails.otherIds.crn, offenderDetails, inmateDetails),
-                  ),
-                ),
-              ),
-            )
+            .bodyAsListOfObjects<Cas1PlacementRequestSummary>()
+
+          assertThat(response).hasSize(10)
```
danhumphreys-moj, line 456: No need to assert contents, since this is done during GET tests.

PR #4117, opened 2025-08-12
CAS-1726 Add validations to archive premises

muhammad-elabdulla:
This [PR CAS-1726](https://dsdmoj.atlassian.net/browse/CAS-1726) is to:
- Prevent archive a premises if the archiving date is before premises start date
- Prevent archive a premises when its clashes with an earlier archive premises end date


gregkhawkins (approved):
LGTM

PR #4122, opened 2025-08-12
CAS-1702 Change filtering by premises status

muhammad-elabdulla:
This [PR CAS-1702](https://dsdmoj.atlassian.net/browse/CAS-1702) is to change the way that filter premises by status


vbala-moj (approved):
LGTM

PR #4132, opened 2025-08-13
Feature/cas 1792 v2 voidbedspaces put

daveawc:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/api/model/UpdateLostBed.kt
```diff
@@ -1,30 +1,9 @@
...
-
-  @Schema(example = "null", required = true, description = "")
-  @get:JsonProperty("reason", required = true) val reason: java.util.UUID,
-
-  @Schema(example = "null", description = "")
-  @get:JsonProperty("referenceNumber") val referenceNumber: kotlin.String? = null,
-
-  @Schema(example = "null", description = "")
-  @get:JsonProperty("notes") val notes: kotlin.String? = null,
+  val startDate: java.time.LocalDate,
```
gregkhawkins, line 4: shall we just go with `val startDate: LocalDate` and the include `import java.time.LocalDate`? And same for other fields too

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/controller/Cas3v2VoidBedspaceController.kt
```diff
@@ -86,4 +88,63 @@ class Cas3v2VoidBedspaceController(
     val result = extractEntityFromCasResult(voidBedspace)
     return ResponseEntity.status(HttpStatus.CREATED).body(cas3VoidBedspacesTransformer.toCas3VoidBedspace(result))
   }
+
+  @PutMapping("/v2/premises/{premisesId}/bedspaces/{bedspaceId}/void-bedspaces/{voidBedspaceId}")
+  fun premisesPremisesIdLostBedsLostBedIdPut(
```
gregkhawkins, line 93: shall we call this `updateVoidBedspace()`?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/service/Cas3v2VoidBedspaceService.kt
```diff
@@ -71,4 +61,77 @@ class Cas3v2VoidBedspaceService(
...
+        this.referenceNumber = referenceNumber
+        this.notes = notes
+      },
+    )
+
+    return success(updatedVoidBedspaceEntity)
+  }
+}
+
+private fun CasResultValidatedScope<Cas3VoidBedspaceEntity>.validateVoidBedspaceDetails(
```
gregkhawkins, line 114: nice!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/service/Cas3v2VoidBedspaceService.kt
```diff
@@ -71,4 +61,77 @@ class Cas3v2VoidBedspaceService(
...
+  }
+  if (bedspace.startDate != null && voidBedspaceStartDate.isBefore(bedspace.startDate)) {
+    "$.startDate" hasValidationError "voidStartDateBeforeBedspaceStartDate"
+  }
+  if (bedspace.endDate != null && voidBedspaceEndDate.isAfter(bedspace.endDate)) {
+    "$.endDate" hasValidationError "voidEndDateAfterBedspaceEndDate"
+  }
+  if (reason == null) {
+    "$.reason" hasValidationError "doesNotExist"
+  }
```
gregkhawkins, line 134: assume this is largely lift & shift so probably do not need to worry but notice that the `updateVoidBedspace` call this function (as well as the create). So, what happens if someone tries to update a cancelled bedspace? Should we let them and potentially set the `cancellationDate` to null for this case? Or just have validation against this happening? 
Perhaps this is such and edge case we do not need to cater for it also :)
Finally, if this is `lift & shift`, then in the spirit of not creating more work for ourselves in the pursuit of perfection ;) we could always add a ticket to analyse/add-this-in later if we are worried about it

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/unit/service/v2/Cas3v2VoidBedspaceTest.kt
```diff
@@ -137,4 +137,132 @@ class Cas3v2VoidBedspaceTest {
...
+    fun `updateVoidBedspace returns FieldValidationError with correct param to message map when invalid parameters supplied`() {
+      every { cas3VoidBedspaceReasonRepository.findByIdOrNull(any()) } returns null
+
+      val result = cas3v2VoidBedspaceService.updateVoidBedspace(
+        voidBedspace,
+        voidBedspaceStartDate = LocalDate.now().plusDays(10),
+        voidBedspaceEndDate = LocalDate.now().plusDays(2),
+        voidBedspaceReason.id,
+        "new ref number",
+        "some new notes",
```
gregkhawkins, line 197: perhaps for these String values we should use `named arguments` for these Strings (even though it's pretty obvious what you are setting). For example:
```
val result = cas3v2VoidBedspaceService.updateVoidBedspace(
        voidBedspace,
        voidBedspaceStartDate = voidBedspaceStartDate,
        voidBedspaceEndDate = voidBedspaceEndDate,
        reasonId = voidBedspaceReason.id,
        referenceNumber = "new ref number",
        notes = "some new notes",
      )
```
I guess this is a style thing so feel free to ignore - the IDE does give you hints afterall. Just a suggestion!

gregkhawkins (approved):
LGTM - couple of comments

PR #4140, opened 2025-08-14
APS-2577 - Backfill automatic placement applications

davidatkinsuk:


danhumphreys-moj (approved):
LGTM - can we sanity check in pre-prod before deploying to prod?

davidatkinsuk:
> LGTM - can we sanity check in pre-prod before deploying to prod?

Of course, i've got a bit of testing planned

PR #4141, opened 2025-08-14
CAS-1625 Add validation to archive bedspace

muhammad-elabdulla:
This [PR CAS-1625](https://dsdmoj.atlassian.net/browse/CAS-1625) is to extend the validations to archive bedspace to include:
- when archive bedspace is before bedspace start date
- when archive date clashes with previous archive date


rashadmughal (approved):
LGTM

PR #4155, opened 2025-08-15
Change case notes endpoint from GET to POST

vbala-moj:
[https://dsdmoj.atlassian.net/browse/APS-2683](url)

A new role has been added to the case notes service clients, 'ROLE_PRISONER_CASE_NOTES__RO<unk>, to access it from all environments.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/client/BaseHMPPSClient.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/client/BaseHMPPSClient.kt
@@ -20,7 +20,7 @@ import java.util.concurrent.atomic.AtomicInteger
 
 abstract class BaseHMPPSClient(
   private val webClientConfig: WebClientConfig,
-  private val objectMapper: ObjectMapper,
+  internal val objectMapper: ObjectMapper,
   private val webClientCache: WebClientCache,
 ) {
 
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/client/prisonsapi/CaseNotes.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/client/prisonsapi/CaseNotes.kt
@@ -3,9 +3,8 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.client.prisonsapi
 import java.time.LocalDateTime
 
 data class CaseNotesPage(
-  val totalElements: Int,
-  val totalPages: Int,
-  val number: Int,
+  val hasCaseNotes: Boolean,
+  val metadata: PageMetaData,
   val content: List<CaseNote>,
 )
 
@@ -32,4 +31,22 @@ data class CaseNoteAmendment(
   val creationDateTime: LocalDateTime,
   val authorName: String,
   val additionalNoteText: String,
+)
+
+data class CaseNotesRequest(
+  val includeSensitive: Boolean,
+  val typeSubTypes: Set<TypeSubTypeRequest> = emptySet(),
+  val occurredFrom: LocalDateTime? = null,
+  val occurredTo: LocalDateTime? = null,
+  val page: Int,
+  val size: Int,
+  val sort: String,
+)
+
+data class TypeSubTypeRequest(val type: String, val subTypes: Set<String> = setOf())
+
+data class PageMetaData(
+  val page: Int,
+  val size: Int,
+  val totalElements: Int,
 )
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/CaseNotesService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/CaseNotesService.kt
@@ -6,11 +6,14 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.CaseNotesClient
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.ClientResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.prisonsapi.CaseNote
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.prisonsapi.CaseNotesPage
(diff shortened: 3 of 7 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/httpmocks/CaseNotesAPI.kt
```diff
@@ -1,16 +1,16 @@
...
-  val fromLocalDateTime = LocalDateTime.of(from, LocalTime.MIN)
+fun IntegrationTestBase.caseNotesAPIMockSuccessfulCaseNotesCall(personIdentifier: String, request: CaseNotesRequest, result: CaseNotesPage) {
+  val requestBodyString = objectMapper.writeValueAsString(request)
 
-  mockSuccessfulGetCallWithJsonResponse(
-    url = "/case-notes/$nomsNumber?startDate=$fromLocalDateTime&page=$page&size=30",
+  mockSuccessfulPostCallWithJsonResponse(
+    url = "/search/case-notes/$personIdentifier",
     responseBody = result,
+    requestBody = equalToJson(requestBodyString),
```
vbala-moj, line 14: Using this to convert the json string to stringValuePattern, but getting a 404 instead. Removing the request body completely from the mock results in all tests working fine. Let's have a chat to figure it out.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/CaseNotesTest.kt
```diff
@@ -113,14 +117,24 @@ class CaseNotesTest : IntegrationTestBase() {
...
           CaseNotesPage(
-            totalElements = 4,
-            totalPages = 1,
-            number = 1,
             content = caseNotes,
+            hasCaseNotes = true,
+            metadata = PageMetaData(
+              totalElements = 4,
+              size = 1,
+              page = 1,
```
davidatkinsuk, line 136: as per my other comment, i think this should be page = 0, which i'm pretty sure will trigger an infinite loop in the code as currently written?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/CaseNotesTest.kt
```diff
@@ -113,14 +117,24 @@ class CaseNotesTest : IntegrationTestBase() {
         )
 
         caseNotesAPIMockSuccessfulCaseNotesCall(
-          0,
-          LocalDate.now().minusDays(365),
           offenderDetails.otherIds.nomsNumber!!,
+          CaseNotesRequest(
+            includeSensitive = true,
+            occurredFrom = LocalDateTime.of(LocalDate.now().minusDays(365), LocalTime.MIN),
+            page = 0,
```
davidatkinsuk, line 124: documentation on the api says page should be >= 1 (i guess the old endpoint accepted >= 0?)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/CaseNotesService.kt
```diff
@@ -76,7 +76,7 @@ class CaseNotesService(
           }
         },
       )
-    } while (currentPage != null && currentPage.totalPages > currentPageIndex!! + 1)
+    } while (currentPage.metadata.page > currentPageIndex + 1)
```
davidatkinsuk, line 79: Looking at the API definition it says `page` provides "The current page number"

So I think this would result in an infinite loop?

The metadata does provide `totalElements` which is "The total number of results across all pages"

So I think we need to use `totalElements / pageSize` to figure out if there's more to process?

davidatkinsuk (commented):
I've left a few comments around paging. Could you please confirm that we're sufficiently testing what happens when multiple pages of data are available?

Once these comments are addressed I propose we deploy this branch to test and check we can retrieve case notes (preferably for a record with multiple pages of data - we could maybe force this by configuring our page size to be a low number in the test environment?)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/client/prisonsapi/CaseNotes.kt
```diff
@@ -33,3 +32,21 @@ data class CaseNoteAmendment(
...
 )
+
+data class CaseNotesRequest(
+  val includeSensitive: Boolean,
+  val typeSubTypes: Set<TypeSubTypeRequest> = emptySet(),
+  val occurredFrom: LocalDateTime? = null,
+  val occurredTo: LocalDateTime? = null,
+  val page: Int,
+  val size: Int,
+  val sort: String?,
```
davidatkinsuk, line 43: sort is mandatory in the API

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/client/BaseHMPPSClient.kt
```diff
@@ -54,6 +55,11 @@ abstract class BaseHMPPSClient(
 
     val cacheConfig = requestBuilder.preemptiveCacheConfig
 
+    // Ensure Content-Type is set when a body is present and no Content-Type header is provided
+    if (requestBuilder.body != null && !requestBuilder.headers.containsKey(HttpHeaders.CONTENT_TYPE)) {
+      requestBuilder.headers.set(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_JSON_VALUE)
+    }
```
davidatkinsuk, line 61: i think it may be preferable to define the headers in `CaseNotesClient` to ensure we don't inadvertently change headers on other requests being made e.g. add something like this in the call to `postRequest`?

```
headers = HttpHeaders().apply { set(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_JSON_VALUE) }
```

davidatkinsuk (commented):
Just one comment from me, looks great.

vbala-moj:
> Looks good, thanks for adding the integration test for paging. Holding approval until we've have results from testing in the `test` environment

Testing completed and the new end point is now working.

danhumphreys-moj:
Approving since this PR has already been reviewed and the new endpoints have been tested in the TEST environment.

PR #4156, opened 2025-08-15
CAS-1729 Should be able to archive/unarchive premises/bedspace in the same day

muhammad-elabdulla:
This PR CAS-1729 is to enable the user to archive/unarchive premises/bedspace in the same day


rashadmughal (approved):
LGTM

PR #4159, opened 2025-08-18
Use self hosted runners

rashadmughal:


rashadmughal:
This is not needed, the github runners are free for open source projects, therefore zero cost to us.

PR #4187, opened 2025-08-20
Stop emitting placement app withdrawn event

davidatkinsuk:
This is in preperation to backfill this event when backfilling placement applications automatic


danhumphreys-moj (approved):
Excellent! LGTM!

PR #4201, opened 2025-08-22
Backfill placement app domain event fix

davidatkinsuk:
There is a single placement request in the produciton database with more than one `APPROVED_PREMISES_MATCH_REQUEST_WITHDRAWN` domain event which causes the backfill job to fail.

The duplicate domain event are identical so it doesn’t matter which is used by the backfill job. This commit updates the query to only return one domain event


muhammad-elabdulla (approved):
LGTM

PR #4203, opened 2025-08-28
CAS-1914 Not return archive/unarchive canceled domain events in the archive history

muhammad-elabdulla:
This [PR CAS-1914](https://dsdmoj.atlassian.net/browse/CAS-1914) is to:
- Add new column cas3_cancelled_at to the domain events table
- Update the domain event when the user cancel it
- Don't return the archive/unarchive premises/bedspace cancelled domain events in the archive history

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/DomainEventEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/DomainEventEntity.kt
@@ -104,10 +104,10 @@ interface DomainEventRepository : JpaRepository<DomainEventEntity, UUID> {
     """
     SELECT d
     FROM DomainEventEntity d
-    WHERE d.cas3BedspaceId in :cas3BedspaceIds and d.type in :bedspaceDomainEventTypes
+    WHERE d.cas3BedspaceId in :cas3BedspaceIds and d.type in :bedspaceDomainEventTypes and d.cas3CancelledAt is null
     """,
   )
-  fun findBedspacesDomainEventsByType(cas3BedspaceIds: List<UUID>, bedspaceDomainEventTypes: List<String>): List<DomainEventEntity>
+  fun findBedspacesActiveDomainEventsByType(cas3BedspaceIds: List<UUID>, bedspaceDomainEventTypes: List<String>): List<DomainEventEntity>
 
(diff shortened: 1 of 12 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/DomainEventEntity.kt
```diff
@@ -155,6 +155,8 @@ data class DomainEventEntity(
   val type: DomainEventType,
   val occurredAt: OffsetDateTime,
   val createdAt: OffsetDateTime,
+  @Column(name = "cas3_cancelled_at")
+  val cas3CancelledAt: OffsetDateTime?,
```
vbala-moj, line 159: Do we need a cas3 prefix here?

vbala-moj: But this column could be reused by everyone, so my preference is to use `'canceledAt' `(the same as `'occuredAt' and 'createdAt')` instead.

vbala-moj: Although I don't like the idea of using CAS-specific columns (e.g., `cas1SpaceBookingId`), cancellation is a common operation and does not need to be tied specifically to a CAS service.

muhammad-elabdulla: This is for specific case in CAS3 that the user want to cancel a domain event. 
The other CASs they use the domain event to emit a event so I don't think they will need to cancel it later

muhammad-elabdulla: I am with make things more generic and not tide it to a specific CAS but my understanding of domain events this scenario is exceptional this why I named the column with cas3 prefix

vbala-moj: Though I prefer the generic approach, since we’ve already adopted CAS-specific fields, I have no issue approving this!

vbala-moj (approved):
Though I prefer the generic approach, since we’ve already adopted CAS-specific fields, I have no issue approving this!

PR #4212, opened 2025-09-02
Add ClientCachingOAuth2AuthorizedClientService

vbala-moj:
https://dsdmoj.atlassian.net/browse/APS-2712

### **Custom `OAuth2AuthorizedClientService` Implementation**
This implementation introduces a custom `OAuth2AuthorizedClientService` that caches `client_credentials` tokens in-memory for a specific system user (`SYSTEM_USERNAME`).

- Tokens are always fetched on behalf of a system user, so the cache is keyed by `OAuth2AuthorizedClientId(clientId, SYSTEM_USERNAME)`.
- Prevents redundant token re-fetching by caching tokens in a `ConcurrentHashMap`.
- The cache is in-memory and will be cleared on application restart.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/Application.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/Application.kt
@@ -3,6 +3,8 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi
 import org.springframework.boot.autoconfigure.SpringBootApplication
 import org.springframework.boot.runApplication
 
+const val SYSTEM_USERNAME = "COMMUNITY_ACCOMMODATION_API"
+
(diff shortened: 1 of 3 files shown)


danhumphreys-moj (approved):
I have looked at this and discussed briefly with @vbala-moj.

This is a variation of an OAuth caching service we currently use elsewhere within HMPPS (e.g. in the [incentives](https://github.com/ministryofjustice/hmpps-incentives-api/blob/main/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/config/ClientCachingOAuth2AuthorizedClientService.kt) API).

So an existing pattern is being followed (albeit with slight modifications).

The service class implements a Spring security interface and overrides the necessary functions, therefore no requirement for unit tests for the new functions we are adding.

cc @williamfalconeruk @maccalsa

PR #4235, opened 2025-09-08
CAS-1882 Return only active Referral Rejection Reasons

muhammad-elabdulla:
This PR CAS-1882 is to return only active Referral Rejection Reasons


rashadmughal (approved):
LGTM

PR #4238, opened 2025-09-08
CAS-1968 Change startDate property to be nullable in Cas3Bedspace

muhammad-elabdulla:
This PR CAS-1968 change the start date property temporary to nullable until we run the migration process then all bedspaces will have createdAt date. Then will set the startDate property back to be not nullable


rashadmughal (approved):
LGTM

PR #4246, opened 2025-09-10
CAS-1967 Make createdAt not nullable for premises entity

muhammad-elabdulla:
This [PR CAS-1967](https://dsdmoj.atlassian.net/browse/CAS-1967) is to: 
- Make createdAt property not nullable for premises entity 
- Change the start date property for CAS3Premises to be not nullable

**Note:** I removed the integration tests for the migration job that set the createdAt field for premises, as it has already run in production and is no longer relevant—since all premises now have a createdAt date.


rashadmughal (approved):
LGTM

PR #4268, opened 2025-09-15
CAS3 V2 Update Bedspace API

gregkhawkins:
**New endpoint delivered**

Screenshot 2025-09-15 at 15 15 01 (2)

**Background**

Latest PR in a larger piece of work delivering new `CAS3` versions of pre-existing endpoints in the API. The reasons we are doing this are:
1. To deliver endpoint separation for all `CAS3` endpoints (in the `Bedspace model refactor` area)
2. To deliver service separation for all `CAS3` services (in the `Bedspace model refactor` area)
3. To deliver data model separation so we end yup with distinct `CAS3` tables (in the`Bedspace model refactor` area)

Here is the new data model:

new premises tables with cas3 void bedspace

**PR includes**
1. Delivery of the new `PUT /cas3/v2/premises/{premisesId}/bedspaces/{bedspaceId}`
2. This endpoint was modelled on the `PUT /cas3/premises/{premisesId}/bedspaces/{bedspaceId}` and so all functionality / integration tests and unit tests have been ported over and refactored to use the new data models

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/controller/Cas3v2BedspaceController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/controller/Cas3v2BedspaceController.kt
@@ -5,12 +5,14 @@ import org.springframework.http.ResponseEntity
 import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.bind.annotation.PathVariable
 import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.PutMapping
 import org.springframework.web.bind.annotation.RequestBody
 import org.springframework.web.bind.annotation.RequestMapping
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas3.jpa.entity.Cas3PremisesEntity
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/controller/Cas3v2BedspaceController.kt
```diff
@@ -66,6 +69,21 @@ class Cas3v2BedspaceController(
     return ResponseEntity.ok(result)
   }
 
+  @Transactional
```
rashadmughal, line 72: Can the `@transactional` be moved down to the controller? I don't expect the `getAndCheckUserCanViewPremises` to be part of the txn

rashadmughal (approved):
LGTM, one comment

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/controller/Cas3v2BedspaceController.kt
gregkhawkins: we don't need the annotation at all in fact as the service only does one `write` related call (i.e. `cas3BedspacesRepository.save(bedspace)`)

PR #4270, opened 2025-09-15
CAS-1975 Don't return upcoming properties in online premises in premises search

muhammad-elabdulla:
This PR CAS-1975 is to not return upcoming properties when search for online premises


rashadmughal (approved):
LGTM

PR #4272, opened 2025-09-16
CAS-1985 Save premises Id for bedspace archive/unarhice bedspace domain events

muhammad-elabdulla:
This PR CAS-1985 is to save premises Id for bedspace archive/unarhice bedspace domain events


rashadmughal (approved):
LGTM

PR #4275, opened 2025-09-16
CAS-1977 Migration job to populate created date for cas3 bedspaces

muhammad-elabdulla:
This PR is to populate created date for cas3 bedspaces when created date is null


rashadmughal (approved):
LGTM

PR #4277, opened 2025-09-16
updated isPremisesScheduleToUnarchive function to accept a LocalDate

daveawc:
updated the parameter type to be a LocalDate instead of the entity, so it can be reused with V2


gregkhawkins (approved):
LGTM

PR #4292, opened 2025-09-19
refactored cas3premises service to remove duplicated code

daveawc:


muhammad-elabdulla (changes requested):
This approach is doing two unnecessary calls to the database if the region does not exist

Its good to have common function for validation but only for fields that don't implies calling the database

PR #4321, opened 2025-09-29
updating models and entities and adding transformer for cas3v2 premises

daveawc:


gregkhawkins (commented):
Looking good. Could of minors and a question :)

PR #4324, opened 2025-09-29
CAS-2003 Migration jobfor archive/unarchive CAS3 domain events

muhammad-elabdulla:
This PR CAS-2003 is create a migration job to populate transaction Id for archive/unarchive CAS3 domain events

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/migration/Cas3UpdateArchiveUnarchiveDomainEventTransactionJob.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/migration/Cas3UpdateArchiveUnarchiveDomainEventTransactionJob.kt
@@ -1,0 +1,64 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.cas3.migration
+
+import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.DomainEventRepository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.DomainEventType.CAS3_BEDSPACE_ARCHIVED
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/migration/Cas3UpdateArchiveUnarchiveDomainEventTransactionJob.kt
```diff
@@ -0,0 +1,64 @@
...
+      migrationLogger.info("Updating CAS3 archive/unarchive domain events with premises Ids ${premisesIds.map { it }}")
+      var lastDomainEventPremisesId: UUID? = null
+      var lastDomainEventCreatedAt: OffsetDateTime? = null
+      var currentTransactionId: UUID? = null
+
+      domainEvents.forEach { domainEvent ->
+        val isNewPremises = lastDomainEventPremisesId != domainEvent.cas3PremisesId
+        val isSameTransaction = (
+          lastDomainEventCreatedAt != null &&
+            domainEvent.createdAt <= lastDomainEventCreatedAt.plusMinutes(3) &&
```
rashadmughal, line 44: 3 minutes seems quite a large window - should probably be 10s or so?

rashadmughal (approved):
LGTM but one comment.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/migration/Cas3UpdateArchiveUnarchiveDomainEventTransactionJob.kt
muhammad-elabdulla: I agree but just in case I checked the existing archive/unarchive domain events and there are not close in time

PR #4333, opened 2025-10-01
Feature/CAS-1951 premises put

daveawc:
Lift and shift and slight refactor. 
* Added extra validation to check a premises belongs to a pdu, as there was nothing previously, so I think maybe you could update a premises in a different PDU if you provided the PDU for your user.
* Also combined all existing validation unit tests into fewer tests. 
* changed the getBedspaceTotals methods to do the calculation in the service, and return as a pair, to save the extra db call

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/controller/Cas3PremisesController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/controller/Cas3PremisesController.kt
@@ -185,6 +185,7 @@ class Cas3PremisesController(
     return ResponseEntity.ok(result)
   }
 
+  @Deprecated("This endpoint will be removed in v2. Use /cas3/v2/premises")
   @PostMapping(
     "/premises",
     consumes = [MediaType.APPLICATION_JSON_VALUE],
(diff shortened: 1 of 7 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/jpa/entity/Cas3PremisesCharacteristicEntity.kt
```diff
@@ -14,7 +14,7 @@ import java.util.UUID
 @Repository
 interface Cas3PremisesCharacteristicRepository : JpaRepository<Cas3PremisesCharacteristicEntity, UUID> {
   @Query("select c from Cas3PremisesCharacteristicEntity c where c.id in (:ids) and c.isActive = true")
-  fun findActiveCharacteristicsByIdIn(ids: List<UUID>): List<Cas3PremisesCharacteristicEntity>
+  fun findActiveCharacteristicsByIdIn(ids: List<UUID>): MutableList<Cas3PremisesCharacteristicEntity>
```
gregkhawkins, line 17: another option here leave this is final (i.e. `List` not `MutableList`) to avoid leaks when others re-use this and then just do this when setting onto the `Cas3PremisesEntity.characteristics` later on:
```
validatedPremisesCharacteristics.toMutableList()
```

gregkhawkins (approved):

PR #4340, opened 2025-10-02
chore: Use v2 of NVD feed for OWASP checks

froddd:
This updates the OWASP security workflow to use the v2 version of the NVD API.

### DIFF
--- a/.github/workflows/security_owasp.yml
+++ b/.github/workflows/security_owasp.yml
@@ -8,5 +8,6 @@ jobs:
     name: Kotlin security OWASP dependency check
     uses: ministryofjustice/hmpps-github-actions/.github/workflows/security_owasp.yml@v2 # WORKFLOW_VERSION
     with:
+      nvd_feed_version: '2'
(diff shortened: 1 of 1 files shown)


davidatkinsuk:
we should merge this change as part of the upgrade to hmpps-gradle-spring-boot v9, as detailed [here](https://github.com/ministryofjustice/hmpps-gradle-spring-boot/blob/main/release-notes/9.x.md)

froddd:
Looks like the OWASP action is already using V2 of the feed -- has the default been updated upstream? https://github.com/ministryofjustice/hmpps-approved-premises-api/actions/runs/24760742947

davidatkinsuk:
Ah! I did check upstream and it defaults to v1, but it looks like the change on this PR has been made via some other PR as we are setting v2 on main - https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/main/.github/workflows/security_owasp.yml

I'll close this PR. Thanks for looking!

PR #4347, opened 2025-10-03
CAS3 V2 Bedspace occupancy report

gregkhawkins:
**PR includes:**
* Refactor the `GET /cas3/reports/{reportName}` endpoint for the `bedOccupancy` report type
* uses the `cas3-reports-with-new-bedspace-model-tables-enabled` feature-flag to determine whether to execute a query against the new `Bedspace model refactor` tables (or to execute against the tables currently queried in production when feature-flag off)
* executes the old or new query accordingly
* test coverage for the report against the new tables


vbala-moj (approved):
Just one minor comment.

PR #4350, opened 2025-10-06
CAS-2046 Fix issue can archive bedspace when a void is cancelled

muhammad-elabdulla:
This [PR CAS-2046](https://dsdmoj.atlassian.net/browse/CAS-2046) is to fix an issue when check if a bedspace can archive and there are a void that is cancelled


rashadmughal (approved):
LGTM

PR #4406, opened 2025-10-21
Chore/cas 2002 flyways to drop unwanted cas 2 columns

besscerule:


davidatkinsuk:
@besscerule is this still required?

PR #4536, opened 2026-01-19
updating restore job

daveawc:


davidatkinsuk:
closing this, but we'll retain the branch. an alternatively approach is to not try and backup the users tables and instead apply the seed job after migration. We could put the seed CSV in a secret

PR #4603, opened 2026-03-24
Spring Boot 4 prep: Update hibernate json annotations

rashadmughal:
Reverts ministryofjustice/hmpps-approved-premises-api#4602


rashadmughal:
Not needed now.

PR #4620, opened 2026-04-15
Chore: update E2E and Generate types workflows

froddd:
This PR updates the E2E workflow to point to `main` on the UI repositories, as [SHA pinning is not required for worklfows](https://mojdt.slack.com/archives/C08SV6MR2P7/p1774868094624769?thread_ts=1774630861.424619&cid=C08SV6MR2P7). This ensures the E2E tests run from this repository are the latest.

A new workflow also updates the approach to running the 'Generate types' workflows for all UIs, moving to the 'uses' pattern instead of requiring a custom GitHub action call.


danhumphreys-moj (approved):
Much better using `uses` than having to manually generate a token to call a Github REST action.  Nice work Fred!!

PR #4623, opened 2026-04-16
Update swagger-ui to fix slack notifications.

rashadmughal:


.github/workflows/security_owasp.yml
```diff
@@ -1,19 +1,18 @@
 name: Security OWASP dependency check
-
 on:
   workflow_dispatch:
   schedule:
-    - cron: "51 3 * * MON-FRI" # Every weekday at 03:51 UTC
+    - cron: "8 4 * * MON-FRI" # Every weekday
```
froddd: Scheduled time doesn't need changing -- this would now run at the same time as the SAS API one which could be confusing. And the comment for expected time is useful!

froddd:
As discussed on Slack, this isn't actually needed at all -- workflows don't need pinning: https://mojdt.slack.com/archives/C08SV6MR2P7/p1774868094624769?thread_ts=1774630861.424619&cid=C08SV6MR2P7

danhumphreys-moj (approved):

PR #4604, opened 2026-03-26
CAS-290: Use spring problem

rashadmughal:

PR #4627, opened 2026-04-20
FM 488 Add unit tests for `ExceptionHandling` class

danhumphreys-moj:

This PR adds some missing unit tests for the global `ExceptionHandling` class.  This class currently takes API exceptions and converts them in to Zalando Problems.  There is upcoming work in https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/4604 to migrate from using [Zalando Spring Problem](https://github.com/zalando/problem-spring-web) to [Spring Framework Spring Problem](https://docs.spring.io/spring-framework/reference/web/webmvc/mvc-ann-rest-exceptions.html) which itself is a requirement for upgrading from Spring Boot 3 to [Spring Boot 4.](https://dsdmoj.atlassian.net/browse/FM-488)

Although integration tests are currently passing PR https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/4604, the addition of unit tests covering the global exception handling gives additional confidence and helps reduce the risk that the work done to migrate away from Zalando Spring Problem to use the default Spring Problem library is likely to cause any issues.  These unit tests will be updated as part of the forthcoming PR.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/problem/ExceptionHandlingTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/problem/ExceptionHandlingTest.kt
@@ -1,0 +1,317 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.unit.problem
+
+import com.fasterxml.jackson.databind.JsonMappingException
+import com.fasterxml.jackson.databind.exc.MismatchedInputException
+import com.fasterxml.jackson.databind.json.JsonMapper
+import com.fasterxml.jackson.databind.node.ArrayNode
+import com.fasterxml.jackson.databind.node.ObjectNode
+import io.mockk.every
+import io.mockk.mockk
+import io.mockk.verify
+import org.assertj.core.api.Assertions.assertThat
+import org.hibernate.exception.JDBCConnectionException
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.RegisterExtension
+import org.springframework.core.MethodParameter
+import org.springframework.http.HttpInputMessage
+import org.springframework.http.HttpMethod
+import org.springframework.http.HttpStatus
+import org.springframework.http.converter.HttpMessageNotReadableException
+import org.springframework.security.authentication.AuthenticationCredentialsNotFoundException
+import org.springframework.web.bind.MissingRequestHeaderException
+import org.springframework.web.bind.MissingServletRequestParameterException
+import org.springframework.web.context.request.NativeWebRequest
+import org.springframework.web.method.annotation.MethodArgumentTypeMismatchException
+import org.springframework.web.servlet.resource.NoResourceFoundException
+import org.springframework.web.util.ContentCachingRequestWrapper
+import org.zalando.problem.Status
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.BadRequestProblem
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.ExceptionHandling
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.ForbiddenProblem
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.NotFoundResourceProblem
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.ParamDetails
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.ServiceUnavailableProblem
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.UnauthenticatedProblem
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.problem.UnhandledExceptionProblem
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.DeserializationValidationService
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.SentryService
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.unit.util.LoggerExtension
+import java.sql.SQLException
+import org.springframework.security.access.AccessDeniedException as SpringAccessDeniedException
+
+class ExceptionHandlingTest {
+  private val mockJsonMapper = mockk<JsonMapper>()
+  private val mockDeserializationValidationService = mockk<DeserializationValidationService>()
+  private val mockSentryService = mockk<SentryService>(relaxed = true)
+
+  val exceptionHandling = ExceptionHandling(mockJsonMapper, mockDeserializationValidationService, mockSentryService)
+
+  @RegisterExtension
+  var loggerExtension: LoggerExtension = LoggerExtension()
+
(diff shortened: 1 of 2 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/problem/ExceptionHandlingTest.kt
```diff
@@ -0,0 +1,255 @@
...
+class ExceptionHandlingTest {
+  private val mockJsonMapper = mockk<JsonMapper>()
+  private val mockDeserializationValidationService = mockk<DeserializationValidationService>()
+  private val mockSentryService = mockk<SentryService>()
+
+  val exceptionHandling = ExceptionHandling(mockJsonMapper, mockDeserializationValidationService, mockSentryService)
+
+  @RegisterExtension
+  var loggerExtension: LoggerExtension = LoggerExtension()
+
```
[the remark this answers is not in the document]
danhumphreys-moj: Added:

```
  @AfterEach
  fun tearDown() {
    unmockkAll()
  }
```

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/problem/ExceptionHandlingTest.kt
```diff
@@ -0,0 +1,255 @@
...
+  @Test
+  fun `Returns UnauthenticatedProblem when exception is AuthenticationCredentialsNotFoundException and exception captured in Sentry`() {
+    val throwable = AuthenticationCredentialsNotFoundException("")
+    val statusType = Status.UNAUTHORIZED
+
+    mockkStatic(Sentry::class)
+    val problem = exceptionHandling.toProblem(throwable, statusType)
```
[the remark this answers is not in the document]
danhumphreys-moj: Added stub for `Sentry.captureException(any())`

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/problem/ExceptionHandlingTest.kt
```diff
@@ -0,0 +1,255 @@
...
+  }
+
+  @Test
+  fun `Returns UnhandledExceptionProblem problem when exception is unhandled (IllegalStateException), logs error and exception captured in Sentry`() {
+    val throwable = IllegalStateException()
+    val statusType = Status.INTERNAL_SERVER_ERROR
+
+    mockSentry()
+    val problem = exceptionHandling.toProblem(throwable, statusType)
+
```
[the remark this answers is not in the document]
danhumphreys-moj: Added stub for `Sentry.captureException(any())`

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/problem/ExceptionHandlingTest.kt
```diff
@@ -0,0 +1,348 @@
...
+  @RegisterExtension
+  var loggerExtension: LoggerExtension = LoggerExtension()
+
+  @Test
+  fun `Returns UnauthenticatedProblem when exception is AuthenticationCredentialsNotFoundException and exception captured in Sentry`() {
+    val throwable = AuthenticationCredentialsNotFoundException("")
+    val statusType = Status.UNAUTHORIZED
+
+    mockkStatic(Sentry::class)
+    every { Sentry.captureException(any()) } returns SentryId.EMPTY_ID
```
davidatkinsuk: An observation, up to you if you want to action it :) 

* I think ideally we should be sending all sentry traffic through the `SentryService`
* That then removes the need to mock a static, which can be a little clunky because you have to remember to unmock the static
* And then we can consider using relaxed mocks which will return a sensible default, meaning you don't need to setup a response unless you really care about it (in this case i don't think we do)

To use a relaxed mock you can do something like

```
private val mockSentryService = mockk<SentryService>(relaxed = true)
```

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/problem/ExceptionHandlingTest.kt
```diff
@@ -0,0 +1,348 @@
...
+    assertThat(problem!!.type).isEqualTo(expectedProblem.type)
+    assertThat(problem.title).isEqualTo(expectedProblem.title)
+    assertThat(problem.status).isEqualTo(expectedProblem.status)
+    assertThat(problem.detail).isEqualTo("Invalid type for query parameter id expected int")
+
+    verify { Sentry.captureException(throwable) }
+  }
+
+  @Test
+  fun `Returns ServiceUnavailableProblem problem when exception is cause chain of JDBCConnectionException and exception captured in Sentry (twice)`() {
```
davidatkinsuk: there does seem to be a legitimate bug in the code with `captureException()` being called multiple times. I don't _think_ this is deliberate and i guess sentry is naturally deduplicating them (as we're not seeing duplicate alerts). Shall we remove the duplicate calls, maybe in a different PR?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/problem/ExceptionHandlingTest.kt
danhumphreys-moj: Added a new [commit](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/4627/changes/3b6b6e1a2deb118714f7c89c98b42f78b537ba63) to replace the `Sentry` call with `sentryService` in `ExceptionHandling`.

This then means we can remove the need to mock the static `Sentry.captureException()` call, remove the the `@AfterAll` tear down function in the unit tests and use:

`private val mockSentryService = mockk<SentryService>(relaxed = true)`

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/problem/ExceptionHandlingTest.kt
danhumphreys-moj: Yes I've removed the duplicate Sentry call and updated the test in this PR in a new [commit](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/4627/changes/2275e1fb70f7f1a59190eab0f9b14b58ad38c223).  Seemed to make sense to do it here, and all integration tests pass so ok.

danhumphreys-moj: Will rebase and resolve merge conflict in https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/4604 once this PR is merged and deployed.

davidatkinsuk (approved):
Good work. LGTM 👍

rashadmughal:
LGTM

PR #4630, opened 2026-04-21
FM-505 Add `approved_premises.allow_new_space_bookings`

davidatkinsuk:
This property is used to indicate if a new space booking can be created for a given premises. It has two functional impacts:

1. Premises where this value is `false` will not be returned by the space search endpoint
2. Any attempts to create a space booking where this value is `false` will result in a validation error


vbala-moj (approved):
LGTM

PR #4631, opened 2026-04-21
FM-491 Bulk update OOSB

vbala-moj:
https://dsdmoj.atlassian.net/browse/FM-491

- Created a new seed job `Cas1UpdateOutOfServiceBedsByPremisesJob` to bulk update active Out-of-Service Beds (OOSB) for a specific premises via CSV.
- The job updates the end date, notes, and reason for all active OOSBs at the given premises.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/seed/Cas1UpdateOutOfServiceBedsByPremisesJob.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/seed/Cas1UpdateOutOfServiceBedsByPremisesJob.kt
@@ -1,0 +1,72 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.cas1.seed
+
+import org.slf4j.LoggerFactory
+import org.springframework.data.repository.findByIdOrNull
+import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas1OutOfServiceBedReasonRepository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.seed.SeedException
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.seed.SeedJob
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/seed/Cas1UpdateOutOfServiceBedsByPremisesJob.kt
```diff
@@ -0,0 +1,70 @@
...
+    "notes",
+    "reason_id",
+  ),
+) {
+
+  private val log = LoggerFactory.getLogger(this::class.java)
+
+  override fun deserializeRow(columns: Map<String, String>) = Cas1UpdateOutOfServiceBedsByPremisesCsvRow(
+    premisesId = UUID.fromString(columns["premises_id"]!!.trim()),
+    endDate = LocalDate.parse(columns["end_date"]!!.trim()),
```
davidatkinsuk: I think we should allow for nullable values for the end date to keep this seed job flexible? If an end date isn't defined in the CSV we should instead use the end date from the most recent existing revision (which may itself be null)?

I think testing this functionality with a unit test only is fine

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/seed/Cas1UpdateOutOfServiceBedsByPremisesJob.kt
```diff
@@ -0,0 +1,70 @@
...
+    notes = columns["notes"].trimToNull(),
+    reasonId = UUID.fromString(columns["reason_id"]!!.trim()),
```
davidatkinsuk: I think we should maybe allow for nullable values for reasonId. If one isn't defined in the CSV we should instead use the reason from the most recent existing revision? I suspect in cases where we want to set an end date on all beds it's possible that the existing reasons aren't all the same?

I think testing this functionality with a unit test only is fine

davidatkinsuk (approved):
Looks good! There's a few proposals from me to maximum flexibility of the seed job

PR #4633, opened 2026-04-22
FM-495: Use spring problem

danhumphreys-moj:

Now that additional unit and integration test coverage has been added, this PR removes the dependency on Zalando Spring Problem so we can use the default Spring Problem libraries which are part of the Spring Framework.

This PR now uses the `@ExceptionHandler` annotations and associated handle methods.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/controller/Cas2AssessmentsController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/controller/Cas2AssessmentsController.kt
@@ -7,7 +7,6 @@ import org.springframework.web.bind.annotation.PathVariable
 import org.springframework.web.bind.annotation.PostMapping
 import org.springframework.web.bind.annotation.PutMapping
 import org.springframework.web.bind.annotation.RequestBody
-import org.zalando.problem.AbstractThrowableProblem
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas2.jpa.entity.Cas2ApplicationNoteEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas2.jpa.entity.Cas2StatusUpdateEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas2.model.Cas2ApplicationNote
@@ -129,7 +128,7 @@ class Cas2AssessmentsController(
     is AuthorisableActionResult.Success -> result.entity
   }
 
-  private fun throwProblem(problem: AbstractThrowableProblem): Unit = throw problem
+  private fun throwProblem(problem: RuntimeException): Unit = throw problem
 
   private fun <EntityType : Any> processValidation(validationResult: ValidatableActionResult<EntityType>): Any = when (validationResult) {
     is ValidatableActionResult.GeneralValidationError -> throwProblem(BadRequestProblem(errorDetail = validationResult.message))
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/controller/Cas2AssessmentsController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/controller/Cas2AssessmentsController.kt
@@ -123,17 +123,15 @@ class Cas2AssessmentsController(
     assessmentId: UUID,
     result: AuthorisableActionResult<ValidatableActionResult<EntityType>>,
   ): Any = when (result) {
-    is AuthorisableActionResult.NotFound -> throwProblem(NotFoundProblem(assessmentId, "Cas2Application"))
-    is AuthorisableActionResult.Unauthorised -> throwProblem(ForbiddenProblem())
+    is AuthorisableActionResult.NotFound -> throw(NotFoundProblem(assessmentId, "Cas2Application"))
+    is AuthorisableActionResult.Unauthorised -> throw(ForbiddenProblem())
     is AuthorisableActionResult.Success -> result.entity
   }
 
-  private fun throwProblem(problem: RuntimeException): Unit = throw problem
-
   private fun <EntityType : Any> processValidation(validationResult: ValidatableActionResult<EntityType>): Any = when (validationResult) {
-    is ValidatableActionResult.GeneralValidationError -> throwProblem(BadRequestProblem(errorDetail = validationResult.message))
-    is ValidatableActionResult.FieldValidationError -> throwProblem(BadRequestProblem(invalidParams = validationResult.validationMessages.mapValues { ParamDetails(it.value) }))
-    is ValidatableActionResult.ConflictError -> throwProblem(ConflictProblem(id = validationResult.conflictingEntityId, conflictReason = validationResult.message))
+    is ValidatableActionResult.GeneralValidationError -> throw(BadRequestProblem(errorDetail = validationResult.message))
+    is ValidatableActionResult.FieldValidationError -> throw(BadRequestProblem(invalidParams = validationResult.validationMessages.mapValues { ParamDetails(it.value) }))
+    is ValidatableActionResult.ConflictError -> throw(ConflictProblem(id = validationResult.conflictingEntityId, conflictReason = validationResult.message))
     is ValidatableActionResult.Success -> validationResult.entity
   }
 }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/BadRequestProblem.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/BadRequestProblem.kt
@@ -1,31 +1,33 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.problem
 
 import com.fasterxml.jackson.annotation.JsonIgnore
-import org.zalando.problem.AbstractThrowableProblem
-import org.zalando.problem.Exceptional
-import org.zalando.problem.Status
+import org.springframework.http.HttpStatus
+import org.springframework.http.ProblemDetail
 
 class BadRequestProblem(
   @JsonIgnore val invalidParams: Map<String, ParamDetails>? = null,
   @JsonIgnore val errorDetail: String? = null,
-) : AbstractThrowableProblem(null, "Bad Request", Status.BAD_REQUEST, errorDetail ?: "There is a problem with your request") {
-  override fun getCause(): Exceptional? = null
+) : RuntimeException(listOfNotNull("Bad Request", errorDetail, invalidParams).joinToString(": ")) {
 
-  override val message: String
-    get() = listOfNotNull(this.title, this.detail, this.invalidParams).joinToString(": ")
-
-  override fun getParameters(): MutableMap<String, Any> = mutableMapOf(
-    "invalid-params" to (
-      invalidParams?.map {
-        ParamError(
-          propertyName = it.key,
-          errorType = it.value.errorType,
-          entityId = it.value.entityId,
-          value = it.value.value,
-        )
-      } ?: emptyList()
-      ),
-  )
+  fun toProblemDetail(): ProblemDetail {
+    val detail = errorDetail ?: "There is a problem with your request"
+    val problemDetail = FlattenedProblemDetail(HttpStatus.BAD_REQUEST, detail)
+    problemDetail.title = "Bad Request"
+    if (invalidParams != null) {
+      problemDetail.setProperty(
+        "invalid-params",
+        invalidParams.map {
+          ParamError(
+            propertyName = it.key,
+            errorType = it.value.errorType,
+            entityId = it.value.entityId,
+            value = it.value.value,
+          )
+        },
+      )
+    }
+    return problemDetail
+  }
 }
 
 data class ParamDetails(
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/BadRequestProblem.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/BadRequestProblem.kt
@@ -11,7 +11,20 @@ class BadRequestProblem(
 
   override fun toProblemDetail(): ProblemDetail {
     val detail = errorDetail ?: "There is a problem with your request"
-    val problemDetail = FlattenedProblemDetail(HttpStatus.BAD_REQUEST, detail)
+    val problemDetail = InvalidParamsProblemDetail(
+      HttpStatus.BAD_REQUEST,
+      detail,
+      invalidParams?.flatMap { (field, paramDetails) ->
+        listOf(
+          ParamError(
+            propertyName = field,
+            errorType = paramDetails.errorType,
+            entityId = paramDetails.entityId,
+            value = paramDetails.value,
+          ),
+        )
+      },
+    )
     problemDetail.title = "Bad Request"
     if (invalidParams != null) {
       problemDetail.setProperty(
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/BadRequestProblem.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/BadRequestProblem.kt
@@ -7,9 +7,9 @@ import org.springframework.http.ProblemDetail
 class BadRequestProblem(
   @JsonIgnore val invalidParams: Map<String, ParamDetails>? = null,
   @JsonIgnore val errorDetail: String? = null,
-) : RuntimeException(listOfNotNull("Bad Request", errorDetail, invalidParams).joinToString(": ")) {
+) : ApiProblem(listOfNotNull("Bad Request", errorDetail, invalidParams).joinToString(": ")) {
 
-  fun toProblemDetail(): ProblemDetail {
+  override fun toProblemDetail(): ProblemDetail {
     val detail = errorDetail ?: "There is a problem with your request"
     val problemDetail = FlattenedProblemDetail(HttpStatus.BAD_REQUEST, detail)
     problemDetail.title = "Bad Request"
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/ConflictProblem.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/ConflictProblem.kt
@@ -1,9 +1,13 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.problem
 
-import org.zalando.problem.AbstractThrowableProblem
-import org.zalando.problem.Exceptional
-import org.zalando.problem.Status
+import org.springframework.http.HttpStatus
+import org.springframework.http.ProblemDetail
 
-class ConflictProblem(id: Any, conflictReason: String) : AbstractThrowableProblem(null, "Conflict", Status.CONFLICT, "$conflictReason: $id") {
-  override fun getCause(): Exceptional? = null
+class ConflictProblem(id: Any, conflictReason: String) : RuntimeException("$conflictReason: $id") {
+
+  val msg = "$conflictReason: $id"
+
+  fun toProblemDetail(): ProblemDetail = ProblemDetail.forStatusAndDetail(HttpStatus.CONFLICT, message ?: msg).apply {
+    title = "Conflict"
+  }
 }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/ConflictProblem.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/ConflictProblem.kt
@@ -3,11 +3,9 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.problem
 import org.springframework.http.HttpStatus
 import org.springframework.http.ProblemDetail
 
-class ConflictProblem(id: Any, conflictReason: String) : RuntimeException("$conflictReason: $id") {
+class ConflictProblem(id: Any, conflictReason: String) : ApiProblem("$conflictReason: $id") {
 
-  val msg = "$conflictReason: $id"
-
-  fun toProblemDetail(): ProblemDetail = ProblemDetail.forStatusAndDetail(HttpStatus.CONFLICT, message ?: msg).apply {
+  override fun toProblemDetail(): ProblemDetail = ProblemDetail.forStatusAndDetail(HttpStatus.CONFLICT, message).apply {
     title = "Conflict"
   }
 }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/ExceptionHandling.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/ExceptionHandling.kt
@@ -204,13 +204,21 @@ class ExceptionHandling(
     return ResponseEntity.status(HttpStatus.SERVICE_UNAVAILABLE).contentType(APPLICATION_PROBLEM_JSON).body(problem.toProblemDetail())
   }
 
+  @ExceptionHandler(UnauthenticatedProblem::class)
+  fun handleUnauthenticatedProblemException(ex: UnauthenticatedProblem): ResponseEntity<ProblemDetail> {
+    sentryService.captureException(ex)
+    return ResponseEntity.status(HttpStatus.UNAUTHORIZED).contentType(APPLICATION_PROBLEM_JSON).body(
+      UnauthenticatedProblem().toProblemDetail(),
+    )
+  }
+
   @ExceptionHandler(Throwable::class)
   fun handleGenericException(ex: Throwable): ResponseEntity<ProblemDetail> {
     sentryService.captureException(ex)
     // Check if JDBCConnectionException is in the cause chain
     if (isTypeInThrowableChain(ex, JDBCConnectionException::class.java)) {
       val problem = ServiceUnavailableProblem(detail = "Error acquiring a database connection")
-      return ResponseEntity.status(HttpStatus.SERVICE_UNAVAILABLE).body(problem.toProblemDetail())
+      return ResponseEntity.status(HttpStatus.SERVICE_UNAVAILABLE).contentType(APPLICATION_PROBLEM_JSON).body(problem.toProblemDetail())
     }
 
     log.error("Unhandled exception type, returning generic 500 response", ex)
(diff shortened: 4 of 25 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/FlattenedProblemDetail.kt
```diff
@@ -0,0 +1,14 @@
...
+import org.springframework.http.HttpStatus
+import org.springframework.http.ProblemDetail
+
+class FlattenedProblemDetail(status: HttpStatus, detail: String) : ProblemDetail(status.value()) {
+  init {
+    this.detail = detail
+  }
+
+  @JsonAnyGetter
+  fun getFlattenedProperties(): Map<String, Any>? = properties
```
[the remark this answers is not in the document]
danhumphreys-moj: This is exactly what is happening,`invalid-params` appears in properties and in the root of the response, see [here](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/4633/changes#diff-46a1a20d8f83ef43ca970225f609953e4d6b06b68bfd4073e4d9645da1e73718R629).

Appearing in the root of the response is what is happening now, and if the UI depends on it being there this is what we should keep, using FlattenedProblemDetail is what adds it at the root level.  But including `invalid-params` in the `properties` attribute is the right technical solution as per the spec.

Conclusion - although duplicated, it seems right to use FlattenedProblemDetail to maintain consistency with how we are doing things now and the way we are returning the invalid parameters to the client, which also avoids any need to update the UI or investigate further impact.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/ExceptionHandling.kt
```diff
@@ -135,13 +125,96 @@ class ExceptionHandling(
...
+    return ResponseEntity.status(HttpStatus.SERVICE_UNAVAILABLE).contentType(APPLICATION_PROBLEM_JSON).body(problem.toProblemDetail())
+  }
+
+  @ExceptionHandler(Throwable::class)
+  fun handleGenericException(ex: Throwable): ResponseEntity<ProblemDetail> {
+    sentryService.captureException(ex)
+    // Check if JDBCConnectionException is in the cause chain
+    if (isTypeInThrowableChain(ex, JDBCConnectionException::class.java)) {
+      val problem = ServiceUnavailableProblem(detail = "Error acquiring a database connection")
+      return ResponseEntity.status(HttpStatus.SERVICE_UNAVAILABLE).body(problem.toProblemDetail())
```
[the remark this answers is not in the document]
danhumphreys-moj: Changed this to:

`return ResponseEntity.status(HttpStatus.SERVICE_UNAVAILABLE).contentType(APPLICATION_PROBLEM_JSON).body(problem.toProblemDetail())`

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/ExceptionHandling.kt
```diff
@@ -8,109 +8,99 @@ import com.fasterxml.jackson.databind.node.ObjectNode
...
-      .withTitle("Bad Request")
-
+  ): ResponseEntity<ProblemDetail> {
     when (exception.cause) {
       is MismatchedInputException -> {
         val mismatchedInputException = exception.cause as MismatchedInputException
 
         val requestBody = request.getNativeRequest(ContentCachingRequestWrapper::class.java)
-        val jsonTree = jsonMapper.readTree(String(requestBody.contentAsByteArray))
+        val jsonTree = jsonMapper.readTree(String(requestBody!!.contentAsByteArray))
```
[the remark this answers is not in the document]
danhumphreys-moj: Not proposing to do this - to maintain compatibility with existing logic.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/ExceptionHandling.kt
```diff
@@ -8,109 +8,99 @@ import com.fasterxml.jackson.databind.node.ObjectNode
...
-    if (throwable is MethodArgumentTypeMismatchException) {
-      return BadRequestProblem(
-        errorDetail = "Invalid type for query parameter ${throwable.parameter.parameterName} expected ${throwable.parameter.parameterType.name}",
-      )
-    }
+  @ExceptionHandler(AccessDeniedException::class)
+  fun handleAccessDeniedException(ex: AccessDeniedException): ResponseEntity<ProblemDetail> {
+    Sentry.captureException(ex)
+    return ResponseEntity.status(HttpStatus.FORBIDDEN).contentType(APPLICATION_PROBLEM_JSON).body(ForbiddenProblem().toProblemDetail())
+  }
```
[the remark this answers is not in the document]
danhumphreys-moj: We no longer use `Sentry.captureException(ex)` we now use `sentryService.captureException(ex)`.  This is an old comment.  Resolving.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/ExceptionHandling.kt
```diff
@@ -8,109 +8,99 @@ import com.fasterxml.jackson.databind.node.ObjectNode
...
 
 @ControllerAdvice
+@Suppress("TooManyFunctions")
 class ExceptionHandling(
   private val jsonMapper: JsonMapper,
   private val deserializationValidationService: DeserializationValidationService,
   private val sentryService: SentryService,
-) : ProblemHandling,
-  MessageNotReadableAdviceTrait {
+) {
```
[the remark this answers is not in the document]
danhumphreys-moj: Old comment, resolving.  Has been addressed now.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/ExceptionHandling.kt
```diff
@@ -135,13 +125,96 @@ class ExceptionHandling(
...
+  fun handleGenericException(ex: Throwable): ResponseEntity<ProblemDetail> {
+    sentryService.captureException(ex)
+    // Check if JDBCConnectionException is in the cause chain
+    if (isTypeInThrowableChain(ex, JDBCConnectionException::class.java)) {
+      val problem = ServiceUnavailableProblem(detail = "Error acquiring a database connection")
+      return ResponseEntity.status(HttpStatus.SERVICE_UNAVAILABLE).body(problem.toProblemDetail())
+    }
+
+    log.error("Unhandled exception type, returning generic 500 response", ex)
+    throw ex
```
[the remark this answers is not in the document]
danhumphreys-moj: This change has already been made now, resolving.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/ExceptionHandling.kt
```diff
@@ -8,109 +8,99 @@ import com.fasterxml.jackson.databind.node.ObjectNode
...
+  }
 
-    log.error("Unhandled exception type, returning generic 500 response", throwable)
+  @ExceptionHandler(MissingRequestHeaderException::class)
+  fun handleMissingRequestHeaderException(ex: MissingRequestHeaderException): ResponseEntity<ProblemDetail> {
+    Sentry.captureException(ex)
+    val problem = BadRequestProblem(errorDetail = "Missing required header ${ex.headerName}")
+    logBadRequestProblem(problem)
+    return ResponseEntity.status(HttpStatus.BAD_REQUEST).contentType(APPLICATION_PROBLEM_JSON).body(problem.toProblemDetail())
+  }
```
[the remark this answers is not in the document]
danhumphreys-moj: Old comment, resolving. Has been addressed now.  We now use `sentryService`.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/ExceptionHandling.kt
```diff
@@ -8,109 +8,99 @@ import com.fasterxml.jackson.databind.node.ObjectNode
...
-    if (isTypeInThrowableChain(throwable, JDBCConnectionException::class.java)) {
-      return ServiceUnavailableProblem(
-        detail = "Error acquiring a database connection",
-      )
-    }
+  @ExceptionHandler(NoResourceFoundException::class)
+  fun handleNoResourceFoundException(ex: NoResourceFoundException): ResponseEntity<ProblemDetail> {
+    Sentry.captureException(ex)
+    return ResponseEntity.status(HttpStatus.NOT_FOUND).contentType(APPLICATION_PROBLEM_JSON).body(NotFoundResourceProblem().toProblemDetail())
+  }
```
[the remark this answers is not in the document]
danhumphreys-moj: Old comment, resolving. Has been addressed now. We now use sentryService.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/ExceptionHandling.kt
```diff
@@ -8,109 +8,99 @@ import com.fasterxml.jackson.databind.node.ObjectNode
...
-    if (throwable is MissingServletRequestParameterException) {
-      return BadRequestProblem(
-        errorDetail = "Missing required query parameter ${throwable.parameterName}",
-      )
-    }
+  @ExceptionHandler(AuthenticationCredentialsNotFoundException::class)
+  fun handleAuthenticationCredentialsNotFoundException(ex: AuthenticationCredentialsNotFoundException): ResponseEntity<ProblemDetail> {
+    sentryService.captureException(ex)
+    return ResponseEntity.status(HttpStatus.UNAUTHORIZED).contentType(APPLICATION_PROBLEM_JSON).body(UnauthenticatedProblem().toProblemDetail())
+  }
```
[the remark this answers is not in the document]
danhumphreys-moj: Have added this:
```
  @ExceptionHandler(UnauthenticatedProblem::class)
  fun handleUnauthenticatedProblemException(ex: UnauthenticatedProblem): ResponseEntity<ProblemDetail> {
    sentryService.captureException(ex)
    return ResponseEntity.status(HttpStatus.UNAUTHORIZED).contentType(APPLICATION_PROBLEM_JSON).body(
      UnauthenticatedProblem().toProblemDetail())
  }
```

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ExceptionHandlingTest.kt
```diff
@@ -531,9 +561,19 @@ class ExceptionHandlingTest : InitialiseDatabasePerClassTestBase() {
...
         expected = """
           {        
-            "title" : "Bad Request",
             "status" : 400,
             "detail" : "There is a problem with your request",
+            "type" : "about:blank",
+            "title" : "Bad Request",
+            "instance" : "/deserialization-test/array",
+            "properties" : {
+              "invalid-params" : [{
```
danhumphreys-moj, line 565: `invalid-params` appears twice in the JSON see discussion [here](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/4633#discussion_r3152282347).  This is the most appropriate place for it and right according to the spec, however we have been including `invalid-params` in the root of the json response as per [here](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/4633/changes#diff-46a1a20d8f83ef43ca970225f609953e4d6b06b68bfd4073e4d9645da1e73718L537) below.  Leaving it in twice seems the most appropriate thing to do.  This meets the spec and guidance but also means we don't need to change existing logic and potentially client (UI).

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ExceptionHandlingTest.kt
```diff
@@ -255,15 +255,19 @@ class ExceptionHandlingTest : InitialiseDatabasePerClassTestBase() {
         .uri("/authentication-credentials-not-found-exception")
         .header("Authorization", "Bearer $jwt")
         .exchange()
-        .returnResult<String>()
+        .expectBody()
+        .returnResult()
```
danhumphreys-moj: I don't understand why I've had to change:
```
.returnResult<String>()
```
to 
```
.expectBody()
.returnResult()
```
for **some** of these responses  - unless anyone has any idea?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/ExceptionHandling.kt
```diff
@@ -8,107 +8,99 @@ import com.fasterxml.jackson.databind.node.ObjectNode
...
 
-    // We throw this instead of an InternalServerErrorProblem to avoid the Sentry Alert
-    // being raised again by [ExceptionHandling.logInternalServerErrorProblem]
-    return UnhandledExceptionProblem(
-      detail = "There was an unexpected problem",
+  @ExceptionHandler(MethodArgumentTypeMismatchException::class)
+  fun handleMethodArgumentTypeMismatchException(ex: MethodArgumentTypeMismatchException): ResponseEntity<ProblemDetail> {
+    sentryService.captureException(ex)
+    val problem = BadRequestProblem(
+      errorDetail = "Invalid type for query parameter ${ex.parameter.parameterName} expected ${ex.parameter.parameterType.name}",
```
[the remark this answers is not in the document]
danhumphreys-moj: Not proposing to do this, since attempting to make minimum changes and re-use existing logic as much as possible.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/ExceptionHandling.kt
```diff
@@ -133,13 +125,97 @@ class ExceptionHandling(
...
+    // Check if JDBCConnectionException is in the cause chain
+    if (isTypeInThrowableChain(ex, JDBCConnectionException::class.java)) {
+      val problem = ServiceUnavailableProblem(detail = "Error acquiring a database connection")
+      return ResponseEntity.status(HttpStatus.SERVICE_UNAVAILABLE).body(problem.toProblemDetail())
     }
 
-    return ResponseEntity<Problem>(responseBuilder.build(), HttpStatus.BAD_REQUEST)
+    log.error("Unhandled exception type, returning generic 500 response", ex)
+    val problem = UnhandledExceptionProblem(detail = "There was an unexpected problem")
+    return ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR).body(problem.toProblemDetail())
```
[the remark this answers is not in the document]
danhumphreys-moj: Not proposing to do this, since attempting to make minimum changes and re-use existing logic as much as possible.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/controller/Cas2AssessmentsController.kt
```diff
@@ -129,7 +128,7 @@ class Cas2AssessmentsController(
     is AuthorisableActionResult.Success -> result.entity
   }
 
-  private fun throwProblem(problem: AbstractThrowableProblem): Unit = throw problem
+  private fun throwProblem(problem: RuntimeException): Unit = throw problem
```
davidatkinsuk: shall we just remove this function, it's doesn't appear to add any value and would less code to throw directly in the calling code i think?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/BadRequestProblem.kt
```diff
@@ -1,31 +1,33 @@
...
-import org.zalando.problem.Status
+import org.springframework.http.HttpStatus
+import org.springframework.http.ProblemDetail
 
 class BadRequestProblem(
   @JsonIgnore val invalidParams: Map<String, ParamDetails>? = null,
   @JsonIgnore val errorDetail: String? = null,
-) : AbstractThrowableProblem(null, "Bad Request", Status.BAD_REQUEST, errorDetail ?: "There is a problem with your request") {
-  override fun getCause(): Exceptional? = null
+) : RuntimeException(listOfNotNull("Bad Request", errorDetail, invalidParams).joinToString(": ")) {
```
davidatkinsuk: I know it's not strictly required, but there may be some structural value in introducing a common super type for our problems that extend runtime exception that all problems extend

that can also define an abstract function `fun toProblemDetail(): ProblemDetail`?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/ExceptionHandling.kt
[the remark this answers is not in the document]
rashadmughal: We should fix this in future and do the null check.

rashadmughal (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/ConflictProblem.kt
```diff
@@ -1,9 +1,13 @@
...
+import org.springframework.http.HttpStatus
+import org.springframework.http.ProblemDetail
 
-class ConflictProblem(id: Any, conflictReason: String) : AbstractThrowableProblem(null, "Conflict", Status.CONFLICT, "$conflictReason: $id") {
-  override fun getCause(): Exceptional? = null
+class ConflictProblem(id: Any, conflictReason: String) : RuntimeException("$conflictReason: $id") {
+
+  val msg = "$conflictReason: $id"
+
+  fun toProblemDetail(): ProblemDetail = ProblemDetail.forStatusAndDetail(HttpStatus.CONFLICT, message ?: msg).apply {
```
davidatkinsuk: I propose dropping `msg` and using `message!!`. This is because we know the message is non null as we pass it into the RuntimeException constructor.

This comment applies to other problem classes that define a `msg` val

(with that said, there may be some kotlin/java integration i don't understand taht means this doesn't work - it's hard to trace).

If we were to use a common supertype as mentioned on another comment we could guarantee the message is non null in the constructor and expose it to subtypes as a non-nullable value?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ExceptionHandlingTest.kt
```diff
@@ -460,13 +484,16 @@ class ExceptionHandlingTest : InitialiseDatabasePerClassTestBase() {
...
             "status" : 400,
-            "detail" : "Expected an array but got an object"
+            "detail" : "Expected an array but got an object",
+            "instance" : "/deserialization-test/array",
+            "properties" : null
           }
           """,
       )
-      assertThat(validationResult.responseHeaders.contentType?.toString()).isEqualTo("application/json")
+      assertThat(validationResult.responseHeaders.contentType?.toString()).isEqualTo("application/problem+json")
```
davidatkinsuk, line 491: do we know if this change in content type is going to cause an issue in the UI? (same goes for other tests where this has changed)

Have we tested UIs problem handling (probably easier said than done?)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/FlattenedProblemDetail.kt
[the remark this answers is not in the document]
davidatkinsuk: Could we please add some documentation to the class to explain how it adds to both levels and why (legacy stuff). I think it can be mostly be a copy paste of the above?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/FlattenedProblemDetail.kt
```diff
@@ -0,0 +1,14 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.problem
+
+import com.fasterxml.jackson.annotation.JsonAnyGetter
+import org.springframework.http.HttpStatus
+import org.springframework.http.ProblemDetail
+
+class FlattenedProblemDetail(status: HttpStatus, detail: String) : ProblemDetail(status.value()) {
```
davidatkinsuk: Given this is only used by `BadRequestProblem` can i propose we make this less generic in an attempt to make its purpose a bit clearer

E.g.

1. Rename to `InvalidParamsProblemDetail`
2. Add an additional constructor argument - `val invalidParams: Map<String, ParamDetails>? = null`
3. Change getFlattenedProperties() as follows

```
@JsonAnyGetter
fun getAdditionalTopLevelProperties() = mapOf("invalid-params", invalidParams)
```

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/ForbiddenProblem.kt
```diff
@@ -1,15 +1,19 @@
...
+  val title = "Forbidden"
+
+  fun toProblemDetail(): ProblemDetail = ProblemDetail.forStatusAndDetail(HttpStatus.FORBIDDEN, message ?: msg).apply {
+    title = this@ForbiddenProblem.title
+    status = HttpStatus.FORBIDDEN.value()
+    detail = msg
+  }
+
+  override val message: String?
+    get() = "$title: $msg"
```
davidatkinsuk: i don't think this is required? It doesn't seem to materialise anywhere. 

same comment for other places we override message

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ExceptionHandlingTest.kt
davidatkinsuk: I tried using `.returnResult<String>()` locally (on this branch) and it worked fine. Here's what i have...

```
    @Test
    fun `Returns 401 unauthenticated when AuthenticationCredentialsNotFoundException is thrown`() {
      val jwt = jwtAuthHelper.createValidAuthorizationCodeJwt()

      val validationResult = webTestClient.get()
        .uri("/authentication-credentials-not-found-exception")
        .exchange()
        .returnResult<String>()

      assertJsonEquals(
        actual = validationResult.responseBody.blockFirst(),
        expected = """
          {
            "type" : "about:blank",
            "title" : "Unauthenticated",
            "status" : 401,
            "detail" : "A valid HMPPS Auth JWT must be supplied via bearer authentication to access this endpoint",
            "instance" : "/authentication-credentials-not-found-exception",
            "properties" : null
          }
          """,
      )

      assertThat(validationResult.responseHeaders.contentType?.toString()).isEqualTo("application/problem+json")
    }
```

davidatkinsuk (commented):
Functionally looks good but there are a few comment i'd like to discuss / resolve before merging if that's ok.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/controller/Cas2AssessmentsController.kt
danhumphreys-moj: Removed this function now and replaced calling code to use `throw()` instead of `throwProblem()`.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/BadRequestProblem.kt
danhumphreys-moj: Added an abstract ApiProblem super class and problem classes now extend it:

```
package uk.gov.justice.digital.hmpps.approvedpremisesapi.problem

import org.springframework.http.ProblemDetail

abstract class ApiProblem(override val message: String) : RuntimeException(message) {
  abstract fun toProblemDetail(): ProblemDetail?
}
```

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ExceptionHandlingTest.kt
danhumphreys-moj: Argh.  Don't know what happened here.  You're right.  I've reverted back to using `.returnResult<String>()`.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/ConflictProblem.kt
danhumphreys-moj: Have refactored sub classes to use non null `message` inherited from super class `ApiProblem`.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/FlattenedProblemDetail.kt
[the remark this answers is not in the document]
danhumphreys-moj: Added some javadoc to `FlattenedProblemDetail`.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/FlattenedProblemDetail.kt
danhumphreys-moj: Made these changes in a recent commit [here](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/4633/changes/703fc71d377999cb09b5bdc25ad730deaca97ba1).

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/problem/ForbiddenProblem.kt
danhumphreys-moj: I have removed this from this problem type (and others) and updated dependent unit tests.

The reason being that since we are now inheriting non null message from the super type if we leave this custom getter in place we end up with:

```
  override val message: String?
    get() = "$title: $message"
```

Which causes recursive issues (getting message gets message etc).  I thought it better to follow the logic of having sub classes use a non null message attribute from the super class at the sacrifice of updating some unit tests.  I would think this has no impact on functionality and communication with UI.

One for discussion potentially.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ExceptionHandlingTest.kt
danhumphreys-moj: Tested this with @davidatkinsuk this morning.  See here for test evidence - https://dsdmoj.atlassian.net/browse/FM-495?focusedCommentId=767265

PR #4636, opened 2026-04-22
Simplify renovate config

davidatkinsuk:
We currently override various hmpps renovate settings despite the inherited settings being identical

Further more, we are not making use of the `github-actions` manager because we override the enabled managers array.

This commit removes unneccessary overrides


daveawc (approved):
👍

PR #4639, opened 2026-04-23
SAS-467: Referral history API

rashadmughal:
Adds additional fields to the “external referrals / referral history” API responses across CAS1/CAS2/CAS2v2/CAS3, including referral metadata and placement/booking details.

Changes:

Extend referral history response models with rejection reason, LA area, PDU, referred-by, and placement details.
Populate the new fields in CAS1/CAS3 transformers using latest booking/space-booking lookups.
Update integration tests to assert the expanded response shapes.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/transformer/Cas2ApplicationsTransformer.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/transformer/Cas2ApplicationsTransformer.kt
@@ -7,6 +7,8 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ApplicationSta
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ServiceType
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas2.jpa.entity.Cas2ApplicationEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas2.jpa.entity.Cas2ApplicationSummaryEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas2.jpa.entity.Cas2StatusUpdateEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas2.jpa.entity.Cas2StatusUpdateNonAssignable
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas2.model.Cas2Application
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas2.model.Cas2ApplicationSummary
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas2.model.Cas2ReferralHistory
@@ -84,13 +86,31 @@ class Cas2ApplicationsTransformer(
 
   fun transformJpaToCas2ReferralHistory(
     jpa: Cas2ApplicationEntity,
-  ) = Cas2ReferralHistory(
-    id = jpa.assessment!!.id,
-    applicationId = jpa.id,
-    type = ServiceType.CAS2,
-    createdAt = jpa.submittedAt!!.toInstant(),
-    status = jpa.statusUpdates!!.first().label,
-  )
+  ): Cas2ReferralHistory {
+    val latestStatusUpdate = getReferralHistoryStatus(jpa)
+    val rejectionReason = latestStatusUpdate
+      ?.takeIf { it.label in listOf(Cas2StatusUpdateNonAssignable.REFERRAL_CANCELLED.label, Cas2StatusUpdateNonAssignable.REFERRAL_WITHDRAWN.label) }
+      ?.label
+
+    val omu = jpa.referringPrisonCode?.let { offenderManagementUnitRepository.findByPrisonCode(it) }
+    val placementAddress = omu?.prisonName ?: jpa.referringPrisonCode
+
+    return Cas2ReferralHistory(
+      id = jpa.assessment!!.id,
+      applicationId = jpa.id,
+      type = ServiceType.CAS2,
+      createdAt = jpa.submittedAt!!.toInstant(),
+      status = jpa.statusUpdates!!.first().label,
+      referralRejectionReason = rejectionReason,
+      localAuthorityArea = placementAddress,
+      pdu = jpa.preferredAreas,
+      referredBy = jpa.createdByUser.name,
+      placementAddress = placementAddress,
+      placementStatus = latestStatusUpdate?.label,
+    )
+  }
+
+  private fun getReferralHistoryStatus(jpa: Cas2ApplicationEntity): Cas2StatusUpdateEntity? = jpa.statusUpdates?.firstOrNull()
 
   private fun getStatus(entity: Cas2ApplicationEntity): ApplicationStatus {
     if (entity.submittedAt !== null) {
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2v2/transformer/Cas2v2ApplicationsTransformer.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2v2/transformer/Cas2v2ApplicationsTransformer.kt
@@ -12,6 +12,8 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Person
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ServiceType
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas2.jpa.entity.Cas2ApplicationEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas2.jpa.entity.Cas2ApplicationSummaryEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas2.jpa.entity.Cas2StatusUpdateNonAssignable
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.OffenderManagementUnitRepository
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.PersonInfoResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.PersonTransformer
 import java.util.UUID
@@ -24,6 +26,7 @@ class Cas2v2ApplicationsTransformer(
   private val statusUpdateTransformer: Cas2v2StatusUpdateTransformer,
   private val timelineEventsTransformer: Cas2v2TimelineEventsTransformer,
   private val cas2v2AssessmentsTransformer: Cas2v2AssessmentsTransformer,
+  private val offenderManagementUnitRepository: OffenderManagementUnitRepository,
 ) {
 
   fun transformJpaToApi(jpa: Cas2ApplicationEntity, personInfo: PersonInfoResult): Cas2v2Application = transformJpaAndFullPersonToApi(jpa, personTransformer.transformModelToPersonApi(personInfo))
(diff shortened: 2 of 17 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas1/Cas1AssessmentTransformer.kt
```diff
@@ -73,13 +76,29 @@ class Cas1AssessmentTransformer(
...
 
-  fun transformDomainToApiCas1ReferralHistory(entity: ApprovedPremisesAssessmentEntity): Cas1ReferralHistory = Cas1ReferralHistory(
-    id = entity.id,
-    applicationId = entity.application.id,
-    createdAt = entity.createdAt.toInstant(),
-    status = getStatusForCas1Assessment(entity),
-    type = ServiceType.CAS1,
-  )
+  fun transformDomainToApiCas1ReferralHistory(entity: ApprovedPremisesAssessmentEntity): Cas1ReferralHistory {
+    val application = entity.application as ApprovedPremisesApplicationEntity
```
davidatkinsuk: could use `entity.cas1Application()` instead

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas1SpaceBookingEntity.kt
```diff
@@ -299,6 +300,15 @@ interface Cas1SpaceBookingRepository : JpaRepository<Cas1SpaceBookingEntity, UUI
 
   fun findAllByApplication(application: ApplicationEntity): List<Cas1SpaceBookingEntity>
 
+  @Query(
+    "SELECT * FROM bookings WHERE application_id = :applicationId AND service = 'approved-premises' ORDER BY created_at DESC LIMIT 1",
```
davidatkinsuk: we use the cas1_space_bookings table so this won't ever return anything. I think we need to consider integration test coverage too as this is a new query.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/Cas1SpaceBookingEntity.kt
```diff
@@ -299,6 +300,15 @@ interface Cas1SpaceBookingRepository : JpaRepository<Cas1SpaceBookingEntity, UUI
...
+    nativeQuery = true,
+  )
+  fun findCas1BookingEntity(
```
davidatkinsuk: could this please be renamed `findLatestCas1SpaceBooking`()?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas1/Cas1AssessmentTransformer.kt
```diff
@@ -73,13 +76,29 @@ class Cas1AssessmentTransformer(
...
+
+    val latestBooking = cas1SpaceBookingRepository.findCas1BookingEntity(application.id, ServiceName.approvedPremises.value)
```
davidatkinsuk: I'd rather we had a function on the `Cas1SpaceBookingService` to return the latest placement to try and start removing use of non reference data repositories to be exclusive to domain-relevant services. I appreciate the pattern used here is typical of the code base but it would be good to start shifting business logic like this out of transformers and into services that should own it.

Are we actually looking for the 'current' booking here rather than the most recently created? Someone can have a series of future bookings (e.g. ROTLs) and this logic may pick up a future booking when they're currently resident somewhere?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/transformer/Cas2ApplicationsTransformer.kt
```diff
@@ -89,13 +90,29 @@ class Cas2ApplicationsTransformer(
...
-    status = jpa.statusUpdates!!.first().label,
-  )
+  ): Cas2ReferralHistory {
+    val latestStatusUpdate = jpa.statusUpdates?.firstOrNull()
+    val rejectionReason = latestStatusUpdate
+      ?.takeIf { it.label in listOf(Cas2StatusUpdateNonAssignable.REFERRAL_CANCELLED.label, Cas2StatusUpdateNonAssignable.REFERRAL_WITHDRAWN.label) }
+      ?.label
+
+    val omu = jpa.referringPrisonCode?.let { offenderManagementUnitRepository.findByPrisonCode(it) }
+    val placementAddress = omu?.prisonName ?: jpa.referringPrisonCode
```
davidatkinsuk, line 96: do we expect an entry to not exist? If not i think this should error?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/transformer/Cas2ApplicationsTransformer.kt
```diff
@@ -89,13 +90,29 @@ class Cas2ApplicationsTransformer(
...
+
+    return Cas2ReferralHistory(
+      id = jpa.assessment!!.id,
+      applicationId = jpa.id,
+      type = ServiceType.CAS2,
+      createdAt = jpa.submittedAt!!.toInstant(),
+      status = latestStatusUpdate?.label ?: "",
+      referralRejectionReason = rejectionReason,
+      localAuthorityArea = omu?.prisonName ?: jpa.referringPrisonCode,
```
davidatkinsuk: this repeats logic from line 100. Maybe we could move this definition to below line 100, or just use placementAddress here?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/transformer/Cas2ApplicationsTransformer.kt
```diff
@@ -89,13 +89,26 @@ class Cas2ApplicationsTransformer(
...
+      ?.takeIf { it.label in listOf("Referral cancelled", "Referral withdrawn") }
+      ?.label
+
+    return Cas2ReferralHistory(
+      id = jpa.assessment!!.id,
+      applicationId = jpa.id,
+      type = ServiceType.CAS2,
+      createdAt = jpa.submittedAt!!.toInstant(),
+      status = latestStatusUpdate?.label ?: "",
+      referralRejectionReason = rejectionReason,
```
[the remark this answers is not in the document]
davidatkinsuk: I agree, we really shouldn't be using empty strings in the absence of something. Either we need to update the data model or we need an explicit status defined when status is unknown. Have we established a pattern for this?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2v2/transformer/Cas2v2ApplicationsTransformer.kt
```diff
@@ -89,13 +92,29 @@ class Cas2v2ApplicationsTransformer(
...
   fun transformJpaToCas2v2ReferralHistory(
     jpa: Cas2ApplicationEntity,
-  ) = Cas2v2ReferralHistory(
-    id = jpa.assessment!!.id,
-    applicationId = jpa.id,
-    type = ServiceType.CAS2v2,
-    createdAt = jpa.submittedAt!!.toInstant(),
-    status = jpa.statusUpdates!!.first().label,
-  )
+  ): Cas2v2ReferralHistory {
```
davidatkinsuk, line 92: this code is almost identical to the CAS2 version, but with a different return type. Presumably we can look at sharing some of this logic given this stuff is derived from `Cas2ApplicationEntity`. Maybe we can move some logic into functions on the entity?

I appreciate that this is an existing pattern, but would be good to try and reduce copy-paste going forward if possible.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2v2/transformer/Cas2v2ApplicationsTransformer.kt
```diff
@@ -89,13 +89,26 @@ class Cas2v2ApplicationsTransformer(
...
+    val rejectionReason = latestStatusUpdate
+      ?.takeIf { it.label in listOf("Referral cancelled", "Referral withdrawn") }
+      ?.label
+
+    return Cas2v2ReferralHistory(
+      id = jpa.assessment!!.id,
+      applicationId = jpa.id,
+      type = ServiceType.CAS2v2,
+      createdAt = jpa.submittedAt!!.toInstant(),
+      status = latestStatusUpdate?.label ?: "",
```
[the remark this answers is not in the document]
davidatkinsuk: I agree, see my prior comment on the cas2 transformer

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2v2/transformer/Cas2v2ApplicationsTransformer.kt
```diff
@@ -89,13 +92,29 @@ class Cas2v2ApplicationsTransformer(
...
+    val placementAddress = omu?.prisonName ?: jpa.referringPrisonCode
+
+    return Cas2v2ReferralHistory(
+      id = jpa.assessment!!.id,
+      applicationId = jpa.id,
+      type = ServiceType.CAS2v2,
+      createdAt = jpa.submittedAt!!.toInstant(),
+      status = latestStatusUpdate?.label ?: "",
+      referralRejectionReason = rejectionReason,
+      localAuthorityArea = omu?.prisonName ?: jpa.referringPrisonCode,
```
davidatkinsuk: see prior comment on cas2 transformer

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/transformer/Cas3AssessmentTransformer.kt
```diff
@@ -78,13 +80,29 @@ class Cas3AssessmentTransformer(
...
 
-  fun transformAssessmentToCas3ReferralHistory(a: TemporaryAccommodationAssessmentEntity): Cas3ReferralHistory = Cas3ReferralHistory(
-    id = a.id,
-    applicationId = a.application.id,
-    createdAt = a.createdAt.toInstant(),
-    status = a.deriveAssessmentStatus(),
-    type = ServiceType.CAS3,
-  )
+  fun transformAssessmentToCas3ReferralHistory(a: TemporaryAccommodationAssessmentEntity): Cas3ReferralHistory {
+    val application = a.typedApplication<TemporaryAccommodationApplicationEntity>()
```
davidatkinsuk, line 84: shall we add a function onto the assessment like we have to get the cas1 application? Although i'm a little confused why the relationship isn't typed to a CAS3 application, but don't worry about that :)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/transformer/Cas3AssessmentTransformer.kt
```diff
@@ -78,13 +80,29 @@ class Cas3AssessmentTransformer(
...
+
+    val latestBooking = bookingRepository.findCas3BookingEntity(application.id, ServiceName.temporaryAccommodation.value)
```
davidatkinsuk: as per cas1, i'd prefer we had a function on a CAS3 booking service to retrieve the latest (or current?) booking

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BookingEntity.kt
```diff
@@ -115,6 +116,15 @@ interface BookingRepository : JpaRepository<BookingEntity, UUID> {
 
   fun findAllByApplication(application: ApplicationEntity): List<BookingEntity>
 
+  @Query(
+    "SELECT * FROM bookings WHERE application_id = :applicationId AND service = :service ORDER BY created_at DESC LIMIT 1",
```
davidatkinsuk, line 120: I think this needs to use a cas3 bookings table as bookings is now defunct? I think we need to consider integration test coverage too as this is a new query.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BookingEntity.kt
```diff
@@ -115,6 +116,15 @@ interface BookingRepository : JpaRepository<BookingEntity, UUID> {
...
+    nativeQuery = true,
+  )
+  fun findCas3BookingEntity(
```
davidatkinsuk: rename to find latest?

davidatkinsuk (commented):
Some comments from me, noting that i'm not as familiar with CAS2/3 and SAS context so I may have missed something.

Going forward would it be possible to split changes into commits where there's a clear functional delineation e.g. a commit for each CAS?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/BookingEntity.kt
davidatkinsuk: Ignore me! i was confused and didn't realise that `Cas3Entity` was backed by the `bookings` table still. Strictly speaking you shouldn't need to filter on service because there will only be CAS3 entries in there

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2v2/transformer/Cas2v2ApplicationsTransformer.kt
daveawc: also, is cas2v2 still being merged into cas2? so might make sense to do it now rather than create more tech debt for later (I think we've already merged the two models in SAS)

maccalsa (approved):
Logically this looks fine I agree with David, if we could at least break the code down vertically in separate commits, that would help. I find if you do this a quick highlight the purpose of each commit in the description helps.

LGTM

PR #4640, opened 2026-04-23
FM-508 - Upgrade to latest spring boot 3 version

davidatkinsuk:
By upgrading the `gradle-spring-boot` plugin to the latest version 9.* version, we get the following dependency upgrades:

* spring boot 3.5.13
* kotlin 2.3.10

The spring bootup grade allows us to subsequently upgrade to JDK 25.

This commit also includes an upgrade to the latest versions of hmpps SQS and Spring Boot Starter libraries to ensure the latest versions built against spring 3.x are used.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,11 +1,10 @@
 
 plugins {
-  // this adds the appinsights agent (see AppInsightsConfigManager.kt)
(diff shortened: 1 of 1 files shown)


daveawc (approved):
nice

PR #4642, opened 2026-04-24
FM-491: Site Closure Seed Job

vbala-moj:
Added a seed job for site closures.
- Closes available beds up to the specified closure date
- Updates the OOSDB end date to match the closure date
- Disables new space bookings for the premises

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/seed/Cas1ClosePremisesSeedJob.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/seed/Cas1ClosePremisesSeedJob.kt
@@ -1,0 +1,81 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.cas1.seed
+
+import org.slf4j.LoggerFactory
+import org.springframework.data.repository.findByIdOrNull
+import org.springframework.stereotype.Component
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/seed/Cas1ClosePremisesSeedJob.kt
```diff
@@ -0,0 +1,84 @@
...
+class Cas1ClosePremisesSeedJob(
+  private val approvedPremisesRepository: ApprovedPremisesRepository,
+  private val bedRepository: BedRepository,
+  private val cas1OutOfServiceBedService: Cas1OutOfServiceBedService,
+) : SeedJob<Cas1ClosePremisesSeedJob.CsvRow>(
+  requiredHeaders = setOf(
+    "premises_id",
+    "closure_date",
+    "notes",
+    "reason_id",
```
davidatkinsuk: The `reason` on an OOSB is there to describe why the bed is out of service. So when setting the end date i think it's unlikely we'll ever want to change it. So in this context, I propose we just drop this argument and keep the existing value?

davidatkinsuk (approved):
Happy for this to be merged to test in pre-prod but could we possibly add an integration test (i'd prefer this over a unit test in this case). Could maybe that on a separate PR?

davidatkinsuk (approved):
IT's look perfect. thank you 👍

PR #4645, opened 2026-04-24
FM-511 - Use HMPPS base Temurin 21 docker image

davidatkinsuk:
This commit updates the pipeline and docker build config to make use of the [hmpps provided temurin 21 image](https://github.com/ministryofjustice/hmpps-base-container-images).

This allows us to simplify our Dockerfile because we are duplicating some configuration already applied in the hmpps base image e.g.

* setting the timezone
* various calls to apt-get to update libraries

This commit also better aligns the project with the hmpps-template build configuration, using the [hmpps template project Dockerfile](https://github.com/ministryofjustice/hmpps-template-kotlin/blob/main/Dockerfile) as the basis for updates. To achieve this we're now building the application JAR  outside of the Dockerfile.

Ideally we’d reuse the JAR produced by our existing build steps but

1. We don’t currently use the hmpps verify step (which includes the upload build artifact option we require)
2. It would remove the ability to concurrently build the docker image alongside code verification (slowing down the build)

These issues should be addressable in a future commit, but the work is too substantial for this upgrade.

We have also had to change the ENV VAR used to set JVM arguments following guidance in the [hmpps image documentation](https://github.com/ministryofjustice/hmpps-base-container-images/tree/main/images/eclipse-temurin). In making this change we have discovered that the previous `JAVA_OPTS` env var was not being applied, and as such the default Max Heap Size configuration was being applied, which is 25% of the available memory. This commit modifies the max heap size to reflect what was actually being used, providing an opportunity to now reduce our container's available memory (in a future commit).


.github/workflows/security_veracode_pipeline_scan.yml
```diff
@@ -6,6 +6,19 @@ on:
     - cron: "51 3 * * MON-FRI" # Every weekday at 03:51 UTC
 
 jobs:
+  create-artifact:
+    name: Build the artifact for veracode to scan
+    uses: ministryofjustice/hmpps-github-actions/.github/workflows/gradle_verify.yml@v2 # WORKFLOW_VERSION
+    with:
+      gradle-command: 'BUILD_NUMBER=veracode ./gradlew assemble'
+      upload-build-artifacts: true
```
danhumphreys-moj, line 14: As I can see these veracode scheduled jobs scan the jar file for vulnerabilities, rather than the source code (c.f. OWASP).

Does this new job build the API jar and upload it to a store - then presumably it's downloaded in the next  `security-veracode-pipeline-scan` job via `veracode_artifact_name`.

I'm curious how this worked before and why we are changing the logic for these cron scheduled jar scanning jobs?

.github/workflows/security_veracode_policy_scan.yml
```diff
@@ -6,6 +6,19 @@ on:
     - cron: "36 3 * * 1" # Every Monday at 03:36 UTC
 
 jobs:
+  create-artifact:
+    name: Build the artifact for veracode to scan
+    uses: ministryofjustice/hmpps-github-actions/.github/workflows/gradle_verify.yml@v2 # WORKFLOW_VERSION
+    with:
+      gradle-command: 'BUILD_NUMBER=veracode ./gradlew assemble'
+      upload-build-artifacts: true
```
danhumphreys-moj, line 14: Similarly here.

.github/workflows/security_veracode_policy_scan.yml
```diff
@@ -6,6 +6,19 @@ on:
...
+      build-artifact-name: 'build-results'
+      upload-test-artifacts: false
+      java-version: "21"
```
danhumphreys-moj, line 17: Will the plan be to use the Java 25 image after this and build with Java 21, then upgrade Java from 21 to 25?  Making build versions and image versions ultimately both 25.

.github/workflows/docker_build.yml
```diff
@@ -5,17 +5,38 @@ on:
...
+  app_version:
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.app_version.outputs.version }}
+    steps:
+      - id: app_version
+        name: Application version creators
+        uses: ministryofjustice/hmpps-github-shared-actions/.github/actions/build-test-and-deploy/create_app_version@fddc5f11d6a8ce9b82a06fe76268835bfd99bcd4 # v1
+
+  gradle_assemble:
```
danhumphreys-moj, line 20: Building of jar now done here within a pipeline job rather than in Dockerfile and then jar uploaded to Github Artifacts.  Presumably a forthcoming PR will use this uploaded artifact?

rashadmughal (approved):
If it builds and runs and deploys to dev we should be OK?

davidatkinsuk:
> If it builds and runs and deploys to dev we should be OK?

I think so, only concern is getting the memory settings correct but it should be pretty obvious if they're not!

danhumphreys-moj (approved):
Some comments - though I'm confused as to how we're actually using artifacts/images as it stands and what the intention is.  I'll reach out to @davidatkinsuk for understanding.

.github/workflows/docker_build.yml
davidatkinsuk: Ideally we would do that. It's worth noting that this change doesn't modify behaviour wrt. reusing JARs. Our currently workflows do not use the same JAR generated in the verify/test step when building the docker image (because these two things run concurrently)

Reusing the JAR from the verify/test step would slow down the build a little.

.github/workflows/security_veracode_pipeline_scan.yml
davidatkinsuk: > Does this new job build the API jar and upload it to a store - then presumably it's downloaded in the next security-veracode-pipeline-scan job via veracode_artifact_name.

That's correct

> I'm curious how this worked before and why we are changing the logic for these cron scheduled jar scanning jobs?

If no pre built file is defined via the `veracode_package_artifact_name` and `veracode_artifact_name` arguments, the vericode steps would build the Dockerfile. This used to work because the calls to `./gradlew` were in the Dockerfile, but this is no longer the case

.github/workflows/security_veracode_policy_scan.yml
davidatkinsuk: see prior response

.github/workflows/security_veracode_policy_scan.yml
davidatkinsuk: I'm planning on using Java 25 to both compile and run the code. We could break it down into separate steps, but i feel like that may cause more issues than just compiling and running using Java 25 at the same time (which isn't a huge change to implement)

daveawc (approved):

PR #4647, opened 2026-04-24
FM-495 Add additional integration test coverage for global exception handling

danhumphreys-moj:

Added additional integration test coverage that explicitly tests JSON responses for global exception handling.

This is extra risk mitigation work that increases our confidence the upcoming migration from Zalando to Spring Frameworks will proceed smoothly and was also recommended during a recent "All Dev Checkin" meeting.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ExceptionHandlingTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ExceptionHandlingTest.kt
@@ -1,24 +1,35 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.integration
 
+import com.fasterxml.jackson.databind.JsonNode
+import com.fasterxml.jackson.databind.node.ObjectNode
+import com.fasterxml.jackson.module.kotlin.jsonMapper
 import org.assertj.core.api.Assertions.assertThat
 import org.hibernate.exception.JDBCConnectionException
 import org.junit.jupiter.api.Nested
 import org.junit.jupiter.api.Test
+import org.springframework.http.HttpMethod
 import org.springframework.http.ResponseEntity
+import org.springframework.security.authentication.AuthenticationCredentialsNotFoundException
 import org.springframework.test.web.reactive.server.returnResult
+import org.springframework.web.bind.MissingServletRequestParameterException
 import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.bind.annotation.PostMapping
 import org.springframework.web.bind.annotation.RequestBody
+import org.springframework.web.bind.annotation.RequestHeader
 import org.springframework.web.bind.annotation.RequestParam
 import org.springframework.web.bind.annotation.RestController
+import org.springframework.web.method.annotation.MethodArgumentTypeMismatchException
+import org.springframework.web.servlet.resource.NoResourceFoundException
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.InvalidParam
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ValidationError
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.util.assertJsonEquals
 import java.sql.SQLException
 import java.time.Instant
 import java.time.LocalDate
 import java.time.LocalDateTime
 import java.time.OffsetDateTime
 import java.util.UUID
+import org.springframework.security.access.AccessDeniedException as SpringAccessDeniedException
 
 class ExceptionHandlingTest : InitialiseDatabasePerClassTestBase() {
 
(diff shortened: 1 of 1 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ExceptionHandlingTest.kt
```diff
@@ -1,18 +1,29 @@
...
 import org.springframework.web.bind.annotation.PostMapping
 import org.springframework.web.bind.annotation.RequestBody
 import org.springframework.web.bind.annotation.RequestParam
 import org.springframework.web.bind.annotation.RestController
+import org.springframework.web.method.annotation.MethodArgumentTypeMismatchException
+import org.springframework.web.servlet.resource.NoResourceFoundException
+import org.springframework.security.access.AccessDeniedException as SpringAccessDeniedException
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.InvalidParam
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ValidationError
+import java.io.File
```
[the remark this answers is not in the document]
danhumphreys-moj: Gone now.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ExceptionHandlingTest.kt
```diff
@@ -321,6 +332,204 @@ class ExceptionHandlingTest : InitialiseDatabasePerClassTestBase() {
...
   }
+
+  @Test
+  fun `Returns 401 unauthenticated when AuthenticationCredentialsNotFoundException is thrown`() {
+    val jwt = jwtAuthHelper.createValidAuthorizationCodeJwt()
+
+    val validationResult = webTestClient.get()
+      .uri("/authentication-credentials-not-found-exception")
+      .header("Authorization", "Bearer $jwt")
+      .exchange()
```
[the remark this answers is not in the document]
danhumphreys-moj: I don't think there is a need to do this, the JSON is explicitly tests so the status checks are inherent in these.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ExceptionHandlingTest.kt
```diff
@@ -321,6 +332,204 @@ class ExceptionHandlingTest : InitialiseDatabasePerClassTestBase() {
...
+    val prettyJsonResponse = """
+      {
+        "title" : "Unauthenticated",
+        "status" : 401,
+        "detail" : "A valid HMPPS Auth JWT must be supplied via bearer authentication to access this endpoint"
+      }
+    """.trimIndent()
+    val compactJsonResponse = jsonMapper.writeValueAsString(jsonMapper.readTree(prettyJsonResponse))
+
+    assertThat(validationResult).isEqualTo(compactJsonResponse)
```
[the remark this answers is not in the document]
danhumphreys-moj: The intention with these tests is to explicitly test the JSON response.  The plan is that these tests are added now, then we will remove the dependence on Zalando Problem, then we can check that the response back to the client when these exceptions are raised is the same.  Seeing the JSON explicitly in these tests makes this super clear in the next PR.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ExceptionHandlingTest.kt
```diff
@@ -321,6 +332,204 @@ class ExceptionHandlingTest : InitialiseDatabasePerClassTestBase() {
...
+
+    val validationResult = webTestClient.get()
+      .uri("/method-argument-type-mismatch-exception")
+      .header("Authorization", "Bearer $jwt")
+      .exchange()
+      .returnResult<JsonNode>()
+      .responseBody
+      .blockFirst()
+
+    val validationResultWithDummyTimestamp = jsonMapper.writeValueAsString((validationResult as? ObjectNode)?.put("timestamp", "2026-04-24T08:11:54.563+00:00"))
```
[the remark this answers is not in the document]
danhumphreys-moj: I think this is ok and have left this as fixing the timestamp to a hard coded value.  I have changed the logic in the one test where a timestamp is returned in the JSON to check that the timestamp exists before replacing it (and to `error` if it doesn't exist).

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ExceptionHandlingTest.kt
```diff
@@ -321,6 +332,204 @@ class ExceptionHandlingTest : InitialiseDatabasePerClassTestBase() {
...
+        "timestamp" : "2026-04-24T08:11:54.563+00:00",
+        "status" : 400,
+        "error" : "Bad Request",
+        "message" : "Method parameter 'requiredProperty': Failed to convert value of type 'java.lang.String' to required type 'int'",
+        "path" : "/method-argument-type-mismatch-exception"
+      }
+    """.trimIndent()
+    val compactJsonResponse = jsonMapper.writeValueAsString(jsonMapper.readTree(prettyJsonResponse))
+
+    assertThat(validationResultWithDummyTimestamp).isEqualTo(compactJsonResponse)
```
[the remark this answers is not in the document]
danhumphreys-moj: This is what we want for these tests, we are making various changes and these explicit JSON String comparisons will make it obvious if there are any changes here regarding the handling of global exceptions.  Having the String JSON verbatim in these tests means we can check the differences (if there are any) in the forthcoming migration from Zalando.  We are checking here that what is returned to the client when an exception is thrown and captured by the handler remains the same before and after migration.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ExceptionHandlingTest.kt
```diff
@@ -233,6 +243,397 @@ class ExceptionHandlingTest : InitialiseDatabasePerClassTestBase() {
...
+    fun `Returns 401 unauthenticated when AuthenticationCredentialsNotFoundException is thrown`() {
+      val jwt = jwtAuthHelper.createValidAuthorizationCodeJwt()
+
+      val validationResult = webTestClient.get()
+        .uri("/authentication-credentials-not-found-exception")
+        .header("Authorization", "Bearer $jwt")
+        .exchange()
+        .returnResult<String>()
+        .responseBody
+        .blockFirst()
```
davidatkinsuk: you can replace this chain of calls in all tests with this extension function - `.bodyAsObject<String>()`

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ExceptionHandlingTest.kt
```diff
@@ -233,6 +243,397 @@ class ExceptionHandlingTest : InitialiseDatabasePerClassTestBase() {
...
+      {
+        "title" : "Bad Request",
+        "status" : 400,
+        "detail" : "Missing required query parameter requiredProperty",
+        "invalid-params": []
+      }
+      """.trimIndent()
+      val compactJsonResponse = jsonMapper.writeValueAsString(jsonMapper.readTree(prettyJsonResponse))
+
+      assertThat(validationResult).isEqualTo(compactJsonResponse)
```
davidatkinsuk: There's an existing function called `assertJsonEquals`. I'm not sure that will do the job here and in other tests? I suspect it may give more helpful errors if comparison fails?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ExceptionHandlingTest.kt
```diff
@@ -233,6 +243,397 @@ class ExceptionHandlingTest : InitialiseDatabasePerClassTestBase() {
...
+      val prettyJsonResponse = """
+      {
+        "title" : "Unauthenticated",
+        "status" : 401,
+        "detail" : "A valid HMPPS Auth JWT must be supplied via bearer authentication to access this endpoint"
+      }
+      """.trimIndent()
+      val compactJsonResponse = jsonMapper.writeValueAsString(jsonMapper.readTree(prettyJsonResponse))
+
+      assertThat(validationResult).isEqualTo(compactJsonResponse)
```
rashadmughal: Can you add assertions for the content type header returned from the service? It should be the correct problem type.

rashadmughal (changes requested):

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ExceptionHandlingTest.kt
```diff
@@ -233,6 +243,397 @@ class ExceptionHandlingTest : InitialiseDatabasePerClassTestBase() {
...
+
+      val prettyJsonResponse = """
+      {
+        "timestamp" : "2026-04-24T08:11:54.563+00:00",
+        "status" : 400,
+        "error" : "Bad Request",
+        "message" : "Method parameter 'requiredProperty': Failed to convert value of type 'java.lang.String' to required type 'int'",
+        "path" : "/method-argument-type-mismatch-exception"
+      }
+      """.trimIndent()
```
davidatkinsuk, line 385: probably not one for this PR but this doesn't appear to use the common problem structure?

davidatkinsuk (approved):
looks good. a few minor proposals from me.

I guess we should probably remove the unit tests you previously added (as they won't work with the rewritten exception handler?)

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ExceptionHandlingTest.kt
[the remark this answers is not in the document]
davidatkinsuk: I think co-pilot is hinting at the same issue i'm mentioned in my comment - the issue is comparing JSON as a straight string comparison can produce difficult to parse errors. If you compare them as 1st class JSON objects failures can be reported better e.g. `attribute a value 'x' doesn't match expected value 'y'`. My comment points at a function you can use to do that. You can still pass the expected JSON in as a String

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ExceptionHandlingTest.kt
davidatkinsuk: (this would then 'solve' the timestamp problem because there won't be one anymore :) )

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ExceptionHandlingTest.kt
danhumphreys-moj: Added the content type assertions: 
`assertThat(validationResult.responseHeaders.contentType?.toString()).isEqualTo("application/problem+json")`
However not all tests return `application/problem+json`, some return `application/json`.  I've left these as they are for now so we can remove the dependency initially and investigate later.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ExceptionHandlingTest.kt
danhumphreys-moj: I haven't made this change since we need `validationResult.responseHeaders` to assert the type returned now.  Would have made this change if it wasn't for this.  Thanks.

davidatkinsuk: makes sense, header check is more important 👍

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ExceptionHandlingTest.kt
danhumphreys-moj: I've updated the tests to use `assertJsonEquals`.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ExceptionHandlingTest.kt
danhumphreys-moj: Yes, this is an example where the return type is `json` instead of `problem+json`.  The test asserts this for now:

`assertThat(validationResult.responseHeaders.contentType?.toString()).isEqualTo("application/json")`

Hopefully we can fix once we remove the dependency on Zalando.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/ExceptionHandlingTest.kt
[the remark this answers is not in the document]
danhumphreys-moj: I have used `assertJsonEquals` instead so this should resolve this conversation.

rashadmughal (approved):

danhumphreys-moj:
> looks good. a few minor proposals from me.
> 
> I guess we should probably remove the unit tests you previously added (as they won't work with the rewritten exception handler?)

@davidatkinsuk We haven't re-written the exception handler yet - this is my next PR to look at now that the unit and integration tests are in: https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/4604

PR #4656, opened 2026-04-28
FM-480 Remove Cas2v2IntegrationTestBase.kt

vbala-moj:
https://dsdmoj.atlassian.net/browse/FM-480 (PR-1)

- Removed Cas2v2IntegrationTestBase.kt and purged all legacy CAS2V2 repository/factory references from IntegrationTestBase.kt.
- Updated all 11 CAS2V2 integration test files to extend IntegrationTestBase directly and use unified models.
- Replaced legacy Cas2v2ApplicationEntity and Cas2v2UserEntity usage in tests with Cas2ApplicationEntity and Cas2UserEntity.


daveawc (approved):
👍

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/integration/Cas2ReportsTest.kt
```diff
@@ -465,13 +467,14 @@ class Cas2ReportsTest : IntegrationTestBase() {
...
-      cas2v2ApplicationEntityFactory.produceAndPersist {
+      cas2ApplicationEntityFactory.produceAndPersist {
         withId(event4.applicationId!!)
         withCreatedByUser(user2v2)
         withCrn(event4Details.personReference.crn.toString())
         withNomsNumber(event2Details.personReference.noms)
         withData("{}")
         withReferringPrisonCode("NEW")
+        withServiceOrigin(Cas2ServiceOrigin.BAIL)
       }
```
davidatkinsuk, line 478: am i correct in saying that the intention of this entity is to make sure it bail entries don't appear in the report? 

And this test change would have ideally been made when everything was moved into the cas2 database tables (but understandably was missed)?

vbala-moj: I think so. I believe it was because the test was using the CAS2V2 version of the factory.

davidatkinsuk: great, thank you

PR #4664, opened 2026-04-29
FM-495 Bump `javers-core` dependency to version 7.10.0

danhumphreys-moj:
Bump `javers-core` dependency to version 7.10.0


davidatkinsuk (approved):
LGTM 👍

PR #4667, opened 2026-04-29
FM-480 Remove CAS2V2 report repositories

vbala-moj:
Refactored CAS2 report repositories to include CAS2V2 report fields and accept service origin as a parameter 
Removed duplicate CAS2V2 report repositories


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/jpa/entity/Cas2SubmittedApplicationReport.kt
```diff
@@ -25,28 +25,27 @@ interface Cas2SubmittedApplicationReportRepository : JpaRepository<DomainEventEn
         TO_CHAR(events.occurred_at,'YYYY-MM-DD"T"HH24:MI:SS') AS submittedAt,
         TO_CHAR(applications.created_at, 'YYYY-MM-DD"T"HH24:MI:SS') AS startedAt,
```
davidatkinsuk, line 26: i'm slightly concerned that the v2 version was forcing these into UTC, which is no longer the case. That is - once this is merged dates in BST will appear to 'shift' by an hour in the report. Do you agree?

(I also find it hard to believe that the user _wants_ these values as UTC, we may have to confirm that)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2v2/jpa/entity/Cas2v2SubmittedApplicationReportRepository.kt
```diff
@@ -1,57 +0,0 @@
...
-        CAST(events.data -> 'eventDetails' ->> 'conditionalReleaseDate' as DATE) AS conditionalReleaseDate,
-        TO_CHAR(events.occurred_at AT TIME ZONE 'UTC','YYYY-MM-DD"T"HH24:MI:SS') AS submittedAt,
-        TO_CHAR(applications.created_at AT TIME ZONE 'UTC', 'YYYY-MM-DD"T"HH24:MI:SS') AS startedAt,
-        applications.application_origin as applicationOrigin,
-        applications.service_origin as serviceOrigin,
-        CAST(applications.bail_hearing_date as DATE) as bailHearingDate
-      FROM domain_events events
-      INNER JOIN cas_2_applications applications ON events.application_id = applications.id and applications.service_origin = 'BAIL'     
-      WHERE events.type = 'CAS2_APPLICATION_SUBMITTED'
-        AND applications.submitted_at IS NOT NULL
```
davidatkinsuk, line 33: I guess we're in agreement that this line is redundant (it doesn't appear in the other version, and should be set presumably if there's a 'CAS2_APPLICATION_SUBMITTED' event?)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/jpa/entity/Cas2SubmittedApplicationReport.kt
```diff
@@ -25,28 +25,27 @@ interface Cas2SubmittedApplicationReportRepository : JpaRepository<DomainEventEn
...
         applications.application_origin as applicationOrigin,
-        applications.service_origin as serviceOrigin,
-        CAST(applications.bail_hearing_date as DATE) as bailHearingDate,
+        TO_CHAR(applications.bail_hearing_date, 'YYYY-MM-DD') as bailHearingDate,
```
davidatkinsuk: could you please explain this was changed from a `date` to a `string`. I must be missing something :)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/jpa/entity/Cas2UnsubmittedApplicationsReport.kt
```diff
@@ -16,18 +16,17 @@ interface Cas2UnsubmittedApplicationsReportRepository : JpaRepository<Cas2Applic
         applications.noms_number AS personNoms,
         to_char(applications.created_at, 'YYYY-MM-DD"T"HH24:MI:SS') AS startedAt,
         users.username AS startedBy,
-        applications.application_origin AS applicationOrigin,
-        applications.service_origin AS serviceOrigin
+        applications.application_origin AS applicationOrigin
       FROM cas_2_applications applications
-      JOIN cas_2_users users ON users.id = applications.created_by_cas2_user_id and users.user_type = 'NOMIS' and users.service_origin = 'HDC'
+      JOIN cas_2_users users ON users.id = applications.created_by_cas2_user_id and users.user_type = 'NOMIS' and users.service_origin = :serviceOrigin
```
davidatkinsuk: in cas2v2 this is 

`LEFT JOIN cas_2_users users ON users.id = applications.created_by_cas2_user_id`

with no constraint on the user type (which is now being constrained to NOMIS user types).

My instinct is remove the following:

`and users.user_type = 'NOMIS' and users.service_origin = :serviceOrigin`

because we're already filtering on the application service origin in the `WHERE`

but i'd like to understand why `users.user_type = 'NOMIS'` was here in the first place. Could you please summarise the different user types that can access both cas2 and cas2 bail and which types can submit applications (i'm yet to look into CAS2 in detail)

davidatkinsuk: this was a LEFT JOIN in cas2v2. Are we happy for it to now be an INNER join. I'm guessing _yes_ because `created_by_cas2_user_id` is non nullable, where as `created_by_user_id` which i guess this query used at one point was nullable?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2v2/integration/Cas2v2ReportsTest.kt
```diff
@@ -553,7 +553,7 @@ class Cas2v2ReportsTest : IntegrationTestBase() {
           personCrn = application2.crn,
           applicationOrigin = application2.applicationOrigin,
           personNoms = application2.nomsNumber.toString(),
-          startedAt = application2.createdAt.toString().split(".").first() + 'Z',
+          startedAt = application2.createdAt.toString().split(".").first(),
```
davidatkinsuk, line 556: ideally we wouldn't see this change without user agreement, see my prior comment on UTC

davidatkinsuk (commented):
This is great work. I have a few comments and concerns, mostly because i'm not too familiar with CAS2

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/jpa/entity/Cas2SubmittedApplicationReport.kt
vbala-moj: Two data models handled this differently—CAS2 as a string and CAS2V2 as a date—so I standardised the field and added a date conversion in one model.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/jpa/entity/Cas2SubmittedApplicationReport.kt
davidatkinsuk: As agreed on the standup this change is fine, especially as we're dropping the 'Z' in the report output so it should be clear that this has changed to end users.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2v2/integration/Cas2v2ReportsTest.kt
vbala-moj: As per the prior comment, we agreed to go head with this change.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/jpa/entity/Cas2SubmittedApplicationReport.kt
davidatkinsuk: I think i'd prefer we did it the other way around because i _think_ it's less code and it avoids another conversion (currently we do Date in DB -> String from query -> LocalDate for bail)

Could we instead do this, which I think is simpler?

1. SQL produces a `Date`
2. That's represented as a `LocalDate` in the `Cas2SubmittedApplicationReportRow`
3. For the CAS2 version of the report we'll then need to convert that into a String (presumably 'YYYY-MM-DD'), but we can remove the function you added to convert strings to Date for CAS2 bail

I'm happy to be convinced otherwise though!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/jpa/entity/Cas2UnsubmittedApplicationsReport.kt
davidatkinsuk: cheers!

vbala-moj: I agree that the user type check is no longer needed since we are already filtering by service origin.

Regarding the different user types, I’m also not very familiar with CAS2, but based on my understanding:

CAS2 (HDC) – Originally intended for NOMIS users only, until service origin was introduced.
CAS2 (BAIL) – Supports all user types.

enum class Cas2UserType(val authSource: String) {
  DELIUS("delius"),
  NOMIS("nomis"),
  EXTERNAL("auth"),
}

davidatkinsuk (approved):
LGTM 👍

PR #4670, opened 2026-04-30
Add full request/response logging in ITs

davidatkinsuk:
Once we've upgraded to spring 4 we can enable the jackson 3 based pretty printer using 

```
import org.springframework.context.annotation.Bean
import org.springframework.context.annotation.Configuration
import org.zalando.logbook.HttpLogFormatter
import org.zalando.logbook.json.JsonHttpLogFormatter


@Configuration
class LogbookConfig {

  @Bean
  fun httpLogFormatter(): HttpLogFormatter {
    return JsonHttpLogFormatter()
  }

}
```


daveawc (approved):
👍

PR #4671, opened 2026-04-30
Improve naming of ObjectMapperFactory

davidatkinsuk:
`ObjectMapperFactory.createRuntimeLikeObjectMapper` is now called `JsonMapperFactory. createJackson2JsonMapper`

‘runtime’ can be a little confusing because there are other json mappers (currently) used at runtime

we’ve also clarified this is a jackson 2 mapper to disambigurate which is being used whilst migrating to jackson 3


daveawc (approved):
👍

PR #4674, opened 2026-05-01
Add ability to run admin jobs without curl

davidatkinsuk:
Curl is no longer available in our docker base image.

Curl was previously used to invoke administration endpoints that were limited to calls originating from localhost (e.g seed/migration).

This commit adds a small Kotlin client for each of the administration endpoints and updates the admin scripts to make use of this client to call the endpoints instead. This is a temporary solution to allow us to continue running admin tasks.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cmd/InvokeAdminJobEndpoint.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cmd/InvokeAdminJobEndpoint.kt
@@ -1,0 +1,53 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.cmd
+
+import com.fasterxml.jackson.databind.json.JsonMapper
+import java.net.URI
+import java.net.http.HttpClient
+import java.net.http.HttpRequest
+import java.net.http.HttpResponse
+import java.time.Duration
+import kotlin.system.exitProcess
+
+@SuppressWarnings("MagicNumber", "TooGenericExceptionCaught")
+object InvokeAdminJobEndpoint {
+
+  fun invokeEndpoint(
+    path: String,
+    map: Map<String, String>,
+  ) {
+    try {
+      val client = HttpClient.newBuilder()
+        .connectTimeout(Duration.ofSeconds(10))
+        .build()
+
+      val jsonBody = JsonMapper().writeValueAsString(map)
+
+      val url = "http://127.0.0.1:8080/$path"
+      println("Invoking $url with arguments $map")
+
+      val request = HttpRequest.newBuilder()
+        .uri(URI.create(url))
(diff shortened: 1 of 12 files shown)


vbala-moj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cmd/InvokeAdminJobEndpoint.kt
```diff
@@ -0,0 +1,53 @@
...
+import com.fasterxml.jackson.databind.json.JsonMapper
+import java.net.URI
+import java.net.http.HttpClient
+import java.net.http.HttpRequest
+import java.net.http.HttpResponse
+import java.time.Duration
+import kotlin.system.exitProcess
+
+@SuppressWarnings("MagicNumber", "TooGenericExceptionCaught")
+object InvokeAdminJobEndpoint {
```
danhumphreys-moj, line 12: Minor: Object.method naming could be `AdminJob.invokeEndpoint(endpoint)`

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cmd/RunSeedFromExcelDirectoryJobCommand.kt
```diff
@@ -0,0 +1,21 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.cmd
+
+import kotlin.system.exitProcess
+
+object RunSeedFromExcelDirectoryJobCommand {
```
danhumphreys-moj, line 5: Minor: It feels there is definite scope for refactoring this functionality which is common across the _Run*Job.kt_ files into a single object (or alternative).

danhumphreys-moj (commented):
This is excellent @davidatkinsuk!  Great stuff - glad it works and no change to the user.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cmd/InvokeAdminJobEndpoint.kt
davidatkinsuk: Good idea. i'm going to rename it to `AdminJobClient`

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cmd/RunSeedFromExcelDirectoryJobCommand.kt
davidatkinsuk: I did try but it become a bit contrived. I think most lines here are unique to each runner but i agree if we start adding more we should revisit

PR #4679, opened 2026-05-06
FM-496 Align dev and test memory usage

davidatkinsuk:
During the spring boot 4 upgrade we discovered that there was insufficient heap memory allocated in dev for spring boot 4 to start. Once it had started heap memory usage settled down to a similar level as that used by spring boot 3.

This commit increases the heap size in `dev` to `800mb`, matching the current usage from `test`. This commit also decreases the total RAM in test to 2GB, as 3GB was not required.


daveawc (approved):
👍

PR #4682, opened 2026-05-06
Bump `generic-prometheus-alerts` dependency to version 1.17

danhumphreys-moj:
HMPPS [Kotlin template project](https://github.com/ministryofjustice/hmpps-template-kotlin/blob/431132cc82ac32f82c131171a9c3652af2a0be00/helm_deploy/hmpps-template-kotlin/Chart.yaml#L11) now uses version 1.17 so we should upgrade.


davidatkinsuk (approved):
LGTM although i don't know why renovate didn't pick this up 🤷

PR #4686, opened 2026-05-07
Add seed job to correct CAS3 postcodes.

daveawc:
This PR creates a seed job so a postcode can be updated by providing the cas3premisesId and the new value in a csv file.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/api/model/SeedFileType.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/api/model/SeedFileType.kt
@@ -47,4 +47,5 @@ enum class SeedFileType(@get:JsonValue val value: String) {
   approvedPremisesCancelOutOfServiceBeds("approved_premises_cancel_out_of_service_beds"),
   approvedPremisesUpdateOutOfServiceBeds("approved_premises_update_out_of_service_beds"),
   approvedPremisesClosePremises("approved_premises_close_premises"),
+  cas3UpdatePremisesPostcode("cas3_update_premises_postcode"),
 }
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/seed/Cas3UpdatePremisesPostcodeSeedJob.kt
```diff
@@ -0,0 +1,36 @@
...
+    cas3PremisesId = UUID.fromString(columns["cas3PremisesId"]),
+    postcode = columns["postcode"]!!.trim(),
+  )
+
+  override fun processRow(row: Cas3UpdatePostcodeRow) {
+    val id = row.cas3PremisesId
+    val postcode = row.postcode
+    log.info("Updating postcode for CAS3 Premises with id: $id")
+    cas3PremisesRepository.findByIdOrNull(id)?.let {
+      it.postcode = postcode
```
davidatkinsuk: for traceability would you mind logging out the current and new postcode. Could maybe fold in the existing log line on 25 so we have easy to read logs?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/api/model/SeedFileType.kt
```diff
@@ -47,4 +47,5 @@ enum class SeedFileType(@get:JsonValue val value: String) {
   approvedPremisesCancelOutOfServiceBeds("approved_premises_cancel_out_of_service_beds"),
   approvedPremisesUpdateOutOfServiceBeds("approved_premises_update_out_of_service_beds"),
   approvedPremisesClosePremises("approved_premises_close_premises"),
+  Cas3UpdatePremisesPostcode("cas3_update_premises_postcode"),
```
davidatkinsuk: is there a reason to uppercase the first character. I know it's probably technically correct/to kotlin convention but lower case would be more consistent? I may have missed a discussion here :)

davidatkinsuk (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/api/model/SeedFileType.kt
daveawc: nah, just convention, I'll lowercase it

davidatkinsuk (approved):
double plus good 👍 👍

PR #4696, opened 2026-05-08
Fix create_api_summary_csv

davidatkinsuk:
This script didn’t work with the latest API definition because `set -e` is set, and regex checks were being used outside of a control flow, returning a non zero exit code


daveawc (approved):
👍

danhumphreys-moj (approved):

PR #4666, opened 2026-04-29
FM-496 - Spring Boot 4 Upgrade (not squashed)

davidatkinsuk:

PR #4699, opened 2026-05-11
FM-496 - Upgrade to Spring Boot 4

davidatkinsuk:
## Summary

This PR upgrades the API to spring boot 4. An 'non squashed' version of the PR is available for reference [here](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/4666)

## Force nullability on implementation of ConverterFactory

Annotations like @Nullable on a Java generic bound do not propagate into Kotlin’s type system for type parameters. Because the Converter interface has been updated to use this construct Kotlin can no longer infer nullabliity

```
public interface Converter<S, T extends @Nullable Object> {
	T convert(S source);
}
```

## replace SpykBean with MockkSpykBean

This is required for mockk upgrade, see https://github.com/Ninja-Squad/springmockk/blob/master/README.md#migrating-to-version-5x

## Update spring jackson configuration

This also removes the `serialization` configuration for jackson 3 because it is no longer supported

the `spring.jackson` configuration now applies to jackson3, not jackson2.

`SerializationFeature.WRITE_DATES_AS_TIMESTAMPS now known as DateTimeFeature.WRITE_DATES_AS_TIMESTAMPS has been changed to false to serialize dates as ISO-8601 strings.`

for this reason, we remove the `WRITE_DATES_AS_TIMESTAMPS` config for jackson 3

Jackson 2 configuration has been added as `spring.jackson2`, matching the pre spring-boot-4 upgrade configuration.

## Ensure JPA Attributes are serializable

The JPA specification requires that the entity attributes are Serializable`

## Handle Jackson3 Exception Handling

Before the spring boot 4 upgrade Jackson 2 was used by Spring MVC, meaning our exception handling responding to unmarshalling issues deal with Jackson 2 classes.

Spring Boot 4 uses jackson 3 for Spring MVC. This commit updates exception handling to work solely using Jackson3 for exception handling.

It also simplifies how we use Spring’s `ProblemDetail` type, which is now correctly placing properties at the top level of the returned JSON instead of in a ‘properties’ attribute. This is mostly likely due to the upgrade to Jackson 3. This allows us to simplify the expected problem JSON in our integration tests, making the JSON match the returned values we saw from Zalando Problem before moving to Spring Problem

## Method Argument Mismatch now produces a problem

Before upgrading to spring boot 4 `MethodArgumentTypeMismatchException` would return a generic error. It now returns a JSON Problem, which is preferred.

## Use Jackson 3 for API `Any` types

As of Spring Boot 4 API marshalling is handled by Jackson 3.

In many places JSON ‘blobs’ are represented in our API model with an ‘Any’ type:

For example, in `ApprovedPremisesApplication` we have

`val document: Any? = null,`

This is populated as follows:

1. Pull JSON from the database as a String
2. Use Jackson to unmarshall this is into an internal Jackson Type
3. Set this value on the `Any` property
4. Spring Web MVC marshalls the `Any` value into JSON in the response

This works fine when the version of Jacksonused in step 2 and 4 are the same. But if Step 2 uses Jackson2 and Step 4 uses Jackson3 (which is the configuration we have after upgrading to Spring 4), Jackson3 doesn’t recognise the Jackson2 native value in the `Any` property and falls back to default serialization, resulting in something like the following:

`{"array":false,"bigDecimal":false,"bigInteger":false,"binary":false,"boolean":false,"containerNode":true,"double":false,"empty":true,"float":false,"floatingPointNumber":false,"int":false,"integralNumber":false,"long":false,"missingNode":false,"nodeType":"OBJECT","null":false,"number":false,"object":true,"pojo":false,"short":false,"textual":false,"valueNode":false} `

For this reason we’ve upgraded all places that use Jackson to populate the `Any` value to use Jackson 3.

Other uses of Jackson will upgraded in subsequent commits

## Pin openapi starter version working version

https://github.com/springdoc/springdoc-openapi/pull/3256 significantly changed our generated schema, making it incompatible with the typescript generators and in some places it was incorrect. We're pinning version 3.0.2 until a new version is available reverting this change, as proposed by https://github.com/springdoc/springdoc-openapi/pull/3276


danhumphreys-moj (approved):
LGTM!

PR #4703, opened 2026-05-12
Use hmpps owasp supression list

davidatkinsuk:
Before this commit we define our own owasp supression file that overwrote the hmpps version. Following guidance from the [hmpps gradle plugin](https://github.com/ministryofjustice/hmpps-gradle-spring-boot/blob/main/README.md#owasp-dependency-check-and-suppression-files) we have changed how this is configured to supplement the baseline supressions


daveawc (approved):
👍

PR #4716, opened 2026-05-18
Simplify renovate config

davidatkinsuk:
Before this commit the renovate config was blocking major updates for 30 days. Furthermore there was additional package rules that did not align with hmpps rules.

This commit reverts to renovate config to the baseline hmpps configuration to determine if any of the custom configuration is required anymore

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,36 +1,4 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
-  "extends": ["github>ministryofjustice/hmpps-renovate-config:jvm"],
-  "assigneesFromCodeOwners": true,
-  "packageRules": [
-    {
-      "matchUpdateTypes": ["patch"],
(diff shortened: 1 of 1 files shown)


renovate.json
rashadmughal, line 1: I think we need to keep these packagerules? I'm sure they were part of the cyber security changes.

davidatkinsuk: we're inheriting from the hmpps baseline config to which the 7 days package rule was added last month

https://github.com/ministryofjustice/hmpps-renovate-config/blob/main/jvm.json

- i don't believe we should have any rule less than 7 days, so that needs removing
- we also have a rule of 30 days, which seems unnecessary (and blocked a postgres upgrade), so that needs removing

I can't see anything else in the remaining config that i think needs retaining wrt. cyber security upgrades. The only recent change i can see is changing the baseline config from node to jvm (oops)

rashadmughal (approved):

PR #4723, opened 2026-05-18
Remove ‘cas3 validate departure in future’ feature

davidatkinsuk:
This feature was added over a year ago and never disabled. This commit removes the feature and its corresponding feature flag


daveawc (approved):
👍

PR #4724, opened 2026-05-18
FM-367 Use nonprod severity for nonprod environments

davidatkinsuk:
This will ensure that non prod prometheus alerts will be sent to the non prod slack channel


daveawc (approved):
👍

PR #4727, opened 2026-05-18
FM-367 Add SQS Queue Alerting

davidatkinsuk:
Configure SQS Queue Monitoring for all environments, ensuring we get alerted if

* there are at least 10 messages in a queue
* there is a message in the queue for more than 10 minutes


daveawc (approved):
Nice one!

PR #4731, opened 2026-05-19
Remove unneccessary qualifiers

davidatkinsuk:


rashadmughal (approved):
LGTM - but I've just done a quick scan, if the tests pass I think it should be OK

davidatkinsuk:
> LGTM - but I've just done a quick scan, if the tests pass I think it should be OK

cheers - i agree it's all from static analysis so should be fine. i did a similar scan

PR #4733, opened 2026-05-19
Add ‘experimental’ role and permission

davidatkinsuk:
We already have the concept of experimental permissions. These are permissions that will never be available to a user in a production environment. This provides us with a way to safely test functionality that is not yet ready for production.

This commit adds an ‘experimental’ permission and role which removes the need for us to add specific experimental permissions when testing new functionality. It also provides us with an experimental permission that _isn’t_ available to the janitor user.

In time it’s likely that this single permission named `experimental` will replace any functon-specific experimental permissions.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/UserRole.kt
```diff
@@ -189,5 +202,11 @@ enum class UserRole(val service: ServiceName, val cas1ApiValue: ApprovedPremises
     fun getAllRolesForPermission(permission: UserPermission) = entries.filter { it.permissions.contains(permission) }
 
     fun getAllRolesExcept(vararg rolesToExclude: UserRole) = UserRole.entries.filter { !rolesToExclude.contains(it) }
+
+    fun getRolesWithAllPermissions() = listOf(CAS1_JANITOR, CAS1_EXPERIMENTAL)
```
danhumphreys-moj, line 206: Function name not strictly true since `CAS1_JANITOR` role does not include `CAS1_EXPERIMENTAL` permission.

`getRolesWithAllProdPermissions()`?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/UserRole.kt
```diff
@@ -189,5 +202,11 @@ enum class UserRole(val service: ServiceName, val cas1ApiValue: ApprovedPremises
...
+
+    fun List<UserRole>.removeRolesWithAllPermissions() = this - getRolesWithAllPermissions().toSet()
```
danhumphreys-moj: `getRolesWithAllPermissions()` returns List<UserRole>, is `toSet()` required here?

danhumphreys-moj (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/UserRole.kt
davidatkinsuk: that was an intellij suggestion for optimisation which frankly i didn't understand :)

davidatkinsuk: https://stackoverflow.com/a/70612578

arguably not really required in this scenario, especially given the size of the list, but it makes intellij happy

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/UserRole.kt
davidatkinsuk: good thinking. done

PR #4734, opened 2026-05-19
CAS2 Bail Report Tidy Up

davidatkinsuk:
**1. Move Cas2v2 App Status finder**

This was previously in a reporting package, which is incorrect.

This commit also removes two unused functions on the `Cas2v2PersistedApplicationStatusFinder`

**2. Simplify Report Service Logic**
**3. move Report Models into Report Service**


daveawc (approved):
👍

PR #4744, opened 2026-05-19
FM-592 align openapi config with hmpps template

davidatkinsuk:
Miscellaneous changes to our open api configuration to align with the hmpps kotlin template and allow us to start adding the default template tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,14 @@ Authorization: Bearer {the JWT}
 
 ## Coding Notes
 
+## Divergence from the hmpps kotlin template
+
+HMPPS provide a [kotlin template project](https://github.com/ministryofjustice/hmpps-template-kotlin)
+
+Whilst we should remain aligned with the template, there are some areas that we diverge
+
+* We use pattern matching rules to apply security, where as the sample [ResourceSecurityTest](https://github.com/ministryofjustice/hmpps-template-kotlin/blob/main/src/test/kotlin/uk/gov/justice/digital/hmpps/templatepackagename/integration/ResourceSecurityTest.kt) expects use of @PreAuthorize annotations. For this reason we do not include this test. AuthTest provides from top level authentication tests, and each individual end test ensures a token is required with a suitable role
+
 ### Load tests
 
 * Load tests are managed in the [hmpps-load-testing-probation](https://github.com/ministryofjustice/hmpps-load-testing-probation) repo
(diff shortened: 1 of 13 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/OpenApiDocsTest.kt
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  @Disabled
+  @Test
+  fun `the swagger json don't contain any duplicate methods`() {
+    // Methods in resource classes with the same name end up with operationIds that have _1 and _2 etc. in the name.
+    // When the code is then generated from the api docs we refer to the endpoint by operationId. If a new method is
+    // added or one removed then the _1 / _2 etc. can change order and thus we end up calling a completely different
+    // endpoint next time the code is generated. This test then prevents that from happening by ensuring all endpoints
+    // have method names / operation ids.
```
daveawc, line 59: 😲

README.md
```diff
@@ -52,6 +52,14 @@ Authorization: Bearer {the JWT}
 
 ## Coding Notes
 
+## Divergence from the hmpps kotlin template
```
daveawc, line 55: It's s bit of a refactor, but I think we _could_ make use of the `@PreAuthorize` annotations, as you can call service functions and do the internal authorisation as part of the `@PreAuthorize`. I've not looked into it too much or how it could be applied, but I think we can do something like:

```kotlin
  @GetMapping("/applications/all")
  @PreAuthorize(value = """hasRole('PROBATION') and @userAccessService.hasRole("CAS1_ASSESSOR")""")
  fun getAllApplications(
    @RequestParam page: Int?,
```

and then 

```kotlin
  fun hasRole(role: String) : Boolean {
    return userService.getUserForRequest().hasRole(UserRole.valueOf(role))
  }

```
I think we'd need to look into when the auth service sets the delius principalc compared to when the `@PreAuthorise` runs though 🤔

daveawc (approved):
Nice!

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/OpenApiDocsTest.kt
davidatkinsuk: i figured it's best to leave the template comments in

README.md
davidatkinsuk: I agree it may be worth doing, although for CAS1/3 it's always be the probation role. I'll keep it in mind 👍

PR #4749, opened 2026-05-20
feature/SAS-598-Refactor-CAS1-suitability-end-point-to-include-accommodation-info

besscerule:
Fleshed out placement histories with the premises and also included a flag to say whether it is the suitable one.

Eventually the top level requestForPlacement and placement statuses will be deprecated as SAS will use the flag in the placementHistories to find the correct statuses and premises - seems neater than duplicating them at the top level.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/api/model/Cas1SuitableApplication.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/api/model/Cas1SuitableApplication.kt
@@ -16,4 +16,15 @@ data class Cas1PlacementHistory(
   val dateApplied: LocalDate,
   val requestForPlacementStatus: RequestForPlacementStatus,
   val placementStatus: Cas1SpaceBookingStatus?,
+  val premises: SuitablePremisesDto?,
+  var isSuitable: Boolean,
+)
+
+data class SuitablePremisesDto(
+  val startDate: LocalDate?,
+  val endDate: LocalDate?,
+  val addressLine1: String,
+  val addressLine2: String?,
+  val town: String?,
+  val postcode: String,
 )
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/external/Cas1ExternalApplicationService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/external/Cas1ExternalApplicationService.kt
@@ -3,10 +3,12 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas1.external
 import org.springframework.stereotype.Service
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Cas1PlacementHistory
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.Cas1SuitableApplication
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.SuitablePremisesDto
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApprovedPremisesApplicationEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApprovedPremisesApplicationRepository
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.ApprovedPremisesApplicationStatus
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.results.CasResult
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas1.Cas1PremisesService
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas1.Cas1RequestForPlacementService
 import java.time.LocalDate
 
@@ -15,6 +17,8 @@ import java.time.LocalDate
 class Cas1ExternalApplicationService(
   private val approvedPremisesApplicationRepository: ApprovedPremisesApplicationRepository,
   private val cas1RequestForPlacementService: Cas1RequestForPlacementService,
+  private val cas1PremisesService: Cas1PremisesService,
+
 ) {
   fun getSuitableApplicationByCrn(crn: String): Cas1SuitableApplication? = approvedPremisesApplicationRepository.findByCrn(crn)
     .maxWithOrNull(
@@ -33,14 +37,35 @@ class Cas1ExternalApplicationService(
               dateApplied = rfp.statusSetDate,
               requestForPlacementStatus = rfp.status,
               placementStatus = null,
+              premises = null,
+              isSuitable = false,
             ),
           )
         } else {
           rfp.placements.map { placement ->
+            println(placement)
+            println("hello")
+            println(placement.premises.id)
+            println(placement.premises.name)
+
+            val premises = cas1PremisesService.findPremisesById(placement.premises.id)
+              ?.let {
+                SuitablePremisesDto(
+                  startDate = placement.expectedArrivalDate,
+                  endDate = placement.expectedDepartureDate,
+                  postcode = it.postcode,
+                  addressLine1 = it.addressLine1,
+                  addressLine2 = it.addressLine2,
+                  town = it.town,
+                )
+              }
+
             Cas1PlacementHistory(
               dateApplied = requireNotNull(placement.statusSetDate),
               requestForPlacementStatus = rfp.status,
               placementStatus = placement.status,
+              premises = premises,
+              isSuitable = false,
             )
           }
         }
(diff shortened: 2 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/api/model/Cas1SuitableApplication.kt
```diff
@@ -9,11 +9,20 @@ data class Cas1SuitableApplication(
   val applicationStatus: ApprovedPremisesApplicationStatus,
   val requestForPlacementStatus: RequestForPlacementStatus?,
   val placementStatus: Cas1SpaceBookingStatus?,
-  val placementHistories: List<Cas1PlacementHistory>,
```
davidatkinsuk, line 12: did you mean to remove placementHistories?  Is that now an internal class only? If yes can we move it to `Cas1ExternalApplicationService.kt‎` and make it private?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/api/model/Cas1SuitableApplication.kt
```diff
@@ -9,11 +9,20 @@ data class Cas1SuitableApplication(
...
 )
 
 data class Cas1PlacementHistory(
   val dateApplied: LocalDate,
   val requestForPlacementStatus: RequestForPlacementStatus,
-  val placementStatus: Cas1SpaceBookingStatus?,
+  val placement: Cas1SpaceBookingShortSummary?,
+)
+
+data class PremisesDto(
```
davidatkinsuk: Could we possibly come up with a more specific name for this to disambiguate from our many other Premises data classes?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/external/Cas1ExternalApplicationService.kt
```diff
@@ -52,12 +54,29 @@ class Cas1ExternalApplicationService(
...
+            }
+        }
       Cas1SuitableApplication(
         id = application.id,
         applicationStatus = application.status,
         requestForPlacementStatus = currentPlacementHistory?.requestForPlacementStatus,
-        placementStatus = currentPlacementHistory?.placementStatus,
-        placementHistories = placementHistories,
+        placementStatus = currentPlacementHistory?.placement?.status,
+        premises = premises,
```
davidatkinsuk: given this is only provided for upcoming placements, could we better name the field to reflect this? What's the harm in always including it? Are we baking some SAS logic into this endpoint by excluding it?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/external/Cas1ExternalApplicationService.kt
```diff
@@ -52,12 +54,29 @@ class Cas1ExternalApplicationService(
...
+            .find { it.id == upcomingPlacementHistory.placement?.id }?.premises
+            ?.let {
+              PremisesDto(
+                startDate = upcomingPlacementHistory.placement?.expectedArrivalDate,
+                endDate = upcomingPlacementHistory.placement?.expectedDepartureDate,
+                postcode = it.postcode,
+                addressLine1 = it.addressLine1,
+                addressLine2 = it.addressLine2,
+                town = it.town,
+              )
```
davidatkinsuk: could we move this into a function to make this main function a bit smaller?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/external/Cas1ExternalApplicationService.kt
```diff
@@ -52,12 +54,29 @@ class Cas1ExternalApplicationService(
         placementHistories.lastOrNull { it.dateApplied >= today }
           ?: placementHistories.firstOrNull { it.dateApplied < today }
 
+      val premises = currentPlacementHistory
+        .takeIf { it?.placement?.status == Cas1SpaceBookingStatus.UPCOMING }
+        ?.let { upcomingPlacementHistory ->
+          application.placementRequests
+            .flatMap { it.spaceBookings }
+            .find { it.id == upcomingPlacementHistory.placement?.id }?.premises
```
davidatkinsuk: If Cas1PlacementHistory is now an internal class, we could just add the PremisesEntity into it to saving having to 'refind' the space booking here?

davidatkinsuk (commented):
Some comments from me

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/external/Cas1ExternalApplicationService.kt
```diff
@@ -33,31 +34,51 @@ class Cas1ExternalApplicationService(
...
       }.sortedByDescending { it.dateApplied }
 
       val today = LocalDate.now()
 
-      val currentPlacementHistory =
+      val suitablePlacement =
         placementHistories.lastOrNull { it.dateApplied >= today }
           ?: placementHistories.firstOrNull { it.dateApplied < today }
 
+      suitablePlacement?.isSuitable = true
```
davidatkinsuk, line 80: instead of mutating a field could we not make it a `val` and use `.copy(isSuitable=true)`?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/external/Cas1ExternalApplicationService.kt
```diff
@@ -33,31 +34,51 @@ class Cas1ExternalApplicationService(
...
               placementStatus = null,
+              premises = null,
+              isSuitable = false,
             ),
           )
         } else {
           rfp.placements.map { placement ->
+            val premises = application.placementRequests
+              .flatMap { it.spaceBookings }
+              .find { it.id == placement.id }?.premises
```
davidatkinsuk: I think this is going to trigger lazy loading for each relationship traversal (e.g. spaceBookings and placementRequests). Therefore, i think it'll be more efficient to use the `Cas1PremisesService` to get the premises using something like

`cas1PremisesServices.findPremisesById(placement.premises.id)`

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/external/Cas1ExternalApplicationService.kt
```diff
@@ -33,31 +34,51 @@ class Cas1ExternalApplicationService(
...
           }
         }
       }.sortedByDescending { it.dateApplied }
 
       val today = LocalDate.now()
 
-      val currentPlacementHistory =
+      val suitablePlacement =
         placementHistories.lastOrNull { it.dateApplied >= today }
           ?: placementHistories.firstOrNull { it.dateApplied < today }
```
davidatkinsuk, line 78: I feel like SAS should maybe be doing this - i think it has all the data is needs to make this decision? This Endpoint/API is then just about providing a summary of info without having to understand what SAS is doing with it. Something to consider going forward!

davidatkinsuk (approved):
Some thoughts, but LGTM 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/external/Cas1ExternalApplicationService.kt
besscerule: would that persist the isSuitable true in the placementHistories list?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/external/Cas1ExternalApplicationService.kt
besscerule: you have a point, I just wanted to make it backwards compatible so I can go and fix it on SAS and then undo it here

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/external/Cas1ExternalApplicationService.kt
davidatkinsuk: oh yes good point. I still don't love the mutation but i suspect it'll be non trivial to get rid of it. resolved.

PR #4753, opened 2026-05-20
Premises Search Optimisation

davidatkinsuk:
The premises search was taking upto 3 seconds in production.

This commit makes a simple optimisation to the query to get it down to around 1 second.

The lateral join wasn’t required because we were retaining all local restrictions that match the premises (lateral would be useful if we wanted to select a specific restriction based on order, for example)


danhumphreys-moj (approved):
LGTM

PR #4755, opened 2026-05-20
Remove duplicate apType on assessment accept

davidatkinsuk:
Before this commit the `apType` on the `Cas1AssessmentAcceptance` type was defined in two places:

1. As `apType: ApType?`
2. As `requirements.apType: ApType`

The former was only used when creating the domain event. The later was persisted to the `placementRequest.placementRequirements`

This commit removes `apType: ApType?` and replaces its usage with the non-nullable `requirements.apType: ApType`


daveawc (approved):
👍

PR #4759, opened 2026-05-21
Refresh dependency check supressions

davidatkinsuk:
Remove stale entries

Add DOMPurify entries, aligned with hmpps gradle plugin rationale (https://github.com/ministryofjustice/hmpps-gradle-spring-boot/blob/main/src/main/resources/dps-gradle-spring-boot-suppressions.xml). We use a different version so have different CVEs

Use variables to remove version duplication in build file


daveawc (approved):
👍

PR #4760, opened 2026-05-21
Optimise Space Search SQL

davidatkinsuk:
The space search query currently has a combinatorial exploision issue due to the amount of joins being applied.

This commit changes how we collate room and premises characteristics for each premises by using a subquery to collate the information int an array, and returning that to the ‘main’ query. This reduces the number of top level joins being applied, significantly increasing the SQL execution time (from around 1 second to 0.1 seconds)


daveawc (approved):
Nice!

danhumphreys-moj (approved):
LGTM - changes/improvements in SQL.  No change to tests required so functionality remains the same.

davidatkinsuk:
> LGTM - changes/improvements in SQL. No change to tests required so functionality remains the same.

FYI i did do some sanity testing against pre-prod data diff-ing a CSV dump of results before and after SQL change. But I agree our tests are very thorough in this area so technically that wasn't required

PR #4778, opened 2026-05-26
FM-603 - Add CAS2 backfill application cohorts migration job and tests

danhumphreys-moj:
Ref: https://dsdmoj.atlassian.net/browse/FM-603

Adds a migration job to populate `cohort` based on `applicationOrigin`.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/migration/Cas2BackfillApplicationCohortJob.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/migration/Cas2BackfillApplicationCohortJob.kt
@@ -1,0 +1,47 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.cas2.migration
+
+import org.slf4j.LoggerFactory
+import org.springframework.stereotype.Component
+import org.springframework.transaction.support.TransactionTemplate
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ApplicationOrigin
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas2.jpa.entity.Cas2ApplicationRepository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas2.jpa.entity.Cas2Cohort
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas2.jpa.entity.Cas2IdApplicationOriginEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.migration.MigrationJob
+
+@Component
+class Cas2BackfillApplicationCohortJob(
+  private val cas2ApplicationRepository: Cas2ApplicationRepository,
+  private val transactionTemplate: TransactionTemplate,
+) : MigrationJob() {
+  override val shouldRunInTransaction = false
+  private val log = LoggerFactory.getLogger(this::class.java)
+
+  @SuppressWarnings("MagicNumber")
+  override fun process(pageSize: Int) {
+    val applications = cas2ApplicationRepository.findIdAndApplicationOriginByCohortIsNull()
+
+    log.info("Found ${applications.size} applications without a cohort to backfill")
+
+    applications.forEach { application ->
+      transactionTemplate.executeWithoutResult {
+        updateCohort(application)
+      }
+      Thread.sleep(50)
+    }
+
+    log.info("Backfilled ${applications.size} applications")
+  }
+
+  private fun updateCohort(application: Cas2IdApplicationOriginEntity) {
+    cas2ApplicationRepository.updateCohort(
+      application.id,
+      when (application.applicationOrigin) {
+        ApplicationOrigin.courtBail -> Cas2Cohort.COURT_BAIL.name
+        ApplicationOrigin.prisonBail -> Cas2Cohort.PRISON_BAIL.name
+        ApplicationOrigin.homeDetentionCurfew -> Cas2Cohort.HDC.name
+        ApplicationOrigin.other -> throw IllegalStateException("Unexpected application origin: ${application.applicationOrigin}")
+      },
+    )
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/migration/Cas2BackfillApplicationCohortJob.kt
```diff
@@ -0,0 +1,33 @@
...
+
+    applications.forEach { application ->
+      application.cohort = when (application.applicationOrigin) {
+        ApplicationOrigin.courtBail -> Cas2Cohort.COURT_BAIL
+        ApplicationOrigin.prisonBail -> Cas2Cohort.PRISON_BAIL
+        ApplicationOrigin.homeDetentionCurfew -> Cas2Cohort.HDC
+        ApplicationOrigin.other -> throw IllegalStateException("Unexpected application origin: ${application.applicationOrigin}")
+      }
+    }
+    cas2ApplicationRepository.saveAll(applications)
```
davidatkinsuk: because we have 13772 applications i'm slightly concerned about doing this update in a single transaction, for a few reasons:

1. it's going to be quite a large transaction, i think we may have resource issues
2. because we don't have version control on the entity (e.g. optimistic locking) any change that comes in on the application during this operation will be overwritten by the state of the transaction, which is a snapshot of the application when this job started running

We could avoid (2) by adding a version column and enable locking, but i think it may be better if we instead [update the single field directly](https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/88869cd87f99bdc1b04e81a2574c7ccd05b659d1/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApplicationEntity.kt#L236) for each iteration, [using a new transaction for each iteration](https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/8734f29180cf7085119c4199251a0cfd214ad9b6/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/migration/Cas1BackfillUserApArea.kt#L29).

This also means if the migration job fails part way through it will continue where it left off

davidatkinsuk (commented):
Just one comment from me regarding the scope of the transaction

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/migration/Cas2BackfillApplicationCohortJob.kt
danhumphreys-moj: Yes good idea.

Have added 2 x new repository functions:
```
  @Query(
    "SELECT id, application_origin FROM cas_2_applications WHERE cohort IS NULL",
    nativeQuery = true,
  )
  fun findIdAndApplicationOriginByCohortIsNull(): List<Cas2IdApplicationOriginEntity>

  @Modifying
  @Query("UPDATE cas_2_applications SET cohort = :cohort WHERE id = :id", nativeQuery = true)
  fun updateCohort(id: UUID, cohort: String)
  ```
  
  New entity with reduced properties to help reduce memory requirements:
  ```
  @Entity
@Table(name = "cas_2_applications")
data class Cas2IdApplicationOriginEntity(
    @Id
    val id: UUID,
    @Enumerated(EnumType.STRING)
    var applicationOrigin: ApplicationOrigin,
)
```

Updated `process` to call update function on each record:
```
  override fun process(pageSize: Int) {
    val applications = cas2ApplicationRepository.findIdAndApplicationOriginByCohortIsNull()

    log.info("Found ${applications.size} applications without a cohort to backfill")

    applications.forEach { application ->
      transactionTemplate.executeWithoutResult {
        updateCohort(application)
      }
      Thread.sleep(50)
    }
```

And new private fun to update, considered splitting out the when into a dedicated look-up but I understand Kotlin when's are sufficiently efficient that this isn't required:

```
  private fun updateCohort(application: Cas2IdApplicationOriginEntity) {
    cas2ApplicationRepository.updateCohort(application.id, when (application.applicationOrigin) {
      ApplicationOrigin.courtBail -> Cas2Cohort.COURT_BAIL.name
      ApplicationOrigin.prisonBail -> Cas2Cohort.PRISON_BAIL.name
      ApplicationOrigin.homeDetentionCurfew -> Cas2Cohort.HDC.name
      ApplicationOrigin.other -> throw IllegalStateException("Unexpected application origin: ${application.applicationOrigin}")
    })
  }
```

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/migration/Cas2BackfillApplicationCohortJob.kt
```diff
@@ -0,0 +1,33 @@
...
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ApplicationOrigin
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas2.jpa.entity.Cas2ApplicationRepository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas2.jpa.entity.Cas2Cohort
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.migration.MigrationJob
+
+@Component
+class Cas2BackfillApplicationCohortJob(
+  private val cas2ApplicationRepository: Cas2ApplicationRepository,
+) : MigrationJob() {
+  override val shouldRunInTransaction = true
```
davidatkinsuk: I think this needs changing to false as we're managing the transactions ourselves per row

danhumphreys-moj: Done!

davidatkinsuk (approved):
Looks great 👍

PR #4781, opened 2026-05-27
Fix open api groups

davidatkinsuk:
Recently we added `paths-to-exclude` entry in the `application.yml` to exclude the `queue-admin/**` urls from all open api documents. This overwrote our custom exclude rules defined in `OpenApiConfiguration`. This commit removes that entry, and instead adds it into the `OpenApiConfiguration`. It also adds a regression test to capture similar breakages in the future


daveawc (approved):
Nice find.

PR #4786, opened 2026-05-28
FM-586 Use correct release type in `PlacementRequest`

davidatkinsuk:
https://dsdmoj.atlassian.net/browse/FM-586

Before this commit the value of `PlacementRequest.releaseType` was always derived from the corresponding applicaton. This meant that if the release type was modified for a subsequent request for placement, the wrong type would be returned.

This commit changes that logic to source the release type from the corresponding placement application, falling back to the application’s release type if a placement application doesn’t exist

I test this locally, screenshots are on the ticket

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/seed/Cas1StartupScript.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/seed/Cas1StartupScript.kt
@@ -51,6 +51,12 @@ class Cas1StartupScript(
 
     createApplication(
       deliusUserName = "AP_USER_TEST_1",
+      crn = "Y006208",
+      state = Cas1ApplicationSeedService.ApplicationState.AUTHORISED,
+    )
+
+    createApplication(
+      deliusUserName = "AP_USER_TEST_1",
       crn = "X698338",
       state = Cas1ApplicationSeedService.ApplicationState.EXPIRED_AFTER_AUTHORISATION,
     )
@@ -63,7 +69,7 @@ class Cas1StartupScript(
     crn: String,
     state: Cas1ApplicationSeedService.ApplicationState,
   ) {
-    seedLogger.info("Creating application with state $state for CRN X320741")
+    seedLogger.info("Creating application with state $state for CRN $crn")
     try {
       cas1ApplicationSeedService.createApplication(
         deliusUserName = deliusUserName,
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/seed/Cas1StartupScript.kt
```diff
@@ -49,6 +49,12 @@ class Cas1StartupScript(
       state = Cas1ApplicationSeedService.ApplicationState.EXPIRED_BEFORE_SUBMISSION,
     )
 
+    createApplication(
+      deliusUserName = "AP_USER_TEST_1",
+      crn = "Y006208",
+      state = Cas1ApplicationSeedService.ApplicationState.AUTHORISED,
+    )
```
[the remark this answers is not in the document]
davidatkinsuk: clever a.i.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/PlacementRequestTransformer.kt
```diff
@@ -41,7 +41,7 @@ class PlacementRequestTransformer(
     risks = risksTransformer.transformDomainToApi(jpa.application.riskRatings!!, jpa.application.crn),
     applicationId = jpa.application.id,
     assessmentId = jpa.assessment.id,
-    releaseType = jpa.application.releaseType!!.apiType,
+    releaseType = jpa.placementApplication?.releaseType?.apiType ?: jpa.application.releaseType!!.apiType,
```
danhumphreys-moj, line 44: I think this logic is right in two scenarios:
1. When an application doesn't have a placement application i.e. placement application is null and ,
2. When an application does have a placement application, but the release type hasn't changed, in which case placement application is not null but placement_application.release_type is null.

In either case we want to revert back to the application release type.

danhumphreys-moj (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/PlacementRequestTransformer.kt
davidatkinsuk: Agreed. And if we get the chance to run the backfill job I was planning on running just before I left last year, there would never be a scenario where `placementRequest.placementApplication` is null, so we can then remove the 'else' part

PR #4787, opened 2026-05-28
FM-557 Remove `Cas2DummyController`

danhumphreys-moj:
Now that the dependent types `OASysSections` and `OASysSupportingInformationQuestion` are no longer shared types we can remove this dummy controller without fear of breaking the HDC UI typecheck.

See: https://github.com/ministryofjustice/hmpps-community-accommodation-tier-2-ui/pull/1095


davidatkinsuk (approved):
LGTM 👍

PR #4793, opened 2026-05-29
FM-650 Allow Risk & Health OAsys info > 6 months for CAS1

davidatkinsuk:
This commit updates the following two endpoints to remove the limit on OASys assessment age:

* /people/{crn}/oasys/risks-to-individual
* /people/{crn}/oasys/health-details

This is done by introducing an OASys `SuitabilityStrategy` with two options:

* AllowAll
* CompletedInLastSixMonths (default)

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/OASysSuitabilityService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/OASysSuitabilityService.kt
@@ -9,12 +9,22 @@ class OASysSuitabilityService(
   private val clock: Clock,
   private val sentryService: SentryService,
 ) {
-  fun isSuitable(info: OASysAssessmentDates): Boolean {
+  fun isSuitable(
+    assessmentDates: OASysAssessmentDates,
+    strategy: SuitabilityStrategy,
+  ) = when (strategy) {
+    SuitabilityStrategy.AllowAll -> true
+    SuitabilityStrategy.CompletedInLastSixMonths -> limitToSixMonths(assessmentDates)
+  }
+
+  private fun limitToSixMonths(
+    assessmentDates: OASysAssessmentDates,
+  ): Boolean {
     val sixMonthThreshold = OffsetDateTime.now(clock).minusMonths(6)
 
-    val crn = info.crn
-    val initiationDate = info.initiationDate
-    val dateCompleted = info.dateCompleted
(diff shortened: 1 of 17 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/OASysSuitabilityService.kt
```diff
@@ -32,4 +42,9 @@ class OASysSuitabilityService(
     val initiationDate: OffsetDateTime,
     val dateCompleted: OffsetDateTime?,
   )
+
+  enum class SuitabilityStrategy {
```
danhumphreys-moj, line 46: nice

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/OASysServiceTest.kt
```diff
@@ -42,6 +43,7 @@ class OASysServiceTest {
     val INITIATION_DATE: OffsetDateTime = OffsetDateTime.parse("2007-12-03T10:15:30+01:00")
     val COMPLETION_DATE: OffsetDateTime = OffsetDateTime.parse("2007-12-04T10:15:30+01:00")
     val APPLICABILITY_INFO: OASysAssessmentDates = OASysAssessmentDates(CRN, INITIATION_DATE, COMPLETION_DATE)
+    val DEFAULT_SUITABILITY_STRATEGY = SuitabilityStrategy.CompletedInLastSixMonths
```
danhumphreys-moj, line 46: Rename to `SIX_MONTHS_SUITABILITY_STRATEGY` to improve test readability?  Actually is this required?  We could use `SuitabilityStrategy.CompletedInLastSixMonths` explicitly in the tests?

danhumphreys-moj (approved):
LGTM!

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/OASysServiceTest.kt
davidatkinsuk: In this case i'm trying to indicate that the default strategy is being used when calling `oasysApplicabilitySevice` because one isn't defined on the call into `OAsysService`. I can see an argument to specify it explicitly in each function call but it felt like noise in the tests

PR #4794, opened 2026-05-29
Update controllers to dynamically build resource URIs using request base path

danhumphreys-moj:
Update controllers to dynamically build resource URIs using request base path


davidatkinsuk:
@danhumphreys-moj I think we can close this one?

PR #4795, opened 2026-05-29
Audit POC - Envers

vbala-moj:
https://dsdmoj.atlassian.net/browse/FM-612

Added Spring Data (Hibernate)Envers for auditing data changes

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -27,6 +27,7 @@ dependencies {
   implementation("org.springframework.boot:spring-boot-starter-webflux")
   implementation("org.springframework.boot:spring-boot-starter-data-jpa")
   implementation("org.springframework.retry:spring-retry")
+  implementation("org.springframework.data:spring-data-envers")
   implementation("io.hypersistence:hypersistence-utils-hibernate-71:3.15.2")
   implementation("org.hibernate.orm:hibernate-spatial")
   implementation("org.hibernate.orm:hibernate-jcache")
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -93,6 +93,7 @@ dependencies {
   testImplementation("com.ninja-squad:springmockk:5.0.1")
   testImplementation("org.springframework.boot:spring-boot-webtestclient")
   testImplementation("org.zalando:logbook-spring-boot-starter:4.0.4")
+  testImplementation("org.springframework.security:spring-security-test")
 
   testImplementation("io.swagger.parser.v3:swagger-parser:2.1.43") {
     exclude(group = "io.swagger.core.v3")
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ReleaseActionEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ReleaseActionEntity.kt
@@ -2,19 +2,9 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity
 
 import jakarta.persistence.Entity
 import jakarta.persistence.Id
-import jakarta.persistence.JoinColumn
-import jakarta.persistence.ManyToOne
 import jakarta.persistence.Table
 import org.hibernate.envers.Audited
-import org.springframework.data.jpa.repository.JpaRepository
-import org.springframework.data.repository.history.RevisionRepository
-import org.springframework.stereotype.Repository
 import java.util.UUID
-
-@Repository
-interface ReleaseActionRepository :
-  JpaRepository<ReleaseActionEntity, UUID>,
-  RevisionRepository<ReleaseActionEntity, UUID, Int>
 
 @Entity
 @Table(name = "release_action")
@@ -26,8 +16,4 @@ class ReleaseActionEntity(
   var description: String,
   var actionCadence: String,
   var otherInformation: String? = null,
-
-  @ManyToOne
-  @JoinColumn(name = "release_plan_id")
-  var releasePlan: ReleasePlanEntity,
 )
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ReleaseActionEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ReleaseActionEntity.kt
@@ -12,7 +12,9 @@ import org.springframework.stereotype.Repository
 import java.util.UUID
 
 @Repository
-interface ReleaseActionRepository : JpaRepository<ReleaseActionEntity, UUID>, RevisionRepository<ReleaseActionEntity, UUID, Int>
+interface ReleaseActionRepository :
+  JpaRepository<ReleaseActionEntity, UUID>,
+  RevisionRepository<ReleaseActionEntity, UUID, Int>
 
 @Entity
 @Table(name = "release_action")
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ReleaseActionEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ReleaseActionEntity.kt
@@ -1,0 +1,17 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.cas1
+
+import jakarta.persistence.Entity
+import jakarta.persistence.Id
+import jakarta.persistence.Table
+import java.util.UUID
+
+@Entity
+@Table(name = "release_action")
+class ReleaseActionEntity(
+  @Id
+  val id: UUID,
+
+  var description: String,
+  var whenField: String,
+  var otherInformation: String? = null,
+)
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ReleaseActionEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ReleaseActionEntity.kt
@@ -1,17 +1,31 @@
-package uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.cas1
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity
 
 import jakarta.persistence.Entity
 import jakarta.persistence.Id
+import jakarta.persistence.JoinColumn
+import jakarta.persistence.ManyToOne
 import jakarta.persistence.Table
+import org.hibernate.envers.Audited
+import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.data.repository.history.RevisionRepository
+import org.springframework.stereotype.Repository
 import java.util.UUID
+
+@Repository
+interface ReleaseActionRepository : JpaRepository<ReleaseActionEntity, UUID>, RevisionRepository<ReleaseActionEntity, UUID, Int>
 
 @Entity
 @Table(name = "release_action")
+@Audited
 class ReleaseActionEntity(
   @Id
   val id: UUID,
 
   var description: String,
-  var whenField: String,
+  var actionCadence: String,
   var otherInformation: String? = null,
+
+  @ManyToOne
+  @JoinColumn(name = "release_plan_id")
+  var releasePlan: ReleasePlanEntity,
 )
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ReleasePlanEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ReleasePlanEntity.kt
@@ -7,6 +7,7 @@ import jakarta.persistence.ManyToOne
 import jakarta.persistence.OneToMany
 import jakarta.persistence.Table
 import org.hibernate.envers.Audited
+import org.hibernate.envers.RelationTargetAuditMode
 import org.springframework.data.jpa.repository.JpaRepository
 import org.springframework.data.repository.history.RevisionRepository
 import org.springframework.stereotype.Repository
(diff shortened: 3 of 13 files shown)


build.gradle.kts
```diff
@@ -92,6 +93,7 @@ dependencies {
   testImplementation("com.ninja-squad:springmockk:5.0.1")
   testImplementation("org.springframework.boot:spring-boot-webtestclient")
   testImplementation("org.zalando:logbook-spring-boot-starter:4.0.4")
+  testImplementation("org.springframework.security:spring-security-test")
```
vbala-moj, line 96: This is only for this POC, as I need the user from the security context, and this approach provides that easily.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/RevInfo.kt
```diff
@@ -0,0 +1,69 @@
...
+  /*
+   *Added a few convenience functions to make it easier to work with the revision info as hibernate only
+   * support longs for any field marked as revision timestamp
+   */
+  fun getRevisionInstant(): Instant = Instant.ofEpochMilli(timestamp ?: 0)
+
+  fun getRevisionTime(): LocalDateTime = LocalDateTime.ofInstant(
+    Instant.ofEpochMilli(timestamp ?: 0),
+    ZoneId.systemDefault(),
+  )
```
vbala-moj, line 68: Not needed at all, but included purely for our understanding purposes.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/EnversPocTest.kt
```diff
@@ -0,0 +1,218 @@
...
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ReleasePlanEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ReleasePlanRepository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.RevInfo
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.AuditService
+import java.time.LocalDate
+import java.time.LocalTime
+import java.util.UUID
+
+class EnversPocTest : IntegrationTestBase() {
+
```
vbala-moj, line 22: Let me know if you want to test any other scenarios.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ReleasePlanEntity.kt
```diff
@@ -0,0 +1,50 @@
...
+
+@Entity
+@Table(name = "release_plan")
+@Audited
+class ReleasePlanEntity(
+  @Id
+  val id: UUID,
+
+  @ManyToOne
+  @Audited(targetAuditMode = RelationTargetAuditMode.NOT_AUDITED)
```
vbala-moj, line 35: This is needed to inform Envers that we are not interested in auditing this relationship if it is not part of auditing.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ReleaseActionEntity.kt
```diff
@@ -26,8 +16,4 @@ class ReleaseActionEntity(
   var description: String,
   var actionCadence: String,
   var otherInformation: String? = null,
-
-  @ManyToOne
-  @JoinColumn(name = "release_plan_id")
-  var releasePlan: ReleasePlanEntity,
 )
```
vbala-moj, line 19: Ideally, we should use a unidirectional approach for this requirement.
________________
**Pros:**

Simpler audit history.
Fewer revisions caused by association changes.
Easier querying of historical versions of the parent entity and its related child entities.

**Cons:**

Requires a separate Envers join audit table. ( Please see ReleasePlan for reasons.)
______________

The appropriate strategy should be selected based on the specific requirements and auditing needs.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ReleasePlanEntity.kt
```diff
@@ -42,9 +44,10 @@ class ReleasePlanEntity(
...
 
   @OneToMany(
-    mappedBy = "releasePlan",
     cascade = [CascadeType.ALL],
     orphanRemoval = true,
   )
+  @JoinColumn(name = "release_plan_id", nullable = false)
+  @AuditJoinTable(name = "release_plan_release_action_aud")
   var releaseActions: MutableList<ReleaseActionEntity> = mutableListOf(),
 )
```
vbala-moj, line 53: Why does Envers require a join audit table?

No foreign key on the parent side:
In a unidirectional mapping, only the child entity contains the foreign key. The parent table does not maintain a direct column for the relationship (this is standard Hibernate behavior for such mappings).
Avoiding audit record bloat:
Without a separate audit join table, changes to fields on the parent entity could unintentionally trigger redundant audit entries for all associated child entities, leading to unnecessary duplication and noise in the audit history.
Independent relationship history tracking:
The join audit table enables Envers to track association changes—such as additions or removals from a collection—independently of changes to the child entity itself. This ensures accurate history of relationship evolution without conflating it with attribute-level changes.

https://discourse.hibernate.org/t/auditjointable-in-envers/3445/2

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AuditService.kt
```diff
@@ -41,4 +57,163 @@ class AuditService(
...
+        action.id to getReleaseActionRevisions(action.id)
+      }
+    val allRevisions = planRevisionsById.values.flatten() + actionRevisionsById.values.flatten()
+
+    return allRevisions
+      .groupBy { it.metadata.revisionNumber.orElse(0) }
+      .map { (revisionNumber, revisions) ->
+        buildTimeline(revisionNumber!!, revisions, planRevisionsById, actionRevisionsById)
+      }
+      .sortedByDescending { it.occurredAt }
```
vbala-moj, line 85: My initial attempt at building the timeline worked well for most scenarios, but it failed when trying to retrieve the timeline for a deleted release plan. The root cause is that the approach itself is fundamentally flawed.

This implementation relies on loading revision history from the current (live) entities. Once an entity is deleted, however, there is no longer a way to retrieve its revisions because the entity ID is no longer available through the live data model.

As a result, this approach cannot reliably support audit history for deleted entities and needs to be reconsidered. (**See AuditTimelineService**) 

See **"Track Full Lifecycle of Release Plan and Actions"** in **EnversPoctest.kt** for the expected output and behavior of this logic.

src/main/resources/db/migration/all/20260527131658__add_audit_tables.sql
```diff
@@ -57,3 +60,13 @@ CREATE TABLE release_action_aud
...
+
+CREATE TABLE release_plan_release_action_aud
+(
+    rev             INTEGER NOT NULL,
+    release_plan_id UUID    NOT NULL,
+    id              UUID    NOT NULL,
+    revtype         SMALLINT,
+    PRIMARY KEY (rev, release_plan_id, id),
+    CONSTRAINT fk_release_plan_release_action_aud_revinfo FOREIGN KEY (rev) REFERENCES revinfo (id)
+);
```
vbala-moj, line 72: This is the new audit join table required when using a unidirectional relationship. Hibernate can generate this table automatically when schema generation is enabled. However, since we use Flyway for database migrations, the table must be created manually and explicitly referenced in the entity mapping.

src/main/resources/application.yml
```diff
@@ -46,6 +46,7 @@ spring:
     hibernate:
       ddl-auto: none
     properties:
+      org.hibernate.envers.store_data_at_delete: true
```
vbala-moj, line 49: This issue took me quite some time to debug. I could see the delete revision present during debugging, but it never appeared in the timeline. There are a couple of reasons for this behavior.

1. **Envers does not store most field values for delete operations by default.**
   Since the entity is being removed, Envers considers storing the full state unnecessary unless explicitly configured otherwise. As a result, the delete revision only contains the Release Plan ID and revision metadata, while most other fields are stored as `null`.

2. **My second approach (`AuditTimelineService`) builds timeline entries using the Space Booking ID.**
   While this is generally the identifier we want to query by, it does not work for delete revisions because Envers does not store the Space Booking foreign key in the delete revision. Even if auditing is enabled on the Space Booking entity, Envers still does not persist foreign key values for deleted entities by default.

If we want delete revisions to appear in the timeline, we have a couple of options:

### Option 1: Configure Envers to store entity data on delete

By enabling this Envers setting, the full entity state is stored for delete revisions, making them available for timeline reconstruction.

### Option 2: Implement custom logic

**Pros:**

* No additional storage required for delete revisions.
* Works well if the timeline only needs to display previous values for presentation purposes.

**Cons:**

* Increases query and reconstruction complexity.
* Requires additional logic to resolve historical relationships and entity state.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/EnversPocTest.kt
```diff
@@ -94,67 +93,90 @@ class EnversPocTest : IntegrationTestBase() {
...
+     *
+     * --------------------------
+     */
+
+    val timelines = auditService.getAuditTimeLineForSpaceBooking(spaceBooking)
+
+    println("--- Audit Timeline ---")
+    timelines.forEach { println(it.render()) }
+    println("--------------------------")
   }
```
vbala-moj, line 180: Uses Approach 1: Building the timeline from live entities

This approach works well for create and update operations but does not reliably support delete operations, as deleted entities can no longer be resolved from the live data model.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/EnversPocTest.kt
```diff
@@ -194,6 +216,102 @@ class EnversPocTest : IntegrationTestBase() {
...
+     * - Created release plan with description 'Initial Plan'
+     *
+     * Release Actions
+     * - Created release action with description 'Action 1'
+     */
+
+    println("--- Audit Timeline ---")
+    timelines.forEach { println(it.render()) }
+    println("--------------------------")
+  }
```
vbala-moj, line 313: Uses Approach 2: Building the timeline from revisions

This approach builds the timeline directly from the audit revisions rather than relying on the current state of live entities. As a result, it works consistently for create, update, and delete operations, providing complete lifecycle tracking even after an entity has been removed.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AuditTimelineService.kt
```diff
@@ -0,0 +1,249 @@
...
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ReleasePlanEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.RevInfo
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.AuditTimeLine
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.AuditTimeLineSection
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.EventType
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.RevisionSnapshot
+import java.util.UUID
+
+@Service
+class AuditTimelineService(
```
vbala-moj, line 22: Better approach for building the timeline

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AuditTimelineService.kt
```diff
@@ -0,0 +1,249 @@
...
+    val snapshots = loadSnapshots(spaceBooking)
+
+    return buildTimeline(snapshots)
+  }
+
+  private fun loadSnapshots(
+    spaceBooking: Cas1SpaceBookingEntity,
+  ): List<RevisionSnapshot> {
+    val planEvents = loadPlanEvents(spaceBooking)
+    val actionEvents = loadActionEvents(planEvents.map { it.entityId }.toSet())
```
vbala-moj, line 42: This implementation fetches all revisions associated with a Space Booking ID and uses them to build the corresponding timeline.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AuditTimelineService.kt
```diff
@@ -0,0 +1,249 @@
...
+        """
+        SELECT DISTINCT id
+        FROM release_plan_release_action_aud
+        WHERE release_plan_id IN (:releasePlanIds)
+        """.trimIndent(),
+      )
+      .setParameter("releasePlanIds", releasePlanIds)
+      .resultList
+      .filterIsInstance<UUID>()
+
```
vbala-moj, line 77: Without a join audit table, the approach relies on loading all revisions and applying a filter on the Release Plan ID.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/AuditTimelineService.kt
```diff
@@ -0,0 +1,249 @@
...
+      timeline += AuditTimeLine(
+        revisionNumber = rev!!,
+        occurredAt = first.timestamp,
+        userName = first.user,
+        title = "Release plan updated",
+        sections = sections,
+      )
+    }
+
+    return timeline.sortedByDescending { it.revisionNumber }
```
vbala-moj, line 182: I have not yet grouped all plans and their revisions by revision.

davidatkinsuk:
closing this to clear down the pending PR list. branch will be retained for future ref. when we get to this work

PR #4797, opened 2026-05-29
FM-676 - Remove generated code for /cas1/placement-requests

davidatkinsuk:
This is to unblock development task https://dsdmoj.atlassian.net/browse/FM-676


daveawc (approved):
👍

PR #4803, opened 2026-06-01
FM-695 Reduce OASys calls in CAS2 HDC

davidatkinsuk:
Before this commit we were fetching the NeedDetails for every OAsys request, and using it populate metadata in the OAsys response. This is unneccessary because the fields required to populate that metadata are available on almost all responses from OASys (as represented by the `AssessmentInfo` abstract class).

This commit updates the `Cas2PeopleController` to make use of these metadata fields, allowing us to remove unneccessary upstream calls to get `NeedsDetail` for every OAsys request.


helm_deploy/values-prod.yaml
```diff
@@ -98,7 +98,7 @@ generic-service:
     USER-ALLOCATIONS_RULES_ESAP-ASSESSMENTS_ALLOCATE-TO-USER: RACHELBENJAMINNPS
 
     FEATURE-FLAGS_CAS2-SQS-LISTENER-ENABLED: true
-    FEATURE-FLAGS_CAS2-OASYS-USE-NEW-QUESTIONS: false
+    FEATURE-FLAGS_CAS2-OASYS-USE-NEW-QUESTIONS: true
```
danhumphreys-moj, line 101: What impact of this being false in prod has this had and what impact does it have making this true now?

danhumphreys-moj (approved):
LGTM

helm_deploy/values-prod.yaml
davidatkinsuk: See this commit log for an explanation https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/4803/changes/7841ee8da9ab645cc471e4dc1e077548e4a22a8f

Also see this discussion https://mojdt.slack.com/archives/C08C440KMM4/p1778764624154629

danhumphreys-moj: Thanks for the references @davidatkinsuk

PR #4804, opened 2026-06-02
FM-696 Remove unneccessary CAS3 OASys Calls

davidatkinsuk:
Before this commit we were fetching the `OffenceDetails` for every OAsys request, and using it populate metadata in the OAsys response. This is unneccessary because the fields required to populate that metadata are available on almost all responses from OASys (as represented by the `AssessmentInfo` abstract class).

This commit updates the `Cas3PeopleController` to make use of these metadata fields, allowing us to remove unneccessary upstream calls to get `OffenceDetails ` for every OAsys request.


danhumphreys-moj (approved):
LGTM!

PR #4809, opened 2026-06-02
FM-629 Tidy up and improve Cas2v2 Report Tests

davidatkinsuk:
This PR makes various changes to `Cas2v2ReportsTest` to reduce the amount of repeated boiler plate code, pulling focus to the important parts of the test data being configured.

It also ensures that HDC applications are excluded from all reports


daveawc (approved):
👍

PR #4812, opened 2026-06-03
Audit POC - Javers

vbala-moj:
From my initial Javers PoC, it seems much easier to generate the timeline JSON with minimal engineering effort. Unlike Envers, which primarily tracks entity revisions, Javers tracks object-level changes and relationships within the object graph.

Based on what I've seen so far, Javers appears to be a better fit if our primary goal is to provide users with a timeline of changes. It can capture creates, updates, deletes, and collection changes with significantly less custom logic than the Envers approach.

In my PoC, I performed a simple exercise by porting a timeline record from SAS. SAS is primarily interested in timelines at the individual entity level, whereas our requirement is to generate timelines at the Space Booking level.

I haven't spent much time implementing that part yet, but based on what I've seen so far, it appears to require some additional development to set up a CommitPropertiesProvider so that Space Booking context can be associated with each commit and later used as a queryable property when retrieving timeline entries.

Assuming that approach works as expected, we should be able to retrieve all changes related to a Space Booking without having to reconstruct the timeline from entity relationships, which is one of the areas where the Envers solution becomes more complex.

********************************
I like Envers from an ORM and persistence perspective, as it integrates naturally with Hibernate and provides automatic auditing with minimal setup. However, when it comes to the complex queries and aggregation logic required to build a user-facing timeline, Javers provides much more out-of-the-box functionality.

As a result, I'm increasingly leaning towards Javers for this particular use case. From a UI perspective, it aligns more closely with our requirements and significantly reduces the amount of custom code needed to generate timeline events.

The main limitation with Envers, in my view, is the additional effort required whenever we introduce a new entity to audit or need to expose new types of changes in the timeline. This typically involves adding new audit queries and updating aggregation logic. In contrast, with Javers, this is largely handled automatically, which reduces ongoing maintenance overhead.
*******************************

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/api/model/TimeLineRecord.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/api/model/TimeLineRecord.kt
@@ -1,42 +1,10 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model
 
 import com.fasterxml.jackson.annotation.JsonFormat
+import com.fasterxml.jackson.annotation.JsonInclude
 import java.time.Instant
 
-/**
- * {
- *   "author": "audit_user",
- *   "commitDate": "2026-06-03T06:35:03Z",
- *   "sections": [
- *     {
- *       "entityType": "RELEASE_PLAN",
- *       "changes": [
- *         {
- *           "changeType": "UPDATE",
- *           "field": "description",
- *           "oldValue": "Initial Plan",
- *           "newValue": "Updated Plan",
- *           "message": "Description changed from 'Initial Plan' to 'Updated Plan'"
- *         }
- *       ]
- *     },
- *     {
- *       "entityType": "RELEASE_ACTION",
- *       "changes": [
- *         {
- *           "changeType": "CREATE",
- *           "message": "Created release action with description 'Action 2'"
- *         },
- *         {
- *           "changeType": "DELETE",
- *           "message": "Deleted release action with description 'Action 1'"
- *         }
- *       ]
- *     }
- *   ]
- * }
- */
-
+@JsonInclude(JsonInclude.Include.NON_NULL)
 data class TimelineRecord(
   val author: String,
 
@@ -58,8 +26,10 @@ data class TimelineSection(
 enum class AuditEntityType {
   RELEASE_PLAN,
   RELEASE_ACTION,
+  MONITORING_INFORMATION,
 }
 
+@JsonInclude(JsonInclude.Include.NON_NULL)
 data class TimelineChange(
 
   val changeType: ChangeType,
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/api/model/TimeLineRecord.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/api/model/TimeLineRecord.kt
@@ -1,0 +1,41 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model
+
+import com.fasterxml.jackson.annotation.JsonFormat
+import com.fasterxml.jackson.annotation.JsonInclude
+import java.time.Instant
+
+data class TimelineRecord(
+  val type: AuditRecordType,
+  val author: String,
+  @field:JsonFormat(
+    shape = JsonFormat.Shape.STRING,
+    pattern = "yyyy-MM-dd'T'HH:mm:ss'Z'",
+    timezone = "UTC",
+  )
+  val commitDate: Instant,
+  val changes: List<FieldChange>,
+  @field:JsonInclude(JsonInclude.Include.NON_NULL)
+  val extraInformation: Map<String, String>? = null,
+)
+
+enum class AuditRecordType {
+  CREATE,
+  UPDATE,
+  DELETE,
+}
+
+interface FieldChange {
+  var field: String
+  var value: String?
+}
+
+data class UpdateFieldChange(
+  override var field: String,
+  override var value: String?,
+  var oldValue: String?,
+) : FieldChange
+
+data class CreateFieldChange(
+  override var field: String,
+  override var value: String?,
+) : FieldChange
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/api/model/TimeLineRecord.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/api/model/TimeLineRecord.kt
@@ -1,41 +1,86 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model
 
 import com.fasterxml.jackson.annotation.JsonFormat
-import com.fasterxml.jackson.annotation.JsonInclude
 import java.time.Instant
 
+/**
+ * {
+ *   "author": "audit_user",
+ *   "commitDate": "2026-06-03T06:35:03Z",
+ *   "sections": [
+ *     {
+ *       "entityType": "RELEASE_PLAN",
+ *       "changes": [
+ *         {
+ *           "changeType": "UPDATE",
+ *           "field": "description",
+ *           "oldValue": "Initial Plan",
+ *           "newValue": "Updated Plan",
+ *           "message": "Description changed from 'Initial Plan' to 'Updated Plan'"
+ *         }
+ *       ]
+ *     },
+ *     {
+ *       "entityType": "RELEASE_ACTION",
+ *       "changes": [
+ *         {
+ *           "changeType": "CREATE",
+ *           "message": "Created release action with description 'Action 2'"
+ *         },
+ *         {
+ *           "changeType": "DELETE",
+ *           "message": "Deleted release action with description 'Action 1'"
+ *         }
+ *       ]
+ *     }
+ *   ]
+ * }
+ */
+
 data class TimelineRecord(
-  val type: AuditRecordType,
   val author: String,
+
   @field:JsonFormat(
     shape = JsonFormat.Shape.STRING,
     pattern = "yyyy-MM-dd'T'HH:mm:ss'Z'",
     timezone = "UTC",
   )
   val commitDate: Instant,
-  val changes: List<FieldChange>,
-  @field:JsonInclude(JsonInclude.Include.NON_NULL)
-  val extraInformation: Map<String, String>? = null,
+
+  val sections: List<TimelineSection>,
 )
 
-enum class AuditRecordType {
+data class TimelineSection(
+  val entityType: AuditEntityType,
+  val changes: List<TimelineChange>,
+)
+
+enum class AuditEntityType {
+  RELEASE_PLAN,
+  RELEASE_ACTION,
+}
+
+data class TimelineChange(
+
+  val changeType: ChangeType,
+
+  /**
+   * Message for UI.
+   */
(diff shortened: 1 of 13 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/api/model/TimeLineRecord.kt
```diff
@@ -0,0 +1,41 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model
+
+import com.fasterxml.jackson.annotation.JsonFormat
+import com.fasterxml.jackson.annotation.JsonInclude
+import java.time.Instant
+
+data class TimelineRecord(
```
vbala-moj, line 8: Copied from SAS 😀

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ReleasePlanEntity.kt
```diff
@@ -0,0 +1,53 @@
...
+import org.javers.core.metamodel.annotation.Entity as JaversEntity
+import org.javers.core.metamodel.annotation.TypeName
+import org.javers.spring.annotation.JaversSpringDataAuditable
+import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.stereotype.Repository
+import java.time.LocalTime
+import java.util.UUID
+
+@Repository
+@JaversSpringDataAuditable
```
vbala-moj, line 25: @JaversSpringDataAuditable is the only annotation required to enable JaVers to record commit-level changes.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ReleaseActionEntity.kt
```diff
@@ -0,0 +1,21 @@
...
+import jakarta.persistence.Entity
+import jakarta.persistence.Id
+import jakarta.persistence.Table
+import org.javers.core.metamodel.annotation.Entity as JaversEntity
+import org.javers.core.metamodel.annotation.TypeName
+import java.util.UUID
+
+@Entity
+@JaversEntity
+@TypeName("ReleaseAction")
```
vbala-moj, line 13: Required to instruct JaVers to record changes to this entity and its child entities.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/JaversTimelineService.kt
```diff
@@ -0,0 +1,37 @@
...
+
+  private fun ChangesByCommit.toTimelineRecord(): TimelineRecord {
+    val commit = getCommit()
+    return TimelineRecord(
+      type = AuditRecordType.UPDATE,
+      author = commit.author,
+      commitDate = commit.commitDateInstant,
+      changes = get().map { it.toString() }.map { CreateFieldChange(field = "Change", value = it) },
+      extraInformation = mapOf("commitId" to commit.id.toString()),
+    )
```
vbala-moj: Simple function for POC but we can build on this to record create, update, and delete commits as entries in an audit timeline.

src/main/resources/db/migration/all/20260603121000__add_javers_tables.sql
```diff
@@ -0,0 +1,59 @@
+CREATE SEQUENCE jv_global_id_pk_seq START WITH 1 INCREMENT BY 1;
+CREATE SEQUENCE jv_commit_pk_seq START WITH 1 INCREMENT BY 1;
+CREATE SEQUENCE jv_snapshot_pk_seq START WITH 1 INCREMENT BY 1;
+
+CREATE TABLE jv_global_id
```
vbala-moj, line 5: Copied from the internet!

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ReleasePlanEntity.kt
```diff
@@ -0,0 +1,76 @@
...
+    name = "monitoring_information",
+    joinColumns = [JoinColumn(name = "release_plan_id")],
+  )
+  var monitoringInformation: MutableList<MonitoringInformation> = mutableListOf(),
+)
+
+@TypeName("MonitoringInformation")
+@Embeddable
+@Table(name = "monitoring_information")
+class MonitoringInformation(
```
vbala-moj, line 70: Used this as a valueobject.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ReleasePlanEntity.kt
```diff
@@ -0,0 +1,76 @@
...
+  var otherInformation: String?,
+
+  @OneToMany(
+    cascade = [CascadeType.ALL],
+    orphanRemoval = true,
+  )
+  @JoinColumn(name = "release_plan_id", nullable = false)
+  var releaseActions: MutableList<ReleaseActionEntity> = mutableListOf(),
+
+  @ElementCollection(fetch = FetchType.LAZY)
```
vbala-moj, line 59: When a value object represents a collection, it must be mapped to a separate table. If it is not a collection, its fields are persisted within the parent table.

If the value object's field names conflict with fields in the parent entity, you must override the column names to avoid duplicate column mappings. This can be done either by using `@Column(name = "...")` within the value object itself or by applying `@AttributeOverrides` in the parent entity. Otherwise, Hibernate will report a duplicate column name mapping error.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ReleasePlanEntity.kt
```diff
@@ -0,0 +1,76 @@
...
+  @CollectionTable(
+    name = "monitoring_information",
+    joinColumns = [JoinColumn(name = "release_plan_id")],
+  )
```
vbala-moj, line 63: You only need to use `@CollectionTable` when the table name or join column name defined in Flyway differs from Hibernate's default naming conventions.

If `@CollectionTable` is not specified, Hibernate derives the table and join column names automatically. For example, for a collection field named `monitoringInformation` in the `ReleasePlan` entity, Hibernate may expect a table name such as `release_plan_monitoring_information`. If the actual table name in the database is different, you must explicitly configure it using `@CollectionTable`.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/TimelineMapperV2.kt
```diff
@@ -0,0 +1,131 @@
...
+import org.javers.core.diff.changetype.PropertyChange
+import org.javers.core.diff.changetype.ValueChange
+import org.javers.core.diff.changetype.container.ContainerChange
+import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ChangeType
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.TimelineChange
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.TimelineRecordV2
+
+@Component
+class TimelineMapperV2 {
```
vbala-moj, line 16: [TimelineRecordV2.json](https://github.com/user-attachments/files/28697169/TimelineRecordV2.json)

Ideally, we want a solution like this. The main challenge is creating a generic mapper, which would require us to rely on JaVers' built-in change types and diff model. There is a learning curve involved, but it should be achievable with some focused effort.

The main unknown is how JaVers represents diffs when multiple types of operations (for example, additions, updates, and removals) are included in a single commit. We'll need to investigate and debug the generated diffs to understand the structure and behavior before we can build a reliable generic mapper around them.

[javers-raw-output.txt](https://github.com/user-attachments/files/28697567/javers-raw-output.txt)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/TimeLineMapper.kt
```diff
@@ -0,0 +1,105 @@
...
+import org.javers.core.diff.changetype.ValueChange
+import org.javers.core.diff.changetype.container.ContainerChange
+import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.AuditEntityType
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ChangeType
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.TimelineChange
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.TimelineSection
+
+@Component
+class TimeLineMapper {
```
vbala-moj, line 18: [TimelineRecordV1.json](https://github.com/user-attachments/files/28697178/TimelineRecordV1.json)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/config/JaversConfig.kt
```diff
@@ -0,0 +1,33 @@
...
+
+    override fun provideForCommittedObject(savedDomainObject: Any): Map<String, String> = propertiesFor(savedDomainObject)
+
+    override fun provideForDeletedObject(deletedDomainObject: Any): Map<String, String> = propertiesFor(deletedDomainObject)
+
+    private fun propertiesFor(domainObject: Any): Map<String, String> = when (domainObject) {
+      is ReleasePlanEntity -> releasePlanCommitProperties(domainObject)
+      else -> emptyMap()
+    }
+  }
```
vbala-moj, line 28: This is required for each entity type that will be queried through JaVers. Each entity type must be explicitly configured so that JaVers can correctly identify and query snapshots for that specific entity.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/TimelineMapperV2.kt
```diff
@@ -0,0 +1,131 @@
...
+     * For UPDATE commit using collections
+     */
+
+    /*    is ContainerChange<*> -> TimelineChange(
+          changeType = ChangeType.UPDATE,
+          message = "${fieldLabel().capitalize()} changed",
+          field = fieldLabel(),
+          oldValue = left.toTimelineValue(),
+          newValue = right.toTimelineValue(),
+        )*/
```
vbala-moj, line 78: These outputs are quite noisy, and unless we specifically need to use these diffs, I think we can safely omit them and use the remaining output to build the mapper.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/TimelineMapperV2.kt
```diff
@@ -0,0 +1,131 @@
...
+
+    is ValueChange ->
+      if (left == null || right == null) {
+        return null
```
vbala-moj, line 82: This is specifically needed to prevent nulls appearing in the message, whether something is being created or deleted.

davidatkinsuk:
closing this to clear down the pending PR list. branch will be retained for future ref. when we get to this work

PR #4813, opened 2026-06-03
FM-708 Rename main and test cas2 to cas2hdc

danhumphreys-moj:
Ref: https://dsdmoj.atlassian.net/browse/FM-708

Right click -> rename - main and test cas2 to cas2hdc

was: src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2

now: src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2hdc

was: src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2

now: src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2hdc


davidatkinsuk (approved):
LGTM

could you update the detekt.yml in the root of the project please, it's the only place that i think also needs updating for the new package

danhumphreys-moj:
> LGTM
> 
> could you update the detekt.yml in the root of the project please, it's the only place that i think also needs updating for the new package

Done this in a new commit.

PR #4817, opened 2026-06-03
Feature/fm 708 package structure updates

danhumphreys-moj:
Ref: https://dsdmoj.atlassian.net/browse/FM-708

Changes in line with: https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/main/doc/architecture/decisions/0029-package-structure.md

This PR updates the CAS2 HDC package structure by moving previously `model`, `reporting.model`, `seed`, and migration classes into new `dto` and `jobs.*` packages, and then updating all main and test code to reference the new package locations.

Changes:

- Move CAS2 HDC API/domain data classes from `cas2hdc.model` / `cas2hdc.reporting.model*` into `cas2hdc.dto*`.
- Move CAS2 HDC seed and migration jobs into `cas2hdc.jobs.seed` and `cas2hdc.jobs.migration`.
- Update controllers, services, transformers, JPA entities, and tests to use the new import paths.


davidatkinsuk (approved):
LGTM 👍

PR #4825, opened 2026-06-05
FM-506 Add `suitability strategy` to CAS1 OASys endpoint

davidatkinsuk:
This allows the UI to request information for assessments older than 6 months

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/dto/OASysAssessmentSuitabilityStrategyDto.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/dto/OASysAssessmentSuitabilityStrategyDto.kt
@@ -1,0 +1,16 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.cas1.dto
+
+import com.fasterxml.jackson.annotation.JsonCreator
+import com.fasterxml.jackson.annotation.JsonValue
+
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/dto/OASysAssessmentSuitabilityStrategyDto.kt
```diff
@@ -0,0 +1,16 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.cas1.dto
+
+import com.fasterxml.jackson.annotation.JsonCreator
+import com.fasterxml.jackson.annotation.JsonValue
+
+enum class OASysAssessmentSuitabilityStrategyDto(@get:JsonValue val value: String) {
```
daveawc, line 6: Does this name need `Dto` on the end?

Also, I guess you're following the patter, but just flagging you can get rid of the jsonvalue/creator stuff as this is a new enum, and just use the actual enum name throughout :)

davidatkinsuk: We've started putting DTO on everything that is exposed to the API to differentiate it from external types

I figured we have so many enums already we should probably just follow the pattern :)

daveawc (approved):
👍

PR #4831, opened 2026-06-08
FM-621 Add active booking validation

vbala-moj:
https://dsdmoj.atlassian.net/browse/FM-621

Prevent creation of a new booking when an active space booking already exists for the placement request.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1SpaceBookingCreateService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1SpaceBookingCreateService.kt
@@ -82,6 +82,10 @@ class Cas1SpaceBookingCreateService(
       "$.placementRequestId" hasValidationError "doesNotExist"
     }
 
+    if (!details.isTransfer && placementRequest?.hasActiveBooking() == true) {
+      "$.placementRequestId" hasValidationError "alreadyHasAnInitialSpaceBooking"
+    }
+
     if (!premises.supportsSpaceBookings) {
       "$.premisesId" hasValidationError "doesNotSupportSpaceBookings"
     }
@@ -168,5 +172,6 @@ class Cas1SpaceBookingCreateService(
     val transferredFrom: TransferInfo?,
     val transferReason: TransferReason?,
     val additionalInformation: String?,
+    val isTransfer: Boolean = transferReason != null,
   )
 }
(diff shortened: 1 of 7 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1SpaceBookingCreateService.kt
```diff
@@ -168,5 +172,6 @@ class Cas1SpaceBookingCreateService(
     val transferredFrom: TransferInfo?,
     val transferReason: TransferReason?,
     val additionalInformation: String?,
+    val isTransfer: Boolean = transferReason != null || transferredFrom != null,
```
vbala-moj: The transfer reason is optional, so relying on it alone is insufficient. We should also verify that transferredFrom != null in the isTransfer logic.

vbala-moj: Alternatively, we could rely solely on transferredFrom, since it is always populated for transfers.

davidatkinsuk: `transferredFrom` is deprecated on the space booking entity. Neither it or `transferType` is set for any space bookings i.e.

```select count(*) from cas1_space_bookings where transferred_from is not null or transfer_type is not null;```

Returns 0

This was added for change request functionality but unfortunately not used when the 'new' transfer functionality was added. Therefore, we should just be using `transferReason` here

davidatkinsuk (approved):
LGTM. Could you please do the following before merging:

* Adjust the `isTransfer` logic according to my comment
* See if you can make a UI change to ensure this error displays in a readable manner. There's a subtask for this - https://dsdmoj.atlassian.net/browse/FM-622

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1SpaceBookingCreateService.kt
vbala-moj: I'll revert the change and update the emergency/planned transfer ITs, as they're currently failing due to the transfer reason being null.

davidatkinsuk: we probably should deprecate the fields on `CreateBookingDetails` and the corresponding `UpdateBookingDetails` too

PR #4832, opened 2026-06-08
Make ReferredBy fields non-nullable.

daveawc:
This PR makes the refferedBy field on the ReferralHistory DTOs non-nullable, as this data comes from non-nullable fields.


davidatkinsuk (approved):
LGTM 👍

PR #4833, opened 2026-06-09
feature/SAS-659-Create-endpoint-on-CAS-to-return-current-CAS1-placement

besscerule:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/external/Cas1ExternalApplicationsController.kt
davidatkinsuk: i appreciate this was already here, but i think it should be enumerated?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/external/Cas1ExternalApplicationService.kt
```diff
@@ -21,68 +22,82 @@ class Cas1ExternalApplicationService(
...
 ) {
-  fun getSuitableApplicationByCrn(crn: String): Cas1SuitableApplication? = approvedPremisesApplicationRepository.findByCrn(crn)
-    .maxWithOrNull(
-      compareBy<ApprovedPremisesApplicationEntity> { suitableStatusesAsc[it.status] }
-        .thenBy { it.submittedAt ?: it.createdAt },
-    )
-    ?.let { application ->
 
-      val rfps = (cas1RequestForPlacementService.getRequestsForPlacementByApplication(application.id, requestingUser = null) as CasResult.Success).value
+  fun getPlacementHistories(applicationId: UUID): List<Cas1PlacementHistory> {
```
davidatkinsuk, line 28: can be private?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/external/Cas1ExternalApplicationService.kt
```diff
@@ -21,68 +22,82 @@ class Cas1ExternalApplicationService(
...
+    Cas1SuitableApplication(
+      id = application.id,
+      applicationStatus = application.status,
+      requestForPlacementStatus = suitablePlacement?.requestForPlacementStatus,
+      placementStatus = suitablePlacement?.placementStatus,
+      premises = suitablePlacement?.premises,
+    )
+  }
+
+  fun getSuitableApplication(crn: String): ApprovedPremisesApplicationEntity? = approvedPremisesApplicationRepository.findByCrn(crn)
```
davidatkinsuk: can be private?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/external/Cas1ExternalApplicationsController.kt
davidatkinsuk: something to consider

* types is 'suitable' or 'arrived'. Given where 'type' is in the URL, it implies that the restriction is applying to the application, but it's actually applying to the placement info being returned within in the response (i.e. all responses are 'suitable')
* I think it may be better implemented as something like

/cases/{crn}/applications?placementRequirement=latest|resident

* taking it one step further, it would be nice if all the placements were returned to SAS and it did the filtering itself? Or this endpoint should defined in terms of placements, not applications

I think at the very least the type 'arrived' should be renamed to 'resident'

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/external/Cas1ExternalApplicationService.kt
```diff
@@ -21,68 +22,82 @@ class Cas1ExternalApplicationService(
...
+    .maxWithOrNull(
+      compareBy<ApprovedPremisesApplicationEntity> { suitableStatusesAsc[it.status] }
+        .thenBy { it.submittedAt ?: it.createdAt },
+    )
 
-      suitablePlacement?.isSuitable = true
+  fun getArrivedApplicationByCrn(crn: String): Cas1SuitableApplication? = getSuitableApplication(crn)?.let { application ->
```
davidatkinsuk: we've discussed this before, but it's possible that the person is currently resident on a placement for an 'older' application. Maybe for the `arrived` type, the way the application is found needs changing? It could just be a case of looking for all space bookings for the CRN and seeing if an have an arrival without departure?

davidatkinsuk (commented):
I have thoughts. sorry.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/external/Cas1ExternalApplicationService.kt
besscerule: unit tests

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/external/Cas1ExternalApplicationService.kt
besscerule: unit tests

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/external/Cas1ExternalApplicationsController.kt
besscerule: there is an argument that maybe this should be a different end point entirely and not linked to the suitable applications

i don't think all placements should be returned because for suitable sometimes you can have just request for placement

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/external/Cas1ExternalApplicationService.kt
besscerule: i think this works

it will find the most suitable application that has an ARRIVED placement or return null - which is what we want

besscerule: ahhh wait

besscerule: i see your point

davidatkinsuk (approved):
LGTM 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/controller/external/Cas1ExternalApplicationsController.kt
```diff
@@ -12,14 +12,18 @@ class Cas1ExternalApplicationsController(
   private val cas1ExternalApplicationService: Cas1ExternalApplicationService,
 ) {
   @PreAuthorize("hasRole('APPROVED_PREMISES__SINGLE_ACCOMMODATION_SERVICE')")
-  @GetMapping("/cases/{crn}/applications/{type}")
-  fun getApplicationsByCrnAndType(
+  @GetMapping("/cases/{crn}/suitable-application")
```
daveawc: if this url is changing, the urls in cas2 and 3 should change as well for consistency. As it will break SAS, would it be better to put through in their own refactoring PRs?

besscerule: yeah i will remove this fro now and add it to a group pr once this work is done

PR #4834, opened 2026-06-09
Suffix `RiskTier` with `Dto`

davidatkinsuk:
We currently have an API and internal Model to represent `PersonRisks`.

This commit is the first step in clearly identifying these two models. For now we only change the `RisksEnvelope` as to avoid clashes with the large CAS2HDC PR


daveawc (approved):
👍

PR #4835, opened 2026-06-09
FM-711 Remove unused CAS2 risk fields

davidatkinsuk:
A `risk` property exists on both the CAS2 HDC and CAS2 Application Summary DTOs, but it is never populated. This PR removes the property.


daveawc (approved):
👍

PR #4836, opened 2026-06-09
Remove explicit use of `PGObject`

davidatkinsuk:
We’ve seen an issue related to PGObject when ugprading to kotlin 2.4.0

```/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/SubjectAccessRequestRepositoryBase.kt:309:90 Type annotation class 'org.checkerframework.checker.nullness.qual.Nullable' of the inferred type is inaccessible. Check the module classpath for missing or conflicting dependencies.```

As `PGObject.toString()` is the same as PGObject.value`, we don’t need to explicitly reference this class

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/SubjectAccessRequestRepositoryBase.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/SubjectAccessRequestRepositoryBase.kt
@@ -1,6 +1,5 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity
 
-import org.postgresql.util.PGobject
 import org.springframework.jdbc.core.namedparam.MapSqlParameterSource
 import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate
 import org.springframework.stereotype.Repository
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/SubjectAccessRequestRepositoryBase.kt [hunk 1]
```diff
@@ -306,7 +305,7 @@ open class SubjectAccessRequestRepositoryBase(val jdbcTemplate: NamedParameterJd
     return toJsonString(result)
   }
 
-  protected fun toJsonString(result: Map<String, Any?>) = (result["json"] as PGobject?)?.value
+  protected fun toJsonString(result: Map<String, Any?>): String? = result["json"]?.toString()
```
danhumphreys-moj, line 308: Curious why it wasn't done like this in the first place. I wonder what the original motivation for using PGobject here was.

danhumphreys-moj: Nice solution however!

danhumphreys-moj (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/SubjectAccessRequestRepositoryBase.kt
davidatkinsuk: I think this may have been me. I guess i inspected what was coming back in the query and decided to cast and use `value`, without appreciating that `toString` would give me the exact same thing. I still think explicitly casting is the better solution as I don't love relying on `toString` behaviour, but this let's us move the PG driver out of the compile classpath and tests would catch issues anyway

PR #4839, opened 2026-06-10
Reduce prod memory requirements

davidatkinsuk:
The headroom is excessive given the heap size


danhumphreys-moj (approved):
LGTM - I know we have been monitoring preprod for a while which has had the same reduced memory values for some time now without issue.

PR #4840, opened 2026-06-10
FM-724 Remove deprecated `SubmitPlacementApplication.placementType`

davidatkinsuk:
This has been long replaced by `releaseType` and `sentenceType`. The UI is no longer populating this value.

Removing this allows us to make releaseType mandatory, removing the need for some validation on this field.


daveawc (approved):
👍

PR #4841, opened 2026-06-10
FM-593 - Use common workflows/actions where possible

davidatkinsuk:
* Use common code ql workflow - This will now run daily, instead of with every PR
* Use common verify workflow for pipeline build gradle
* Use common verify workflow for pipeline ktlint/detekt
* Use common verify workflow for pipeline unit tests
* Use common actions where possible for pipeline integration tests


daveawc (approved):
👍

PR #4842, opened 2026-06-10
FM-710 Use OAuth Token for Snyk Jobs

sajifdeva-moj:
Updated Snyk Workflow to use OAuth tokens instead of API tokens.


davidatkinsuk (approved):
LGTM 👍

PR #4843, opened 2026-06-10
FM-724 remove deprecated placement app placement dates

davidatkinsuk:
This has been long replaced by `requestedPlacementPeriods`. The UI is no longer populating this value.


daveawc (approved):
👍

PR #4844, opened 2026-06-10
FM-728 add cas2 mappings to cas2v2

danhumphreys-moj:
https://dsdmoj.atlassian.net/browse/FM-728

Adds `/cas2` route aliases for existing CAS2v2 endpoints (and security rules) so consumers can access CAS2v2 functionality via `/cas2/**` as well as `/cas2v2/**`.

Changes:

- Update CAS2v2 controller base mappings to accept both `/cas2v2` and `/cas2` (including external routes).
- Add matching Spring Security authorization rules for `/cas2/**` endpoints (mirroring the existing `/cas2v2/**` rules).
- Update CAS2v2 integration tests to call the new `/cas2/**` paths.
- Update OpenAPI config for new cas2
- Remove CasHdc type aliases from OpenAPI config since no longer used (function left since it will be required in a subsequent PR)


davidatkinsuk (approved):
LGMT 👍

PR #4846, opened 2026-06-10
feature/SAS-659-Create-endpoint-on-CAS-to-return-current-CAS3-placement

besscerule:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/api/model/Cas1SuitableApplication.kt
```diff
@@ -9,10 +9,10 @@ data class Cas1SuitableApplication(
   val applicationStatus: ApprovedPremisesApplicationStatus,
   val requestForPlacementStatus: RequestForPlacementStatus?,
   val placementStatus: Cas1SpaceBookingStatus?,
-  val premises: Cas1SuitablePremisesDto?,
+  val premises: Cas1PremisesSummary?,
 )
 
-data class Cas1SuitablePremisesDto(
+data class Cas1PremisesSummary(
```
davidatkinsuk: We already have 3 types named Cas1Premises*Summary, and an additional ApprovedPremises type.

Ideally we'd reuse one of those existing types (Cas1PremisesBasicSummary is close, but you'd need to add some fields. For now can we just indicate this type only exists for the external API somehow (i'm also coming to the thinking that this should really just be called the SAS API as what it returns is quite specific)

Can we please retain the Dto suffix, we're trying to do that for all API types going forward

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/model/Cas3SuitableApplication.kt
```diff
@@ -11,10 +11,10 @@ data class Cas3SuitableApplication(
   val applicationStatus: ApplicationStatus,
   val assessmentStatus: TemporaryAccommodationAssessmentStatus?,
   val bookingStatus: Cas3BookingStatus?,
-  val premises: Cas3SuitablePremisesDto?,
+  val premises: Cas3PremisesSummary?,
 )
 
-data class Cas3SuitablePremisesDto(
+data class Cas3PremisesSummary(
```
davidatkinsuk: As per the previous comment, can this be named specifically for the external API as we already have a bunch of premises API types

Alternatively, consider using `Cas3Premises` although i think that may be a bit overkill in this case

Can we please retain the Dto suffix, we're trying to do that for all API types going forward

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/service/Cas3ApplicationService.kt
```diff
@@ -55,43 +56,29 @@ class Cas3ApplicationService(
...
+  fun getSuitableApplicationByCrn(crn: String): Cas3SuitableApplication? = temporaryAccommodationApplicationRepository.findByCrnOrderByCreatedAtDesc(crn)
+    .firstOrNull {
       val now = OffsetDateTime.now(clock)
       when {
-        application.arrivalDate != null -> application.arrivalDate!! >= now
-        application.personReleaseDate != null -> application.personReleaseDate!! >= now.toLocalDate()
-        else -> application.createdAt >= now.minusMonths(2)
+        it.arrivalDate != null -> it.arrivalDate!! >= now
+        it.personReleaseDate != null -> it.personReleaseDate!! >= now.toLocalDate()
+        else -> it.createdAt >= now.minusMonths(2)
```
davidatkinsuk: this feels like a bit of a magic number. Could we please use a constant to explain why 2 months?

davidatkinsuk (approved):
A few thoughts on naming. LGTM 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/api/model/Cas1SuitableApplication.kt
besscerule: done

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/model/Cas3SuitableApplication.kt
besscerule: done

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/service/Cas3ApplicationService.kt
besscerule: done

davidatkinsuk (approved):
LGTM 👍

PR #4848, opened 2026-06-11
FM-725 CAS2v2 user details with delius info

vbala-moj:
https://dsdmoj.atlassian.net/browse/FM-725

Added `GET /cas2v2/users/{username}` returning user details including Delius probation area when applicable.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2hdc/service/Cas2HdcUserService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2hdc/service/Cas2HdcUserService.kt
@@ -1,10 +1,15 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.cas2hdc.service
 
+import org.springframework.http.HttpStatus
 import org.springframework.stereotype.Service
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas2hdc.jpa.entity.Cas2UserEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas2hdc.jpa.entity.Cas2UserRepository
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas2hdc.jpa.entity.Cas2UserType
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas2v2.model.Cas2ServiceOrigin
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas2v2.model.Cas2v2DeliusUserInfoDto
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas2v2.model.Cas2v2UserDto
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas2v2.model.Cas2v2UserTypeDto
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas2v2.model.ProbationAreaDto
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.ApDeliusContextApiClient
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.ClientResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.ManageUsersApiClient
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2v2/model/Cas2v2UserDto.kt
```diff
@@ -0,0 +1,22 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.cas2v2.model
+
+data class Cas2v2UserDto(
+  val userName: String,
```
davidatkinsuk: can we make this all lower case please (i've never been sure on this one but it's in the dictionary as 'username' so i guess we should try and use that going forward!)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2v2/model/Cas2v2UserDto.kt
```diff
@@ -0,0 +1,22 @@
...
+  val type: Cas2v2UserType,
+  val deliusUserInfo: Cas2v2DeliusUserInfoDto? = null,
+)
+
+enum class Cas2v2UserType {
```
davidatkinsuk: Could we please suffix this with `Dto`

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2hdc/service/Cas2HdcUserService.kt
```diff
@@ -49,6 +54,43 @@ class Cas2HdcUserService(
...
+      deliusUserInfo = deliusUserInfo,
+    )
+  }
+
+  private fun getDeliusUser(username: String): StaffDetail = when (
+    val staffDetailResponse = apDeliusContextApiClient.getStaffDetail(username)
+  ) {
+    is ClientResult.Success -> staffDetailResponse.body
+    is ClientResult.Failure.StatusCode -> when (staffDetailResponse.status) {
+      HttpStatus.NOT_FOUND -> throw NotFoundProblem(username, "DeliusUser")
```
davidatkinsuk: I think the second arg should be "username" because it's meant to specify the type of value provided in the first arg

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2hdc/service/Cas2HdcUserService.kt
```diff
@@ -49,6 +54,43 @@ class Cas2HdcUserService(
     return findOrCreateNomisUser(username = normalisedUsername, userDetail, serviceOrigin)
   }
 
+  fun getCas2v2UserDetails(userName: String): Cas2v2UserDto {
+    val existingUser =
+      cas2UserRepository.findByUsernameAndServiceOrigin(userName, Cas2ServiceOrigin.BAIL)
+        ?: throw NotFoundProblem(userName, "Cas2v2User")
```
davidatkinsuk: I think the second arg should be "username" because it's meant to specify the type of value provided in the first arg

davidatkinsuk (approved):
Just a few minor comments from me. Looks great 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2v2/model/Cas2v2UserDto.kt
vbala-moj: I noticed this, but I’m a bit unsure about adding the `Dto` suffix to the enum. In most cases in the codebase we’ve been using plain enum names rather than `*Dto`.

Even though this enum is used in the controller response, it’s still a value type rather than a DTO in itself.

Happy to change it though — is this something we’re applying consistently going forward?

davidatkinsuk: I agree it's not perfect, but it's better than using the same naming as internal _stuff_ so going forward let's suffix anything being returned via the API with `Dto`. We're also going to move this stuff into `Dto` packages at some point (see https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/main/doc/architecture/decisions/0029-package-structure.md)

PR #4849, opened 2026-06-11
Upgrade hmpps spring boot

davidatkinsuk:
This required some tweaks to our OAuth2 config classes as the spring boot interfaces no longer support nullable values

This also upgraded hibernate from 7.2.12 to 7.4.1.Final. This caused some change in behaviour wrt. deleteall() with FKs that referenced entries in the same table.


daveawc (approved):
nice!

PR #4856, opened 2026-06-12
Move common code into `common` package

davidatkinsuk:


danhumphreys-moj (approved):
LGTM - paired with @davidatkinsuk whilst we did this.

PR #4859, opened 2026-06-12
FM-735 Update Esap AssessmentRule

vbala-moj:
- Add support for default and near-arrival (<= 28 days) allocation usernames.
- Migrate configuration to use `cas1-allocations` secrets across all Helm environments.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/ApplicationEntity.kt
```diff
@@ -443,6 +443,13 @@ class ApprovedPremisesApplicationEntity(
 
   fun getLatestPlacementRequest(): PlacementRequestEntity? = this.placementRequests.maxByOrNull { it.createdAt }
   fun isSubmitted() = submittedAt != null
+  val hasArrivalWithin28Days: Boolean
+    get() = arrivalDate?.toLocalDate()?.let { arrival ->
+      val today = LocalDate.now()
+
+      !arrival.isBefore(today) &&
```
davidatkinsuk: I think we can drop the before check because it should never really happen, and if it does i would assume it should be treated as high priority (i.e. within 28 days)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/allocations/rules/AppealedAssessmentRule.kt
```diff
@@ -20,11 +18,9 @@ class AppealedAssessmentRule(
   override val priority: Int
     get() = 0
 
-  override fun evaluateAssessment(assessmentEntity: AssessmentEntity): UserAllocatorRuleOutcome {
-    val application = assessmentEntity.application
+  override fun evaluateAssessment(assessmentEntity: ApprovedPremisesAssessmentEntity): UserAllocatorRuleOutcome {
```
davidatkinsuk, line 21: nice

davidatkinsuk (approved):
great work! 👍

PR #4864, opened 2026-06-15
FM-723 Add departure notes

vbala-moj:
https://dsdmoj.atlassian.net/browse/FM-723

Included departure notes in the Placement Report and added the corresponding field to the PII list

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/reporting/Cas1PlacementReportRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/reporting/Cas1PlacementReportRepository.kt
@@ -32,6 +32,7 @@ class Cas1PlacementReportRepository(
                 THEN dr_parent.name || ' - ' || dr.name
                 ELSE dr.name
             END AS departure_reason,
+            csb.departure_notes AS departure_notes,
             moc.name AS departure_move_on_category,
             to_char(CAST(csb.non_arrival_confirmed_at as timestamp), 'YYYY-MM-DD"T"HH24:MI:SS"Z"') AS non_arrival_recorded_date_time,
             nar.name AS non_arrival_reason,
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/reporting/Cas1PlacementReportRepository.kt
```diff
@@ -32,6 +32,7 @@ class Cas1PlacementReportRepository(
                 THEN dr_parent.name || ' - ' || dr.name
                 ELSE dr.name
             END AS departure_reason,
+            left(csb.departure_notes, 250) AS departure_notes,
```
vbala-moj: Let me know if this is sufficient or if it needs to be increased.

davidatkinsuk: I guess not truncating this will push the column out? I think maybe we remove the limit for now and add truncation if the end user feeds back that it's too much. I suspect truncating may be clip off what the end user is looking for

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/integration/reporting/Cas1PlacementReportTest.kt
```diff
@@ -212,11 +213,11 @@ class Cas1PlacementReportTest : InitialiseDatabasePerClassTestBase() {
 
           assertThat(actual.size).isEqualTo(5)
 
-          bookingWithCanonicalDepartureDateWithinRange.assertRow(actual[0])
-          bookingWithNonArrivalConfirmedDateWithinRange.assertRow(actual[1])
-          bookingWithCanonicalArrivalDateWithinRange.assertRow(actual[2])
-          bookingWithCanonicalArrivalDateWithinRange.assertRow(actual[3])
-          bookingWithCancellationDateWithinRange.assertRow(actual[4])
+          bookingWithCanonicalDepartureDateWithinRange.assertRow(actual[0], isPii = false)
```
davidatkinsuk: could we maybe rename the new argument `includePii` across this PR? Good spot that we aren't actually asserting on the PII values currently

davidatkinsuk (approved):
LGTM , a few minor comments from me

PR #4870, opened 2026-06-16
FM741  Refactor CAS1 offender entity

vbala-moj:
Removed classes that are no longer needed.


davidatkinsuk (approved):
LGTM although i guess it needs removing from the migration job service too?

vbala-moj:
> LGTM although i guess it needs removing from the migration job service too?

Yes, I was a bit too quick to push the changes.🤦🏻‍♂️

PR #4879, opened 2026-06-17
Move all cas1 controllers and DTOs into `cas1` package

davidatkinsuk:
There are some CAS1-specific types not prefixed with `CAS1` that remain on the `api.models` package. These will be picked up in a subsequent PR

There are also some remaining common controllers that can now be moved to `common.controller`. This will be picked up in a subsequent PR


danhumphreys-moj (approved):
LGTM!  .. and Copilot!

PR #4882, opened 2026-06-18
Move common controllers and models into `common` packages

davidatkinsuk:
Move common controllers into `common.controllers`
Start the process of moving top level `model` entries into better namespaces packages


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/transformer/ApplicationsTransformersTest.kt
```diff
@@ -382,142 +379,6 @@ class ApplicationsTransformersTest {
     assertThat(result.status).isEqualTo(ApplicationStatus.submitted)
   }
 
-  @ParameterizedTest
```
daveawc, line 385: guessing these tests were not needed? just flagging as not related to the pr description 😅

daveawc (approved):
👍, one minor question

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/transformer/ApplicationsTransformersTest.kt
davidatkinsuk: oops. It was for the commit `Remove `ApplicationSummary` and subtypes`. I should have mentioned I removed some orphaned stuff too 👍 thanks for checking!

PR #4884, opened 2026-06-18
FM-742 Create cases entry for all CAS applications.

vbala-moj:
Create a case entry for all new CAS applications or update existing ones if they already exist.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/service/CaseService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/service/CaseService.kt
@@ -1,29 +1,36 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.common.service
 
 import org.springframework.stereotype.Service
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.ApDeliusContextApiClient
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.ClientResult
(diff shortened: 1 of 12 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/service/CaseService.kt
```diff
@@ -41,4 +47,16 @@ class CaseService(
...
+
+  private fun getRiskTier(crn: String): String? = when (val tierResponse = hmppsTierApiClient.getTier(crn)) {
+    is ClientResult.Success -> tierResponse.body.tierScore
+    is ClientResult.Failure -> null
+  }
+
+  private fun getCaseSummary(crn: String): CaseSummary = when (
+    val caseSummariesResponse = apDeliusContextApiClient.getCaseSummaries(listOf(crn))
+  ) {
+    is ClientResult.Success -> caseSummariesResponse.body.cases.first()
```
davidatkinsuk: if the CRN doesn't exist cases will be empty. I think that'll result in `IndexOutOfBoundsException` here. It may be cleaner if we do something like

```
is ClientResult.Success -> {
    if(caseSummariesResponse.body.cases.isEmpty() {
      error("Couldn't not find case for CRN $crn")
    } else {
       caseSummariesResponse.body.cases.first()
    }
}
```

Of course this should never really happen, but the error will be easier to understand this way

davidatkinsuk (approved):
One minor comment from me. Looks good! 👍

PR #4889, opened 2026-06-22
SAS-467: Fix cas1 and cas3 status.

rashadmughal:
Updates the CAS1 and CAS3 external referrals responses to report application-level status (and booking/placement status) rather than assessment status, and expands referral history to include multiple placement/booking history entries per assessment/application.

**Changes:**
- CAS1: `Cas1ReferralHistory` now includes `applicationStatus`, `requestForPlacementStatus`, and a typed `placementStatus`; controller/transformer updated to return multiple history rows via placement histories.
- CAS3: `Cas3ReferralHistory` now includes `applicationStatus` and a typed `bookingStatus`; controller/transformer updated to return multiple history rows via booking lists.
- Booking repository query changed to support returning CAS3 booking history entries.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/transformer/Cas3AssessmentTransformer.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/transformer/Cas3AssessmentTransformer.kt
@@ -84,30 +84,33 @@ class Cas3AssessmentTransformer(
     probationDeliveryUnitName = ase.probationDeliveryUnitName,
   )
 
-  fun transformAssessmentToCas3ReferralHistory(a: TemporaryAccommodationAssessmentEntity): Cas3ReferralHistory {
+  fun transformAssessmentToCas3ReferralHistory(a: TemporaryAccommodationAssessmentEntity): List<Cas3ReferralHistory> {
     val application = a.typedApplication<TemporaryAccommodationApplicationEntity>()
 
-    val latestBooking = bookingRepository.findLatestBookingEntity(application.id, ServiceName.temporaryAccommodation.value)
+    val bookings = bookingRepository.findAllCas3BookingEntity(application.id, ServiceName.temporaryAccommodation.value)
 
-    val placementAddress = latestBooking?.let {
-      listOfNotNull(it.premises.addressLine1, it.premises.town, it.premises.postcode).joinToString(", ")
-    }
-
-    return Cas3ReferralHistory(
+    val referralHistory = Cas3ReferralHistory(
       id = a.id,
       applicationId = application.id,
       createdAt = a.createdAt.toInstant(),
-      status = a.deriveAssessmentStatus(),
+      applicationStatus = application.getStatus(),
       type = ServiceType.CAS3,
       referralRejectionReason = a.referralRejectionReason?.name ?: a.rejectionRationale,
       referralRejectionReasonDetail = a.referralRejectionReasonDetail,
       localAuthorityArea = application.dutyToReferLocalAuthorityAreaName,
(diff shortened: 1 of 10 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas1/Cas1AssessmentTransformer.kt
```diff
@@ -33,7 +34,7 @@ class Cas1AssessmentTransformer(
   private val userTransformer: UserTransformer,
   private val personTransformer: PersonTransformer,
   private val risksTransformer: RisksTransformer,
-  private val cas1SpaceBookingRepository: Cas1SpaceBookingRepository,
+  private val cas1ExternalApplicationService: Cas1ExternalApplicationService,
```
davidatkinsuk: transformers shouldn't be using services directly, instead pass what you need into the transformer function

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas1/Cas1AssessmentTransformer.kt
```diff
@@ -77,30 +81,51 @@ class Cas1AssessmentTransformer(
...
+        applicationStatus = application.status,
+        type = ServiceType.CAS1,
+        referralRejectionReason = entity.rejectionRationale,
+        localAuthorityArea = application.apArea?.name,
+        pdu = application.cruManagementArea?.name,
+        referredBy = transformToStaffDto(application.createdByUser),
+        placementAddress = toPlacementAddress(it.premises),
+        placementStatus = it.placementStatus,
+        requestForPlacementStatus = it.requestForPlacementStatus,
+      )
```
davidatkinsuk: there's quite a bit of duplicate code here. I think it could be removed by creating a `Cas1ReferralHistory` like you do if `placementHistories.isEmpty()`, and then use `copy()` to create versions for each placement (if any exist)?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/transformer/Cas3AssessmentTransformer.kt
```diff
@@ -82,29 +82,27 @@ class Cas3AssessmentTransformer(
...
-  fun transformAssessmentToCas3ReferralHistory(a: TemporaryAccommodationAssessmentEntity): Cas3ReferralHistory {
+  fun transformAssessmentToCas3ReferralHistory(a: TemporaryAccommodationAssessmentEntity): List<Cas3ReferralHistory> {
     val application = a.typedApplication<TemporaryAccommodationApplicationEntity>()
 
-    val latestBooking = bookingRepository.findLatestCas3BookingEntity(application.id, ServiceName.temporaryAccommodation.value)
+    val bookings = bookingRepository.findAllCas3BookingEntity(application.id, ServiceName.temporaryAccommodation.value)
 
-    val placementAddress = latestBooking?.let {
-      listOfNotNull(it.premises.addressLine1, it.premises.town, it.premises.postcode).joinToString(", ")
+    return bookings.map {
```
davidatkinsuk, line 108: if there's no booking we return an empty list (I think). Is that deliberate?

davidatkinsuk (commented):
Some thoughts from me. Going forward can we please split each change into an individual commit (at the very least a commit for the change to the DTOs and a commit for changing to return lists of placements would be helpful to quickly spot what's happened in the diffs). Also please detail on the commit why you're making the change.

davidatkinsuk (approved):
LGTM 👍

PR #4890, opened 2026-06-22
FM-743 Backfill Cases

vbala-moj:
https://dsdmoj.atlassian.net/browse/FM-743

Added migration job to backfill crns missing from the case table.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/jobs/migration/BackfillCasesJob.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/jobs/migration/BackfillCasesJob.kt
@@ -1,0 +1,113 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.common.jobs.migration
+
+import org.slf4j.LoggerFactory
+import org.springframework.stereotype.Component
+import org.springframework.transaction.support.TransactionTemplate
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.ClientResult
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.HMPPSTierApiClient
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.common.dto.BackfillCaseSummaryMigrationDto
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.common.entity.CaseEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.common.entity.CaseRepository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.PersonSummaryInfoResult
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.LaoStrategy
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/jobs/migration/BackfillCasesJob.kt
```diff
@@ -0,0 +1,114 @@
...
+) : MigrationInBatchesJob(migrationLogger, transactionTemplate) {
+  private val log = LoggerFactory.getLogger(this::class.java)
+
+  override val shouldRunInTransaction = false
+
+  @SuppressWarnings("TooGenericExceptionCaught")
+  override fun process(pageSize: Int) {
+    val missingCaseDtos = caseRepository.findAllUniqueCrnsMissingFromCases()
+
+    migrationLogger.info("Found ${missingCaseDtos.size} unique CRNs to backfill.")
```
davidatkinsuk, line 33: i think this could potentially include multiple entries for the same CRNs (e.g. if used for CAS2 and CAS3). Could we do a quick filter here or in SQL to remove duplicates?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/jobs/migration/BackfillCasesJob.kt
```diff
@@ -0,0 +1,114 @@
...
+
+    processInBatches(missingCaseDtos, pageSize) { batch ->
+      val summariesByCrn = try {
+        offenderService.getPersonSummaryInfoResultsInBatches(
+          batch.map { it.crn }.toSet(),
+          LaoStrategy.NeverRestricted,
+        ).associateBy { it.crn }
+      } catch (exception: Exception) {
+        log.error(
+          "Unable to retrieve person summaries for batch, continuing with DTO values",
```
davidatkinsuk, line 43: nice. we should monitor how much this happens

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/jobs/migration/BackfillCasesJob.kt
```diff
@@ -0,0 +1,114 @@
...
+        summaryName =
+          "${summaryResult.summary.name.forename} ${summaryResult.summary.name.surname}"
+            .uppercase()
+        summaryNomsNumber = summaryResult.summary.nomsId
+      }
+
+      is PersonSummaryInfoResult.Success.Restricted -> {
+        summaryName = null
+        summaryNomsNumber = summaryResult.nomsNumber
+      }
```
davidatkinsuk, line 76: this should never happen given the use of `LaoStrategy.NeverRestricted`. we should throw an exception and stop processing in this case because something wrong has happened

davidatkinsuk (approved):
Looks great. A few thoughts from me. It'll be interesting to see what gets logged when trying in preprod/prod given how old some of these records we're pulling from are.

PR #4894, opened 2026-06-23
Add Inbox Functionality to handle domain events

davidatkinsuk:
This is currently disabled as the SQS listener is not yet configured, and there are no event handlers

This is a lightly modified version of the SAS inbox event handler


danhumphreys-moj (approved):
LGTM

PR #4896, opened 2026-06-24
FM-792 `/cas2/users/{username}` duplicate user fix

davidatkinsuk:
update `/cas2/users/{username}` to return the current user in a way that will not fail if there are multiple entries in the cas2 users table with the same username (but different types)

This is a quick fix to work around a production issue. It will be followed by a new endpoint that explicitly returns the dto for the current user and this endpoint will be removed.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/integration/Cas2v2UsersTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/integration/Cas2v2UsersTest.kt
@@ -26,13 +26,14 @@ class Cas2v2UsersTest : InitialiseDatabasePerClassTestBase() {
     }
 
     @Test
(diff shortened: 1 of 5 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/integration/Cas2v2UsersTest.kt
```diff
@@ -69,4 +70,33 @@ class Cas2v2UsersTest : InitialiseDatabasePerClassTestBase() {
       assertThat(response?.deliusUserInfo).isNull()
     }
   }
+
+  @Test
+  fun `Does not fail if there are multiple users with the same username`() {
```
danhumphreys-moj, line 75: nice!

danhumphreys-moj (approved):
LGTM!

PR #4899, opened 2026-06-24
FM-768 Update `cases.tier` on tier changed domain events

davidatkinsuk:
This PR introduces a domain event listener for the `tier.calculation.changed` domain event, updating the `cases.tier` value where we have a corresponding record for the CRN

Any failures retreiving an updatng the tier are propagated back to the domain event dispatcher as an exception, which will be alerted. Under normal operation this should never happen because:

a) the CRN should exist because we just received a domain event about it
b) we’re only updating CRNs we’re interested in, so there shouldn’t any database issues

The most likely cause of this error is failure to retrieve the tier from the API due to a temporary issue. In this case replay tooling will be required, which will be added by a subsequent commit (e.g. replay failed domain events)

Some additional tweaks were made to the inbox event dispatcher:

* Rename processing status `NOT_PROCESSED` to `IGNORED`. This provides clarity on what this status means.
* Remove event handler response status `FAILED`, favouring use of exceptions to report failures

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/domainevent/listener/HmppsDomainEventListener.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/domainevent/listener/HmppsDomainEventListener.kt
@@ -1,0 +1,63 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.common.domainevent.listener
+
+import io.awspring.cloud.sqs.annotation.SqsListener
+import org.slf4j.LoggerFactory
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
+import org.springframework.stereotype.Component
+import tools.jackson.databind.json.JsonMapper
+import tools.jackson.module.kotlin.readValue
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.common.jpa.InboxEventEntity
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.common.jpa.ProcessedStatus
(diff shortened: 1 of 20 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/domainevent/listener/HmppsDomainEventListener.kt
```diff
@@ -0,0 +1,63 @@
...
+          createdAt = Instant.now(),
+          processedStatus = ProcessedStatus.PENDING,
+          processedAt = null,
+          payload = snsMessage.message,
+        ),
+      )
+    } catch (e: Exception) {
+      log.error("Exception caught in HmppsDomainEventListener", e)
+      sentryService.captureException(e)
+      throw e
```
vbala-moj, line 60: Should we be concerned about duplicate deliveries here? Since SQS provides at-least-once delivery, is there a risk of persisting the same event multiple times if it's redelivered?

vbala-moj (approved):
LGTM, just one minor observation.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/domainevent/listener/HmppsDomainEventListener.kt
davidatkinsuk: That's a really good point that i hadn't considered. The documentation for the `InboxEventHandler` says handling should be idempotent, so this should be fine if it occurs. If we find it causes issues (e.g. happens a lot) we could consider persisting the message ID in the `inbox_events` table and ignoring duplicates

PR #4902, opened 2026-06-25
FM-758 Add CAS2 email service

vbala-moj:
https://dsdmoj.atlassian.net/browse/FM-758

- Add CAS2 email service to encapsulate email sending logic.
- Add feature flag to control sending of new CAS2 email notifications.
- Delegate email delivery through EmailNotificationService


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/service/Cas2EmailService.kt
```diff
@@ -0,0 +1,28 @@
...
+    templateId: String,
+    personalisation: Map<String, *>,
+    cas2Application: Cas2ApplicationEntity,
+  ) {
+    emailNotificationService.sendEmail(
+      recipientEmailAddress = recipientEmailAddress,
+      templateId = templateId,
+      personalisation = personalisation,
+      replyToEmailId = notifyConfig.emailAddresses.cas2ReplyToId,
+      reference = cas2Application.id.toString(),
```
vbala-moj, line 25: At the moment we're only using cas2Application.id to populate the Notify reference. Would it be simpler to pass the application ID (or reference) instead of the entire entity, unless we expect additional application-derived behaviour in this service?

davidatkinsuk: fair point. I think stick with the entire entity in case we want to start pulling additional stuff from it in the future (this follows the pattern used on CAS1)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/service/Cas2ApplicationEmailService.kt
```diff
@@ -0,0 +1,39 @@
...
+  private val featureFlagService: FeatureFlagService,
+  private val cas2EmailService: Cas2EmailService,
+  private val cas2UserService: Cas2UserService,
+  @Value("\${url-templates.frontend.cas2v2.submitted-application-overview}") private val submittedApplicationUrlTemplate: String,
+) {
+
+  fun applicationSubmitted(cas2Application: Cas2ApplicationEntity) {
+    if (!featureFlagService.getBooleanFlag("isr-email-changes-enabled")) return
+
+    val recipientEmailAddress = cas2UserService.getUserForRequest().email ?: return
```
davidatkinsuk: I know they'll be the same thing, but i think it may make the intention clearer if you used `cas2Application.createdByUser.email` here instead?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/service/Cas2ApplicationService.kt
```diff
@@ -317,6 +318,8 @@ class Cas2ApplicationService(
 
     sendEmailApplicationSubmitted(user, application)
```
davidatkinsuk: shall we rename this to 'sendAssessorEmailApplicationSubmitted'? (can do on the PR updating this specific email). Otherwise it reads a little confusing given the next line.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/service/Cas2ApplicationEmailService.kt
```diff
@@ -0,0 +1,39 @@
...
+    val submittedAt = cas2Application.submittedAt
+
+    cas2EmailService.sendEmail(
+      recipientEmailAddress = recipientEmailAddress,
+      templateId = Cas2NotifyTemplates.CAS2_APPLICATION_SUBMITTED,
+      personalisation = mapOf(
+        "cohort" to cas2Application.cohort?.name,
```
davidatkinsuk: Some of the enum names don't match the values in the ticket. Shall we add a new property to the enumeration to store the `emailLabel` (or think of a better property name :) )

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/service/Cas2ApplicationEmailService.kt
```diff
@@ -0,0 +1,39 @@
...
+        "crn" to cas2Application.crn,
+        "timeApplicationReceived" to submittedAt?.format(DateTimeFormatter.ofPattern("HH:mm")).orEmpty(),
+        "dateApplicationReceived" to submittedAt?.format(DateTimeFormatter.ofPattern("dd/MM/yyyy")).orEmpty(),
```
davidatkinsuk: I think instead of using `?` and `orEmpty` we should use `!!` to make expectations clear (submittedAt should never be null)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/service/Cas2ApplicationEmailService.kt
```diff
@@ -0,0 +1,39 @@
...
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.config.Cas2NotifyTemplates
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.FeatureFlagService
+import java.time.format.DateTimeFormatter
+
+@Service
+class Cas2ApplicationEmailService(
+  private val featureFlagService: FeatureFlagService,
+  private val cas2EmailService: Cas2EmailService,
+  private val cas2UserService: Cas2UserService,
+  @Value("\${url-templates.frontend.cas2v2.submitted-application-overview}") private val submittedApplicationUrlTemplate: String,
```
davidatkinsuk: if you inject this as a `UrlTemplate` instead of `String` you can then use the `resolve` function which removes the need to prefix the arg with `#`

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/integration/Cas2v2SubmissionTest.kt
```diff
@@ -898,6 +898,8 @@ class Cas2v2SubmissionTest : IntegrationTestBase() {
 
           Assertions.assertThat(responseStatuses.count { it.value() == 200 }).isEqualTo(1)
           Assertions.assertThat(responseStatuses.count { it.value() == 400 }).isEqualTo(9)
+
+          Assertions.assertThat(emailAsserter.assertEmailsRequestedCount(2))
```
davidatkinsuk, line 904: nice

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/unit/service/Cas2ApplicationEmailServiceTest.kt
```diff
@@ -0,0 +1,135 @@
...
+
+    every { mockFeatureFlagService.getBooleanFlag("isr-email-changes-enabled") } returns true
+    every {
+      mockCas2EmailService.sendEmail(
+        any(),
+        any(),
+        any(),
+        any(),
+      )
+    } returns Unit
```
davidatkinsuk: if you use a relaxed mockk, you don't need to setup this mock as I believe it'll return Unit by default

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/unit/service/Cas2ApplicationEmailServiceTest.kt
```diff
@@ -0,0 +1,135 @@
...
+          "nacroReferenceId" to applicationId.toString(),
+          "viewSubmittedApplicationUrl" to "http://frontend/applications/$applicationId",
+        ),
+        cas2Application = application,
+      )
+    }
+  }
+
+  @Test
+  fun `applicationSubmitted handles null cohort and submittedAt`() {
```
davidatkinsuk: as per previous comments i don't think we need this test

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/service/Cas2ApplicationEmailService.kt
davidatkinsuk: I think instead of using `?` we should use `!!` to make expectations clear (it should never be null)

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/unit/service/Cas2EmailServiceTest.kt
```diff
@@ -0,0 +1,62 @@
...
+    val personalisation = mapOf("key" to "value")
+    val user = Cas2UserEntityFactory().produce()
+    val application = Cas2ApplicationEntityFactory()
+      .withCreatedByUser(user)
+      .produce()
+
+    val cas2ReplyToId = "reply-to-id"
+
+    every { mockNotifyConfig.emailAddresses.cas2ReplyToId } returns cas2ReplyToId
+    every {
```
davidatkinsuk: see previous comment on relaxed mockks

davidatkinsuk (approved):
Very neat! Just a few minor comments from me

PR #4905, opened 2026-06-25
FM-759 CAS2 bail application assess email

vbala-moj:
Add new template for CAS2 bail application to assess email.

Refactor submit application to send emails based on feature flag.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/service/Cas2ApplicationEmailService.kt
```diff
@@ -3,37 +3,62 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.cas2.service
...
+
+  val Cas2Cohort.sla: String
+    get() = when (this) {
+      Cas2Cohort.COURT_BAIL,
+      Cas2Cohort.ATCR,
+      -> "1 hour"
+
+      Cas2Cohort.PRISON_BAIL -> "1 working day"
+
+      else -> "3 working days"
```
vbala-moj: If we expect 'sla' to be used across the codebase, it might be worth moving it onto Cas2Cohort itself as an enum property instead of keeping it as an extension property.

davidatkinsuk: I don't think we need it anywhere else, but i still think it may be a good idea regardless to make it quick and unambiguous to see what the SLA is. Would you mind doing that and calling it 'assessmentSla?'

davidatkinsuk (approved):
One comment from me, looks great!

PR #4910, opened 2026-06-26
FM-760 Add CAS2 status update email

vbala-moj:
Added a new email template for CAS2 application status updates.

Update the service to send emails using the new template.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/service/Cas2ApplicationStatusUpdateEmailService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/service/Cas2ApplicationStatusUpdateEmailService.kt
@@ -1,0 +1,46 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.cas2.service
+
+import org.springframework.beans.factory.annotation.Value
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas2/service/Cas2ApplicationStatusUpdateEmailService.kt
```diff
@@ -0,0 +1,46 @@
...
+
+@Service
+class Cas2ApplicationStatusUpdateEmailService(
+  private val cas2EmailService: Cas2EmailService,
+  @Value("\${url-templates.frontend.cas2v2.application-overview}") private val applicationOverviewUrlTemplate: UrlTemplate,
+) {
+
+  fun statusUpdate(cas2Application: Cas2ApplicationEntity, status: Cas2StatusUpdateEntity) {
+    val recipientEmailAddress = cas2Application.createdByUser.email ?: return
+    val submittedAt = cas2Application.submittedAt ?: return
```
davidatkinsuk: given status updates are made on assessment, i think this will always be set so should use `requireNotNull` instead?

davidatkinsuk (approved):
Looks good, one minor comment from me 👍

PR #4911, opened 2026-06-26
Harden github actions and update renovate

davidatkinsuk:
* Use latest template renovate config
* Enable renovate dashboard
* Harden use of actions/checkout


bobmeredith (approved):
LGTM

PR #4918, opened 2026-06-29
FM-821 Remove Legacy Cas3BookingRepository and Cas3BookingSearchService

muhammad-elabdulla:
This [PR FM-821](https://dsdmoj.atlassian.net/browse/FM-821) is to remove Cas3BookingRepository and Cas3BookingSearchService, as they are no longer used by the service. They have been replaced by Cas3v2BookingRepository and Cas3v2BookingSearchService, respectively.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/migration/Cas3UpdateBookingOffenderNameJob.kt
```diff
@@ -33,7 +31,7 @@ class Cas3UpdateBookingOffenderNameJob(
     try {
       while (hasNext) {
         migrationLogger.info("Getting page $page for max page size $pageSize")
-        slice = cas3BookingRepository.findAllTemporaryAccommodationBookings(BookingEntity::class.java, PageRequest.of(page - 1, pageSize))
+        slice = bookingRepository.findAllTemporaryAccommodationBookings(BookingEntity::class.java, PageRequest.of(page - 1, pageSize))
```
davidatkinsuk: this appears to now be using `BookingRepository`. Should it not be using `Cas3v2BookingRepository` and as such be using the `Cas3BookingEntity`?

muhammad-elabdulla: You're right. I was about to change it

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/jpa/entity/Cas3BookingEntity.kt
```diff
@@ -196,6 +197,19 @@ interface Cas3v2BookingRepository : JpaRepository<Cas3BookingEntity, UUID> {
     pageable: Pageable?,
   ): Page<Cas3v2BookingSearchResult>
 
+  @Query(
+    """
+      SELECT * FROM  bookings b 
+      WHERE b.service='temporary-accommodation'
```
davidatkinsuk, line 203: i think we can remove this constraint. I just checked prod using

```select count(*), service from bookings group by service;```

and all bookings have 'temporary-accommodation' as the service value

 (I guess `service` could be deprecated in a subsequent PR?)

File: src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/jpa/entity/Cas3BookingEntity.kt [hunk 3]
```diff
@@ -196,6 +197,19 @@ interface Cas3v2BookingRepository : JpaRepository<Cas3BookingEntity, UUID> {
...
+      ORDER BY b.crn
+      """,
+    nativeQuery = true,
+  )
+  fun <T : Cas3BookingEntity> findAllBookings(
+    type: Class<T>,
+    pageable: Pageable?,
+  ): Slice<Cas3BookingEntity>
```
davidatkinsuk, line 211: Do we need the generics and type value given it's always `Cas3BookingEntity`? I guess we could probably just use a function name and let spring data figure out the query, something like:

`fun findAllOrderByCrn(): Slice<Cas3BookingEntity>`

davidatkinsuk (approved):
A few comments to consider. Great to see chunks of code being removed 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/jpa/entity/Cas3BookingEntity.kt
muhammad-elabdulla: do we need to rename bookings table to cas3_bookings at the end of the refactoring?

davidatkinsuk: yes i think we should (along with linked tables that are now cas3 only)

PR #4922, opened 2026-06-29
FM-824 Remove CAS1 dependency on BookingRepository

muhammad-elabdulla:
This [PR FM-824](https://dsdmoj.atlassian.net/browse/FM-824) is to remove CAS1 dependency on BookingRepository


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1BookingDomainEventService.kt
```diff
@@ -173,16 +171,8 @@ class Cas1BookingDomainEventService(
...
-          null
-        } else {
-          bookingId
-        },
-        cas1SpaceBookingId = if (bookingChangedInfo.isSpaceBooking) {
-          bookingId
-        } else {
-          null
-        },
+        bookingId = null,
```
davidatkinsuk: This may be picked upon on a subsequent commit already but i believe we can remove `SaveCas1DomainEvent.bookingId` after this commit? Also I think we can remove `SaveCas1DomainEventWithPayload.bookingId`

davidatkinsuk (approved):
Looks good, just one potential improvement from me 👍

davidatkinsuk (approved):
Perfect, thank you 👍

PR #4924, opened 2026-06-30
SAS-615: Open up CAS3 referrer view referral access

david-mcdowall:
This PR allows users with the `CAS3_REFERRER` role to view a referral created by another user (previously they could only view a referral created by themselves) returned via the `GET /applications/{applicationId}` endpoint. 

**Background**
SAS is building deeplinks into CAS3 to allow probation practitioners to view a specific referral (either current or historic). After implementing this feature, we discovered the users are only able to view the referral if they have the `CAS3_REFERRER` role and the referral was created by themselves - otherwise they would see 'You do not have permission to complete this action' page in CAS3. 

After some discussion, it was agreed that we should open up the access to allow referrers to view any referral - including ones created by others. 

**This PR**

- Refactors CAS3 access checks into two specific functions - `userCanViewCas3Application` and `userCanViewCas3Assessment` (`userCanViewTemporaryAccommodationApplicationCreatedBySomeoneElse` removed)
- `userCanViewCas3Application` allows access to `CAS3_REFERRER` role (needed for SAS users) alongside `CAS3_ASSESSOR` role
- Region lock is kept as is (can only view referrals in your probation region)
- Additional test coverage / new function name refactors

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserAccessService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserAccessService.kt
@@ -103,7 +103,7 @@ class UserAccessService(
     user: UserEntity,
     application: TemporaryAccommodationApplicationEntity,
   ): Boolean = userCanAccessRegion(user, ServiceName.temporaryAccommodation, application.probationRegion.id) &&
-    user.hasRole(UserRole.CAS3_ASSESSOR) &&
+    user.hasAnyRole(UserRole.CAS3_ASSESSOR, UserRole.CAS3_REFERRER) &&
     application.submittedAt != null
 
   fun userCanReallocateTask(user: UserEntity): Boolean = when (requestContextService.getServiceForRequest()) {
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserAccessService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserAccessService.kt
@@ -118,7 +118,7 @@ class UserAccessService(
     else -> false
   }
 
-  fun userCanViewAssessment(user: UserEntity, assessment: AssessmentEntity): Boolean = assessment is ApprovedPremisesAssessmentEntity
+  fun userCanViewAssessment(assessment: AssessmentEntity): Boolean = assessment is ApprovedPremisesAssessmentEntity
 
   fun userCanViewCas3Assessment(user: UserEntity, assessment: TemporaryAccommodationAssessmentEntity): Boolean = user.hasRole(UserRole.CAS3_ASSESSOR) &&
     userCanAccessRegion(user, ServiceName.temporaryAccommodation, (assessment.application as TemporaryAccommodationApplicationEntity).probationRegion.id)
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1AssessmentService.kt
```diff
@@ -151,7 +151,7 @@ class Cas1AssessmentService(
     val assessment = approvedPremisesAssessmentRepository.findByIdOrNull(assessmentId)
       ?: return CasResult.NotFound("AssessmentEntity", assessmentId.toString())
 
-    if (!userAccessService.userCanViewAssessment(user, assessment)) {
+    if (!userAccessService.userCanViewAssessment(assessment)) {
```
david-mcdowall, line 154: this was only being used for CAS3 user access check - which is now in its own `userCanViewCas3Assessment` method (`user` param is now dead code and removed)

davidatkinsuk: good spot

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/UserAccessService.kt
```diff
@@ -116,16 +118,10 @@ class UserAccessService(
...
-
-    is TemporaryAccommodationAssessmentEntity ->
-      user.hasRole(UserRole.CAS3_ASSESSOR) &&
-        userCanAccessRegion(user, ServiceName.temporaryAccommodation, (assessment.application as TemporaryAccommodationApplicationEntity).probationRegion.id)
+  fun userCanViewAssessment(assessment: AssessmentEntity): Boolean = assessment is ApprovedPremisesAssessmentEntity
 
-    else -> false
-  }
+  fun userCanViewCas3Assessment(user: UserEntity, assessment: TemporaryAccommodationAssessmentEntity): Boolean = user.hasRole(UserRole.CAS3_ASSESSOR) &&
+    userCanAccessRegion(user, ServiceName.temporaryAccommodation, (assessment.application as TemporaryAccommodationApplicationEntity).probationRegion.id)
```
muhammad-elabdulla, line 124: If you add:
`&& assessment.application.submittedAt != null`

then you don't need to use `userCanViewCas3Application` in `getAssessmentAndValidate`.

In all cases, an assessment is only created when an application is submitted, so submittedAt should never be null.

david-mcdowall: @muhammad-elabdulla I don't think we can remove `userCanViewCas3Application` from `getAssessmentAndValidate` because `getAssessmentAndValidate` is also used by the `Cas3TimelineController` which is called by the UI on:
- `GET /review-and-assess/:id/summary`
- `GET /referrals/:id/full`

If we remove it then any referrer accessing `GET /referrals/:id/full` hits the `getAssessmentAndValidate` method (via the timeline call) and it returns 403 unauthorised - the user will see 'You do not have permission to complete this action' (even for their own referral)

muhammad-elabdulla (approved):
LGTM

PR #4926, opened 2026-06-30
FM-825 Change CAS3 Domain Events to use new CAS3 entities

muhammad-elabdulla:
This [PR FM-825](https://dsdmoj.atlassian.net/browse/FM-825) is to change CAS3 Domain Events to use new CAS3 entities


davidatkinsuk (approved):
LGTM 👍

PR #4931, opened 2026-07-01
FM-820: Remove Tier service call

vbala-moj:
- Removed the Tier service call when creating the application-submitted domain event.
- Added the `toMappa()` extension function to `CaseDetail` to map MAPPA details.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1ApplicationDomainEventService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1ApplicationDomainEventService.kt
@@ -67,10 +67,6 @@ class Cas1ApplicationDomainEventService(
           )
       }
 
-    val risks = offenderRisksService.getPersonRisks(application.crn)
-
-    val mappaLevel = risks.mappa.value?.level
-
     val staffDetails = when (val staffDetailsResult = apDeliusContextApiClient.getStaffDetail(username)) {
       is ClientResult.Success -> staffDetailsResult.body
       is ClientResult.Failure -> staffDetailsResult.throwException()
@@ -80,6 +76,8 @@ class Cas1ApplicationDomainEventService(
       is ClientResult.Success -> caseDetailResult.body
       is ClientResult.Failure -> caseDetailResult.throwException()
     }
+
+    val mappaLevel = offenderRisksService.run { caseDetail.toMappa() }.value ?.level
 
     domainEventService.saveApplicationSubmittedDomainEvent(
       SaveCas1DomainEvent(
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/client/deliuscontext/CaseDetail.kt
```diff
@@ -14,6 +17,21 @@ data class CaseDetail(
...
+  }
+
+  return RiskWithStatus(
+    value = Mappa(
+      "CAT ${detail.categoryDescription}/LEVEL ${detail.levelDescription}",
+      detail.lastUpdated.toLocalDate(),
+    ),
+  )
+}
+
```
vbala-moj: Could we have toMappa() return just Mappa instead?
```
fun CaseDetail.toMappa(): Mappa? =
    mappaDetail?.let { detail ->
        val category = detail.categoryDescription ?: return null
        val level = detail.levelDescription ?: return null

        Mappa(
            "CAT $category/LEVEL $level",
            detail.lastUpdated.toLocalDate(),
        )
    }
```

davidatkinsuk: i would just delegate to the existing `OffenderRiskService.toMappa`() and remove this duplicated code.

davidatkinsuk: (this reuse of an 'internal' function isn't ideal but the risks implementation needs a rework at some point so i think it's acceptable)

davidatkinsuk (approved):
Looks great, one comment from me on removing duplicated code

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/client/deliuscontext/CaseDetail.kt
danhumphreys-moj: I see @davidatkinsuk commented above but I had started reviewing so thought I would comment anyway.  I was wondering whether nullable Mappa is in effect a property of data class `CaseDetail` rather than something returned from a extension function?

```
data class CaseDetail(
  val case: CaseSummary,
  val offences: List<Offence>,
  val registrations: List<Registration>,
  val mappaDetail: MappaDetail?,
  val sentences: List<Sentence>,
  val personalContacts: List<PersonalContact> = emptyList(),
  val mappa: Mappa? = mappaDetail?.let { Mappa(
    "CAT ${it.categoryDescription}/LEVEL ${it.levelDescription}",
    it.lastUpdated.toLocalDate()
  ) }
)
```

danhumphreys-moj: I see `categoryDescription` and `levelDescription` can be nullable so checks on these would be required, so above is not quite right but nevertheless.

davidatkinsuk: > I was wondering whether nullable Mappa is in effect a property of data class CaseDetail rather than something returned from a extension function?

That would work but i think we did that for all extension functions on CaseDetail it could get a bit bloated

vbala-moj: I’m trying to use the existing toMappa member extension without changing it:

`val mappaLevel = offenderRisksService.run { caseDetail.toMappa() }.value?.level
`
However, I’m struggling to mock it in unit tests because it isn’t a standard method call.

Would you recommend refactoring it into a regular function on OffenderRiskService, or moving it to a top-level function instead, so it’s easier to call and mock? Or is there a preferred alternative approach we should follow?

davidatkinsuk: I think either

1. don't bother mocking and call the real code in the test
2. convert into a regular function to allow you to mock it

just go with path of least resistance given how messy this risk stuff is :)

PR #4939, opened 2026-07-02
FM-838 Remove Legacy CAS3 reports

muhammad-elabdulla:
This [PR FM-838](https://dsdmoj.atlassian.net/browse/FM-838) is to:

- Remove the legacy CAS3 reporting code.
- Remove the `cas3-reports-with-new-bedspace-model-tables-enabled` feature flag, as it is now enabled in all environments.


rashadmughal (approved):
Approved - but please address the copilot comments.

PR #4941, opened 2026-07-02
Add UI url to Cas1ReferralHistory.kt

daveawc:
This PR adds a link for a CAS1 application to the referral history response.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/dto/Cas1ReferralHistory.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/dto/Cas1ReferralHistory.kt
@@ -16,5 +16,5 @@ data class Cas1ReferralHistory(
   val referredBy: Cas1StaffDto,
   val placementAddress: String?,
   val placementStatus: String?,
-  val applicationUrl: String,
+  val uiUrl: String,
 )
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/dto/Cas1ReferralHistory.kt
```diff
@@ -16,4 +16,5 @@ data class Cas1ReferralHistory(
   val referredBy: Cas1StaffDto,
   val placementAddress: String?,
   val placementStatus: String?,
+  val uiUrl: String,
```
gregkhawkins, line 19: why not just url?

gregkhawkins (approved):
Approved with a question

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/dto/Cas1ReferralHistory.kt
daveawc: I asked 2 AIs for some naming suggestions and this was top on both, to make it clear it's user facing and to distinguish it from an api url 😅

gregkhawkins: haha, that's good enough for me!

PR #4943, opened 2026-07-03
Remove CAS1 Assessments API Generated code

davidatkinsuk:
As part of this change `crnOrName` was removed from the `getAssessmentsForUser` API because it is not used


daveawc (approved):
👍

PR #4948, opened 2026-07-06
FM-819 Add tierV2 and tierV3

vbala-moj:
https://dsdmoj.atlassian.net/browse/FM-819

 Removed the `tier` column and added `tierV2` and `tierV3` columns.
 Removed `tier` from the case entity and added `tierV2`.
 Refactored the code to remove `tier` references and populate `tierV2` data.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/entity/CaseEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/entity/CaseEntity.kt
@@ -1,14 +1,18 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.common.entity
 
+import jakarta.persistence.Column
 import jakarta.persistence.Entity
 import jakarta.persistence.Id
 import jakarta.persistence.PreUpdate
 import jakarta.persistence.Table
 import jakarta.persistence.Version
+import org.hibernate.annotations.JdbcTypeCode
+import org.hibernate.type.SqlTypes
 import org.springframework.data.jpa.repository.JpaRepository
 import org.springframework.data.jpa.repository.Query
 import org.springframework.stereotype.Repository
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.common.dto.BackfillCaseSummaryMigrationDto
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.common.entity.model.Tier
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.asHibernateProxy
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.getHibernateClass
 import java.time.OffsetDateTime
@@ -71,9 +75,11 @@ data class CaseEntity(
    * will consider any LAO restrictions
    */
   var name: String,
-  var tier: String?,
   val createdAt: OffsetDateTime,
   var lastUpdatedAt: OffsetDateTime,
+  @Column(name = "tier_v2", columnDefinition = "jsonb")
+  @JdbcTypeCode(SqlTypes.JSON)
+  var tierV2: Tier?,
   @Version
   var version: Long = 1,
 ) {
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/service/CaseService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/service/CaseService.kt
@@ -2,6 +2,8 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.common.service
 
 import org.slf4j.LoggerFactory
 import org.springframework.stereotype.Service
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas1.dto.TierDto
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas1.dto.TierVersionDto
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.ApDeliusContextApiClient
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.ClientResult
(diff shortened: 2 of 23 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/entity/CaseEntity.kt
```diff
@@ -71,9 +75,11 @@ data class CaseEntity(
    * will consider any LAO restrictions
    */
   var name: String,
-  var tier: String?,
   val createdAt: OffsetDateTime,
   var lastUpdatedAt: OffsetDateTime,
+  @Column(name = "tier_v2", columnDefinition = "jsonb")
+  @JdbcTypeCode(SqlTypes.JSON)
```
davidatkinsuk, line 81: good to see we're not using hypersistence for this anymore 👍 (relates to https://dsdmoj.atlassian.net/browse/FM-432)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/service/CaseService.kt
```diff
@@ -24,19 +29,39 @@ class CaseService(
...
+    val fullName =
+      "${caseSummary.name.forename} ${caseSummary.name.surname}"
+        .uppercase()
+        .trim()
+
+    fun toTier() = riskTier?.let {
+      Tier(
+        tierScore = it.tierScore,
+        calculationId = it.calculationId ?: UUID.randomUUID(),
+        calculationDate = it.calculationDate ?: LocalDateTime.now(),
```
davidatkinsuk: I think these defaults are redundant as the API response values should be non null? regardless, we shouldn't be guessing at these values if they don't exist

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/service/CaseService.kt
```diff
@@ -45,33 +70,43 @@ class CaseService(
...
       name = it.name,
-      tier = it.tier,
       createdAt = it.createdAt,
       lastUpdatedAt = it.lastUpdatedAt,
+      tier = it.tierV2?.let {
+        TierDto(
+          tierScore = it.tierScore,
+          calculationDate = it.calculationDate,
+          provisional = it.provisional,
+          version = TierVersionDto.valueOf(it.version.name),
```
davidatkinsuk, line 70: can we convert this to a `when` statement please so we're forced to consider any new types added (Although that is unlikely!)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/service/CaseService.kt
```diff
@@ -45,33 +70,43 @@ class CaseService(
...
 
-    case.tier = tier
+    case.tierV2 = Tier(
+      tierScore = tier.tierScore,
+      calculationId = tier.calculationId,
+      calculationDate = tier.calculationDate,
+      changeReason = tier.changeReason,
+      provisional = null,
+      version = TierVersion.V2,
+    )
```
davidatkinsuk: I think converting the upstream Tier response into our JPA/JSON model is duplicated above in an internal function `toTier`. Could this be made a private extension function and shared to remove duplication?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/service/CaseService.kt
```diff
@@ -45,33 +70,43 @@ class CaseService(
...
     caseRepository.save(case)
 
     log.info("Have updated tier for $crn to $tier")
     return true
   }
 
-  private fun getRiskTierOrNull(crn: String): String? = when (val tierResponse = hmppsTierApiClient.getTier(crn)) {
-    is ClientResult.Success -> tierResponse.body.tierScore
+  private fun getRiskTierOrNull(crn: String): uk.gov.justice.digital.hmpps.approvedpremisesapi.client.hmppstier.Tier? = when (val tierResponse = hmppsTierApiClient.getTier(crn)) {
```
davidatkinsuk, line 101: These import clashes are annoying. It would be nice if we had a standard naming convention for our upstream data types (not to fix now, but worth considering for the future).

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/integration/TierCalculationChangedTest.kt
```diff
@@ -45,15 +46,25 @@ class TierCalculationChangedTest : IntegrationTestBase() {
   fun `case exists, update tier, inbox event is PROCESSED`() {
     caseEntityFactory.produceAndPersist {
       withCrn(CRN)
-      withTier(OLD_TIER)
+      withTierV2(
+        Tier(
```
davidatkinsuk: you could use the new `TierFactory` for this? (and subsequent tier construction)

davidatkinsuk (approved):
Looking great, a few minor comments from me

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/service/CaseService.kt
vbala-moj: Since we’re mapping tierV2 and tierV3 explicitly I don’t think a when would add extra protection against future changes — happy to be corrected if I’m missing something.

davidatkinsuk: If we added a new version to `TierVersion` but not the `TierVersionDto` this code would fail, but we'd only find out about it at runtime (unless we add tests). This is, of course, highly unlikely so i'm not too bothered about changing it

PR #4951, opened 2026-07-06
FM-853 Stop returning change requests on space bookings and placement requests

davidatkinsuk:
Change request functionality was developed but never used

This PR is motivated by upcoming changes to `PersonSummary` to include tier information. This change has allowed us to remove `PersonTransformer.transformPersonInfoResultToPersonSummary` which would have require updates to add tier, despite it never being actively used in the corresponding API response


daveawc (approved):
👍

PR #4957, opened 2026-07-06
Add uiUrl property to suitable applications

daveawc:
This PR adds a uiUrl property to the cas1 and cas3 suitable applications fields, so that SAS can give deep links to the user.


david-mcdowall (approved):
LGTM

PR #4958, opened 2026-07-06
FM-847 Rollback adding version to PersonRisks RiskTier

davidatkinsuk:
created this if required, don't merge


davidatkinsuk:
this wasn't required

PR #4972, opened 2026-07-08
FM-852 Add live tier to person and person summary

davidatkinsuk:
This PR adds tier to the following API Types:

* `FullPerson`
* `RestrictedPerson`
* `FullPersonSummary`
* `RestrictedPersonSummary`

The tier is populated from the corresponding value in `cases.tier_v2`. If no entry exists in cases, or no tier is set, the tier will be set to null

This PR does not support providing tier for CAS2. There are currently no plans to support tier in CAS2. A lower priority ticket will be created to address this in the backend for consistency in what the API returns.


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/OffenderServiceTest.kt
```diff
@@ -1023,9 +1214,13 @@ class OffenderServiceTest {
 
       assertThat(results).hasSize(750)
       (0..749).forEach {
+        println(it)
```
muhammad-elabdulla: Do you need this line of code or was it only for local testing?

muhammad-elabdulla (approved):
LGTM

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/OffenderServiceTest.kt
davidatkinsuk: Good catch! i'll remove it

PR #4974, opened 2026-07-08
Add filtering and sorting for live tier v2 to CRU Dashboard (`GET /placement-requests`)

davidatkinsuk:
This tier adds API parameters to support filtering and sorting on a person's live tier for the `GET /placement-requests` API.

The parameter naming is unfortunately similar to existing parameters. These will be deprecated and removed once we've moved to live tiers. To avoid confusion i've added documentation to these parameters.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/api/model/PlacementRequestSortField.kt
```diff
@@ -12,14 +13,19 @@ enum class PlacementRequestSortField(@get:JsonValue val value: String) {
   applicationSubmittedAt("application_date"),
   requestType("request_type"),
   personName("person_name"),
+
+  @Schema(description = "Sort on the tier captured when the application was created")
```
muhammad-elabdulla: This can be @Parameter instead of @Schema?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/Cas1PlacementRequestTest.kt
```diff
@@ -614,6 +615,36 @@ class Cas1PlacementRequestTest : IntegrationTestBase() {
...
+            listOf(
+              transformNotMatchedPlacementRequestJpaToApiSummary(
+                placementRequestB6,
+                PersonInfoResult.Success.Full(offender2Details.otherIds.crn, offender2Details, inmate2Details),
+              ),
+            ),
+          ),
+        )
+    }
+
```
muhammad-elabdulla, line 662: Do we need to consider adding a test for when the personTier does not match it will return an empty list?

muhammad-elabdulla (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/api/model/PlacementRequestSortField.kt
davidatkinsuk: Good call. This wasn't actually appearing in the swagger anywhere. I've made some changes so it appears in the swagger both this type and on the parameter itself

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/Cas1PlacementRequestTest.kt
davidatkinsuk: I think the test partially covers this because there are placement requests with a different tier that aren't returned. i've a general purpose 'no results' test

PR #4975, opened 2026-07-08
FM-872 Remove Legacy DateChangeEntity

muhammad-elabdulla:
This [PR FM-872](https://dsdmoj.atlassian.net/browse/FM-872) is to remove Legacy DateChangeEntity and delete `date_changes` table


src/main/resources/db/migration/all/20260708164523__drop_date_changes_table.sql
```diff
@@ -0,0 +1,2 @@
+DROP TABLE IF EXISTS date_changes;
```
davidatkinsuk, line 1: can you please rename to `archive_data_changes` instead of dropping. please use this approach going forward

davidatkinsuk (approved):
Just one comment from me. It's great to see all this code being removed 👍

src/main/resources/db/migration/all/20260708164523__drop_date_changes_table.sql
muhammad-elabdulla: Sure I dropped the table as it didn't contain any data

davidatkinsuk: nevertheless, i don't want tables being dropped directly like this as it's a very difficult state to revert from if mistakes are made

PR #4980, opened 2026-07-09
FM-819 BackfillCasesJob for cases table

vbala-moj:
- Implementation of one-off migration job
- Fetches unique CRNs from various application tables
- Backfills missing names, NOMS numbers, and Tiers


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/jobs/migration/BackfillCasesJob.kt
```diff
@@ -25,102 +25,142 @@ class BackfillCasesJob(
...
+  private fun createCase(
+    dto: BackfillCaseSummaryMigrationDto,
+    summaryResult: PersonSummaryInfoResult?,
+  ) {
+    val personDetails = resolvePersonDetails(dto, summaryResult)
+
+    if (personDetails.name == null) {
+      migrationLogger.error("Could not determine name for CRN ${dto.crn}. Skipping.")
       return
     }
```
vbala-moj: Do we still need to do this?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/entity/CaseEntity.kt
```diff
@@ -24,41 +24,67 @@ interface CaseRepository : JpaRepository<CaseEntity, UUID> {
...
+                a.crn,
+                COALESCE(ta.name, ap.name) AS name,
+                a.noms_number,
+                a.created_at
+            FROM applications a
+            LEFT JOIN temporary_accommodation_applications ta
+                ON ta.id = a.id
+            LEFT JOIN approved_premises_applications ap
+                ON ap.id = a.id
+            ),
```
davidatkinsuk, line 56: don't worry about changing it but i suspect these joins are unnecessary as only CAS1/CAS3 use this tbale

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/jobs/migration/BackfillCasesJob.kt
davidatkinsuk: as agreed let's make `cases.name` nullable and set to null if we can't resolve the name. We'll review the impact of this once backfill has ran

davidatkinsuk (approved):
This is great, one comment from me in response to your comment.

Could you also please enable FEATURE-FLAGS_INCLUDE-TIER-V3 in all environments as the API for this is now available in Prod

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/entity/CaseEntity.kt
vbala-moj: OK. Initially, I didn't have these joins, but that means I have to make the `name` field `NULL` since we only store that field in the child tables.
But since we're making the `name` field nullable, these joins are no longer needed.

davidatkinsuk: Yes sorry that makes perfect sense. Ignore me!

PR #4981, opened 2026-07-09
FM-839

danhumphreys-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/dto/Cas1NewApplication.kt
```diff
@@ -0,0 +1,22 @@
+package uk.gov.justice.digital.hmpps.approvedpremisesapi.cas1.dto
+
+import com.fasterxml.jackson.annotation.JsonProperty
+import io.swagger.v3.oas.annotations.media.Schema
+
+data class Cas1NewApplication(
+
+    @get:JsonProperty("crn", required = true)
```
davidatkinsuk, line 8: I don't think we need any of these jsonProperty annotations

PR #4990, opened 2026-07-10
FM-875 Publish inbox metrics if 0 results

davidatkinsuk:
Before this commit if there were no entries in the inbox with a given status a metric wasn’t published for that status


muhammad-elabdulla (approved):
LGTM

PR #4999, opened 2026-07-14
FM-787 Reduce transaction scope on backfill cases

davidatkinsuk:
* move batching logic into the `BackfillCasesJob`, instead of inheriting
* this allows us to move transaction scope from batch to individual cases
* reduce delay between each batch to 500L
* remove flushes on save because transaction will commit after save

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/jobs/migration/BackfillCasesJob.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/jobs/migration/BackfillCasesJob.kt
@@ -23,8 +23,8 @@ class BackfillCasesJob(
   private val offenderService: OffenderService,
   private val hmppsTierApiClient: HMPPSTierApiClient,
   private val migrationLogger: MigrationLogger,
-  transactionTemplate: TransactionTemplate,
-) : MigrationInBatchesJob(migrationLogger, transactionTemplate) {
+  private val transactionTemplate: TransactionTemplate,
+) : MigrationJob() {
 
   override val shouldRunInTransaction = false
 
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/jobs/migration/BackfillCasesJob.kt
```diff
@@ -93,7 +111,7 @@ class BackfillCasesJob(
       lastUpdatedAt = OffsetDateTime.now()
     }
 
-    caseRepository.saveAndFlush(existingCase)
+    caseRepository.save(existingCase)
```
vbala-moj, line 114: On second thought, this entity is managed by the `findBy()` call and will be flushed with the updated changes, so calling `save()` again is unnecessary. What do you think? There’s no harm in keeping it there, though.

davidatkinsuk: i've _always_ struggled with this implicit behaviour - i think it's nice to put the [unnecessary] save in, even if just to show intention

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/jobs/migration/BackfillCasesJob.kt
```diff
@@ -33,21 +34,38 @@ class BackfillCasesJob(
...
+    processBatch: (List<T>) -> Unit,
+  ) {
+    items.chunked(batchSize).forEachIndexed { index, batch ->
+      migrationLogger.info("Processing batch ${index + 1} of ${items.size / batchSize}...")
+      processBatch(batch)
+
+      if (index < items.chunked(batchSize).lastIndex) {
+        Thread.sleep(500L)
+      }
+    }
```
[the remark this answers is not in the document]
davidatkinsuk: done

vbala-moj (approved):

PR #5006, opened 2026-07-15
FM-808-removed-legacy-endpoint-POST-booking-not-made

sajifdeva-moj:
# Context
Main - https://dsdmoj.atlassian.net/browse/FM-676
Subticket - https://dsdmoj.atlassian.net/browse/FM-808

# Changes in this PR
Removes POST /placement-requests/{id}/booking-not-made from the API and the now unused NewBookingNotMade type


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/Cas1SimpleApiClient.kt
```diff
@@ -319,7 +319,7 @@ class Cas1SimpleApiClient {
     integrationTestBase: IntegrationTestBase,
     placementRequestId: UUID,
     cruMemberJwt: String,
-    body: NewBookingNotMade,
+    body: Cas1NewBookingNotMade,
   ) {
     integrationTestBase.webTestClient.post()
       .uri("/placement-requests/$placementRequestId/booking-not-made")
```
danhumphreys-moj: This PR removes endpoint `/placement-requests/$placementRequestId/booking-not-made` so why aren't integration tests for `Cas1PlacementMatchingOutcomesV2ReportTest.kt` failing in the pipeline? 🤔

They are for this branch locally (with a 404 Not Found exception) which is what I would expect to happen.  Changing the uri to be `/cas1/placement-requests/$placementRequestId/booking-not-made` then means tests pass locally.

sajifdeva-moj: Good catch. I thought I'd updated the URI as part of this work, but clearly I missed that one.

Not too sure why the integration tests passed in the pipeline though, as I'd have expected them to fail if the old endpoint is no longer available.

danhumphreys-moj: Issue with the pipeline not running all integration tests identified and fixed in PR https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/5025.

danhumphreys-moj (approved):
Great stuff!

PR #5012, opened 2026-07-17
FM-907 Use common handle AssessmentInfo Response

muhammad-elabdulla:
This [PR FM-907](https://dsdmoj.atlassian.net/browse/FM-907) is to use common handle AssessmentInfo Response


vbala-moj (approved):
Nice

PR #5019, opened 2026-07-20
FM-917 Remove legacy properties and update types in Cas3VoidBedspace

muhammad-elabdulla:
This [PR FM-917](https://dsdmoj.atlassian.net/browse/FM-917) is to:

Update the Cas3VoidBedspaceEntity model to:

- Remove the legacy bed property.
- Change the premises property type to Cas3PremisesEntity

Also:

- Delete Cas3VoidBedspaceCancellationEntity, as void cancellations are now handled by the cancellationDate and cancellationNotes properties in Cas3VoidBedspaceEntity.
- Delete the Cas3VoidBedspaceJob migration job, as it is no longer needed and was only used as part of the migration to the new tables.


rashadmughal (approved):
LGTM

PR #5023, opened 2026-07-21
FM-921 Remove TemporaryAccommodationPremisesEntity

muhammad-elabdulla:
This [PR FM-921](https://dsdmoj.atlassian.net/browse/FM-921) is to remove TemporaryAccommodationPremisesEntity from the code base as it was replaced with Cas3PremisesEntity


src/main/resources/db/migration/all/20260721110712000__rename_table_to_archive_temporary_accommodation_premises.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE IF EXISTS temporary_accommodation_premises RENAME TO archived_temporary_accommodation_premises;
```
davidatkinsuk: can we please remove the `IF EXISTS` from this query, it should always exist. This gives us some protection if old migrations are accidentally re-ran (they'll fail fast if things don't look correct)

davidatkinsuk (approved):
LGTM 👍

PR #5029, opened 2026-07-22
FM-890 Add tier-based duration calculation logic for CAS1 placement requests…

danhumphreys-moj:
Ref: https://dsdmoj.atlassian.net/browse/FM-890

Add tier-based duration calculation logic for CAS1 placement requests, including V2 and V3 tier versions. Update service, controller, tests, and mock setups accordingly.

**Changes:**
- Update `Cas1RequestForPlacementService.defaultDurations` to calculate durations using live tier version (V2/V3), with expanded V3 rule set.
- Update controller + integration/unit tests to call the new duration calculation entrypoint and cover additional scenarios.
- Extend test givens/factories to populate `sentenceType` for applications to support the new rules.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/controller/Cas1RequestsForPlacementController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/controller/Cas1RequestsForPlacementController.kt
@@ -29,7 +29,7 @@ class Cas1RequestsForPlacementController(
     @RequestParam("apType") apType: ApType,
     @RequestParam("sentenceType") sentenceType: SentenceTypeOption,
   ): ResponseEntity<Cas1RequestsForPlacementDurationsCalculationResponseDto> {
-    val result = cas1RequestForPlacementService.defaultDurations(apType)
+    val result = cas1RequestForPlacementService.defaultDurations(applicationId, apType)
 
     return ResponseEntity.ok(extractEntityFromCasResult(result))
   }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/controller/Cas1RequestsForPlacementController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/controller/Cas1RequestsForPlacementController.kt
@@ -29,7 +29,7 @@ class Cas1RequestsForPlacementController(
     @RequestParam("apType") apType: ApType,
     @RequestParam("sentenceType") sentenceType: SentenceTypeOption,
   ): ResponseEntity<Cas1RequestsForPlacementDurationsCalculationResponseDto> {
-    val result = cas1RequestForPlacementService.defaultDurations(applicationId, apType)
+    val result = cas1RequestForPlacementService.defaultDurations(applicationId, apType, sentenceType.value)
 
     return ResponseEntity.ok(extractEntityFromCasResult(result))
   }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1RequestForPlacementService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1RequestForPlacementService.kt
@@ -3,14 +3,16 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas1
 import org.springframework.stereotype.Component
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.ApType
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.RequestForPlacement
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.SentenceTypeOption
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas1.dto.Cas1RequestsForPlacementDurationsCalculationResponseDto
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas1.dto.TierVersionDto
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.common.results.CasResult
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.common.service.CaseService
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApprovedPremisesApplicationEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.Cas1SpaceBookingRepository
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.PlacementApplicationEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.PlacementRequestEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.UserEntity
-import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.ApplicationService
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.RequestForPlacementTransformer
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.cas1.Cas1SpaceBookingTransformer
 import java.time.Period
@@ -18,13 +20,14 @@ import java.util.UUID
 
 @Component
 class Cas1RequestForPlacementService(
-  private val applicationService: ApplicationService,
+  private val applicationService: Cas1ApplicationService,
   private val cas1PlacementApplicationService: Cas1PlacementApplicationService,
   private val placementRequestService: Cas1PlacementRequestService,
   private val requestForPlacementTransformer: RequestForPlacementTransformer,
   private val cas1WithdrawableService: Cas1WithdrawableService,
   private val cas1SpaceBookingRepository: Cas1SpaceBookingRepository,
   private val cas1SpaceBookingTransformer: Cas1SpaceBookingTransformer,
+  private val caseService: CaseService,
 ) {
   fun getRequestsForPlacementByApplication(applicationId: UUID, requestingUser: UserEntity?): CasResult<List<RequestForPlacement>> {
     val application = applicationService.getApplication(applicationId)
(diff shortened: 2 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/controller/Cas1RequestsForPlacementController.kt
```diff
@@ -29,7 +29,7 @@ class Cas1RequestsForPlacementController(
     @RequestParam("apType") apType: ApType,
     @RequestParam("sentenceType") sentenceType: SentenceTypeOption,
   ): ResponseEntity<Cas1RequestsForPlacementDurationsCalculationResponseDto> {
```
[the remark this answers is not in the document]
danhumphreys-moj: Will leave for now since we are reviewing this and due to make some changes to this logic.  Sentence type was originally required to be a query parameter but can be retrieved from the application - hence it isn't currently passed to the service.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1RequestForPlacementService.kt
```diff
@@ -42,19 +45,96 @@ class Cas1RequestForPlacementService(
...
-      ApType.normal,
-      ApType.rfap,
-      ApType.mhapStJosephs,
-      ApType.mhapElliottHouse,
-      -> Cas1RequestsForPlacementDurationsCalculationResponseDto(Period.ofWeeks(12).days, maxDurationDays = null)
+    val application = applicationService.getApplication(applicationId)
+      ?: return CasResult.NotFound("Application", applicationId.toString())
+    val liveTier = (caseService.getCase(application.crn) ?: return CasResult.NotFound("Case", application.crn)).tier
+    val liveTierVersion = liveTier?.version
+      ?: return CasResult.NotFound("Version for live tier associated with case CRN", application.crn)
```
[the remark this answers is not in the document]
danhumphreys-moj: This is fine for now, live version is analogous to tier.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1RequestForPlacementService.kt
```diff
@@ -42,19 +45,96 @@ class Cas1RequestForPlacementService(
...
+                )
+              } else if (liveTier.tierScore == "C" || liveTier.tierScore == "D") {
+                CasResult.Success(
+                  Cas1RequestsForPlacementDurationsCalculationResponseDto(Period.ofWeeks(8).days, maxDurationDays = null),
+                )
+              } else if (liveTier.tierScore in listOf("E", "F", "G")) {
+                CasResult.GeneralValidationError("Cannot calculate duration for ap type $apType, sentence type ${application.sentenceType}, tier score ${liveTier.tierScore}")
+              } else {
+                CasResult.GeneralValidationError("Cannot calculate duration for ap type $apType, sentence type ${application.sentenceType}, tier score ${liveTier.tierScore}")
+              }
```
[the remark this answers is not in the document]
danhumphreys-moj: Yes the errors here are the same here.  We're currently looking at removing sentence type so will look at this in the next PR.  We could change this logic to use `when` which could be exhaustive.  It would be good if the tier 3 scores were an enumeration here too since we know they are fixed to A - G.

danhumphreys-moj:
Noticed we use sentenceType: String quite a lot instead of sentenceType: SentenceTypeOption which could probably used instead.

Also v3 tier scores could be an enumeration since they are fixed to A - G which would make duration logic more readable (along with `when`s).

The single `defaultDurations()` method needs refactoring but I understand we're still looking at requirements so at least we have an initial version to work with.  Potentially we could have separate private validation functions per AP type which would help simplify things somewhat.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/controller/Cas1RequestsForPlacementController.kt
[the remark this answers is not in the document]
davidatkinsuk: agreed

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1RequestForPlacementService.kt
```diff
@@ -42,19 +45,96 @@ class Cas1RequestForPlacementService(
...
-      ApType.pipe -> Cas1RequestsForPlacementDurationsCalculationResponseDto(Period.ofWeeks(26).days, maxDurationDays = null)
-      ApType.esap -> Cas1RequestsForPlacementDurationsCalculationResponseDto(Period.ofWeeks(52).days, maxDurationDays = null)
-      ApType.normal,
-      ApType.rfap,
-      ApType.mhapStJosephs,
-      ApType.mhapElliottHouse,
-      -> Cas1RequestsForPlacementDurationsCalculationResponseDto(Period.ofWeeks(12).days, maxDurationDays = null)
+    val application = applicationService.getApplication(applicationId)
+      ?: return CasResult.NotFound("Application", applicationId.toString())
+    val liveTier = (caseService.getCase(application.crn) ?: return CasResult.NotFound("Case", application.crn)).tier
```
davidatkinsuk: I don't think we should return a 404 here because the case crn isn't in the URL. Instead I think this is a generic error because it should never happen

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1RequestForPlacementService.kt
```diff
@@ -42,19 +45,96 @@ class Cas1RequestForPlacementService(
...
+    val liveTierVersion = liveTier?.version
+      ?: return CasResult.NotFound("Version for live tier associated with case CRN", application.crn)
+
+    return when (liveTierVersion) {
```
davidatkinsuk, line 60: i'm not sure if it would make things more readable if we have a private function for each tier version calculation?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1RequestForPlacementService.kt
```diff
@@ -42,19 +45,96 @@ class Cas1RequestForPlacementService(
...
+          ApType.mhapStJosephs,
+          ApType.mhapElliottHouse,
+          -> Cas1RequestsForPlacementDurationsCalculationResponseDto(Period.ofWeeks(12).days, maxDurationDays = null)
+        },
+      )
+
+      TierVersionDto.V3 -> {
+        when (apType) {
+          ApType.pipe -> CasResult.Success(
+            Cas1RequestsForPlacementDurationsCalculationResponseDto(Period.ofWeeks(26).days, maxDurationDays = null),
```
davidatkinsuk: maybe worth having a private function to build the success response to remove some duplication e.g.

```fun createDurationCalculation(period: Period, maxDurationDays: Period?): CasResult.Success<Response> etc.```

File: src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1RequestForPlacementService.kt [hunk 7]
```diff
@@ -42,19 +45,96 @@ class Cas1RequestForPlacementService(
...
+          }
+
+          ApType.normal,
+          ApType.rfap,
+          -> if (application.isWomensApplication == true) {
+            CasResult.Success(
+              Cas1RequestsForPlacementDurationsCalculationResponseDto(Period.ofWeeks(16).days, maxDurationDays = null),
+            )
+          } else {
+            if (application.sentenceType == SentenceTypeOption.life.value || application.sentenceType == SentenceTypeOption.ipp.value) {
```
davidatkinsuk: we should be using the sentence type passed to the controller for this (this is required to support additional requests for placements where sentence type may change). It's likely we're removing sentence type from the calculation but as things will change probably best to fix this now just in case it is still required

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1RequestForPlacementService.kt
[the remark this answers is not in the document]
davidatkinsuk: yes i agree - functionally works fine so will leave as is to avoid unnecessary rework

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas1/Cas1RequestForPlacementServiceTest.kt
```diff
@@ -259,41 +269,690 @@ class Cas1RequestForPlacementServiceTest {
...
+      )
+      fun `returns duration 12 weeks when apType is normal or mhapElliottHouse or mhapStJosephs or rfap`(apType: String) {
+        val application = ApprovedPremisesApplicationEntityFactory()
+          .withCreatedByUser(
+            UserEntityFactory().withProbationRegion(
+              ProbationRegionEntityFactory().produce(),
+            )
+              .produce(),
+          )
+          .produce()
```
davidatkinsuk, line 339: i think you could use `ApprovedPremisesApplicationEntityFactory().withDefaults().produce()` in all places we create an application? (i suspect this was existing code)

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas1/Cas1RequestForPlacementServiceTest.kt
```diff
@@ -259,41 +269,690 @@ class Cas1RequestForPlacementServiceTest {
...
+        ],
+      )
+      fun `returns duration 26 weeks when apType is mhap st josephs or mhap elliott house and mens`(apType: String) {
+        val application = ApprovedPremisesApplicationEntityFactory()
+          .withCreatedByUser(
+            UserEntityFactory().withProbationRegion(
+              ProbationRegionEntityFactory().produce(),
+            )
+              .produce(),
+          )
```
davidatkinsuk: could we explicitly set `.withIsWomensApplication(false)` on this (and anywhere else where a rule is for men only)

davidatkinsuk (approved):
Thanks for the exhaustive testing. I think all my comments are just around style and reuse 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1RequestForPlacementService.kt
danhumphreys-moj: Yes I did wonder what error to return here and landed on a 404.  Have changed this now to general validation error.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1RequestForPlacementService.kt
danhumphreys-moj: Yes - I think this will re-jig the logic so will save it for the next iteration.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1RequestForPlacementService.kt
danhumphreys-moj: Yes nice - have added this which helps reduce duplication:

```
  private fun createDurationCalculation(period: Period, maxDurationDays: Int?): CasResult.Success<Cas1RequestsForPlacementDurationsCalculationResponseDto> {
    return CasResult.Success(Cas1RequestsForPlacementDurationsCalculationResponseDto(period.days, maxDurationDays))
  }
```

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1RequestForPlacementService.kt
danhumphreys-moj: Have changed logic throughout to use sentence type passed as query param and not application sentence type.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas1/Cas1RequestForPlacementServiceTest.kt
danhumphreys-moj: Yes - have added this default application method which reduces boilerplate.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas1/Cas1RequestForPlacementServiceTest.kt
danhumphreys-moj: Yes - have added `.withIsWomensApplication(false)` for all mens tests.

danhumphreys-moj:
@davidatkinsuk would you mind giving this another once over?  I've addressed the comments in a new commit to make reviewing easier: https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/5029/changes/f1352c96dcdd1f025d659b9aa02f69c2993f0617

Reason for the re-review request is that the sentence type passed to the controller is used exclusively in the default duration logic now (rather than the application sentence type).

Integration tests consciously use different application sentence types to that used in the query param.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/controller/Cas1RequestsForPlacementController.kt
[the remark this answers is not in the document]
danhumphreys-moj: Following [this comment](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/5029#discussion_r3629597658), sentence type passed to the controller as a query parameter is now exclusively used to determine default placement duration logic, rather than application sentence type.

davidatkinsuk:
> f1352c9

i'm happy with this

PR #5031, opened 2026-07-22
FM-922 Ensure case exists

vbala-moj:
Updated the CAS3 booking service to ensure the case exists when creating a new booking.


davidatkinsuk (approved):
LGTM 👍

PR #5035, opened 2026-07-23
FM-922: Backfill Cases Job

vbala-moj:
Updated the backfill cases job query to use the CRN from the CAS3 booking, as a booking can exist without an associated application.


davidatkinsuk (approved):
LGTM 👍

PR #5043, opened 2026-07-24
FM-945 Move BedEntity to CAS1

muhammad-elabdulla:
This [PR FM-945](https://dsdmoj.atlassian.net/browse/FM-945)  is to move the BedEntity into the CAS1 namespace and rename it to Cas1BedEntity, as CAS3 has its own premises entity Cas3BedspacesEntity.


src/main/resources/db/migration/all/20260724114520000__delete_temporary_accommodation_beds.sql
```diff
@@ -0,0 +1,15 @@
+-- Delete all beds belonging to temporary-accommodation CAS3 premises from beds table
+ALTER TABLE cas3_void_bedspaces DROP COLUMN IF EXISTS bed_id;
+
+ALTER TABLE bed_moves RENAME TO archive_bed_moves;
+ALTER INDEX IF EXISTS bed_moves_pkey RENAME TO archive_bed_moves_pkey;
+
+DELETE FROM beds
+WHERE room_id IN (
+  SELECT id FROM rooms WHERE premises_id IN (
+    SELECT id FROM premises WHERE service = 'temporary-accommodation'
```
davidatkinsuk, line 10: in a previous PR we renamed the corresponding entity for `premises` to `Cas1PremisesBaseEntity`, but this query suggests that there are still CAS3 premises in the `premises` table? Is that correct? are we planning on removing these entries (and corresponding rooms?)

src/main/resources/db/migration/all/20260724115914000__rename_beds_to_cas1_beds.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE beds RENAME TO cas1_beds;
+
+-- Align the index/primary-key names on the (now renamed) cas1_beds table
+ALTER INDEX IF EXISTS beds_pkey RENAME TO cas1_beds_pkey;
+ALTER INDEX IF EXISTS beds_code_key RENAME TO cas1_beds_code_key;
+ALTER INDEX IF EXISTS beds_room_id_idx RENAME TO cas1_beds_room_id_idx;
```
davidatkinsuk: Can we remove `IF EXISTS` in these please as they should exist

src/main/resources/db/migration/all/20260724114520000__delete_temporary_accommodation_beds.sql
```diff
@@ -0,0 +1,15 @@
+-- Delete all beds belonging to temporary-accommodation CAS3 premises from beds table
+ALTER TABLE cas3_void_bedspaces DROP COLUMN IF EXISTS bed_id;
+
+ALTER TABLE bed_moves RENAME TO archive_bed_moves;
+ALTER INDEX IF EXISTS bed_moves_pkey RENAME TO archive_bed_moves_pkey;
```
davidatkinsuk: can we remove `IF_EXISTS` from this please as it should exist

src/main/resources/db/migration/all/20260724114520000__delete_temporary_accommodation_beds.sql
```diff
@@ -0,0 +1,15 @@
+-- Delete all beds belonging to temporary-accommodation CAS3 premises from beds table
+ALTER TABLE cas3_void_bedspaces DROP COLUMN IF EXISTS bed_id;
```
davidatkinsuk: can we remove IF_EXISTS from this please as it should exist

davidatkinsuk (approved):
Looks good. A few minor comments from me 👍

src/main/resources/db/migration/all/20260724114520000__delete_temporary_accommodation_beds.sql
muhammad-elabdulla: Yes the approach is to remove all the rooms and premises that belong to CAS3

src/main/resources/db/migration/all/20260724115914000__rename_beds_to_cas1_beds.sql
muhammad-elabdulla: Indexes are not as sensitive as tables, so I don't see any issue with checking whether they exist before renaming them

davidatkinsuk: Database structure resulting from migrations should be deterministic. If an index is missing this script should fail because something bad has happened, or there's a mistake in this script. Therefore, we should drop `IF EXISTS`

src/main/resources/db/migration/all/20260724114520000__delete_temporary_accommodation_beds.sql
davidatkinsuk: OK. So ideally we shouldn't have renamed `PremisesEntity` to `Cas1PremisesBaseEntity` until after this change and CAS3 rooms and premises had also been removed? I just want to make sure i understand this correctly. I think in the short term it's fine as long as we are removing all those entries in subsequent PRs

PR #5046, opened 2026-07-24
FM-876 Delete old inbox events

vbala-moj:
Added a scheduled job to delete inbox events older than 90 days.


davidatkinsuk (approved):
perfect. thanks

PR #5049, opened 2026-07-27
FM-945 Rename beds table in migration scripts

muhammad-elabdulla:
This PR FM-945 is to rename beds table in migration scripts


davidatkinsuk (approved):
LGTM 👍

PR #5052, opened 2026-07-27
FM-906 CAS3 Expose OASys Last Updated date

muhammad-elabdulla:
This [PR FM-906](https://dsdmoj.atlassian.net/browse/FM-906) is to expose OASys last updated date in CAS3


davidatkinsuk (approved):
LGTM 👍

PR #5056, opened 2026-07-28
FM-926 Check tier on app creation

vbala-moj:
Check tier availability when creating a new application and throw an error if the tier cannot be retrieved.


muhammad-elabdulla (approved):
LGTM

PR #5058, opened 2026-07-28
FM-958 Move RoomEntity to CAS1

muhammad-elabdulla:
This [PR FM-958](https://dsdmoj.atlassian.net/browse/FM-958) is to move RoomEntity to CAS1


davidatkinsuk (approved):
LGTM 👍

PR #5059, opened 2026-07-28
FM-958 Delete from room_characteristics table all rows belonging to CAS3

muhammad-elabdulla:
This PR is to delete from room_characteristics table all rows belonging to CAS3


davidatkinsuk (dismissed):
LGTM 👍

davidatkinsuk (approved):
LGTM 👍

PR #5061, opened 2026-07-29
FM-967 Move Premises table to CAS1

muhammad-elabdulla:
This PR is to move Premises table to CAS1


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/jpa/entity/cas1/Cas1PremisesBaseEntity.kt
```diff
@@ -114,7 +114,7 @@ interface ApprovedPremisesRepository : JpaRepository<ApprovedPremisesEntity, UUI
 
 @SuppressWarnings("LongParameterList")
 @Entity
-@Table(name = "premises")
+@Table(name = "cas1_premises")
```
davidatkinsuk: this will result in two tables

`cas1_premises` and `approved_premises`

I think this is a bit confusing. Can I propose we either 

1. rename to `cas1_premises_base` to match the entity _or_ 
2. we leave this table as is until we get the chance to merge the two tables

I agree we should remove the cas3 entries from `premises`

muhammad-elabdulla: I prefer the first approach to rename it to cas1_premises_base

davidatkinsuk (approved):
LGTM 👍

PR #5065, opened 2026-07-29
FM-951: Enable case-insensitive CRN handling

vbala-moj:
Convert CRNs to uppercase to ensure case creation and retrieval are case-insensitive.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/service/CaseService.kt
```diff
@@ -95,11 +98,16 @@ class CaseService(
    * If a case can't be found for a given CRN there will be no corresponding entry in the result
    */
   fun getCases(crns: List<String>): List<CaseDto> {
-    val result = caseRepository.findByCrnIn(crns).map { it.toDto() }
+    val requestedCrns = crns.map(String::uppercase)
```
davidatkinsuk: i guess could call this `normalisedCrns` for consistency with other functions?

davidatkinsuk (approved):
LGTM, just one minor comment

PR #5067, opened 2026-07-30
Avoid unnecessary case updates.

vbala-moj:
Only update Case entities when their data has changed.

Replaced the manual lastUpdatedAt update with Hibernate's @UpdateTimestamp, so it is updated automatically whenever the entity is modified.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/entity/CaseEntity.kt
```diff
@@ -124,11 +125,6 @@ data class CaseEntity(
   var version: Long = 1,
 ) {
 
-  @PreUpdate
-  fun preUpdate() {
-    lastUpdatedAt = OffsetDateTime.now()
-  }
-
```
vbala-moj, line 131: No longer needed since `@UpdateTimestamp` handles this automatically.

muhammad-elabdulla (approved):
LGTM

PR #5072, opened 2026-07-30
feature/SAS-651-CAS1-Provide-extra-box-content

besscerule:
This PR adds the following fields to CAS1Suitable Application and makes sure it gets them from the correct places:

`
  // application info
  val applicationStartedAt: OffsetDateTime?,
  val applicationStartedBy: String?,
  val applicationSubmittedAt: OffsetDateTime?,
  val applicationSubmittedBy: String?,
  val applicationExpiresAt: LocalDate?,

  // assessment info
  val assessmentDecision: String?,
  val assessmentRejectionRationale: String?,

  // request for placement info
  val requestForPlacementDecision: String?,
  val requestForPlacementRejectionReason: String?,
  val requestSubmittedBy: String?,
  val requestSubmittedAt: LocalDate?,
  val withdrawalReason: String?,
  val withdrawalDate: LocalDate?,

  // placement info
  val actualArrivalDate: LocalDate?,
  val actualDepartureDate: LocalDate?,
  val cancellationReason: String?,

// multiple
  val expectedArrivalDate: LocalDate?,
  val duration: Int?,
`

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/api/model/RequestForPlacement.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/api/model/RequestForPlacement.kt
@@ -3,6 +3,8 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model
 import io.swagger.v3.oas.annotations.media.Schema
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.cas1.Cas1RequestedPlacementPeriod
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas1.dto.Cas1SpaceBookingShortSummary
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas1.dto.Cas1StaffDto
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.PlacementApplicationDecision
 import java.time.LocalDate
 
 data class RequestForPlacement(
@@ -16,11 +18,13 @@ data class RequestForPlacement(
   val type: RequestForPlacementType,
   @Schema(description = "Deprecated. Requests for placements only have one set of placement dates, use 'requestedPlacementPeriod' or 'authorisedPlacementPeriod' instead", deprecated = true)
   val placementDates: List<PlacementDates>,
+  val canonicalPlacementPeriod: Cas1RequestedPlacementPeriod,
   val requestedPlacementPeriod: Cas1RequestedPlacementPeriod,
   val authorisedPlacementPeriod: Cas1RequestedPlacementPeriod?,
   val status: RequestForPlacementStatus,
   val statusSetDate: LocalDate,
   val submittedAt: java.time.Instant? = null,
+  val submittedBy: Cas1StaffDto? = null,
   @Schema(description = "If `type` is `\"manual\"`, provides the value of `PlacementApplication.decisionMadeAt`. If `type` is `\"automatic\"` this field provides the value of `PlacementRequest.assessmentCompletedAt`. ")
   val requestReviewedAt: java.time.Instant? = null,
   val document: Any? = null,
@@ -29,4 +33,5 @@ data class RequestForPlacement(
   val releaseType: ReleaseTypeOption? = null,
   val situation: SituationOption? = null,
   var placements: List<Cas1SpaceBookingShortSummary> = emptyList(),
+  val decision: PlacementApplicationDecision?,
 )
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/dto/Cas1SuitableApplication.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/dto/Cas1SuitableApplication.kt
@@ -4,7 +4,6 @@ import io.swagger.v3.oas.annotations.media.Schema
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.RequestForPlacementStatus
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.WithdrawPlacementRequestReason
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.AssessmentDecision
-import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.PlacementApplicationDecision
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.ApprovedPremisesApplicationStatus
 import java.time.LocalDate
 import java.time.OffsetDateTime
@@ -58,7 +57,7 @@ data class Cas1ExternalAssessmentDto(
 
 data class Cas1ExternalRequestForPlacementDto(
   val status: RequestForPlacementStatus?,
-  val decision: PlacementApplicationDecision?,
+  val decision: PlacementApplicationDecisionDto?,
   val rejectionReason: String?,
   val submittedBy: Cas1StaffDto?,
   val submittedAt: LocalDate?,
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/dto/Cas1SuitableApplication.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/dto/Cas1SuitableApplication.kt
@@ -1,9 +1,9 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.cas1.dto
 
 import io.swagger.v3.oas.annotations.media.Schema
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.AssessmentDecision
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.RequestForPlacementStatus
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.WithdrawPlacementRequestReason
-import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.AssessmentDecision
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.ApprovedPremisesApplicationStatus
 import java.time.LocalDate
 import java.time.OffsetDateTime
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/dto/Cas1SuitableApplication.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/dto/Cas1SuitableApplication.kt
@@ -1,17 +1,79 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.cas1.dto
 
+import io.swagger.v3.oas.annotations.media.Schema
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.RequestForPlacementStatus
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.WithdrawPlacementRequestReason
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.AssessmentDecision
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.PlacementApplicationDecision
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.ApprovedPremisesApplicationStatus
 import java.time.LocalDate
+import java.time.OffsetDateTime
 import java.util.UUID
 
+@Schema(description = "Details about the most current application, with associated assessment, request for placement and placement")
 data class Cas1SuitableApplication(
-  val id: UUID,
-  val applicationStatus: ApprovedPremisesApplicationStatus,
-  val requestForPlacementStatus: RequestForPlacementStatus?,
-  val placementStatus: Cas1SpaceBookingStatus?,
-  val premises: Cas1ExternalPremisesDto?,
   val uiUrl: String,
+  val application: Cas1ExternalApplicationDto,
+  val assessment: Cas1ExternalAssessmentDto?,
+  val requestForPlacement: Cas1ExternalRequestForPlacementDto?,
+  val placement: Cas1ExternalPlacementDto?,
+  val placementHistory: List<Cas1PlacementPairDto>,
+
+  @Deprecated("This field will be removed once SAS is updated to use application.id")
+  val id: UUID,
+
+  @Deprecated("This field will be removed once SAS is updated to use application.status")
+  val applicationStatus: ApprovedPremisesApplicationStatus,
+
+  @Deprecated("This field will be removed once SAS is updated to use requestForPlacement.status")
+  val requestForPlacementStatus: RequestForPlacementStatus?,
+
+  @Deprecated("This field will be removed once SAS is updated to use placement.status")
+  val placementStatus: Cas1SpaceBookingStatus?,
+
+  @Deprecated("This field will be removed once SAS is updated to use placement.premises")
+  val premises: Cas1ExternalPremisesDto?,
+)
+
+@Schema(description = "Details about a placement, with it's associated request for placement")
+data class Cas1PlacementPairDto(
+  val requestForPlacement: Cas1ExternalRequestForPlacementDto?,
+  val placement: Cas1ExternalPlacementDto?,
+  val dateApplied: LocalDate,
+)
+
+data class Cas1ExternalApplicationDto(
+  val id: UUID,
+  val status: ApprovedPremisesApplicationStatus,
+  val createdAt: OffsetDateTime,
+  val createdBy: Cas1StaffDto,
+  val submittedAt: OffsetDateTime?,
+  val expiresAt: LocalDate?,
+)
+
+data class Cas1ExternalAssessmentDto(
+  val decision: AssessmentDecision?,
+  val rejectionRationale: String?,
+)
+
+data class Cas1ExternalRequestForPlacementDto(
+  val status: RequestForPlacementStatus?,
+  val decision: PlacementApplicationDecision?,
+  val rejectionReason: String?,
+  val submittedBy: Cas1StaffDto?,
+  val submittedAt: LocalDate?,
+  val withdrawalReason: WithdrawPlacementRequestReason?,
+  val withdrawalDate: LocalDate?,
+  val expectedArrivalDate: LocalDate?,
+  val durationDays: Int?,
+)
+
+data class Cas1ExternalPlacementDto(
+  val status: Cas1SpaceBookingStatus?,
+  val actualArrivalDate: LocalDate?,
+  val actualDepartureDate: LocalDate?,
+  val cancellationReason: String?,
+  val premises: Cas1ExternalPremisesDto?,
 )
 
 data class Cas1ExternalPremisesDto(
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1DomainEventService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1DomainEventService.kt
@@ -8,7 +8,6 @@ import org.springframework.data.repository.findByIdOrNull
 import org.springframework.stereotype.Service
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.cas1.model.ApplicationAssessed
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.cas1.model.ApplicationExpired
-import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.cas1.model.ApplicationExpiredManually
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.cas1.model.ApplicationSubmitted
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.cas1.model.ApplicationSubmittedEnvelope
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.events.cas1.model.ApplicationWithdrawn
@@ -86,7 +85,6 @@ class Cas1DomainEventService(
   fun getBookingKeyWorkerAssignedEvent(id: UUID) = get(id, BookingKeyWorkerAssigned::class)
   fun getApplicationWithdrawnEvent(id: UUID) = get(id, ApplicationWithdrawn::class)
   fun getApplicationExpiredEvent(id: UUID) = get(id, ApplicationExpired::class)
-  fun getApplicationExpiredManuallyEvent(id: UUID) = get(id, ApplicationExpiredManually::class)
   fun getPlacementApplicationWithdrawnEvent(id: UUID) = get(id, PlacementApplicationWithdrawn::class)
   fun getPlacementApplicationAllocatedEvent(id: UUID) = get(id, PlacementApplicationAllocated::class)
   fun getMatchRequestWithdrawnEvent(id: UUID) = get(id, MatchRequestWithdrawn::class)
@@ -281,7 +279,7 @@ class Cas1DomainEventService(
     if (emittable && domainEvent.emit) {
       emit(domainEventEntity)
     } else {
-      log.debug("Not emitting domain event of type $eventType. Type emittable? $emittable. Emit? ${domainEvent.emit}")
+      log.debug("Not emitting domain event of type {}. Type emittable? {}. Emit? {}", eventType, emittable, domainEvent.emit)
     }
   }
 
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/external/Cas1ExternalApplicationService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/external/Cas1ExternalApplicationService.kt
@@ -1,8 +1,6 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas1.external
 
 import org.springframework.stereotype.Service
-import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.RequestForPlacementStatus
-import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.WithdrawPlacementRequestReason
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas1.dto.Cas1ExternalPremisesDto
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas1.dto.Cas1PlacementPairDto
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas1.dto.Cas1SpaceBookingStatus
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/external/Cas1ExternalApplicationService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/external/Cas1ExternalApplicationService.kt
@@ -1,18 +1,20 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas1.external
 
-import org.springframework.beans.factory.annotation.Value
 import org.springframework.stereotype.Service
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.RequestForPlacementStatus
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.WithdrawPlacementRequestReason
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas1.dto.Cas1ExternalPremisesDto
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas1.dto.Cas1PlacementPairDto
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas1.dto.Cas1SpaceBookingStatus
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas1.dto.Cas1SuitableApplication
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.common.results.CasResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApprovedPremisesApplicationEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.ApprovedPremisesApplicationRepository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.PlacementApplicationDecision
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.ApprovedPremisesApplicationStatus
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas1.Cas1PremisesService
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.cas1.Cas1RequestForPlacementService
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.transformer.cas1.external.Cas1ExternalApplicationTransformer
 import java.time.LocalDate
 import java.util.UUID
 
@@ -22,47 +24,38 @@ class Cas1ExternalApplicationService(
   private val approvedPremisesApplicationRepository: ApprovedPremisesApplicationRepository,
   private val cas1RequestForPlacementService: Cas1RequestForPlacementService,
   private val cas1PremisesService: Cas1PremisesService,
-  @Value($$"${url-templates.frontend.application}") private val cas1ApplicationUrlTemplate: String,
+  private val transformer: Cas1ExternalApplicationTransformer,
 ) {
 
   private val mostSuitableApplication = compareBy<ApprovedPremisesApplicationEntity> { suitableStatusesAsc[it.status] }
     .thenBy { it.submittedAt ?: it.createdAt }
 
-  fun getPlacementHistories(applicationId: UUID): List<Cas1PlacementHistory> {
+  fun getPlacementPairs(applicationId: UUID): List<Cas1PlacementPairDto> {
     val rfps = (cas1RequestForPlacementService.getRequestsForPlacementByApplication(applicationId, requestingUser = null) as CasResult.Success).value
 
     return rfps.flatMap { rfp ->
+      val withdrawalDate = if (rfp.isWithdrawn) cas1RequestForPlacementService.getRequestForPlacementWithdrawalDate(rfp) else null
 
       if (rfp.placements.isEmpty()) {
+        val rejectionReason = if (rfp.decision == PlacementApplicationDecision.REJECTED) cas1RequestForPlacementService.getRequestForPlacementRejectionReason(rfp) else null
+
         listOf(
-          Cas1PlacementHistory(
-            dateApplied = rfp.statusSetDate,
-            requestForPlacementStatus = rfp.status,
-            withdrawalReason = rfp.withdrawalReason,
-            placementStatus = null,
-            premises = null,
+          transformer.transformToCas1PlacementPair(
+            rfp = rfp,
+            rejectionReason = rejectionReason,
+            withdrawalDate = withdrawalDate,
           ),
         )
       } else {
         rfp.placements.map { placement ->
-          val premises = cas1PremisesService.findPremisesById(placement.premises.id)
-            ?.let {
-              Cas1ExternalPremisesDto(
-                startDate = placement.actualArrivalDate ?: placement.expectedArrivalDate,
-                endDate = placement.actualDepartureDate ?: placement.expectedDepartureDate,
-                postcode = it.postcode,
-                addressLine1 = it.addressLine1,
-                addressLine2 = it.addressLine2,
-                town = it.town,
-              )
-            }
+          val premisesEntity = cas1PremisesService.findPremisesById(placement.premises.id)
 
-          Cas1PlacementHistory(
-            dateApplied = requireNotNull(placement.statusSetDate),
-            requestForPlacementStatus = rfp.status,
-            withdrawalReason = rfp.withdrawalReason,
-            placementStatus = placement.status,
-            premises = premises,
+          transformer.transformToCas1PlacementPair(
+            rfp = rfp,
+            rejectionReason = null,
+            withdrawalDate = withdrawalDate,
+            placement = placement,
+            premises = premisesEntity,
           )
         }
       }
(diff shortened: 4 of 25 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/Cas1DomainEventService.kt
besscerule, line 97: never used

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas1/Cas1DomainEventServiceTest.kt
```diff
@@ -91,11 +111,6 @@ class Cas1DomainEventServiceTest {
       .entries
       .filter { it.cas == DomainEventCas.CAS1 && it.cas1Info!!.emittable }
```
besscerule, line 93: not used

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/api/model/RequestForPlacement.kt
```diff
@@ -21,6 +21,7 @@ data class RequestForPlacement(
   val status: RequestForPlacementStatus,
   val statusSetDate: LocalDate,
   val submittedAt: java.time.Instant? = null,
+  val submittedBy: String?,
```
davidatkinsuk: SAS introduced a Cas1StaffDto following a discussion about how user info should be passed back to SAS. I think we should use that here? It was defined in https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/4868/changes#diff-a6d504752ab3e58d3d8fcb92efacfb3c89a4a8df8bc9978d4f47ec2dd9643322 following this discussion https://mojdt.slack.com/archives/C03K0HB0LBE/p1781520639764629

We do have something similar we use for timelines in CAS1 called `UserSummary` but that exposes an internal ID

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/dto/Cas1SuitableApplication.kt
```diff
@@ -3,15 +3,44 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.cas1.dto
...
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.ApprovedPremisesApplicationStatus
 import java.time.LocalDate
+import java.time.OffsetDateTime
 import java.util.UUID
 
 data class Cas1SuitableApplication(
   val id: UUID,
+  val uiUrl: String,
+
+  // application info
```
davidatkinsuk: presence of comments suggest we should maybe introduce embedded types to group this stuff?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/dto/Cas1SuitableApplication.kt
```diff
@@ -3,15 +3,44 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.cas1.dto
...
   val applicationStatus: ApprovedPremisesApplicationStatus,
+  val applicationStartedAt: OffsetDateTime?,
+  val applicationStartedBy: String?,
```
davidatkinsuk: as mentioned before can we use `Cas1StaffDto` where-ever we represent a user?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/cas1/Cas1DomainEventServiceTest.kt
davidatkinsuk: good spot

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/external/Cas1ExternalApplicationService.kt
```diff
@@ -31,35 +33,28 @@ class Cas1ExternalApplicationService(
     val rfps = (cas1RequestForPlacementService.getRequestsForPlacementByApplication(applicationId, requestingUser = null) as CasResult.Success).value
 
     return rfps.flatMap { rfp ->
+      val withdrawalDate = if (rfp.isWithdrawn) cas1DomainEventService.getRequestForPlacementWithdrawalDate(rfp.id) else null
```
davidatkinsuk: I think `withdrawalDate` should really go on the `RequestForPlacement` model, and pulled from a first class field (i.e. we should add `withdrawal_date` to `placement_applications` and `placement_requests`, which we'd need to backfill from domain events).

Obviously that's a non trivial piece of work. In the meantime can i propose the following:

**Isolate the logic**
1. Move the getRequestForPlacementWithdrawalDate to RequestForPlacementService
2. Pass in the RequestForPlacement to this function instead of the ID. This at least keeps the logic contained within the RFP code (because as you know the ID can actually refer to multiple entity types :( )

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/external/Cas1ExternalApplicationService.kt
```diff
@@ -31,35 +33,28 @@ class Cas1ExternalApplicationService(
...
 
       if (rfp.placements.isEmpty()) {
+        val rejectionReason = if (rfp.decision == "REJECTED") cas1DomainEventService.getRequestForPlacementRejectionReason(rfp.id) else null
```
davidatkinsuk: As with my previous comment, can we please move the `getRequestForPlacementRejectionReason` function onto the `RequestForPlacementService` with an end goal of making this a first class property?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/api/model/RequestForPlacement.kt
```diff
@@ -29,4 +30,5 @@ data class RequestForPlacement(
   val releaseType: ReleaseTypeOption? = null,
   val situation: SituationOption? = null,
   var placements: List<Cas1SpaceBookingShortSummary> = emptyList(),
+  val decision: String?,
```
davidatkinsuk: this is mapped as a String from an enumerated type, but then we have code in the external transformer like

```(rfp.decision == "REJECTED")```

we should retain the enumeration here so we're not relying on string comparisons

But! i think you could just use `status` instead which has `requestRejected`, meaning 'decision' should be removed?

i've improved status documentation incase it helps here - https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/5111

File: src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/dto/Cas1SuitableApplication.kt [hunk 8]
```diff
@@ -3,15 +3,44 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.cas1.dto
...
+
+  // placement info
   val placementStatus: Cas1SpaceBookingStatus?,
   val premises: Cas1ExternalPremisesDto?,
-  val uiUrl: String,
+  val actualArrivalDate: LocalDate?,
+  val actualDepartureDate: LocalDate?,
+  val cancellationReason: String?,
+
+// multiple
```
davidatkinsuk: i'm not sure what this comment means?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/dto/Cas1SuitableApplication.kt
```diff
@@ -3,15 +3,44 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.cas1.dto
...
 data class Cas1SuitableApplication(
   val id: UUID,
+  val uiUrl: String,
+
+  // application info
   val applicationStatus: ApprovedPremisesApplicationStatus,
+  val applicationStartedAt: OffsetDateTime?,
+  val applicationStartedBy: String?,
+  val applicationSubmittedAt: OffsetDateTime?,
+  val applicationSubmittedBy: String?,
```
davidatkinsuk: as mentioned before can we use Cas1StaffDto where-ever we represent a user?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/dto/Cas1SuitableApplication.kt
```diff
@@ -3,15 +3,44 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.cas1.dto
...
+
+  // assessment info
+  val assessmentDecision: String?,
+  val assessmentRejectionRationale: String?,
+
+  // request for placement info
   val requestForPlacementStatus: RequestForPlacementStatus?,
+  val requestForPlacementDecision: String?,
+  val requestForPlacementRejectionReason: String?,
+  val requestSubmittedBy: String?,
```
davidatkinsuk: as mentioned before can we use Cas1StaffDto where-ever we represent a user?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/dto/Cas1SuitableApplication.kt
```diff
@@ -3,15 +3,44 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.cas1.dto
...
+  val requestSubmittedAt: LocalDate?,
+  val withdrawalReason: String?,
```
davidatkinsuk: i think this comes from an enumeration. I think we should retain that enumeration and let the caller make it user friendly?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/dto/Cas1SuitableApplication.kt
```diff
@@ -3,15 +3,44 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.cas1.dto
...
   val placementStatus: Cas1SpaceBookingStatus?,
   val premises: Cas1ExternalPremisesDto?,
-  val uiUrl: String,
+  val actualArrivalDate: LocalDate?,
+  val actualDepartureDate: LocalDate?,
+  val cancellationReason: String?,
+
+// multiple
+  val expectedArrivalDate: LocalDate?,
+  val duration: Int?,
```
davidatkinsuk: rename to durationDays for clarity?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/dto/Cas1SuitableApplication.kt
```diff
@@ -3,15 +3,44 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.cas1.dto
...
+  // application info
   val applicationStatus: ApprovedPremisesApplicationStatus,
+  val applicationStartedAt: OffsetDateTime?,
+  val applicationStartedBy: String?,
+  val applicationSubmittedAt: OffsetDateTime?,
+  val applicationSubmittedBy: String?,
+  val applicationExpiresAt: LocalDate?,
+
+  // assessment info
+  val assessmentDecision: String?,
```
davidatkinsuk: Retain the enum. Can use `AssessmentDecision`

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/dto/Cas1SuitableApplication.kt
```diff
@@ -3,15 +3,44 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.cas1.dto
...
+  val assessmentRejectionRationale: String?,
+
+  // request for placement info
   val requestForPlacementStatus: RequestForPlacementStatus?,
+  val requestForPlacementDecision: String?,
```
davidatkinsuk: as mentioned in another comment i'm not sure you need this as can use status?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas1/external/Cas1ExternalApplicationTransformer.kt
```diff
@@ -0,0 +1,124 @@
...
+    requestSubmittedAt = rfp.submittedAt?.toLocalDate(),
+    withdrawalReason = if (rfp.isWithdrawn) rfp.withdrawalReason?.name else null,
+    withdrawalDate = withdrawalDate,
+    placementStatus = placement?.status,
+    premises = transformToCas1ExternalPremisesDto(rfp, placement, premises),
+    actualArrivalDate = placement?.actualArrivalDate,
+    actualDepartureDate = placement?.actualDepartureDate,
+    cancellationReason = getCancellationReason(placement),
+    expectedArrivalDate = placement?.expectedArrivalDate ?: rfp.placementDates.firstOrNull()?.expectedArrival,
+    duration = rfp.placementDates.firstOrNull()?.duration,
```
davidatkinsuk: placementDates are deprecated and shouldn't be used (because we no longer have a list)

annoyingly that does mean you'd need to check `authorisedPlacementPeriod` and if that's null use `requestedPlacementPeriod`. I propose we update `RequestForPlacement` to have a `canonicalPlacementPeriod` that will `authorisedPlacementPeriod` if defined or `requestedPlacementPeriod`. You should then use that here

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas1/external/Cas1ExternalApplicationTransformer.kt
```diff
@@ -0,0 +1,124 @@
...
+    dateApplied = placement?.statusSetDate ?: rfp.statusSetDate,
+  )
+
+  private fun transformToCas1ExternalPremisesDto(
+    rfp: RequestForPlacement,
+    placement: Cas1SpaceBookingShortSummary? = null,
+    premises: ApprovedPremisesEntity? = null,
+  ) = premises?.let {
+    Cas1ExternalPremisesDto(
+      startDate = placement?.actualArrivalDate ?: placement?.expectedArrivalDate ?: rfp.placementDates.firstOrNull()?.expectedArrival,
```
davidatkinsuk: see comment above about using `placementDates`. Should instead add and use  `canonicalPlacementPeriod`. Could also add the same to `Cas1SpaceBookingShortSummary` if it helps

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas1/external/Cas1ExternalApplicationTransformer.kt
```diff
@@ -0,0 +1,124 @@
...
+    )
+  }
+
+  fun getCancellationReason(placement: Cas1SpaceBookingShortSummary?): String? = if (placement?.cancellation?.reason?.name == "Other") placement.cancellation.reasonNotes else placement?.cancellation?.reason?.name
+
+  fun getApplicationExpiresAt(application: ApprovedPremisesApplicationEntity): LocalDate? {
+    val latestAssessment = application.getLatestAssessment()
+    val possibleExpiryDate = latestAssessment?.submittedAt?.toLocalDate()?.plusDays(365)
+    return if (latestAssessment?.decision == AssessmentDecision.ACCEPTED) possibleExpiryDate else null
+  }
```
davidatkinsuk: Can this function please be moved into the `Cas1ApplicationService`

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas1/external/Cas1ExternalApplicationTransformer.kt
```diff
@@ -0,0 +1,124 @@
...
+      requestSubmittedAt = suitablePlacement?.requestSubmittedAt,
+      withdrawalReason = suitablePlacement?.withdrawalReason,
+      withdrawalDate = suitablePlacement?.withdrawalDate,
+      placementStatus = suitablePlacement?.placementStatus,
+      premises = suitablePlacement?.premises,
+      actualArrivalDate = suitablePlacement?.actualArrivalDate,
+      actualDepartureDate = suitablePlacement?.actualDepartureDate,
+      cancellationReason = suitablePlacement?.cancellationReason,
+      expectedArrivalDate = suitablePlacement?.expectedArrivalDate ?: application.arrivalDate?.toLocalDate(),
+      duration = suitablePlacement?.duration ?: application.duration,
```
davidatkinsuk: this doesn't consider durations for requests for placements made after the application was submitted. If you could get at the request for placement here that should cover the application or subsequent request for placements. I'm a little lost in the model so will look at this again once all comments addressed

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas1/external/Cas1ExternalApplicationTransformer.kt
```diff
@@ -0,0 +1,124 @@
...
+      requestForPlacementRejectionReason = suitablePlacement?.requestForPlacementRejectionReason,
+    )
+  }
+}
+
+data class Cas1PlacementHistory(
```
davidatkinsuk: can you please succinctly describe in a comment what this represents

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/dto/Cas1SuitableApplication.kt
```diff
@@ -3,15 +3,44 @@ package uk.gov.justice.digital.hmpps.approvedpremisesapi.cas1.dto
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.RequestForPlacementStatus
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.model.ApprovedPremisesApplicationStatus
 import java.time.LocalDate
+import java.time.OffsetDateTime
 import java.util.UUID
 
 data class Cas1SuitableApplication(
```
davidatkinsuk, line 13: can you please describe using a `@Schema(description="")` what this represents

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas1/external/Cas1ExternalApplicationTransformer.kt
```diff
@@ -0,0 +1,124 @@
...
+      addressLine2 = it.addressLine2,
+      town = it.town,
+    )
+  }
+
+  fun getCancellationReason(placement: Cas1SpaceBookingShortSummary?): String? = if (placement?.cancellation?.reason?.name == "Other") placement.cancellation.reasonNotes else placement?.cancellation?.reason?.name
+
+  fun getApplicationExpiresAt(application: ApprovedPremisesApplicationEntity): LocalDate? {
+    val latestAssessment = application.getLatestAssessment()
+    val possibleExpiryDate = latestAssessment?.submittedAt?.toLocalDate()?.plusDays(365)
```
davidatkinsuk: could the 365 be pulled from a constant defined in an `ApplicationRepository` companion object (e.g. CAS1_EXPIRY_AFTER_ASSESSMENT_THRESHOLD_DAYS) and update the query defined on the function `findAllExpiredApplications` to use this constant. That will at least create a little bit of coupling between the duplicated logic, if annotations will allow it :D

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/dto/Cas1SuitableApplication.kt
besscerule: can be from placement, rfp or application info

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/api/model/RequestForPlacement.kt
besscerule: done

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/dto/Cas1SuitableApplication.kt
besscerule: done

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/dto/Cas1SuitableApplication.kt
besscerule: done

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/dto/Cas1SuitableApplication.kt
besscerule: done

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/dto/Cas1SuitableApplication.kt
besscerule: done

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/dto/Cas1SuitableApplication.kt
besscerule: done

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/external/Cas1ExternalApplicationService.kt
besscerule: done

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/external/Cas1ExternalApplicationService.kt
besscerule: done

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/api/model/RequestForPlacement.kt
besscerule: done

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/dto/Cas1SuitableApplication.kt
besscerule: done

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/dto/Cas1SuitableApplication.kt
besscerule: done

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/dto/Cas1SuitableApplication.kt
besscerule: we do need it because the requestForPlacementStatus can be 6 different statuses

davidatkinsuk: in that case can we retain the enumeration? Can use PlacementApplicationDecisionDto

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas1/external/Cas1ExternalApplicationTransformer.kt
besscerule: done

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas1/external/Cas1ExternalApplicationTransformer.kt
besscerule: done

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas1/external/Cas1ExternalApplicationTransformer.kt
besscerule: done

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas1/external/Cas1ExternalApplicationTransformer.kt
besscerule: i'm not sure what this means?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas1/external/Cas1ExternalApplicationTransformer.kt
besscerule: done

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas1/external/Cas1ExternalApplicationTransformer.kt
davidatkinsuk: i think is was getting confused because `suitablePlacement` doesn't necessarily/exclusively represent a `Cas1SpaceBooking` and can also represents a request for placement. 

With that said - I don't think you should fall back onto application.duration in this case because that should already be reflected by a `suitablePlacement` existing. Furthermore in the case where you used `application.duration` wouldn't have any values for the arrival/departure date fields so it may be strange to just provide duration (which represents the duration that would be requested _if_ a request for placement was to be created

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/dto/Cas1SuitableApplication.kt
besscerule: done - but used a slightly different enum

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas1/external/Cas1ExternalApplicationTransformer.kt
besscerule: done

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/dto/Cas1SuitableApplication.kt
besscerule: done

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas1/external/Cas1ExternalApplicationTransformer.kt
besscerule: done and changed the name and moved it somewhere else

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/external/Cas1ExternalApplicationService.kt
```diff
@@ -21,45 +22,38 @@ class Cas1ExternalApplicationService(
...
 
   private val mostSuitableApplication = compareBy<ApprovedPremisesApplicationEntity> { suitableStatusesAsc[it.status] }
     .thenBy { it.submittedAt ?: it.createdAt }
 
-  fun getPlacementHistories(applicationId: UUID): List<Cas1PlacementHistory> {
+  fun getPlacementPairs(applicationId: UUID): List<Cas1PlacementPairDto> {
     val rfps = (cas1RequestForPlacementService.getRequestsForPlacementByApplication(applicationId, requestingUser = null) as CasResult.Success).value
 
     return rfps.flatMap { rfp ->
+      val withdrawalDate = if (rfp.isWithdrawn) cas1RequestForPlacementService.getRequestForPlacementWithdrawalDate(rfp) else null
```
muhammad-elabdulla, line 35: There looks to be a potential N+1 issue in getPlacementPairs.

For each RFP, we're now making a couple of additional domain event lookups per row (getRequestForPlacementWithdrawalDate and getRequestForPlacementRejectionReason), on top of the existing findPremisesById call that's already made for each placement.

On its own, that's not too concerning. However, getReferralsByCrn calls getPlacementPairs inside a flatMap over every assessment for a CRN, so the number of queries scales roughly with assessments × RFPs × (placements + event lookups). For CRNs with a long history, that could result in a significant number of database round trips.

Would it be possible to batch the domain event lookups into a single query keyed by a set of placement/application IDs, and hoist the premises loading out of the loop, rather than querying one row at a time?

It would also be worth adding (or updating) an NPlus1QueriesTest assertion to lock this behaviour down and help prevent regressions in future.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/dto/Cas1SuitableApplication.kt
```diff
@@ -1,17 +1,79 @@
...
+)
+
+data class Cas1ExternalAssessmentDto(
+  val decision: AssessmentDecision?,
+  val rejectionRationale: String?,
+)
+
+data class Cas1ExternalRequestForPlacementDto(
+  val status: RequestForPlacementStatus?,
+  val decision: PlacementApplicationDecision?,
```
muhammad-elabdulla: Can you use PlacementApplicationDecisionDto instead of PlacementApplicationDecision?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/dto/Cas1SuitableApplication.kt
```diff
@@ -1,17 +1,79 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.cas1.dto
 
+import io.swagger.v3.oas.annotations.media.Schema
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.RequestForPlacementStatus
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.WithdrawPlacementRequestReason
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.AssessmentDecision
```
davidatkinsuk: this import should be to the API version of this type `uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.AssessmentDecision` (yes, naming is annoyingly the same!)

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/dto/Cas1SuitableApplication.kt
davidatkinsuk: this should be using the API type which is `PlacementApplicationDecisionDto`

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas1/external/Cas1ExternalApplicationTransformer.kt
```diff
@@ -0,0 +1,123 @@
...
+      startDate = placement?.actualArrivalDate ?: placement?.expectedArrivalDate,
+      endDate = placement?.actualDepartureDate ?: placement?.expectedDepartureDate,
+      postcode = it.postcode,
+      addressLine1 = it.addressLine1,
+      addressLine2 = it.addressLine2,
+      town = it.town,
+    )
+  }
+
+  private fun getCancellationReason(placement: Cas1SpaceBookingShortSummary?): String? = if (placement?.cancellation?.reason?.name == "Other") placement.cancellation.reasonNotes else placement?.cancellation?.reason?.name
```
muhammad-elabdulla, line 83: Can the value of Other be defined as a constant instead of a string in the code?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/external/Cas1ExternalApplicationService.kt
davidatkinsuk: This is a good point but i think it's probably fine as is for a few reasons:

* An application won't ever have _that_ many requests for placements (maybe 10)
* It's unlikely that they'll all be withdrawn
* we have a ticket to move the withdrawal date onto the entity (See https://dsdmoj.atlassian.net/browse/FM-1037) which would resolve this.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas1/external/Cas1ExternalApplicationTransformer.kt
davidatkinsuk: seconded. I guess it would be best placed as a constant on `CancellationReasonRepository` (we already have some constants there with reason ids). Could name it 'CAS1_OTHER_NAME'

simonwilson4moj: I think I've managed to achieve this. Can you review commit https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/5072/commits/7ca3531c26ad9e20e925800be9d23dc392a592c6

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/dto/Cas1SuitableApplication.kt
davidatkinsuk: resolved

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/transformer/cas1/external/Cas1ExternalApplicationTransformer.kt
davidatkinsuk: Updating on this, i still can't see this change. I think it should be a constant on CancellationReasonRepository

simonwilson4moj: Sorry i think i commented on the wrong comment, along with the wrong commit hash! 
Different comment resolved, i've updated the commit hash anyway in my above comment.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/external/Cas1ExternalApplicationService.kt
simonwilson4moj: I've had a go at this nonetheless in the below commit. Feel free to review if you have the time, i'm just looking at updating the tests. If not needed i'll resolve the other comments instead.

https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/5072/commits/d095319bd1cb083e2497848acd4701005afb1d4c

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas1/integration/external/Cas1ExternalApplicationsTest.kt
```diff
@@ -325,7 +326,7 @@ class Cas1ExternalApplicationsTest : IntegrationTestBase() {
                 id = application.id,
               ),
               assessment = Cas1ExternalAssessmentDto(
-                decision = assessment.decision,
+                decision = AssessmentDecisionApi.forValue(assessment.decision.toString()),
```
davidatkinsuk: The API types are lower case but the internal types used in the JPA model are upper case, so this won't work. 

Probably the easiest fix is to add the API mapping into the `AssessmentDecision` class defined in `AssessmentEntity`, similar to how we've mapped to `PlacementApplicationDecisionDto` from `PlacementApplicationDecision`

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/external/Cas1ExternalApplicationService.kt
```diff
@@ -22,47 +22,38 @@ class Cas1ExternalApplicationService(
...
 
       if (rfp.placements.isEmpty()) {
+        val rejectionReason = if (rfp.decision == PlacementApplicationDecision.REJECTED) cas1RequestForPlacementService.getRequestForPlacementRejectionReason(rfp) else null
+
```
davidatkinsuk, line 39: flagging this here but please continue to merge because i don't quite have the answer right now.

I _think_ using the rejection domain event may only work for requests for placements that were made after the application was assessed (that is where rfp.type == manual).

If the request for placement was part of the original application it would have been rejected as part of the application's assessment, and therefore i don't think the domain events backing these will exist and rejection reason would come from that assessment (but that's problematic because the reason could be about the application itself, not the specific request for placement date, for example)

I think the only way to easily figure this out is to do some local testing or static analysis of the various routes. We should then expose these two fields (withdrawal and rejection) on the `RequestForPlacement` model being used here as rfp

We've already done a bit of work to help with this via

https://dsdmoj.atlassian.net/browse/FM-1036
https://dsdmoj.atlassian.net/browse/FM-1035

these two stories add first class properties to the placement application for withdrawal date and decision summary and backfill them from the domain events current used in this code. If we could start populating these for 'automatic' placement applications (those in the original application) we can then pull directly from them and not have to worry about that detail here (we'd need to do a backfill there too)

As always, this comes back to the bi-furcated request for placement model see https://dsdmoj.atlassian.net/browse/FM-613

davidatkinsuk (approved):
LGTM 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/service/cas1/external/Cas1ExternalApplicationService.kt
davidatkinsuk: With all the above said ...

I just checked the code and 'automatic' placement application are only created once the application has been approved, therefore you'll never get a placement_applciation[automated=true] that has been rejected

This may highlight a bigger issue if you're expecting to see requests for placements that are part of the original applications where the original application is yet to be approved, or the original application has been rejected. This would probably be solved via https://dsdmoj.atlassian.net/browse/APS-2548, but we could always hack something together using `placement_applications_automatic` (i'd rather not do that if possible)

PR #5087, opened 2026-08-03
FM-1000 remove deprecated offender service function

davidatkinsuk:
The `OffenderService.getOffenderByCrn` function was used by one caller and returned the deprecated community api `OffenderDetailSummary` type. This PR updates the caller to use the non-deprecated equivalent and removes this function


danhumphreys-moj (approved):
LGTM - thanks for making this easy to review!

PR #5100, opened 2026-08-05
FM-579 Map New OAsys healthConditions property

muhammad-elabdulla:
This PR is to map New OAsys healthConditions property


davidatkinsuk (approved):
LGTM 👍

PR #5102, opened 2026-08-05
FM-738 Map OASys healthConditions field for supporting information

muhammad-elabdulla:
This [PR FM-738](https://dsdmoj.atlassian.net/browse/FM-738) is to map OASys healthConditions field for supporting information


davidatkinsuk (approved):
LGTM 👍

PR #5121, opened 2026-08-07
FM-1029 Make Tier Optional when building Person

muhammad-elabdulla:
This [PR FM-1029](https://dsdmoj.atlassian.net/browse/FM-1029) is to make tier optional when building Person

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/PersonSearchTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/PersonSearchTest.kt
@@ -6,7 +6,6 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.PersonStatus
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.api.model.PersonType
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.prisonsapi.AssignedLivingUnit
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.prisonsapi.InmateStatus
-import uk.gov.justice.digital.hmpps.approvedpremisesapi.common.transformer.toDto
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.factory.TierFactory
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.integration.givens.givenACase
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.integration.givens.givenAUser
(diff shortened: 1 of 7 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/OffenderDetailServiceTest.kt
```diff
@@ -329,13 +339,62 @@ class OffenderDetailServiceTest {
...
+      } returns mockPersonInfoResult1
+
+      val result = service.getPersonInfoResults(
+        crns = setOf(CRN1),
+        laoStrategy = CheckUserAccess(DELIUS_USERNAME),
+        includeTier = false,
+      )
+
+      assertThat(result).hasSize(1)
+      assertThat(result[0]).isEqualTo(mockPersonInfoResult1)
```
danhumphreys-moj, line 388: This test is called `If includeTier is false, don't fetch tier from OffenderService` - shouldn't we be explicitly checking that result doesn't contain tier here?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/PersonSearchTest.kt
```diff
@@ -162,7 +161,7 @@ class PersonSearchTest : IntegrationTestBase() {
                 genderIdentity = "This is a self described identity",
                 prisonName = "HMP Bristol",
                 isRestricted = false,
-                tier = tier.toDto(),
+                tier = null,
```
danhumphreys-moj, line 164: Wondering why tier is now null for this test in `PersonSearchTest` - the default is to include it and nothing else has changed for this test by the looks of it?

danhumphreys-moj (commented):
A couple of comments from me

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/PersonSearchTest.kt
muhammad-elabdulla: This is because I changed the `GET /people/search` API endpoint to call `getPersonInfoResult` with `includeTier` set to false

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/unit/service/OffenderDetailServiceTest.kt
muhammad-elabdulla: In this test, we are mocking the `getPersonSummaryInfoResults` function response to return tier as null. However, this scenario is already covered by the test `single crn with no limited access, includeTier false, does not fetch cases and returns Success with null tier`

danhumphreys-moj (approved):

PR #5125, opened 2026-08-10
Update Copilot Instructions

muhammad-elabdulla:
This [PR FM-1056](https://dsdmoj.atlassian.net/browse/FM-1056) is to update copilot instructions in the project

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,21 +1,521 @@
-# Controller (API) Style Guide
+# hmpps-approved-premises-api — Development Guide
 
-* @Operation annotations should not define an operationId. These are provided on-the-fly in OpenApiConfiguration
-* @PathVariable and @RequestParam annotations should not include redundant values (e.g. where the name of the variable being annotation matches the value defined in the annotation)
-* @RequestParam and @Parameter annotations should not include redundant 'required' entries in annotations because this can be inferred from the kotlin definition (e.g. if the type is suffixed with '?')
-* @Schema annotations should not include empty descriptions. If a description is empty it is redundant and should be removed
-* In the controller function argument list put each argument on a new line
+**What this repo is:** the shared Kotlin/Spring Boot backend service for the Community Accommodation (CAS) products
 
-# Post-change tidy up
+- CAS1 - Approved Premises
+- CAS2 - Short-Term Accommodation
+- CAS2 HDC - Home Detention Curfew
+- CAS3 - Temporary Accommodation
 
-Once you've finished generating code, you should tidy it up and validate as follows:
+It exposes REST APIs consumed by the CAS user interfaces, backed by
+PostgreSQL and emitting domain events to SQS. It is a large codebase built with Gradle on JDK 25.
 
-Run
+This guide documents the conventions and workflows actually used in this
+codebase: project orientation and common commands, the Controller, Service, Data
+(JPA entity / repository) and Transformer layers, request/response models,
+database migrations, domain events, feature flags, testing, and the lint/build
+gates. Follow these patterns when adding or changing code so new work stays
+consistent with what already exists.
 
-`./gradlew ktlintCheck`
+The typical request flow is:
 
-And
+`Controller` -> `Service` (business logic + `CasResult`) -> `Repository` (JPA)
+-> entities returned to the `Service` -> `Transformer` (entity -> API model) ->
+`Controller` returns a `ResponseEntity`.
 
-`./gradlew detekt`
+Service packages are scoped by product: **CAS1** (Approved Premises, `cas1`),
+**CAS2** (`cas2`, `cas2hdc`), **CAS3** (Temporary Accommodation,
+`cas3`). Shared/legacy code lives in the root `approvedpremisesapi` package.
+Prefer the service-scoped types when working in a CAS-specific area.
 
-and fix any issues raised by detekt
+---
+
+## Quick orientation (read first)
+
+* **Stack:** Kotlin (JDK 25 — see `.sdkmanrc`), Spring Boot, Spring Data JPA,
+  PostgreSQL + Flyway, Gradle (`./gradlew`).
+* **Package layout (ADR 0029):** `casx` (`cas1`, `cas2`, `cas2hdc`,
+  `cas3`) each contain `controller`, `dto`, `service`, `jobs/{migration,seed}`,
+  `transformer`, and an entities+repositories package. **Note the entity package
+  differs by service:** CAS1 uses `cas1/entity/**`, whereas CAS2/CAS2hdc/CAS3 use
+  `<service>/jpa/entity/**` — match the convention already used in the module you
+  are editing. Shared/legacy code lives in the root `approvedpremisesapi`
+  package (`config`, `common`, `client`, `cmd`, plus legacy `api`, `jpa`,
+  `service`, `transformer`). New service-specific code belongs under `casx/**`;
+  add sub-packages by domain (reports, applications, assessment, ...).
+* **Request flow:** `Controller` -> `Service` (returns `CasResult<T>`) ->
+  `Repository` (JPA) -> entity -> `Transformer` (entity -> DTO/API model) ->
+  `Controller` returns `ResponseEntity`.
+* **Key files:** `common/results/CasResult.kt`, `util/EntityUtils.kt` (result
+  unwrapping), `config/OAuth2ResourceServerSecurityConfiguration.kt` (endpoint
+  security), `integration/IntegrationTestBase.kt` (integration test base).
+* **Reference docs:** architecture decisions in `doc/architecture/decisions/**`
+  and task guides in `doc/how-to/**` (e.g. `add_new_endpoint.md`,
+  `define_an_api.md`, `best-practice-jpa-entities.md`,
+  `add-a-feature-flag.md`, `modifying_domain_event_schemas.md`). Specialised
+  sub-agents live in `.github/agents/**`.
+
+### Common commands
+
+```bash
+./gradlew ktlintFormat          # auto-fix formatting
+./gradlew ktlintCheck detekt    # lint + static analysis (must pass, no new baseline)
+./gradlew compileKotlin compileTestKotlin -q   # verify structural/cross-file changes
+
+script/test_database            # start test dependencies (Postgres etc.) — run first
+./gradlew test --tests '*Cas1PremisesTest'     # run a focused test locally
+./script/test                   # full suite (slow; prefer CI for the whole run)
+```
+
+---
+
+## Controller (API) Conventions
+
+### General annotation rules
+
+* `@Operation` annotations should not define an `operationId`. These are provided
+  on-the-fly in `OpenApiConfiguration`.
+* `@PathVariable` and `@RequestParam` annotations should not include redundant
+  values (e.g. where the name of the variable being annotated matches the value
+  defined in the annotation).
+* `@RequestParam` and `@Parameter` annotations should not include redundant
+  `required` entries because this can be inferred from the Kotlin definition (e.g.
+  if the type is suffixed with `?`).
+* `@Schema` annotations should not include empty descriptions. If a description is
+  empty it is redundant and should be removed.
+* In the controller function argument list put each argument on a new line.
+
+### Controller declaration
+
+* Prefer the hand-written controller style using the per-service meta-annotation
+  — `@Cas1Controller`, `@Cas3Controller`, etc. These are `@Target(CLASS)`
+  meta-annotations that combine `@RestController` and
+  `@RequestMapping("${api.base-path:}/<service>", produces = [JSON, PROBLEM_JSON])`.
+  See `cas1/controller/Cas1Controller.kt`.
+* The older style implements a generated OpenAPI `*Delegate` interface (e.g.
+  `Cas1ReferenceDataController : ReferenceDataCas1Delegate`). Both coexist; new
+  hand-written controllers should use the meta-annotation style.
+* Use constructor injection (`private val`) for services and transformers.
+* Add a class-level `@Tag(name = "CAS1 Premises")` to group endpoints in Swagger.
+
+### Endpoint methods
+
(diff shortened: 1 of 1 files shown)


.github/copilot-instructions.md
```diff
@@ -1,21 +1,520 @@
-# Controller (API) Style Guide
+# hmpps-approved-premises-api — Development Guide
 
-* @Operation annotations should not define an operationId. These are provided on-the-fly in OpenApiConfiguration
-* @PathVariable and @RequestParam annotations should not include redundant values (e.g. where the name of the variable being annotation matches the value defined in the annotation)
-* @RequestParam and @Parameter annotations should not include redundant 'required' entries in annotations because this can be inferred from the kotlin definition (e.g. if the type is suffixed with '?')
-* @Schema annotations should not include empty descriptions. If a description is empty it is redundant and should be removed
-* In the controller function argument list put each argument on a new line
+**What this repo is:** the shared Kotlin/Spring Boot backend service for the Community Accommodation (CAS) products — Approved
+Premises (CAS1), Short-Term Accommodation (CAS2) and Temporary Accommodation
```
davidatkinsuk: I think we should call out the two CAS2 variants

* CAS1 - Approved Premises
* CAS2 - Short-Term Accommodation
* CAS2 HDC - Home Detention Curfew
* CAS3 - Temporary Accommodation

.github/copilot-instructions.md
```diff
@@ -1,21 +1,520 @@
...
+gates. Follow these patterns when adding or changing code so new work stays
+consistent with what already exists.
 
-Once you've finished generating code, you should tidy it up and validate as follows:
+The typical request flow is:
 
-Run
+`Controller` -> `Service` (business logic + `CasResult`) -> `Repository` (JPA)
+-> entities returned to the `Service` -> `Transformer` (entity -> API model) ->
+`Controller` returns a `ResponseEntity`.
```
davidatkinsuk, line 24: I wonder if this could be a little clearer e.g.

* API Requests are received by `Controllers`
* `Controllers` delegate requests to `Services` that handle business logic, including interacting with `Repositorys` to query for and persist `Entities`
* The `Services` should return `CasResult`, which wraps responses that are then mapped into API models using `Transformers`
* `Controllers` return `ResponseEntity`

.github/copilot-instructions.md
```diff
@@ -1,21 +1,520 @@
...
 
-`./gradlew ktlintCheck`
+Service packages are scoped by product: **CAS1** (Approved Premises, `cas1`),
+**CAS2** (`cas2`, `cas2v2`, `cas2hdc`), **CAS3** (Temporary Accommodation,
```
davidatkinsuk: It should really just be `cas2` and `cas2hdc`, but we still have some tidying up to do

.github/copilot-instructions.md
```diff
@@ -1,21 +1,520 @@
...
 
-* @Operation annotations should not define an operationId. These are provided on-the-fly in OpenApiConfiguration
-* @PathVariable and @RequestParam annotations should not include redundant values (e.g. where the name of the variable being annotation matches the value defined in the annotation)
-* @RequestParam and @Parameter annotations should not include redundant 'required' entries in annotations because this can be inferred from the kotlin definition (e.g. if the type is suffixed with '?')
-* @Schema annotations should not include empty descriptions. If a description is empty it is redundant and should be removed
-* In the controller function argument list put each argument on a new line
+**What this repo is:** the shared Kotlin/Spring Boot backend service for the Community Accommodation (CAS) products — Approved
+Premises (CAS1), Short-Term Accommodation (CAS2) and Temporary Accommodation
+(CAS3). It exposes REST APIs consumed by the CAS user interfaces, backed by
+PostgreSQL and emitting domain events to SQS. It is a large codebase built with Gradle on JDK 25.
```
davidatkinsuk, line 11: I don't think copilot really needs to know this but it may be helpful to add the following for context...

"The project is named `approved-premises-api` because it was originally developed to the support the CAS1 (approved-premises) UI"

.github/copilot-instructions.md
```diff
@@ -1,21 +1,520 @@
...
 
-And
+---
 
-`./gradlew detekt`
+## Quick orientation (read first)
 
-and fix any issues raised by detekt
+* **Stack:** Kotlin (JDK 25 — see `.sdkmanrc`), Spring Boot, Spring Data JPA,
+  PostgreSQL + Flyway, WebFlux, Gradle (`./gradlew`).
```
davidatkinsuk: I think we can drop WebFlux here

.github/copilot-instructions.md
```diff
@@ -1,21 +1,520 @@
...
+* **Package layout (ADR 0029):** `casx` (`cas1`, `cas2`, `cas2v2`, `cas2hdc`,
+  `cas3`) each contain `controller`, `dto`, `service`, `jobs/{migration,seed}`,
```
davidatkinsuk: should drop `cas2v2` from this list

.github/copilot-instructions.md
```diff
@@ -1,21 +1,520 @@
...
+  `transformer`, and an entities+repositories package. **Note the entity package
+  differs by service:** CAS1 uses `cas1/entity/**`, whereas CAS2/CAS2hdc/CAS3 use
+  `<service>/jpa/entity/**` — match the convention already used in the module you
+  are editing. Shared/legacy code lives in the root `approvedpremisesapi`
```
davidatkinsuk, line 42: They should all be `<service>/entity` (as per ADR 029). Could you maybe fix this as part of this PR and then remove this text?

.github/copilot-instructions.md
```diff
@@ -1,21 +1,520 @@
...
+  package (`config`, `common`, `client`, `cmd`, plus legacy `api`, `jpa`,
+  `service`, `transformer`). New service-specific code belongs under `casx/**`;
+  add sub-packages by domain (reports, applications, assessment, ...).
+* **Request flow:** `Controller` -> `Service` (returns `CasResult<T>`) ->
+  `Repository` (JPA) -> entity -> `Transformer` (entity -> DTO/API model) ->
+  `Controller` returns `ResponseEntity`.
```
davidatkinsuk, line 48: this repeats the text at the top. Maybe put the proposed text here instead?

.github/copilot-instructions.md
```diff
@@ -1,21 +1,520 @@
...
+* **Key files:** `common/results/CasResult.kt`, `util/EntityUtils.kt` (result
+  unwrapping), `config/OAuth2ResourceServerSecurityConfiguration.kt` (endpoint
+  security), `integration/IntegrationTestBase.kt` (integration test base).
```
davidatkinsuk, line 51: this list seems a bit arbitrary. I'm not sure if copilot can do without it?

.github/copilot-instructions.md
```diff
@@ -1,21 +1,520 @@
...
+```
+
+---
+
+## Controller (API) Conventions
+
+### General annotation rules
+
+* `@Operation` annotations should not define an `operationId`. These are provided
+  on-the-fly in `OpenApiConfiguration`.
```
davidatkinsuk, line 77: accidental line break here?

.github/copilot-instructions.md
```diff
@@ -1,21 +1,520 @@
...
+* `@PathVariable` and `@RequestParam` annotations should not include redundant
+  values (e.g. where the name of the variable being annotated matches the value
```
davidatkinsuk, line 79: accidental line break here?

(please check the remainder of the file as i think some wrap arounds have become line breaks)

.github/copilot-instructions.md
```diff
@@ -1,21 +1,520 @@
...
+
+* Endpoint security is enforced in code and covered by `AuthTest` — **not** with
+  `@PreAuthorize`/`ResourceSecurityTest` on internal endpoints.
+* Enforce permissions with the user-access service, e.g.
+  `userAccessService.ensureCurrentUserHasPermission(UserPermission.CAS1_PREMISES_VIEW)`
+  (CAS1 uses `Cas1UserAccessService`; CAS3 uses `UserAccessService`).
+* External controllers (`controller/external/**`) are the exception and do use
+  `@PreAuthorize("hasRole(...)")`.
+* Every new endpoint must require an appropriate role/permission and have auth
+  tests.
```
davidatkinsuk, line 125: I think this could be a little clearer and also highlight the varying (and inconsistent!) models we use

* CAS1 uses Roles stored in the database for a given user. Roles have associated permissions. Permissions are checked in code programmatically (checking permissions should be favoured over checking roles). For an example see `Cas1UserAccessService`.
* CAS2 and CAS2 HDC use Roles defined on the given JWT, managed via rules URL mappings defined in `OAuth2ResourceServerSecurityConfiguration`
* CAS3 uses Roles stored in the database for a given user. Roles are checked in code programatically. For an example see `UserAccessService`
* The only exception to this rule is the `external` endpoints provided to support SAS. These use @PreAuthorize("hasRole(...)") to check roles

.github/copilot-instructions.md
```diff
@@ -1,21 +1,520 @@
...
+
+### Current user and result handling
+
+* Get the request user with `userService.getUserForRequest()`.
+* Services return `CasResult<T>` (see below). Unwrap it in the controller with
+  `extractEntityFromCasResult(result)` (from `util/EntityUtils.kt`), which returns
+  the success entity or throws the appropriate problem. Use
+  `ensureEntityFromCasResultIsSuccess(result)` when you only need to assert success
+  without an entity.
```
davidatkinsuk, line 134: This sections mixes two different concepts. 

Maybe just rename to `resultHandling` and remove the user bit?

.github/copilot-instructions.md
```diff
@@ -1,21 +1,520 @@
...
+  (`@Inheritance`).
+* **Always override `equals`, `hashCode`, and `toString` manually.** Data-class
+  defaults cause eager loading and infinite loops across relationships:
+  * `equals`/`hashCode`: base on the `id` plus a few scalar (non-collection)
+    fields — never include lazy collections.
+  * `toString`: minimal, e.g. `"Cas3BookingEntity:$id"`.
+* Primary key: `@Id val id: UUID`, assigned by the application (no
+  `@GeneratedValue`). Prefer `val` for immutable fields; use `var` only where the
+  field is genuinely mutable.
+
```
davidatkinsuk, line 252: Maybe add a final bullet point saying 'See `CaseEntity` for an example'

.github/copilot-instructions.md
```diff
@@ -1,21 +1,520 @@
...
+* Enums: `@Enumerated(EnumType.STRING)`.
+* JSON columns: `@Type(JsonType::class)` on a `String` field.
+* Custom conversions: `@Convert(converter = XConverter::class)`.
+* Dates/times: `LocalDate` for dates, `OffsetDateTime` for timestamps with
+  offset, `Instant` for UTC instants.
+* Optimistic locking: `@Version var version: Long = 1` on contended entities.
+* Auditing fields are explicit (`val createdAt: OffsetDateTime`,
+  `var lastUpdatedAt: OffsetDateTime?`) and set by application logic — there is no
+  `@CreationTimestamp`/`@UpdateTimestamp`. Soft delete via a nullable
+  `deletedAt`/`isActive` filtered in queries.
```
davidatkinsuk: I'd remove this, I think we have a mix of approaches in code and i'm not sure there is one size fits all. Note that CaseEntity uses `@UpdateTimestamp`

.github/copilot-instructions.md
```diff
@@ -1,21 +1,520 @@
...
+* Paginate large result sets with `Pageable`, returning `Page<T>` (needs total
+  count) or `Slice<T>` (lighter, no count).
+* For read-only projections, declare a projection interface (getter methods
+  matching column aliases) and return it from the query.
+* Use `@Lock(LockModeType.PESSIMISTIC_WRITE)` on a dedicated locking method for
+  pessimistic locking.
+
+### Service-scoped vs shared entities
+
+* New, service-specific entities live under `<service>/jpa/entity/**` and are
```
davidatkinsuk: drop `jpa` from the path

.github/copilot-instructions.md
```diff
@@ -1,21 +1,520 @@
...
+
+**Frameworks:**
+
+* **JUnit 5** — `@Test`, `@Nested`, `@ParameterizedTest` with
+  `@MethodSource`/`@EnumSource`/`@CsvSource`.
+* **AssertJ** — `assertThat(...)`, preferred over JUnit `assertEquals` for new code.
+* **MockK** — `mockk()`, `every { } returns`, `verify { }`, `confirmVerified`,
+  `just Runs`.
+* **kfactory** — `Factory<T>` builders under `src/test/.../factory`.
+* **Testcontainers** — real PostgreSQL for integration tests.
```
davidatkinsuk: i don't think we use testcontainers (at least for our integration tests?)

.github/copilot-instructions.md
```diff
@@ -1,21 +1,520 @@
...
+* Tests mirror the main source package structure.
```
davidatkinsuk: this is only really true for unit tests

davidatkinsuk (approved):
Thanks for doing this. I've added some comments that are hopefully of use. It will be interesting to see how accepting co-pilot reviews are of exising code that break some of these rules (which probably _don't_ need highlighting, unless we're rewriting that code)

.github/copilot-instructions.md
muhammad-elabdulla: The original one is clear instructions for Copilot

.github/copilot-instructions.md
muhammad-elabdulla: I don’t think this will make any difference to Copilot

muhammad-elabdulla:
I don’t think copilot reviews will be highly affected changing the old code, as the instructions highlight that there are different approaches. Let’s see how it works and we can improve it as needed

.github/copilot-instructions.md
davidatkinsuk: if copilot can also deal with the new verison i'd prefer that because then it's also more human readable?

PR #5128, opened 2026-08-11
FM-1032 Add additional box content for CAS 3.

danhumphreys-moj:
Ref: https://dsdmoj.atlassian.net/browse/FM-1032

**Changes:**
- Extends `Cas3SuitableApplication` with submitted metadata, rejection reason, provisional-offer date, and previous bookings (incl. cancellation info).
- Updates `Cas3ApplicationService.getSuitableApplicationByCrn` to fetch all bookings for an application (instead of only the latest booking).
- Updates CAS3 transformer and test suites (unit + integration) to accommodate the new shape and booking list input.


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/transformer/Cas3ApplicationTransformer.kt
```diff
@@ -26,16 +28,25 @@ class Cas3ApplicationTransformer(
...
+      assessmentStatus = application.getLatestAssessment()?.deriveAssessmentStatus(),
+      bookingStatus = firstBooking?.status,
+      bookingProvisionalOfferSentDate = null,
+      previousBookings = bookings?.drop(1)?.map { transformToPreviousBookingDto(it) },
+      premises = firstBooking?.premises?.let {
+        transformToCas3PremisesSummary(firstBooking)
+      },
+      uiUrl = cas3ApplicationFullUrlTemplate.replace("#applicationId", application.id.toString()),
+    )
+  }
```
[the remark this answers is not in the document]
danhumphreys-moj: (a) is behaviour we want (b) has been resolved and the date is only set if the booking status is provisional and (c) is behaviour we want since we just want to return the status of the most recent booking regardless of whether it is null or not

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/service/Cas3ApplicationService.kt
```diff
@@ -81,7 +83,7 @@ class Cas3ApplicationService(
     ?.let {
       transformer.transformToCas3SuitableApplication(
         application = it,
-        booking = cas3BookingService.getLatestBooking(it.id),
+        bookings = cas3BookingRepository.findAllCas3BookingEntity(it.id),
       )
```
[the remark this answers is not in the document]
danhumphreys-moj: No action required - the patterns followed in this PR are consistent with those used already used.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/model/Cas3SuitableApplication.kt
```diff
@@ -24,3 +29,13 @@ data class Cas3ExternalPremisesDto(
   val town: String?,
   val postcode: String,
 )
+
+data class PreviousBookingDto(
```
danhumphreys-moj: I'm surprised that more meta data isn't required regarding previous bookings but this structure as coded is in line with the prototype.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/model/Cas3SuitableApplication.kt
```diff
@@ -9,8 +9,13 @@ import java.util.UUID
 data class Cas3SuitableApplication(
   val id: UUID,
   val applicationStatus: ApplicationStatus,
+  val applicationSubmittedDate: LocalDate?,
```
danhumphreys-moj, line 13: It would be good to have some kind of contract test between CAS <-> SAS APIs to ensure we don't break production - unless there are some SAS end to end tests that check we don't?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/service/Cas3ApplicationService.kt
```diff
@@ -81,7 +83,7 @@ class Cas3ApplicationService(
     ?.let {
       transformer.transformToCas3SuitableApplication(
         application = it,
-        booking = cas3BookingService.getLatestBooking(it.id),
+        bookings = cas3BookingRepository.findAllCas3BookingEntity(it.id),
```
danhumphreys-moj: Existing method `getLatestBooking` uses created descending sort order, `findAllCasBookingEntity` also sorts by createdAt descending so the first one of these bookings will be the equivalent of the latest booking - the remainder will be previous bookings,

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/model/Cas3SuitableApplication.kt
danhumphreys-moj, line 26: Booking start and end dates probably shouldn't be in this PremisesDto object - but leaving as is since SAS API expects them here.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/service/Cas3ApplicationService.kt
```diff
@@ -81,7 +83,7 @@ class Cas3ApplicationService(
     ?.let {
       transformer.transformToCas3SuitableApplication(
         application = it,
-        booking = cas3BookingService.getLatestBooking(it.id),
```
muhammad-elabdulla, line 84: getLatestBooking is now referenced only by tests, so it can be removed from the code

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/model/Cas3SuitableApplication.kt
```diff
@@ -9,8 +9,13 @@ import java.util.UUID
...
+  val applicationSubmittedByName: String?,
+  val applicationRejectedReason: String?,
   val assessmentStatus: TemporaryAccommodationAssessmentStatus?,
   val bookingStatus: Cas3BookingStatus?,
```
muhammad-elabdulla, line 17: The ticket specifies that bookingStartDate and bookingEndDate properties that will be included

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/transformer/Cas3ApplicationTransformer.kt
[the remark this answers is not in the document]
muhammad-elabdulla: I think using copilots suggestion `val latestAssessment = application.getLatestAssessment()` would reduce the two calls to one.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/model/Cas3SuitableApplication.kt
```diff
@@ -9,8 +9,13 @@ import java.util.UUID
 data class Cas3SuitableApplication(
```
davidatkinsuk, line 10: there is a similar ticket for CAS1 here - https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/5072/changes#diff-7c90092af02730bfbcf2f7494fe6d6861c71061e2a303b89c6d5af16bdba3b5d

In that ticket the top level response (`Cas1SuitableApplication`) was broken down into smaller data types. I'm not sure if it's worth considering doing that here?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/model/Cas3SuitableApplication.kt
```diff
@@ -9,8 +9,13 @@ import java.util.UUID
...
   val id: UUID,
   val applicationStatus: ApplicationStatus,
+  val applicationSubmittedDate: LocalDate?,
+  val applicationSubmittedByName: String?,
```
davidatkinsuk: can we please use `StaffDto` here (and anywhere else providing user info). The rationale is ideally SAS would be loading the name given the username, but `StaffDto` provides some flexibility. See https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/5072/changes#diff-7c90092af02730bfbcf2f7494fe6d6861c71061e2a303b89c6d5af16bdba3b5d

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/model/Cas3SuitableApplication.kt
```diff
@@ -24,3 +29,13 @@ data class Cas3ExternalPremisesDto(
   val town: String?,
   val postcode: String,
 )
+
+data class PreviousBookingDto(
+  val bookingStatus: Cas3BookingStatus?,
+  val cancellation: PreviousBookingCancellationDto?,
+)
+
+data class PreviousBookingCancellationDto(
```
davidatkinsuk: maybe just call this `Cas3ExternalCancellationDto`

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/model/Cas3SuitableApplication.kt
davidatkinsuk: maybe call this `Cas3ExternalPreviousBookingDto`?

I don't love the `Cas3External` prefix but i think it would be good to separate these types from our UI facing APIs?

See https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/5072/changes#diff-7c90092af02730bfbcf2f7494fe6d6861c71061e2a303b89c6d5af16bdba3b5d

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/transformer/Cas3ApplicationTransformer.kt
```diff
@@ -26,16 +29,25 @@ class Cas3ApplicationTransformer(
...
-  )
+  fun transformToCas3SuitableApplication(application: TemporaryAccommodationApplicationEntity, bookings: List<Cas3BookingEntity>?): Cas3SuitableApplication {
+    val firstBooking = bookings?.firstOrNull()
+
+    return Cas3SuitableApplication(
+      id = application.id,
+      applicationStatus = application.getStatus(),
+      applicationSubmittedDate = application.submittedAt?.toLocalDate(),
+      applicationSubmittedByName = application.createdByUser.name,
+      applicationRejectedReason = application.getLatestAssessment()?.referralRejectionReason?.name,
```
davidatkinsuk: note this change https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/5134

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/transformer/Cas3ApplicationTransformer.kt
```diff
@@ -26,16 +29,25 @@ class Cas3ApplicationTransformer(
...
+      assessmentStatus = application.getLatestAssessment()?.deriveAssessmentStatus(),
```
davidatkinsuk: could pull latest assesssment into a variable at the top of this function?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/integration/external/Cas3ExternalApplicationsTest.kt
```diff
@@ -143,8 +153,13 @@ class Cas3ExternalApplicationsTest : IntegrationTestBase() {
           val suitableApplication = Cas3SuitableApplication(
             id = application.id,
             applicationStatus = ApplicationStatus.submitted,
+            applicationSubmittedDate = application.submittedAt?.toLocalDate(),
+            applicationSubmittedByName = application.createdByUser.name,
+            applicationRejectedReason = null,
             assessmentStatus = null,
             bookingStatus = Cas3BookingStatus.confirmed,
+            bookingProvisionalOfferSentDate = null,
+            previousBookings = emptyList(),
```
muhammad-elabdulla, line 235: Can you add an integration test to cover a provisional booking, rejected assessment reason and previous bookings with cancellations?

muhammad-elabdulla (commented):
Left some comments

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/transformer/Cas3ApplicationTransformer.kt
```diff
@@ -26,16 +29,25 @@ class Cas3ApplicationTransformer(
...
+      bookingStatus = firstBooking?.status,
+      bookingProvisionalOfferSentDate = if (firstBooking?.status == Cas3BookingStatus.provisional) firstBooking.createdAt.toLocalDate() else null,
+      previousBookings = bookings?.drop(1)?.map { transformToPreviousBookingDto(it) },
+      premises = firstBooking?.premises?.let {
```
davidatkinsuk: could this be named something more reflective of what it is? Maybe if it was container by some data class it could make it clearer? I believe this is the 'initial booked premises'?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/model/Cas3SuitableApplication.kt
davidatkinsuk: Not sure if we can get away with renaming this Dto possibly? I think that'll only break compiled code in SAS e.g. `Cas3ExternalInitialBookingDto`?

davidatkinsuk (commented):
Some comments from me

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/transformer/Cas3ApplicationTransformer.kt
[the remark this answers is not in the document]
danhumphreys-moj: Added `val latestAssessment = application.getLatestAssessment()`

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/service/Cas3ApplicationService.kt
danhumphreys-moj: Renamed `findAllCas3BookingEntity` to `findAllBookingEntityCreatedAtDesc`.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/model/Cas3SuitableApplication.kt
danhumphreys-moj: Renamed `Cas3ExternalPremisesDto` to `Cas3ExternalLatestBookingDto`.  It is the latest (i.e. most recent) booking associated with this application.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/service/Cas3ApplicationService.kt
danhumphreys-moj: Good spot!  Removed `getLatestBooking()` and associated unit test class.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/model/Cas3SuitableApplication.kt
danhumphreys-moj: I've removed these fields from the ticket @muhammad-elabdulla.  They can be retrieved from `Cas3ExternalLatestBookingDto` (previously `Cas3ExternalPremisesDto`).

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/model/Cas3SuitableApplication.kt
danhumphreys-moj: Renamed `PreviousBookingDto` to `Cas3ExternalPreviousBookingDto`.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/transformer/Cas3ApplicationTransformer.kt
danhumphreys-moj: Have added `val latestAssessment = application.getLatestAssessment()` now.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/transformer/Cas3ApplicationTransformer.kt
danhumphreys-moj: To avoid confusion I have renamed `firstBooking` variable to `latestBooking` since this is the first one in a list of bookings ordered by created at descending.

```
data class Cas3SuitableApplication(
  val id: UUID,
  val applicationStatus: ApplicationStatus,
  val applicationSubmittedDate: LocalDate?,
  val applicationSubmittedByName: String?,
  val applicationRejectedReason: String?,
  val assessmentStatus: TemporaryAccommodationAssessmentStatus?,
  val bookingStatus: Cas3BookingStatus?,
  val bookingProvisionalOfferSentDate: LocalDate?,
  val previousBookings: List<Cas3ExternalPreviousBookingDto>?,
  val premises: Cas3ExternalLatestBookingDto?,
  val uiUrl: String,
)
```

We now have the latest booking as `Cas3ExternalLatestBookingDto` and a list of previous bookings using `Cas3ExternalPreviousBookingDto`.  I'm not sure I can change `premises` to `latestBooking` without breaking SAS production?  Which is unfortunate.

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/integration/external/Cas3ExternalApplicationsTest.kt
danhumphreys-moj: Additional integration tests added.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/model/Cas3SuitableApplication.kt
danhumphreys-moj: Have changed to use `Cas3StaffDto`.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/transformer/Cas3ApplicationTransformer.kt
davidatkinsuk: thanks for clearing this up. It may be worth adding @Schema(description) onto premises to explain what it is? or add a new better named field `latestBookingPremises`? and deprecate the existing one?

danhumphreys-moj: Have added a schema description to this property.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/model/Cas3SuitableApplication.kt
danhumphreys-moj: Ahh yes I can see they have followed a pattern of deprecating the existing type and properties and added a new one with a better structure.  Would following this approach here have meant deprecating the controller and related service functions and creating new ones?

danhumphreys-moj:
@muhammad-elabdulla please can you re-review this?  It would be good to agree a strategy if we are do make these kind of changes again in the future - part of the refinement of ticket FM-1032 could have been to deprecate the existing return type and create a new one - if the structure of the new type diverges significantly from the existing type.  @davidatkinsuk

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/model/Cas3SuitableApplication.kt
davidatkinsuk: I think it would be a good idea - but maybe could be a follow up PR because that should be quite a simple change in isolation (and easy to review, presumably?).

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/transformer/Cas3ApplicationTransformer.kt
davidatkinsuk: I feel sorry for the github account `schema` because i've tagged them a few times trying to reference an annotation :D

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/model/Cas3SuitableApplication.kt
danhumphreys-moj: I'll have a chat with @simonwilson4moj, I think there might be some follow on work from this PR since I think there should be some more data returned about the previous bookings for an application such as booking dates, premises information etc rather than just booking status and a cancellation reason (if it exists).  We can bring refactoring `Cas3Suitable` application into scope then - and if not we can raise as a dedicated ticket.

davidatkinsuk: sounds good!

muhammad-elabdulla (approved):
LGTM

PR #5132, opened 2026-08-12
FM-1038 Make ensureCaseExist thread safe

vbala-moj:
**For a detailed explanation, please refer to the comments section.**
https://dsdmoj.atlassian.net/browse/FM-1038

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/service/CaseService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/service/CaseService.kt
@@ -1,9 +1,9 @@
 package uk.gov.justice.digital.hmpps.approvedpremisesapi.common.service
 
+import org.hibernate.exception.ConstraintViolationException
 import org.slf4j.LoggerFactory
+import org.springframework.dao.DataIntegrityViolationException
 import org.springframework.stereotype.Service
-import org.springframework.transaction.annotation.Propagation
-import org.springframework.transaction.annotation.Transactional
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.ApDeliusContextApiClient
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.ClientResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.HMPPSTierApiClient
@@ -11,6 +11,7 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.deliuscontext.Cas
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.common.dto.CaseDto
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.common.entity.CaseEntity
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.common.entity.CaseRepository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.common.entity.model.CaseTiers
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.common.entity.model.Tier
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.common.entity.model.TierVersion
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.common.jobs.migration.BackfillCasesJob
@@ -20,8 +21,6 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.FeatureFlagServi
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.FeatureFlagService.Companion.FEATURE_FLAG_INCLUDE_TIER_V3
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.FeatureFlagService.Companion.FEATURE_FLAG_USE_TIER_V3
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.service.SentryService
-import java.time.OffsetDateTime
-import java.util.UUID
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.hmppstier.Tier as UpstreamTier
 
 /**
@@ -46,8 +45,13 @@ class CaseService(
   private val hmppsTierApiClient: HMPPSTierApiClient,
   private val featureFlagService: FeatureFlagService,
   private val sentryService: SentryService,
+  private val casePersistenceService: CasePersistenceService,
 ) {
   private val log = LoggerFactory.getLogger(this::class.java)
+
+  companion object {
+    private const val CASES_CRN_UNIQUE_CONSTRAINT = "cas1_offenders_crn"
+  }
 
   private val includeTierV3: Boolean
     get() = featureFlagService.getBooleanFlag(FEATURE_FLAG_INCLUDE_TIER_V3)
@@ -55,17 +59,36 @@ class CaseService(
   private val useTierV3: Boolean
     get() = featureFlagService.getBooleanFlag(FEATURE_FLAG_USE_TIER_V3)
 
-  @Transactional(propagation = Propagation.REQUIRES_NEW)
+  /**
+   * Each persistence operation runs in its own transaction so concurrent
+   * case creation can be handled safely. If another request inserts the
+   * same case first, the database unique constraint rejects the duplicate
+   * insert, after which the existing row is read and returned.
+   */
   fun ensureCaseExists(crn: String): CaseDto {
     val normalizedCrn = crn.uppercase()
     val caseSummary = getCaseSummary(normalizedCrn)
     val tiers = fetchAvailableTiers(normalizedCrn)
 
-    val caseEntity = caseRepository.findByCrn(caseSummary.crn.uppercase())
-      ?.updateFrom(caseSummary, tiers)
-      ?: newCaseEntity(caseSummary, tiers)
+    val existingCase = casePersistenceService.updateIfExist(normalizedCrn, caseSummary, tiers)
+    if (existingCase != null) {
+      return existingCase.toDto()
+    } else {
+      return try {
+        casePersistenceService.createCase(caseSummary, tiers).toDto()
+      } catch (e: DataIntegrityViolationException) {
+        if (!isCrnDuplicateException(e)) {
+          throw e
+        }
 
-    return caseRepository.saveAndFlush(caseEntity).toDto()
+        log.warn(
+          "Failed to create case for CRN $normalizedCrn; " +
+            "case may have been created concurrently.",
+          e,
+        )
+        casePersistenceService.getCase(normalizedCrn)?.toDto() ?: throw e
+      }
+    }
   }
 
   fun reviseTier(crn: String): Boolean {
@@ -113,39 +136,10 @@ class CaseService(
     return cases
   }
 
-  private data class CaseTiers(
-    val v2: Tier?,
-    val v3: Tier?,
-  )
-
   private fun fetchAvailableTiers(crn: String) = CaseTiers(
     v2 = fetchTierOrNull(crn, TierVersion.V2),
     v3 = if (includeTierV3) fetchTierOrNull(crn, TierVersion.V3) else null,
   )
-
-  private fun CaseEntity.updateFrom(caseSummary: CaseSummary, tiers: CaseTiers): CaseEntity = apply {
-    name = caseSummary.buildName()
-    nomsNumber = caseSummary.nomsId
-    if (tiers.v2 != null) {
-      tierV2 = tiers.v2
-    }
-    if (tiers.v3 != null) {
-      tierV3 = tiers.v3
-    }
-  }
-
-  private fun newCaseEntity(caseSummary: CaseSummary, tiers: CaseTiers) = CaseEntity(
-    id = UUID.randomUUID(),
-    crn = caseSummary.crn.uppercase(),
-    createdAt = OffsetDateTime.now(),
-    lastUpdatedAt = OffsetDateTime.now(),
-    name = caseSummary.buildName(),
-    nomsNumber = caseSummary.nomsId,
-    tierV2 = tiers.v2,
-    tierV3 = tiers.v3,
-  )
-
-  private fun CaseSummary.buildName() = "${name.forename} ${name.surname}".uppercase().trim()
 
   private fun CaseEntity.toDto() = CaseDto(
     crn = crn,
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/service/CaseService.kt
```diff
@@ -46,26 +43,44 @@ class CaseService(
...
-      ?.updateFrom(caseSummary, tiers)
-      ?: newCaseEntity(caseSummary, tiers)
-
-    return caseRepository.saveAndFlush(caseEntity).toDto()
+    ensureCasePersistenceService.findAndUpdate(normalizedCrn, caseSummary, tiers)
+      ?.let { return it.toDto() }
+
+    return try {
+      this@CaseService.ensureCasePersistenceService.createCase(caseSummary, tiers).toDto()
+    } catch (e: DataIntegrityViolationException) {
```
vbala-moj: `DataIntegrityViolationException` is a broad exception type, while we are only interested in handling duplicate-key constraint violations. I am currently working on identifying how to reliably determine whether the exception is specifically caused by the duplicate CRN constraint, and will rethrow any other types of `DataIntegrityViolationException`.

```
Caused by: org.postgresql.util.PSQLException: ERROR: duplicate key value violates unique constraint "cas1_offenders_crn"
  Detail: Key (crn)=(SQJNHJ) already exists.
```

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/service/CaseService.kt
```diff
@@ -46,26 +43,44 @@ class CaseService(
...
+      log.warn(
+        "Failed to create case for CRN $normalizedCrn; " +
+          "case may have been created concurrently.",
+        e,
+      )
+      ensureCasePersistenceService.getCase(normalizedCrn)?.toDto() ?: throw e
+    }
```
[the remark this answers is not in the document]
vbala-moj: Already working on it.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/entity/CaseEntity.kt
```diff
@@ -11,8 +11,10 @@ import org.hibernate.type.SqlTypes
 import org.springframework.data.jpa.repository.JpaRepository
 import org.springframework.data.jpa.repository.Query
 import org.springframework.stereotype.Repository
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.client.deliuscontext.CaseSummary
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.common.dto.BackfillCaseSummaryMigrationDto
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.common.entity.model.Tier
+import uk.gov.justice.digital.hmpps.approvedpremisesapi.common.service.CaseService.CaseTiers
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.jpa.entity.asHibernateProxy
```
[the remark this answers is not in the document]
vbala-moj: What do you think @davidatkinsuk ?

src/test/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/integration/cas1/Cas1ApplicationTest.kt
```diff
@@ -2017,6 +2019,77 @@ class Cas1ApplicationTest : IntegrationTestBase() {
...
+          val barrier = CyclicBarrier(3) // 2 for child threads and 1 for the main thread
+
+          assertThat(caseRepository.findByCrn(crn)).isNull()
+
+          try {
+            val responses = List(2) {
+              executor.submit {
+                barrier.await()
+                webTestClient.post()
+                  .uri("/cas1/applications/create")
```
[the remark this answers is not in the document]
vbala-moj: Will add some timeout.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/entity/CaseEntity.kt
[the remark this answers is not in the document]
vbala-moj: I’ve added `CaseTiers` alongside `Tiers`. I’ll push this change together with the other review comments. Let me know if you have any other suggestions.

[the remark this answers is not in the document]
vbala-moj: Done

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/service/CaseService.kt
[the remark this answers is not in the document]
vbala-moj: I’ve used this simple approach instead, and it works. Let me know if you think it’s better to follow Copilot’s suggestion.

```
  private fun isCrnDuplicateException(
    e: DataIntegrityViolationException,
  ): Boolean = e.cause?.message?.contains(CASES_CRN_UNIQUE_CONSTRAINT) == true
```
The actual error message is this

```
org.springframework.dao.DataIntegrityViolationException: could not execute statement [ERROR: duplicate key value violates unique constraint "cas1_offenders_crn"
  Detail: Key (crn)=(SQJNHJ) already exists.] [insert into cases (created_at,crn,last_updated_at,name,noms_number,tier_v2,tier_v3,version,id) values (?,?,?,?,?,?,?,?,?)]; SQL [insert into cases (created_at,crn,last_updated_at,name,noms_number,tier_v2,tier_v3,version,id) values (?,?,?,?,?,?,?,?,?)]; constraint [cas1_offenders_crn]
```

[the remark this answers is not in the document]
vbala-moj: The error chain is as follows, so we could use the generated sequence as suggested. However, we don’t have access to the `PSQLException`, so I think we should stop at `ConstraintViolationException` instead.

```
org.springframework.dao.DataIntegrityViolationException: could not execute statement [ERROR: duplicate key value violates unique constraint "cas1_offenders_crn"
  Detail: Key (crn)=(SQJNHJ) already exists.]

Caused by: org.hibernate.exception.ConstraintViolationException: could not execute statement [ERROR: duplicate key value violates unique constraint "cas1_offenders_crn"
  Detail: Key (crn)=(SQJNHJ) already exists.]

Caused by: org.postgresql.util.PSQLException: ERROR: duplicate key value violates unique constraint "cas1_offenders_crn"
  Detail: Key (crn)=(SQJNHJ) already exists.
```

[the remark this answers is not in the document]
vbala-moj: Updated as sugested.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/service/EnsureCasePersistenceService.kt
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * In particular, create operations must commit or fail independently
+ * so that unique constraint violations caused by concurrent inserts
+ * can be handled by re-reading the case created by the competing request.
+ *
+ * The transaction boundaries in this class are intentional and should be
+ * reviewed carefully before being changed.
+ */
+@Service
+class EnsureCasePersistenceService(
```
davidatkinsuk: Could we maybe call this `CasePeristenceService` so it sits alongside the `CaseService` in directory listings?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/service/CaseService.kt
```diff
@@ -192,5 +184,11 @@ class CaseService(
     sentryService.captureException(CaseNotFound("Case with CRN $crn not found"))
   }
 
+  private fun isCrnDuplicateException(
+    e: DataIntegrityViolationException,
+  ): Boolean = generateSequence<Throwable>(e) { it.cause }
+    .filterIsInstance<ConstraintViolationException>()
+    .any { it.constraintName == CASES_CRN_UNIQUE_CONSTRAINT }
```
davidatkinsuk, line 193: nice

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/service/EnsureCasePersistenceService.kt
```diff
@@ -0,0 +1,50 @@
...
+
+  @Transactional(propagation = Propagation.REQUIRES_NEW, readOnly = true)
+  fun getCase(crn: String) = caseRepository.findByCrn(crn)
+
+  @Transactional(propagation = Propagation.REQUIRES_NEW)
+  fun createCase(caseSummary: CaseSummary, tiers: CaseTiers): CaseEntity = CaseEntity.createFrom(caseSummary, tiers)
+    .let(caseRepository::saveAndFlush)
+
+  @Transactional(propagation = Propagation.REQUIRES_NEW)
+  fun findAndUpdate(
```
davidatkinsuk: maybe `updateIfExists` as a name may make it clearer that nothing happens if it doesn't exist?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/service/CaseService.kt
```diff
@@ -46,26 +45,48 @@ class CaseService(
...
     val normalizedCrn = crn.uppercase()
     val caseSummary = getCaseSummary(normalizedCrn)
     val tiers = fetchAvailableTiers(normalizedCrn)
 
-    val caseEntity = caseRepository.findByCrn(caseSummary.crn.uppercase())
-      ?.updateFrom(caseSummary, tiers)
-      ?: newCaseEntity(caseSummary, tiers)
-
-    return caseRepository.saveAndFlush(caseEntity).toDto()
+    ensureCasePersistenceService.findAndUpdate(normalizedCrn, caseSummary, tiers)
```
davidatkinsuk: i appreciate it'll be a little more verbose, but i think it may make the shape of this code clearer if we used if/else constructs e.g.

```
val existingCase = ensureCasePersistenceService.findAndUpdate(normalizedCrn, caseSummary, tiers)
if(existingCase != null) {
   return it.toDto()
} else {
   do the rest of the code
}
```

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/service/CaseService.kt
```diff
@@ -46,26 +45,48 @@ class CaseService(
...
+      ?.let { return it.toDto() }
+
+    return try {
+      this@CaseService.ensureCasePersistenceService.createCase(caseSummary, tiers).toDto()
```
davidatkinsuk: is this@CaseService required here?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/entity/CaseEntity.kt
```diff
@@ -124,6 +126,34 @@ data class CaseEntity(
...
+  fun updateFrom(caseSummary: CaseSummary, tiers: CaseTiers): CaseEntity = apply {
+    name = caseSummary.buildName()
+    nomsNumber = caseSummary.nomsId
+    if (tiers.v2 != null) {
+      tierV2 = tiers.v2
+    }
+    if (tiers.v3 != null) {
+      tierV3 = tiers.v3
+    }
+  }
```
davidatkinsuk: I think it may be more consistent with existing code to put these functions in the new persistence service (i.e. services builds entities)? That service class could then embed the `CaseTiers` type? Would that work? I'm inclined to agree with co-pilot on the cross package coupling that may be best avoided if possible (E.g. we're referencing an API type in the entity model)

davidatkinsuk (approved):
A few comments from me. Thanks for iterating to come up with a pretty elegant solution

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/service/EnsureCasePersistenceService.kt
vbala-moj: Initially, I had that as the name, but thinking about it again, I feel it’s a bit misleading. Its functions are quite specific to the `ensureCaseExist` workflow and involve distinct transaction propagation.

Ideally, the `REQUIRED_NEW` propagation isn’t needed for most of the `CaseService` workflow, especially `getCase`, where using it could potentially affect atomicity in  the feature. 

That said, I don’t mind changing the name if you think having it alongside CaseService in the directory listing would be more useful.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/service/CaseService.kt
vbala-moj: That’s how we used to do things, once upon a time 😀

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/service/CaseService.kt
vbala-moj: Not at all, will update.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/entity/CaseEntity.kt
vbala-moj: Yeah, I think that’s a fair point. I agree that having the entity model depend directly on an API type isn’t ideal, and avoiding that cross-package coupling would make the dependency direction cleaner.

I originally had these functions in the persistence service, but later changed my mind as I thought they were more of a domain-specific concern.

I’m happy to move the entity-building functions back into the persistence service and have it own/embed the `CaseTiers` type if that gives us a cleaner separation. I think that’s probably a better direction overall.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/common/service/EnsureCasePersistenceService.kt
davidatkinsuk: I guess this structure has been forced on us by how spring transaction delineation works with annotations (transaction template would solve this but it's a pain to unit test)

Leave the name for now and we'll see if it becomes a cognitive problem going forward.

PR #5133, opened 2026-08-12
FM-1104 Cross-Region Authorization Bypass in CAS3 Premises Update endpoint

muhammad-elabdulla:
This [PR FM-1104 ](https://dsdmoj.atlassian.net/browse/FM-1104) is to prevent users from updating premises in regions they do not belong to.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/service/Cas3PremisesService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/service/Cas3PremisesService.kt
@@ -13,7 +13,6 @@ import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas3.model.CAS3PremisesA
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas3.model.CAS3PremisesUnarchiveEvent
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas3.model.Cas3PremisesArchiveAction
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas3.model.Cas3PremisesStatus
-import uk.gov.justice.digital.hmpps.approvedpremisesapi.cas3.service.Cas3UserAccessService
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.common.results.CasResult
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.common.results.CasResultValidatedScope
 import uk.gov.justice.digital.hmpps.approvedpremisesapi.common.results.CasResultValidatedScope.Companion.NDELIUS_MAX_POSTCODE_LENGTH
@@ -54,7 +53,7 @@ class Cas3PremisesService(
     val premises = cas3PremisesRepository.findByIdOrNull(premisesId)
     if (premises == null) return CasResult.NotFound("Cas3Premises", premisesId.toString())
     premises.postcode = premises.postcode.trim()
-    if (!cas3UserAccessService.currentUserCanViewPremises(premises.probationDeliveryUnit.probationRegion.id)) return CasResult.Unauthorised()
+    if (!cas3UserAccessService.currentUserCanViewOrUpdatePremises(premises.probationDeliveryUnit.probationRegion.id)) return CasResult.Unauthorised()
     return CasResult.Success(premises)
   }
   fun getPremises(premisesId: UUID): Cas3PremisesEntity? = cas3PremisesRepository.findByIdOrNull(premisesId)
@@ -143,6 +142,7 @@ class Cas3PremisesService(
     return success(cas3PremisesRepository.save(premises))
   }
 
+  @SuppressWarnings("ReturnCount")
   fun updatePremises(
     premisesId: UUID,
     reference: String,
@@ -151,7 +151,6 @@ class Cas3PremisesService(
     town: String?,
     postcode: String,
     localAuthorityAreaId: UUID?,
-    probationRegionId: UUID,
     characteristicIds: List<UUID>,
     notes: String?,
     probationDeliveryUnitId: UUID,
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/service/Cas3PremisesService.kt
```diff
@@ -157,7 +157,7 @@ class Cas3PremisesService(
     probationDeliveryUnitId: UUID,
     turnaroundWorkingDays: Int?,
   ): CasResult<Cas3PremisesEntity> = validatedCasResult {
-    val premises = cas3PremisesRepository.findByIdOrNull(premisesId)
+    val premises = cas3PremisesRepository.findByIdAndProbationRegionId(premisesId, probationRegionId)
       ?: return CasResult.NotFound("Cas3Premises", premisesId.toString())
```
davidatkinsuk: I think think this should be `CasResult.Unauthorised`? This will also make it clearer why we're making this check

davidatkinsuk: whilst it's more clunky, would it not be better to use getValidatedPremises() as this makes it clear why we're doing this check? (also i have no idea why that has postcode trim logic in it!?)

muhammad-elabdulla: I’m not sure whether we should return Unauthorised because the premises Id may not exist in the service so in this case is not unauthorised

davidatkinsuk: Fair point. I think that suggests we need to be doing something like `getValidatedPremises()` (or reuse/refactor that function logic) so we can differentiate between not found and unauthorised

muhammad-elabdulla: I think the Trim is because CAS3 had situations where the UI was sending postcodes with spaces.

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/service/Cas3PremisesService.kt
```diff
@@ -157,8 +161,10 @@ class Cas3PremisesService(
     probationDeliveryUnitId: UUID,
     turnaroundWorkingDays: Int?,
   ): CasResult<Cas3PremisesEntity> = validatedCasResult {
-    val premises = cas3PremisesRepository.findByIdOrNull(premisesId)
-      ?: return CasResult.NotFound("Cas3Premises", premisesId.toString())
+    val premises = when (val validatedPremises = getValidatedPremises(premisesId, probationRegionId)) {
+      is CasResult.Success -> validatedPremises.value
+      else -> return validatedPremises
+    }
```
davidatkinsuk: Taking a step back:

As far as i understand the linked ticket, we should be completely ignoring the `Cas3UpdatePremises.probationRegionId` value being passed in, or at least not implicitly trusting it. 

`Cas3UpdatePremises.probationRegionId` is used in the controller for validation, so IMO that usage should  _also_ be removed and instead we should be checking against region on the loaded premises. I appreciate that logically the code may be doing that check right now, but it's not very clear IMO

Instead I propose that we:

0. Undo the refactor of `getValidatedPremises`. That was a bad steer from me because i hadn't appreciated the existing check in there is different because it involves a role.
1. move the current `cas3UserAccessService.currentUserCanAccessRegion` check from the controller into the service, using the region from the loaded premises. If premises can't be found, return CasResult.NotFound.
2. then do the check we're trying to fix as part of this ticket. I propose we add a new function to `cas3UserAccessService` called `currentUserCanUpdatePremises`. Pass in the probation region from the premises into this function and compare to the current user's region
3. Deprecate `Cas3UpdatePremises.probationRegionId` as it should not longer be used by anything

I think this approach should fix the issue and make the checks clearer?

muhammad-elabdulla: There is an option in the UI to select the premises region, so we need to check the value of probationRegionId 
being passed

muhammad-elabdulla: I agree to create a function currentUserCanUpdatePremises will be more clean approach

davidatkinsuk: > There is an option in the UI to select the premises region, so we need to check the value of probationRegionId
being passed

But it doesn't appear to be used by anything else in the code (other than checks) so i'm not sure what's its purpose is? Should we be updating the premise's region maybe (which would be another bug?). maybe i'm missing something in the code

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/service/Cas3PremisesService.kt
```diff
@@ -160,14 +159,21 @@ class Cas3PremisesService(
...
       ?: return CasResult.NotFound("Cas3Premises", premisesId.toString())
 
+    val probationDeliveryUnit = probationDeliveryUnitRepository.findByIdOrNull(probationDeliveryUnitId)
+      ?: return CasResult.NotFound("ProbationDeliveryUnit", probationDeliveryUnitId.toString())
+
+    if (!cas3UserAccessService.currentUserCanViewOrUpdatePremises(premises.probationDeliveryUnit.probationRegion.id) ||
+      !cas3UserAccessService.currentUserCanViewOrUpdatePremises(probationDeliveryUnit.probationRegion.id)
+    ) {
+      return CasResult.Unauthorised()
+    }
```
davidatkinsuk, line 169: perfect 👍

davidatkinsuk (approved):
Great work 👍

PR #5134, opened 2026-08-12
SAS 883: filter default rejection rationale

SaintRhys:
[This](https://dsdmoj.atlassian.net/browse/SAS-883) filters out 'default" rejection rationale.

FE was displaying "default" for rejection reason when there wasn't a `referralRejectionReason` and the fallback `rejectionRationale` was just "default"


SaintRhys:
Has copilot just given up reviewing?

davidatkinsuk:
> Has copilot just given up reviewing?

i think it's just being a bit slow?

src/main/kotlin/uk/gov/justice/digital/hmpps/approvedpremisesapi/cas3/transformer/Cas3AssessmentTransformer.kt
```diff
@@ -96,7 +96,7 @@ class Cas3AssessmentTransformer(
       applicationStatus = application.getStatus(),
       assessmentStatus = a.deriveAssessmentStatus(),
       type = ServiceType.CAS3,
-      referralRejectionReason = a.referralRejectionReason?.name ?: a.rejectionRationale,
+      referralRejectionReason = a.referralRejectionReason?.name ?: a.rejectionRationale?.takeUnless { it == "default" },
```
davidatkinsuk: to try and keep business logic in the service/entity layers, i propose we add the following function to `ReferralRejectionReasonEntity` and use it here:

`fun isDefault() = name == "default"` 

That way it's less likely that someone will break this logic if/when changing the label on that rejection (Which is unlikely but possible i guess!)

davidatkinsuk (approved):
Just one comment from me

PR #5139, opened 2026-08-14
Fix deprecated field documentation and slight refactor

davidatkinsuk:
A few small fixes:

* Fix the reference in a space booking deprecation message
* Rename all booking services in CAS1 to be prefixed with Spacebooking


muhammad-elabdulla (approved):
LGTM

